PR #13, opened 2021-02-23
fix bug in OperatorBase::add_diagonal()

nfehn:
The new diagonal computation introduced in MR !384 (Mg with hanging nodes) (gitlab) introduced a bug. I guess the newly introduced tests have not been run ;)

This PR is a bugfix. 

Independently from this PR, the ``dealii::MatrixFreeTools::compute_diagonal()`` function needs to be extended by another lambda that takes ``ExaDG::OperatorBase::reinit_cell()`` as argument, so that classes derived from ``OperatorBase`` can do the setup that they need, e.g., nonlinear operators.

### DIFF
--- a/include/exadg/operators/operator_base.cpp
+++ b/include/exadg/operators/operator_base.cpp
@@ -400,6 +400,10 @@ OperatorBase<dim, Number, n_components>::add_diagonal(VectorType & diagonal) con
       *matrix_free,
       diagonal,
       [&](auto & integrator) -> void {
+        // TODO this line is currently needed as bugfix, but should be
+        // removed because reinit is now done twice
+        this->reinit_cell(integrator.get_current_cell_index());
+
         integrator.evaluate(integrator_flags.cell_evaluate.value,
                             integrator_flags.cell_evaluate.gradient,
                             integrator_flags.cell_evaluate.hessian);


include/exadg/operators/operator_base.cpp
```diff
@@ -400,6 +400,10 @@ OperatorBase<dim, Number, n_components>::add_diagonal(VectorType & diagonal) con
       *matrix_free,
       diagonal,
       [&](auto & integrator) -> void {
+        // TODO this line is currently needed as bugfix, but should be
+        // removed because reinit is now done twice
+        this->reinit_cell(integrator.get_current_cell_index());
```
peterrum, line 405: You could write:
```cpp
    unsigned int cell_index = numbers::invalid_unsigned_int;
    MatrixFreeTools::compute_diagonal<dim, -1, 0, n_components, Number, VectorizedArray<Number>>(
      *matrix_free,
      diagonal,
      [&](auto & integrator) -> void {
        if(cell_index != integrator.get_current_cell_index())
          {
            cell_index = integrator.get_current_cell_index()
            this->reinit_cell(integrator.get_current_cell_index());
          }
```
Not a perfect solution since we make here the implicit assumption that MF is configured without threading but this is an assumption we make in the rest of the code as well.

nfehn: What is the reason for this change? Why not use cell_index in the line below?

kronbichler: I think it makes more sense to let the user attach a separate `reinit()` function to make this more clear - this way the `MatrixFreeTools` function in deal.II is independent of threading. We could have an empty function as a default.

nfehn: @peterrum If the intention is that we do not need to extend the dealii function, the problem then is that ``this->reinit_cell()`` can be something expensive that needs to be done outside of the loop over all columns (diagonal computation). I think this is not the case for your suggestion, but would really need an extension of ``MatrixFreeTools::compute_diagonal()``

kronbichler: @nfehn I think that's the intent with the `cell_index` variable, to make the code only run once. But it is quite involved to spot this intent (and we send in a counter variable).

nfehn: Obviously, I did not understand it. So maybe this should guide the design ;)

kronbichler (approved):

PR #18, opened 2021-02-25
MG: avoid code redundancy

nfehn:


peterrum (approved):
I will add the assert in a follow-up PR.

Issue #19, opened 2021-02-25
Add Asserts to MultigridPreconditionerBase to exclude invalid multigrid parameters


nfehn:
`p:f:T` in combination with `use_global_coarsening` does not allow h-MG, which should be asserted.

The standard transfer type (`use_global_coarsening = false`) does not allow hanging nodes in general, which should also be asserted.

PR #20, opened 2021-02-25
Add two asserts in MultigridPreconditionerBase

peterrum:


include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -375,8 +376,17 @@ MultigridPreconditionerBase<dim, Number>::do_initialize_dof_handler_and_constrai
   {
     // create coarse grid triangulations only once
     if(data.involves_h_transfer() && coarse_grid_triangulations.empty())
+    {
+      AssertThrow(tria->n_global_levels() == 1 ||
+                    dynamic_cast<parallel::fullydistributed::Triangulation<dim> const *>(tria) ==
+                      nullptr,
+                  ExcMessage(
+                    "A parallel::fullydistributed::Triangulation cannot be refined if the option "
```
nfehn, line 384: Just a minor comment here regarding the message: 'To me, this sounds like the aim of this function would be to refine the `p:f:T`?

I would therefore formulate it from a multigrid perspective: "h-transfer is currently not supported for the option use_global_coarsening in combination withr a p:f:T that contains refinements. Either use a p:f:T without refinements, a p:d:T, or a MultigridType without h-transfer." Is this statement correct?

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -410,6 +420,10 @@ MultigridPreconditionerBase<dim, Number>::do_initialize_dof_handler_and_constrai
   }
   else // can only be used for triangulations without hanging nodes
   {
+    Assert(tria->has_hanging_nodes() == false,
+           ExcMessage(
+             "Hanging nodes are only supported with the option use_global_coarsening enabled."));
```
nfehn, line 425: Independently from this PR, do you think a better name for `use_global_coarsening` would be the name `enable_hanging_nodes` from the perspective of ExaDG users, because the condition in the Assert and also the comment related to the `else` above indicate that this boolean describes something different? Then, I could change this in a subsequent PR.

nfehn:
If not already done, we should check that the test are running successfully with these additional asserts.

nfehn:
closes #19

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
peterrum: Done.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
peterrum: No I don't think so. If it is only about hanging nodes, we would not need any flag and could query the triangulation that information. The flag indicates which transfer implementation should be used. How, we distinguish both implementation is up to discussion; it does not have to be via the bool `use_global_coarsening` but could also be done via an enum.

nfehn: I agree principally, and that's why my suggestion is not `has_hanging_nodes` (which can be answered with yes/no for a given triangulation as you write), but `enable_hanging_nodes`, which reflects that this transfer type can deal with hanging nodes, but also if there are no hanging nodes. Querying the bool parameter from the triangulation would affect the implementation, i.e., it reduces the amount of parameter combinations that are possible. For verification, we found it useful to be able to use both options for the case without hanging nodes.

My point is the following: Currently, the situation might happen that the user writes `use_global_coarsening=true` in the input file, but gets an error message with the suggestion to use a `MultigridType` without h-transfer!? At the same time, the name "global coarsening" insinuates some form of h-transfer in my opinion. This is a contradiction that a general user can not resolve in my opinion. A user of ExaDG does not know that `use_global_coarsening` comes from a MGTransfer implementation in deal.II with a similar name. Even worse, the user can not know that `use_global_coarsening` triggers which implementation to choose for all multigrid types (c, p, h) and not just the h-transfer part of the multigrid algorithm. For the ExaDG user, this boolean is related to enabling hanging nodes in my opinion.

A good documentation of this input parameter would explain the incompatibility of this parameter with p:f:T and refinements (which is not the case currently, see https://github.com/exadg/exadg/blob/965853f349ed30a396b41661ddbf897205de4545/include/exadg/solvers_and_preconditioners/multigrid/multigrid_input_parameters.h#L318). When thinking about how to write this docu so that it is clear to the user, I come to the conclusion that the name `use_global_coarsening` is misleading.

nfehn:
Did the tests succeed?

peterrum:
The tests don't seem to be robust. The following tests fail:
```
The following tests FAILED:
	  6 - tests/diffusive_implicit_time_int.debug (Failed)
	 11 - tests/coupled_bdf2_moving_mesh.debug (Failed)
	 13 - tests/parabolic_inflow.debug (Failed)
	 14 - tests/parabolic_inflow_symmetry_bc.debug (Failed)
	 15 - tests/periodic_bc.debug (Failed)
	 16 - tests/pressure_inflow.debug (Failed)
	 20 - tests/2d.debug (Failed)
	 21 - tests/3d.debug (Failed)
	 22 - solvers_and_preconditioners/elementwise_gmres.debug (Failed)
	 23 - utilities/timer.debug (Failed)
```
due to some rounding errors.

nfehn:
sorry, you need to run in release mode, then everything should pass. Of course, we have to discuss the discrepancy, independently from this PR.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
peterrum: OK.

peterrum:
Release does not look better.
```
The following tests FAILED:
	 10 - tests/steady_navier_stokes_coupled.release (Failed)
	 11 - tests/coupled_bdf2_moving_mesh.release (Failed)
	 13 - tests/parabolic_inflow.release (Failed)
	 14 - tests/parabolic_inflow_symmetry_bc.release (Failed)
	 15 - tests/periodic_bc.release (Failed)
	 16 - tests/pressure_inflow.release (Failed)
	 20 - tests/2d.release (Failed)
	 21 - tests/3d.release (Failed)
	 22 - solvers_and_preconditioners/elementwise_gmres.release (Failed)
	 23 - utilities/timer.release (Failed)

```

nfehn:
you are right, I tested different machines and the tests show different results. 

But I think I can merge now because the changes of this PR would have caused a crash of the simulation and not different numbers.

nfehn (approved):

PR #21, opened 2021-02-26
adjust comments related to new multigrid transfer variant

nfehn:
@peterrum, @bergbauer Related to the discussion in PR #20, I added a comment that explains that `use_global_coarsening` has to be activated in order to deal with hanging nodes. I also voted for renaming this variable (e.g. `enable_hanging_nodes`), but would only do that in case you agree on that.


bergbauer:
> @peterrum, @bergbauer Related to the discussion in PR #20, I added a comment that explains that `use_global_coarsening` has to be activated in order to deal with hanging nodes. I also voted for renaming this variable (e.g. `enable_hanging_nodes`), but would only do that in case you agree on that.

The comment is very helpful. I would keep the name `use_global_coarsening` to indicate which transfer operator and dealii functionality is used.

bergbauer (approved):

peterrum (approved):

PR #24, opened 2021-03-01
fix bug related to *Mg with hanging nodes* MR

nfehn:
The "Mg with hanging nodes* MR did not implement the switch between the two multigrid transfer options for the ConvDiff module. The behavior is therefore unexpected for the user, and can be considered a bug in the current form since it ignores the multigrid parameter `use_global_coarsening`.

The problem is solved by introducing a member function in the base class that needs to be called by all derived classes whenever initializing transfer operators.


peterrum (approved):

peterrum:
This should work once https://github.com/dealii/dealii/pull/11815 is merged.

nfehn:
> This should work once dealii/dealii#11815 is merged.

@peterrum What do you mean? Should I wait here?

PR #28, opened 2021-03-02
reduce compile time for multigrid by forward declarations

nfehn:
as pointed out by @peterrum, compilation of ExaDG takes unnecessarily long when making changes to multigrid ingredients.


peterrum (approved):
Great :100:

kronbichler (approved):

PR #30, opened 2021-03-02
fix bug related to accessing wrong mapping data for global coarsening MG

nfehn:
The "Mg with hanging nodes" merge request did not support global coarsening multigrid for mappings such as `MappingQCache`. This is fixed in the present pull request.

A TODO is added to the code since a correct transfer of mapping information to coarser multigrid levels still needs to be done. Currently, `MappingQGeneric<dim>(1)` is used as default and in order to avoid a crash of the simulation.

### DIFF
--- a/include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
+++ b/include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
@@ -325,7 +325,33 @@ MultigridPreconditionerBase<dim, Number>::initialize_coarse_grid_triangulations(
 
       coarse_grid_triangulations =
         MGTransferGlobalCoarseningTools::create_geometric_coarsening_sequence(*tria);
+
+      // We also need separate mappings for the coarse grid triangulations
+      unsigned int const n_h_levels = coarse_grid_triangulations.size();
+      coarse_grid_mappings.resize(n_h_levels - 1);
+      for(unsigned int h_level = 0; h_level < coarse_grid_mappings.size(); ++h_level)
+      {
+        // TODO
+        coarse_grid_mappings[h_level].reset(new MappingQGeneric<dim>(1));
+      }
     }
+  }
+}
+
+template<int dim, typename Number>
+Mapping<dim> const &
+MultigridPreconditionerBase<dim, Number>::get_mapping(unsigned int const h_level) const
+{
+  if(data.use_global_coarsening)
+  {
+    if(data.involves_h_transfer() && h_level < coarse_grid_mappings.size())
+      return *(coarse_grid_mappings[h_level]);
+    else
+      return *mapping;
+  }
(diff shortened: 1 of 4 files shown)


include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -575,11 +602,14 @@ MultigridPreconditionerBase<dim, Number>::initialize_affine_constraints(
 
   DoFTools::make_hanging_node_constraints(dof_handler, affine_constraints);
   for(auto & it : dirichlet_bc)
-    VectorTools::interpolate_boundary_values(*mapping,
+  {
+    MappingQGeneric<dim> mapping_dummy(MappingQGeneric<dim>(1));
```
kronbichler, line 606:
```suggestion
    MappingQGeneric<dim> mapping_dummy(1);
```
(the same also in the function below.

kronbichler (approved):

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -861,7 +891,8 @@ MultigridPreconditionerBase<dim, Number>::do_initialize_transfer_operators(
   {
     auto tmp = std::make_shared<MGTransferGlobalCoarsening<dim, MultigridNumber, VectorTypeMG>>();
 
-    tmp->reinit(*mapping, matrix_free_objects, constraints, constrained_dofs, dof_index);
+    MappingQGeneric<dim> mapping_dummy(1);
+    tmp->reinit(mapping_dummy, matrix_free_objects, constraints, constrained_dofs, dof_index);
```
nfehn, line 895: @peterrum can I remove Mapping from this interface, or will it be needed in the future?

nfehn: the `reinit()` function is not part of the interface of the base class `MGTransfer`

peterrum: It is not needed. I think we wrote this like this to be consistent with the reinit function of the other transfer operator. I don't mind if you'd remove it.

nfehn: removed it, and also another parameter

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
nfehn: I changed this as suggested

nfehn:
@peterrum do you agree with this version?

peterrum (approved):

PR #32, opened 2021-03-03
Refactoring of MovingMesh/Mapping design

nfehn:
The moving mesh classes provide functionality that is useful also for other problems. A refactoring is required in order to be able to use this function also for multigrid with global coarsening (with the open problem of how to transfer the fine level mapping information to coarser levels). 

The key element of this PR is to derive all the moving mesh classes from `dealii::MappingQCache`, and to introduce another class called `MappingFiniteElement` in between to get rid of the "moving mesh" notion for problems where moving the mesh is not relevant, e.g. multigrid with global coarsening.


peterrum (approved):
I don't like the name of the class that much. In deal.II, we already have an `MappingFE` and it's purpose it is different that of the class here. For me, the class here is more like `MappingFEFields`. To prevent confusions, I would suggest to rename the class. But feel free to ignore my comments if you disagree.

Everything else looks fine!

nfehn:
Yes, I took the freedom to choose names independently from deal.II. I actually don't like the names in deal.II so much (they are also against our naming conventions, i.e., how to read/spell `...FEF...`?). `MappingQGeneric`, `MappingQCache`??? Just don't know what these names should tell me, but I can live with that :) 

I mean, in principle I agree and I am also of the opinion to name things consistently with deal.II, but I think the mapping design in deal.II is not a good example for that. Another example in deal.II would be multigrid and all the names beginning with `MG...` ...

nfehn:
Would you prefer `MappingDofVector`? This is the second option I thought about.

peterrum:
Yes, that would be better (with capital f -> `MappingDoFVector`). But let's wait for other opinions on this matter!

nfehn:
Note that the use of this class has not been safe so far. Fixed this in https://github.com/exadg/exadg/pull/32/commits/06fef7b9ee6b35c01d9ee1d03eab9e56c7a2f757. Derived classes initialized the mapping_q_cache object, but we need to guarantee a correct initialization already in the constructor of `MappingFiniteElement` so that the class can not be misused.

I decided against `MappingDofVector` again because this class provides also the interface with `Function<dim>` and because I don't want to capitalize the f.

nfehn:
I renamed into `MappingDoFVector` here https://github.com/exadg/exadg/pull/32/commits/2bbfabe89b1b6b61e9d3b8b6c086fc30eababb67, since @kronbichler was also not happy with the `FiniteElement` name

include/exadg/grid/mapping_finite_element.h
```diff
@@ -0,0 +1,373 @@
...
+          fe_values.reinit(typename Triangulation<dim>::cell_iterator(cell));
+          std::vector<types::global_dof_index> dof_indices(fe.dofs_per_cell);
+          cell->get_mg_dof_indices(dof_indices);
+
+          // extract displacement and add to original position
+          for(unsigned int i = 0; i < scalar_dofs_per_cell; ++i)
+          {
+            grid_coordinates[i] =
+              fe_values.quadrature_point(this->hierarchic_to_lexicographic_numbering[i]);
+          }
```
nfehn, line 335: @kronbichler Isn't the above loop too much? fe_values/fe_nothing should also not be necessary?

include/exadg/grid/mapping_finite_element.h
```diff
@@ -0,0 +1,373 @@
...
+
+          for(unsigned int i = 0; i < dof_indices.size(); ++i)
+          {
+            std::pair<unsigned int, unsigned int> const id = fe.system_to_component_index(i);
+
+            if(fe.dofs_per_vertex > 0) // FE_Q
+            {
+              grid_coordinates[id.second][id.first] +=
```
nfehn, line 343: @kronbichler Adding into the vector is probably not wrong, but an assignment would be clearer?

include/exadg/grid/mapping_finite_element.h
kronbichler: How do you mean? We need the position of the non-deformed point to which we want to add the displacement vector. So the question is whether we hand in here a displacement vector (against the undeformed state) or absolute coordinates. It seems we have absolute coordinates, in the code above? In that case you are right, we do not need `FEValues` and can simply assign in the loop below.

nfehn: Yes, I think this is the logic that we wanted to have, i.e., no "displacement"-notion for the `..._multigrid()` part of this class.

kronbichler: Yes, then this can go away and we do not need to add below.

nfehn: see https://github.com/exadg/exadg/pull/32/commits/60e8c02efc9f9d2034602344d6dcc4b1db45ba25

kronbichler (approved):

nfehn:
@peterrum Could you check whether I implemented the correct logic for global coarsening / global refinement in the multigrid implementation, see this commit https://github.com/exadg/exadg/pull/32/commits/b6cba99bd1257e017332b5ad4b404af779e4b61d ? Of course there is still the big TODO that needs to be done for the global coarsening case.

peterrum:
> @peterrum Could you check whether I implemented the correct logic for global coarsening / global refinement in the multigrid implementation, see this commit b6cba99 ? Of course there is still the big TODO that needs to be done for the global coarsening case.

Looks correct.

nfehn:
I think I am done now with refactoring. The global coarsening mapping transfer should be done in a separate PR.

peterrum (approved):

nfehn:
the tests also passed on my machine apart from the round-off issue and the slight variation in iteration counts that we recognized before (and that should be addressed elsewhere).

PR #39, opened 2021-03-10
make initialize_multigrid() functions for MappingDoFVector free functions

nfehn:
This PR shows that the `MappingDoFVector::initialize_multigrid()` functions are best implemented as free functions. This PR also shows that we only have to add two functions (`fill_grid_coordinates_vector()` and another `initialize()` function) to `dealii::MappingQCache` to enable all the ALE and mapping-multigrid functionality in ExaDG.

### DIFF
--- a/include/exadg/grid/mapping_dof_vector.h
+++ b/include/exadg/grid/mapping_dof_vector.h
@@ -24,6 +24,7 @@
 
 // deal.II
 #include <deal.II/base/conditional_ostream.h>
+#include <deal.II/base/mg_level_object.h>
 #include <deal.II/dofs/dof_tools.h>
 #include <deal.II/fe/fe_nothing.h>
 #include <deal.II/fe/fe_q.h>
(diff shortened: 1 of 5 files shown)


peterrum (approved):
Looks good!

>This PR also shows that we only have to add two functions (fill_grid_coordinates_vector()

As far as I can tell this function is equivalent with `dealii::VectorTools::get_position_vector()` if it would take `Mapping` as input argument. I have opened a PR for that purpose (https://github.com/dealii/dealii/pull/11880). Once the PR is merged, could you check if it produces the same results or we need to make further modifications.

> and another initialize() function) to dealii::MappingQCache to enable all the ALE and mapping-multigrid functionality in ExaDG.

I am on this.

kronbichler (approved):

kronbichler:
I think the `fill_grid_coordinates_vector` topic should be addressed elsewhere because it is independent of what we do here, so I'm merging now.

PR #41, opened 2021-03-16
Give dof_handler to setup function instead of triangulation

bergbauer:


nfehn:
how did you realize this or why did you only change the function call and not the implementation?

bergbauer:
I had problems with compilation after pulling the newest deal master, in the implementation the `dof_handler` is requested

nfehn:
So this is a bug which we did not detect so far because the triangulation could be interpreted as a DoFHandler before recent changes in deal.II? @kronbichler can you explain this?

peterrum:
> So this is a bug which we did not detect so far because the triangulation could be interpreted as a DoFHandler before recent changes in deal.II?

Implicitly the constructor of `DoFHandler` was called. We labeled the constructor of `DoFHandler` that takes the `Triangulation` as explicit to prevent such issues (inside of the library) last week.

nfehn (approved):

PR #42, opened 2021-03-17
Remove Assert in compute_performance_metrics()

nfehn:
It might happen that the number of linear iterations is 0, e.g., for trivial initial and boundary conditions, in which case the solver should not terminate but instead use reasonable default values.

### DIFF
--- a/include/exadg/poisson/driver.cpp
+++ b/include/exadg/poisson/driver.cpp
@@ -223,7 +223,7 @@ Driver<dim, Number>::print_performance_results(double const total_time) const
 
   unsigned int const N_mpi_processes = Utilities::MPI::n_mpi_processes(mpi_comm);
 
-  double const t_10 = solve_time * n_10 / iterations;
+  double const t_10 = iterations > 0 ? solve_time * double(n_10) / double(iterations) : solve_time;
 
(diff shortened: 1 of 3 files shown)


include/exadg/poisson/driver.cpp
```diff
@@ -223,7 +223,7 @@ Driver<dim, Number>::print_performance_results(double const total_time) const
 
   unsigned int const N_mpi_processes = Utilities::MPI::n_mpi_processes(mpi_comm);
 
-  double const t_10 = solve_time * n_10 / iterations;
+  double const t_10 = iterations > 0 ? solve_time * double(n_10) / double(iterations) : solve_time;
```
bergbauer, line 226:
```suggestion
  double const t_10 = iterations > 0 ? solve_time * n_10 / double(iterations) : solve_time;
```

bergbauer (approved):

include/exadg/poisson/driver.cpp
nfehn: principally, non of the casts is necessary due to the C++ associativity from left to right, but I thought it is better to be explicit about the casts.

PR #45, opened 2021-03-22
Avoid warning regarding overriden virtual functions

kronbichler:
The clang-12 compiler I am using to compile exadg reports warnings when we override virtual functions but do not add the `override` attribute. I tried to fix most of those warnings with this PR. I am not entirely sure if I fixed all already, but since I cannot immediately recheck, I wanted to post it now.

One thing I was undecided about is whether we want to report the `virtual` keyword on those overriden functions. We use both variants within exadg, and I have a slight preference for adding it nonetheless, but I am happy to get convinced the other way around as well.

(I fixed two other small warnings that I will point out by separate posts in a minute so that they don't get overlooked.)

### DIFF
--- a/include/exadg/compressible_navier_stokes/postprocessor/postprocessor.h
+++ b/include/exadg/compressible_navier_stokes/postprocessor/postprocessor.h
@@ -64,11 +64,13 @@ public:
 
   virtual ~PostProcessor();
 
-  virtual void
-  setup(Operator<dim, Number> const & pde_operator);
+  void
+  setup(Operator<dim, Number> const & pde_operator) override;
 
-  virtual void
-  do_postprocessing(VectorType const & solution, double const time, int const time_step_number);
+  void
+  do_postprocessing(VectorType const & solution,
+                    double const       time,
+                    int const          time_step_number) override;
 
 protected:
   // DoF vectors for derived quantities: (p, u, T)
--- a/include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.h
+++ b/include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.h
@@ -101,22 +101,22 @@ public:
   print_iterations() const = 0;
 
   bool
-  print_solver_info() const;
+  print_solver_info() const final;
 
 protected:
-  virtual void
+  void
   allocate_vectors() override;
 
-  virtual void
+  void
   setup_derived() override;
 
-  virtual void
+  void
   read_restart_vectors(boost::archive::binary_iarchive & ia) override;
 
-  virtual void
+  void
   write_restart_vectors(boost::archive::binary_oarchive & oa) const override;
 
-  virtual void
+  void
   prepare_vectors_for_next_timestep() override;
 
   /*
@@ -126,7 +126,7 @@ protected:
   void
   calculate_sum_alphai_ui_oif_substepping(VectorType & sum_alphai_ui,
                                           double const cfl,
-                                          double const cfl_oif);
+                                          double const cfl_oif) final;
 
   void
   move_mesh(double const time) const;
@@ -162,30 +162,27 @@ private:
   initialize_vec_convective_term();
 
   void
-  initialize_oif();
+  initialize_oif() final;
 
   void
-  initialize_solution_oif_substepping(VectorType & solution_tilde_m, unsigned int i);
+  initialize_solution_oif_substepping(VectorType & solution_tilde_m, unsigned int i) final;
 
   void
   update_sum_alphai_ui_oif_substepping(VectorType &       sum_alphai_ui,
                                        VectorType const & u_tilde_i,
-                                       unsigned int       i);
+                                       unsigned int       i) final;
 
   void
   do_timestep_oif_substepping(VectorType & solution_tilde_mp,
                               VectorType & solution_tilde_m,
                               double const start_time,
-                              double const time_step_size);
+                              double const time_step_size) final;
 
   double
-  calculate_time_step_size();
+  calculate_time_step_size() final;
 
   double
-  recalculate_time_step_size() const;
-
-  virtual void
-  solve_steady_problem() = 0;
+  recalculate_time_step_size() const final;
 
   virtual VectorType const &
   get_velocity(unsigned int i /* t_{n-i} */) const = 0;
@@ -200,7 +197,7 @@ private:
   set_pressure(VectorType const & pressure, unsigned int const i /* t_{n-i} */) = 0;
 
   void
-  postprocessing() const;
+  postprocessing() const final;
 
   // Operator-integration-factor splitting for convective term
   std::shared_ptr<OperatorOIF<dim, Number>> convective_operator_OIF;
--- a/include/exadg/time_integration/time_int_bdf_base.h
+++ b/include/exadg/time_integration/time_int_bdf_base.h
@@ -108,7 +108,7 @@ protected:
    * Do one time step including different updates before and after the actual solution of the
    * current time step.
    */
-  virtual void
+  void
   do_timestep_pre_solve(bool const print_header) override;
 
   void
@@ -236,7 +236,7 @@ private:
    * Restart: read solution vectors (has to be implemented in derived classes).
    */
   void
-  do_read_restart(std::ifstream & in);
+  do_read_restart(std::ifstream & in) override;
 
   void
   read_restart_preamble(boost::archive::binary_iarchive & ia);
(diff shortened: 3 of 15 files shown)


include/exadg/incompressible_navier_stokes/postprocessor/line_plot_data.h
```diff
@@ -221,7 +221,7 @@ struct LinePlotData
   }
 
   void
-  print(ConditionalOStream & pcout)
+  print(ConditionalOStream &)
```
kronbichler, line 224: This is one other change to avoid having an unused variable.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_input_parameters.h
```diff
@@ -293,7 +293,7 @@ struct MultigridData
   {
     print_parameter(pcout, "Multigrid type", enum_to_string(type));
 
-    if(type != MultigridType::hMG || type != MultigridType::hcMG || type != MultigridType::chMG)
+    if(!(type == MultigridType::hMG || type == MultigridType::hcMG || type == MultigridType::chMG))
```
kronbichler, line 296: This is an actual change to be looked at - the previous comparison always resulted in `true` because when say the first equality is true, the second must be false by definition.

include/exadg/time_integration/time_int_bdf_base.h
```diff
@@ -220,12 +220,6 @@ class TimeIntBDFBase : public TimeIntBase
   virtual void
   prepare_vectors_for_next_timestep() = 0;
 
-  /*
-   * Solve the current time step.
-   */
-  virtual void
-  solve_timestep() = 0;
-
```
kronbichler, line 228: This looks like a change but the method is already in the base class, as indicated by the compiler warning I got.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_input_parameters.h
nfehn: I think the logical consequence would be to write a short function `involves_p_transfer()`, given that we already have such functions for h-/c-transfer (see a few lines below).

include/exadg/time_integration/time_int_bdf_base.h
nfehn: ah, makes sense.

nfehn:
I do not have a strong opinion regarding `virtual`. In the past, I think I only used this keyword when necessary (following the rule "let's do it when we need it"), but this was more a personal feeling than an expert opinion. Thinking more about it, I would introduce it where we see the possibility to derive from this class and overwrite this function, which makes sense for the postprocessor functions that get overwritten by the postprocessors for certain applications. It seems that we currently do not have that at the moment for the ConvDiff module, but it makes sense nevertheless. 

On the other hand, functions in the time integrator are not meant to be overwritten (they are an implementation of the interface and since these functions can work with shared pointers of a Postprocessor, I currently do not see a need to reimplement such functions. According to my understanding, not using `virtual` indicates that the design is meant to be closed in this regard. Otherwise, I would interpret the `virtual` keyword in the sense that I have to be careful because there might be a derived class.

bergbauer:
I would use `virtual` in a virtual function declaration, `override` when overriding such a function and `final` when a function shouldn't be overwritten anymore. But of course this is only most intuitive for me.

nfehn:
Then we should probably use `final` in the time integrators and in the Laplace operator, and introduce `virtual` only for the Postprocessors or the multigrid preconditioners (where deriving from these classes is forseeable).

bergbauer:
In my opinion also in classes which can be derived from `override` without `virtual` is sufficient. It's short and simple.

kronbichler:
OK, let me remove the `virtual` keyword then and add `final` for the time integrator.

kronbichler:
Have a look at 8d57960 for the new variant. I removed `virtual` in case `override` already specifies the `virtual` function, and marked the time integrator and Laplace operator functions as `final`.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.h
```diff
@@ -126,7 +126,7 @@ class TimeIntBDF : public TimeIntBDFBase<Number>
   void
   calculate_sum_alphai_ui_oif_substepping(VectorType & sum_alphai_ui,
                                           double const cfl,
-                                          double const cfl_oif);
+                                          double const cfl_oif) override final;
```
bergbauer, line 129: I think `final` already implies `override`. Do you want to keep it for readability?

kronbichler: You are probably right, we can skip `override` then. Do you agree @nfehn ?

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.h
```diff
@@ -101,22 +101,22 @@ class TimeIntBDF : public TimeIntBDFBase<Number>
   print_iterations() const = 0;
 
   bool
-  print_solver_info() const;
+  print_solver_info() const override final;
 
 protected:
-  virtual void
+  void
```
nfehn, line 107: I think it is not correct to remove `virtual` here?

include/exadg/compressible_navier_stokes/postprocessor/postprocessor.h
```diff
@@ -64,11 +64,13 @@ class PostProcessor : public PostProcessorBase<dim, Number>
 
   virtual ~PostProcessor();
 
-  virtual void
-  setup(Operator<dim, Number> const & pde_operator);
+  void
+  setup(Operator<dim, Number> const & pde_operator) override;
```
nfehn, line 68: `virtual` actually makes sense for postprocessors

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.h
bergbauer: It is correct, `override` implies `virtual`.

include/exadg/compressible_navier_stokes/postprocessor/postprocessor.h
kronbichler: I would be fine either way, so we need to decide about the style how we declare overriden virtual functions - do it as @bergbauer suggests with simply `override` and no `virtual`, or `virtual void ... override` in the verbose version.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.h
nfehn: We should then probably write only `final`.

include/exadg/compressible_navier_stokes/postprocessor/postprocessor.h
bergbauer: I am also fine with both options but I favor the short version. What is the style in dealii?

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.h
kronbichler: Done in 4e16828

include/exadg/compressible_navier_stokes/postprocessor/postprocessor.h
kronbichler: I checked in deal.II and it seems that it is mostly `virtual ... override`, e.g. here:
https://github.com/dealii/dealii/blob/981214b60b491f465dd8cf04e24989a2d806229c/include/deal.II/multigrid/mg_matrix.h#L78-L85
or here:
https://github.com/dealii/dealii/blob/981214b60b491f465dd8cf04e24989a2d806229c/include/deal.II/dofs/dof_handler_policy.h#L132-L133

nfehn: the `virtual ... override` combination is what I thought of originally

include/exadg/solvers_and_preconditioners/multigrid/multigrid_input_parameters.h
nfehn: done in PR #52

nfehn:
@kronbichler Could you provide the changes not related to virtual/override/final in a separate PR, before merging this one here?

kronbichler:
Let me have a look at the merge conflict and extract the other parts.

kronbichler:
I rebased this on master and squashed all commits into one to reflect the current state we discussed. Ready for re-review.

nfehn:
I think this looks good now. Thanks.

nfehn (approved):

PR #47, opened 2021-03-24
Implement BoomerAMG via PETSc

kronbichler:
This PR implements the BoomerAMG preconditioner from the HYPRE package, available through the deal.II/PETSc interface. I had to implement a second matrix infrastructure for PETSc matrices in the base operator, and then I added another preconditioner class for BoomerAMG. The current interface is somewhat wasteful because:
* we copy the vectors two times, once for `MultigridNumber` -> `TrilinosNumber` and then again from `LinearAlgebra::distributed::Vector` into PETSc's own vector type, because we need the latter for `PETScWrappers::SparseMatrix::vmult`. If we think it is important, we might get around with a single copy.
* ~~if we use a CG solver around multigrid (or GMRES for that matter), we might want to consider to use a Krylov solver directly from PETSc, rather than `SolverCG` from deal.II. This might help to get around the inefficiencies in transitioning data types.~~ I now use PETSc's solver directly.

Given that we only do expensive operations within the wrapped function (apart from vmult tests), I think the first point should not matter too much.

### DIFF
--- a/include/exadg/operators/multigrid_operator.h
+++ b/include/exadg/operators/multigrid_operator.h
@@ -52,113 +52,113 @@ public:
     return pde_operator;
   }
 
-  AffineConstraints<typename Operator::value_type> const &
-  get_affine_constraints() const override
+  virtual AffineConstraints<typename Operator::value_type> const &
+  get_affine_constraints() const
   {
     return pde_operator->get_affine_constraints();
   }
 
-  MatrixFree<dim, Number> const &
-  get_matrix_free() const override
+  virtual MatrixFree<dim, Number> const &
+  get_matrix_free() const
   {
     return pde_operator->get_matrix_free();
   }
 
-  unsigned int
-  get_dof_index() const override
+  virtual unsigned int
+  get_dof_index() const
   {
     return pde_operator->get_dof_index();
   }
 
-  types::global_dof_index
-  m() const override
+  virtual types::global_dof_index
+  m() const
   {
     return pde_operator->m();
   }
 
-  types::global_dof_index
-  n() const override
+  virtual types::global_dof_index
+  n() const
   {
     return pde_operator->n();
   }
 
-  Number
-  el(unsigned int const i, const unsigned int j) const override
+  virtual Number
+  el(unsigned int const i, const unsigned int j) const
   {
     return pde_operator->el(i, j);
   }
 
-  void
-  initialize_dof_vector(VectorType & vector) const override
+  virtual void
+  initialize_dof_vector(VectorType & vector) const
   {
     pde_operator->initialize_dof_vector(vector);
   }
 
-  void
-  vmult(VectorType & dst, VectorType const & src) const override
+  virtual void
+  vmult(VectorType & dst, VectorType const & src) const
   {
     pde_operator->vmult(dst, src);
   }
 
-  void
-  vmult_add(VectorType & dst, VectorType const & src) const override
+  virtual void
+  vmult_add(VectorType & dst, VectorType const & src) const
   {
     pde_operator->vmult_add(dst, src);
   }
 
-  void
-  vmult_interface_down(VectorType & dst, VectorType const & src) const override
+  virtual void
+  vmult_interface_down(VectorType & dst, VectorType const & src) const
   {
     pde_operator->vmult_interface_down(dst, src);
   }
 
-  void
-  vmult_add_interface_up(VectorType & dst, VectorType const & src) const override
+  virtual void
+  vmult_add_interface_up(VectorType & dst, VectorType const & src) const
   {
     pde_operator->vmult_add_interface_up(dst, src);
   }
 
-  void
-  calculate_inverse_diagonal(VectorType & inverse_diagonal_entries) const override
+  virtual void
+  calculate_inverse_diagonal(VectorType & inverse_diagonal_entries) const
   {
     pde_operator->calculate_inverse_diagonal(inverse_diagonal_entries);
   }
 
-  void
-  update_block_diagonal_preconditioner() const override
+  virtual void
+  update_block_diagonal_preconditioner() const
   {
     pde_operator->update_block_diagonal_preconditioner();
   }
 
-  void
-  apply_inverse_block_diagonal(VectorType & dst, VectorType const & src) const override
+  virtual void
+  apply_inverse_block_diagonal(VectorType & dst, VectorType const & src) const
   {
     pde_operator->apply_inverse_block_diagonal(dst, src);
   }
 
 #ifdef DEAL_II_WITH_TRILINOS
-  void
-  init_system_matrix(TrilinosWrappers::SparseMatrix & system_matrix) const override
+  virtual void
+  init_system_matrix(TrilinosWrappers::SparseMatrix & system_matrix) const
   {
     pde_operator->init_system_matrix(system_matrix);
   }
 
-  void
-  calculate_system_matrix(TrilinosWrappers::SparseMatrix & system_matrix) const override
+  virtual void
+  calculate_system_matrix(TrilinosWrappers::SparseMatrix & system_matrix) const
   {
     pde_operator->calculate_system_matrix(system_matrix);
   }
 #endif
 
 #ifdef DEAL_II_WITH_PETSC
-  void
-  init_system_matrix(PETScWrappers::MPI::SparseMatrix & system_matrix) const override
+  virtual void
+  init_system_matrix(PETScWrappers::MPI::SparseMatrix & system_matrix) const
   {
     pde_operator->init_system_matrix(system_matrix);
   }
 
-  void
-  calculate_system_matrix(PETScWrappers::MPI::SparseMatrix & system_matrix) const override
+  virtual void
+  calculate_system_matrix(PETScWrappers::MPI::SparseMatrix & system_matrix) const
   {
     pde_operator->calculate_system_matrix(system_matrix);
   }
--- a/include/exadg/operators/multigrid_operator.h
+++ b/include/exadg/operators/multigrid_operator.h
@@ -150,6 +150,20 @@ public:
   }
 #endif
 
+#ifdef DEAL_II_WITH_PETSC
+  virtual void
+  init_system_matrix(PETScWrappers::MPI::SparseMatrix & system_matrix) const
+  {
+    pde_operator->init_system_matrix(system_matrix);
+  }
+
+  virtual void
+  calculate_system_matrix(PETScWrappers::MPI::SparseMatrix & system_matrix) const
+  {
+    pde_operator->calculate_system_matrix(system_matrix);
+  }
+#endif
+
 private:
   std::shared_ptr<Operator> pde_operator;
 };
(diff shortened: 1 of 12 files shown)


include/exadg/operators/multigrid_operator_base.h
```diff
@@ -22,6 +22,8 @@
 #ifndef OPERATOR_PRECONDITIONABLE_H
 #define OPERATOR_PRECONDITIONABLE_H
 
+#include <deal.II/lac/petsc_parallel_sparse_matrix.h>
```
nfehn, line 25: Change to 
```suggestion
#include <deal.II/lac/petsc_sparse_matrix.h>
```

include/exadg/operators/multigrid_operator.h
```diff
@@ -150,6 +150,20 @@ class MultigridOperator : public MultigridOperatorBase<dim, Number>
   }
 #endif
 
+#ifdef DEAL_II_WITH_PETSC
+  virtual void
+  init_system_matrix(PETScWrappers::MPI::SparseMatrix & system_matrix) const
```
bergbauer, line 155:
```suggestion
  void
  init_system_matrix(PETScWrappers::MPI::SparseMatrix & system_matrix) const override
```
or depending on whatever #45 will settle on

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
```diff
@@ -23,6 +23,8 @@
 #define PRECONDITIONER_AMG
 
 #include <deal.II/lac/la_parallel_vector.h>
+#include <deal.II/lac/petsc_sparse_matrix.h>
+#include <deal.II/lac/petsc_precondition.h>
```
kronbichler, line 27:
```suggestion
#include <deal.II/lac/petsc_precondition.h>
#include <deal.II/lac/petsc_sparse_matrix.h>
```

include/exadg/operators/multigrid_operator.h
nfehn: In my opinion, we should better separate the PRs in the future. Now we have different kinds of changes not only in the previous PR #45 , but also in the present one?

Wouldn't it make more sense to focus here on the AMG aspects, and consider the `virtual` things in the other PR #45, once we have decided there what to do?

kronbichler: We can do that. Should we undo the commit?

nfehn: yes, I think it would be clearer to address these aspects in PR #45

include/exadg/operators/operator_base.cpp
```diff
@@ -742,7 +789,11 @@ OperatorBase<dim, Number, n_components>::calculate_system_matrix(SparseMatrix &
   // communicate overlapping matrix parts
   system_matrix.compress(VectorOperation::add);
 
-  if(!is_dg)
+  if(!is_dg
+#ifdef DEAL_II_WITH_TRILINOS
```
nfehn, line 793: I wonder if this would be easier to read with a second if-statement guarded by `#ifdef`?

include/exadg/solvers_and_preconditioners/multigrid/coarse_grid_solvers.h
```diff
@@ -113,13 +113,19 @@ class MGCoarseKrylov
       preconditioner_trilinos.reset(
         new PreconditionerAMG<Operator, TrilinosNumber>(matrix, additional_data.amg_data));
     }
+    else if(additional_data.preconditioner == MultigridCoarseGridPreconditioner::BoomerAMG)
+    {
+      preconditioner_trilinos.reset(
```
nfehn, line 118: could you please rename into `amg_preconditioner`?

include/exadg/solvers_and_preconditioners/multigrid/coarse_grid_solvers.h
```diff
@@ -318,9 +324,12 @@ class MGCoarseAMG
   typedef LinearAlgebra::distributed::Vector<typename Operator::value_type> VectorTypeMultigrid;
 
 public:
-  MGCoarseAMG(Operator const & op, AMGData data = AMGData())
+  MGCoarseAMG(Operator const & op, bool const use_boomer_amg, AMGData data = AMGData())
   {
-    amg_preconditioner.reset(new PreconditionerAMG<Operator, TrilinosNumber>(op, data));
+    if(use_boomer_amg)
+      amg_preconditioner.reset(new PreconditionerBoomerAMG<Operator, TrilinosNumber>(op, data));
```
nfehn, line 330: please rename into `NumberAMG` and `VectorTypeAMG` (should only affect this file).

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
```diff
@@ -110,6 +112,148 @@ class PreconditionerAMG : public PreconditionerBase<TrilinosNumber>
 
   AMGData amg_data;
 };
+
+
+/*
+ * Wrapper class for BoomerAMG from Hypre
+ */
+template<typename Operator, typename PreconditionerNumber>
```
nfehn, line 120: Can we simplify to just `Number`?

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
```diff
@@ -110,6 +112,148 @@ class PreconditionerAMG : public PreconditionerBase<TrilinosNumber>
...
+    // settings; right now we must skip most parameters because there does not
+    // appear to be a setting available in deal.II
+    PETScWrappers::PreconditionBoomerAMG::AdditionalData boomer_data;
+    boomer_data.symmetric_operator               = amg_data.data.elliptic;
+    boomer_data.strong_threshold                 = 0.25;
+    boomer_data.aggressive_coarsening_num_levels = 2;
+    amg.initialize(system_matrix, boomer_data);
+#else
+    AssertThrow(false, ExcMessage("deal.II is not compiled with PETSc!"));
+#endif
```
nfehn, line 156: Could you shift this part into a private member function `calculate_system_matrix()` to avoid code duplication?

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
```diff
@@ -110,6 +112,148 @@ class PreconditionerAMG : public PreconditionerBase<TrilinosNumber>
...
+      {
+        dst.local_element(i) = ptr[i];
+      }
+
+      ierr = VecRestoreArray(vector_dst, &ptr);
+    }
+#else
+    (void)dst;
+    (void)src;
+    AssertThrow(false, ExcMessage("deal.II is not compiled with PETSC!"));
```
nfehn, line 247: "PETSc" is used in the other Asserts above.

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
```diff
@@ -110,6 +112,148 @@ class PreconditionerAMG : public PreconditionerBase<TrilinosNumber>
...
+      for(PetscInt i = 0; i < local_size; ++i)
+      {
+        ptr[i] = src.local_element(i);
+      }
+
+      ierr = VecRestoreArray(vector_src, &ptr);
+    }
+
+    PETScWrappers::VectorBase petsc_dst(vector_dst);
+    amg.vmult(petsc_dst, PETScWrappers::VectorBase(vector_src));
```
nfehn, line 224: Why does one needs `petsc_dst`? Below, `vector_dst` is again used?

include/exadg/poisson/spatial_discretization/operator.cpp
```diff
@@ -453,6 +456,86 @@ Operator<dim, Number, n_components>::vmult_matrix_based(
...
+    const PetscInt local_size = dst.get_partitioner()->local_size();
+    AssertDimension(local_size, static_cast<unsigned int>(end - begin));
+    for(PetscInt i = 0; i < local_size; ++i)
+    {
+      dst.local_element(i) = ptr[i];
+    }
+
+    ierr = VecRestoreArray(vector_dst, &ptr);
+  }
+}
```
nfehn, line 536: Would it be possible to put this into a free function, say 

```suggestion
template<typename VectorTypeDealii>
void
apply_petsc_operation(VectorTypeDealii &dst, VectorTypeDealii const &src, lambda petsc_operation)
{
// pre copy

 petsc_operation(petsc_dst, petsc_src);

// post copy
}
```
It seems as if the same code is implemented in the AMG preconditioner.

include/exadg/poisson/spatial_discretization/operator.cpp
```diff
@@ -23,6 +23,9 @@
 #include <deal.II/fe/fe_dgq.h>
 #include <deal.II/fe/fe_q.h>
 #include <deal.II/fe/fe_system.h>
+#ifdef DEAL_II_WITH_PETSC
+#  include <deal.II/lac/petsc_parallel_vector.h>
```
bergbauer, line 27:
```suggestion
#  include <deal.II/lac/petsc_vector.h>
```

include/exadg/operators/operator_base.cpp
kronbichler: In my opinion it is best to completely remove this block by letting the `constraint_double` object use the function in deal.II that can put an appropriate diagonal entry into the constrained rows of the matrix. Done in commit 51a8b207

include/exadg/poisson/spatial_discretization/operator.cpp
kronbichler: Yes, that is a good idea, see e2a3cd2e

include/exadg/operators/multigrid_operator_base.h
nfehn: is this already done?

kronbichler:
I now addressed the final comments in the last commit. I had to shift around some methods to make sure also the combination `MultigridCoarseGridSolver::CG` works with boomer amg, but I think things are now valid.

include/exadg/solvers_and_preconditioners/multigrid/coarse_grid_solvers.h
```diff
@@ -159,45 +168,75 @@ class MGCoarseKrylov
...
+                                          additional_data.solver_data.rel_tol);
+          if(additional_data.solver_type == KrylovSolverType::CG)
+          {
+            PETScWrappers::SolverCG solver(solver_control);
+            solver.solve(coarse_operator->system_matrix,
+                         petsc_dst,
+                         petsc_src,
+                         coarse_operator->amg);
+          }
+          else if(additional_data.solver_type == KrylovSolverType::CG)
```
nfehn, line 227: `CG` is written twice!

kronbichler: Fixed in 3f7ce831

nfehn:
Now, I think I am really happy with the code. Thanks!

PR #51, opened 2021-03-25
fix compilation error

nfehn:


bergbauer (approved):

kronbichler:
Try to compile without PETSc wass exactly what was needed after #47, because I did not check that. Thanks!

PR #52, opened 2021-03-25
some more bug fixes and clean-up related to PR #47

nfehn:
I am not completely happy with this solution (better would be an additional `enum AlgebraicMultigridType { ML, BoomerAMG }`), but for now the priority should be to fix the bugs.


kronbichler (approved):
Very good, thanks for fixing up these things.

PR #56, opened 2021-03-26
Add config files for supermuc-ng

bergbauer:


scripts/supermuc-ng/bashrc_supermuc_ng
```diff
@@ -0,0 +1,13 @@
+module load slurm_setup
```
nfehn, line 1: One minor change here: I would name the file just bashrc because Supermuc-NG is already in the name of the directory

bergbauer: `bashrc`or `.bashrc`?

nfehn: `bashrc` is better in my opinion, otherwise the user might not see/find the file. One could also call it `bashrc_add_ons` to highlight that this should be added to the `~/.bashrc`

bergbauer: I like this idea `bashrc_add_ons`!

nfehn (approved):

scripts/supermuc-ng/bashrc_supermuc_ng
bergbauer: Done in commit [a857741](https://github.com/exadg/exadg/pull/56/commits/a857741a23912395cd56a0c62d7659a2dffd3c91)

PR #57, opened 2021-03-30
MG constraints setup: call interpolate_boundary_values only once

kronbichler:
I noticed that the setup of the multigrid preconditioner took a long time for lung simulations with many generation (outlets). As a first improvement, I found out that the `VectorTools::interpolate_boundary_values` was called for every boundary id separately. As that is a rather expensive function because it iterates over the whole mesh in each call (including the coarse artificial cells), this pull request allows me to reduce the time in the setup for 11 generations and 1 global refinement on 16 nodes from 610 seconds to 594 seconds. (I'm still looking into the other big consumers.)


nfehn (approved):
You mentioned other bottlenecks as well, but I think it is best to finalize this PR since it is open for quite a while now.

nfehn:
@kronbichler Is there something else you plan for this PR?

kronbichler:
No, this one is fine now.

PR #61, opened 2021-05-04
Avoid some redundant operations in MultigridAlgorithm

kronbichler:
Since we do not really use the `defect2` variable (and we should nowadays be able to use multiple iterations via `Smoother::step`), let us simply remove it. Also, I found some assignment to zero which is immediately followed by `copy_locally_owned_data_from`, which helps to slightly improve performance.


nfehn (approved):
Great that you identified these issues! Thanks.

PR #62, opened 2021-05-05
create output directory if not already existing

### LINKED ISSUE #55: Create output directory during simulation if not already existing
A common source of error is that the simulation crashes since the output directory specified in the input file does not exist. We should make the software more user friendly and create the directories automatically in this case.

nfehn:
closes #55

provides solution for a common problem as reported in #60

### DIFF
--- a/applications/compressible_navier_stokes/taylor_green/application.h
+++ b/applications/compressible_navier_stokes/taylor_green/application.h
@@ -296,15 +296,16 @@ public:
     pp_data.kinetic_energy_data.calculate                  = true;
     pp_data.kinetic_energy_data.calculate_every_time_steps = 1;
     pp_data.kinetic_energy_data.viscosity                  = DYN_VISCOSITY / RHO_0;
-    pp_data.kinetic_energy_data.filename = this->output_directory + this->output_name;
+    pp_data.kinetic_energy_data.directory                  = this->output_directory;
+    pp_data.kinetic_energy_data.filename                   = this->output_name;
 
     // kinetic energy spectrum
     pp_data.kinetic_energy_spectrum_data.calculate                     = true;
     pp_data.kinetic_energy_spectrum_data.calculate_every_time_steps    = -1;
     pp_data.kinetic_energy_spectrum_data.calculate_every_time_interval = 0.5;
-    pp_data.kinetic_energy_spectrum_data.filename =
-      this->output_directory + this->output_name + "_spectrum";
-    pp_data.kinetic_energy_spectrum_data.degree                     = degree;
+    pp_data.kinetic_energy_spectrum_data.directory                     = this->output_directory;
+    pp_data.kinetic_energy_spectrum_data.filename = this->output_name + "_spectrum";
+    pp_data.kinetic_energy_spectrum_data.degree   = degree;
     pp_data.kinetic_energy_spectrum_data.evaluation_points_per_cell = (degree + 1) * 1;
     pp_data.kinetic_energy_spectrum_data.exploit_symmetry           = false;
 
--- a/applications/incompressible_navier_stokes/orr_sommerfeld/include/perturbation_energy.h
+++ b/applications/incompressible_navier_stokes/orr_sommerfeld/include/perturbation_energy.h
@@ -23,11 +23,11 @@
 #define SOLVERS_INCOMPRESSIBLE_NAVIER_STOKES_APPLICATIONS_ORR_SOMMERFELD_PERTURBATION_ENERGY_H_
 
 // C/C++
-#include <filesystem>
 #include <fstream>
 
 // ExaDG
 #include <exadg/matrix_free/integrators.h>
+#include <exadg/utilities/create_directories.h>
 #include <exadg/utilities/print_functions.h>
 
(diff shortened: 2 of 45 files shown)


nfehn:
still work in progess

nfehn:
ready for review from my side

applications/incompressible_navier_stokes/orr_sommerfeld/include/perturbation_energy.h
```diff
@@ -89,28 +93,35 @@ class PerturbationEnergyCalculator
...
+    matrix_free = &matrix_free_in;
+    dof_index   = dof_index_in;
+    quad_index  = quad_index_in;
+    data        = data_in;
+
+    if(data.calculate)
+    {
+      // create directory if not already existing
+      if(Utilities::MPI::this_mpi_process(mpi_comm) == 0)
+        std::filesystem::create_directories(data.directory);
```
kronbichler, line 117: Do you think we should check the return value of `create_directories` that tells us that the directory was created? We could otherwise throw an assert that we could not create (which might happen when the user gives a directory it does not have permission to write into)?

kronbichler:
I like this approach. It would have saved me some time at various points. :wink:

bergbauer (approved):
By introducing this feature we should finally make C++17 a requirement. See also the discussion in deal [https://github.com/dealii/dealii/pull/12080](https://github.com/dealii/dealii/pull/12080).

applications/compressible_navier_stokes/taylor_green/application.h
```diff
@@ -296,15 +296,16 @@ class Application : public ApplicationBase<dim, Number>
...
     // kinetic energy spectrum
     pp_data.kinetic_energy_spectrum_data.calculate                     = true;
     pp_data.kinetic_energy_spectrum_data.calculate_every_time_steps    = -1;
     pp_data.kinetic_energy_spectrum_data.calculate_every_time_interval = 0.5;
-    pp_data.kinetic_energy_spectrum_data.filename =
-      this->output_directory + this->output_name + "_spectrum";
-    pp_data.kinetic_energy_spectrum_data.degree                     = degree;
+    pp_data.kinetic_energy_spectrum_data.directory                     = this->output_directory;
+    pp_data.kinetic_energy_spectrum_data.filename = this->output_name + "_spectrum";
+    pp_data.kinetic_energy_spectrum_data.degree   = degree;
```
peterrum, line 308: This looks like a lot of work. You could have used `std::filesystem::path(file_name).parent_path().to_string();` or something similar to get to the same result.

nfehn: Good to know. On the other hand, I wanted to unify the code in ExaDG anyways and always use parameters for directory + filename (we currently have a mixture of different styles). I think this style is more intuitive for the user.

applications/incompressible_navier_stokes/orr_sommerfeld/include/perturbation_energy.h
nfehn: I did not think about this. Maybe I assumed that one uses only paths relative to the ExaDG directory (where permissions should not be an issue?), but this is not necessarily the case. In principle, we could of course do something like that. In this case I would introduce a small utility function in ExaDG with `directory` and `mpi_comm` as parameters. We should first agree on how this looks like exactly, so that I have to go through all the changes only once more.

kronbichler: I agree that a utility function that centralizes these two lines plus captures the result would be good. And as to the rights: You never know how a user sets up the things, so any guidance on possible errors might be helpful.

nfehn: I introduced a utility function here https://github.com/exadg/exadg/pull/62/commits/2e02391ebfc841e01c906be0610ab79fed19645e

nfehn: @kronbichler Now we could introduce the error handling also easily in a separate PR. Maybe this would be easier because I don't know what you think of exactly.

kronbichler (approved):
Yes, we can do the check in a follow-up PR.

applications/incompressible_navier_stokes/orr_sommerfeld/include/perturbation_energy.h
kronbichler: I checked the new code with an invalid directory and I think the messages are good enough, so no need for further action:
```
Setup incompressible Navier-Stokes operator ...

... done!
terminate called after throwing an instance of 'std::filesystem::__cxx11::filesystem_error'
  what():  filesystem error: cannot create directories: Permission denied [/output/flow_past_cylinder/]
[mklap5:04334] *** Process received signal ***
[mklap5:04334] Signal: Aborted (6)
[mklap5:04334] Signal code:  (-6)
```

kronbichler: This comes from the automatic exception that gets raised.

PR #67, opened 2021-05-07
Optimize matrix-vector product in OperatorBase

kronbichler:
This is another (modest) optimization I played with during the code analysis last month: We might want to run the vector-zero within `MatrixFree::loop` to save one access to the result vector. In the `apply` function, we can then only manipulate in one of the `dst/src` vectors. To make more clear where the constrained values that we put aside come from, I tried to give them more intuitive names.

### DIFF
--- a/include/exadg/operators/operator_base.cpp
+++ b/include/exadg/operators/operator_base.cpp
@@ -81,7 +81,8 @@ OperatorBase<dim, Number, n_components>::reinit(MatrixFree<dim, Number> const & 
     constrained_indices.clear();
     for(auto i : this->matrix_free->get_constrained_dofs(this->data.dof_index))
       constrained_indices.push_back(i);
-    constrained_values.resize(constrained_indices.size());
+    constrained_values_src.resize(constrained_indices.size());
+    constrained_values_dst.resize(constrained_indices.size());
   }
 
   // set multigrid level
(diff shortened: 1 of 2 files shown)


peterrum (approved):

kronbichler:
> You want to remove dst=0 and therefore need to call matrix_free->loop() with true in the last parameter, is that correct?

Yes.

And the code with constraints does two things: Before the mat-vec, you set the entries of constrained unknowns to zero as that is their correct value for the connections to the neighboring unknowns (but remember the content of the src vector at those places from before; it typically is zero but not always). Then we do the cell integrals. Afterwards, we restore whatever was in the `src` vector at the place of the constrained indices. We also need to take into account that the test functions associated to the Dirichlet-constrained unknowns should not carry a result of the integral that comes out of the cell loop (as those test functions do not technically exist), which is why we delete those results from the `dst` vector at the constrained indices. Instead of using zero, we set them to `1. * src(constrained_indices[i])` because you want to ensure that the matrix remains invertible, with `1.` at the place of the diagonal. The code should get clearer in the `apply` function, the `apply_add` function which I also touched due to different variable names contains one more corner as we need to save/restore the previous content in `dst` as well.

nfehn:
Thanks for the detailed explanations. This should be clear now for me.

There is, however, one thing that I would like to discuss before merging. It looks like we have the wrong design with `apply` and `apply_add` functions, each of them having two parameters `dst` and `src`. Maybe this comes from a time where `MatrixFree` did not provide the possibility to zero the `dst` vector within the `loop()` function? 

Since the changes of this PR come along with code duplication, it appears to be more natural to have one `apply()` function with a third boolean parameter that is passed to `matrix_free->loop()`? This is probably not the right PR to realize changes of design described above, because this should be done consistently within ExaDG and not just pointwise. On the other hand, I would like this aspect to not remain unmentioned/undiscussed, given that it was kind of a long way to remove such subtle code duplications from ExaDG in the past, and that we actually want to avoid them in the future.  Hope that this is not too negative because I really appreciate your optimizations. @kronbichler What do you think?

nfehn:
> What I would like to avoid is to have to implement vmult_add with the help of a vmult, where we first create a temporary vector to hold the mat-vec of vmult and then add that. Or, to be more precise, I would not mind such an implementation if it is almost never used (because then simple code is best), and the performance impact would only disturb me if it is a hot function. (Which I believe vmult_add or our backend apply_add is not, as I see it almost never used, at least at first sight.)

I thought that this kind of problems is resolved by `MatrixFree` (which already has the boolean parameter in its interface), or do you have concerns how to realize the thing with the constraints?

nfehn:
@kronbichler Would you like to open an issue that points to the design improvements of the `apply()` functions, with potentially two different concepts/realizations as you already outlined above?

kronbichler:
Done in #71.

PR #68, opened 2021-05-08
Optimizations for interface to PETSc/BoomerAMG preconditioner

kronbichler:
This is the next set of optimizations. This PR does two things:
- Enable the PETSc infrastructure to run on a communicator with fewer processes, i.e., the communicator for the PETSc matrices can be only a subset of all MPI processes; this forces us to pass in the relevant communicator into `apply_petsc_operation`
- Pre-allocate the vectors needed for the `apply_petsc_operation` for the BoomerAMG preconditioner, to avoid setting them up over and over again.

### DIFF
--- a/include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
+++ b/include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
@@ -235,8 +235,8 @@ private:
 
 #ifdef DEAL_II_WITH_PETSC
   // PETSc vector objects to avoid re-allocation in every vmult() operation
-  mutable Vec petsc_vector_src;
-  mutable Vec petsc_vector_dst;
+  mutable VectorTypePETSc petsc_vector_src;
+  mutable VectorTypePETSc petsc_vector_dst;
 #endif
 };
 } // namespace ExaDG
--- a/include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
+++ b/include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
@@ -146,6 +146,16 @@ public:
     calculate_preconditioner();
   }
 
+  ~PreconditionerBoomerAMG()
+  {
+#ifdef DEAL_II_WITH_PETSC
+    PetscErrorCode ierr = VecDestroy(&petsc_vector_dst);
+    AssertThrow(ierr == 0, ExcPETScError(ierr));
+    ierr = VecDestroy(&petsc_vector_src);
+    AssertThrow(ierr == 0, ExcPETScError(ierr));
+#endif
+  }
+
 #ifdef DEAL_II_WITH_PETSC
   PETScWrappers::MPI::SparseMatrix const &
   get_system_matrix()
@@ -159,7 +169,8 @@ public:
   {
 #ifdef DEAL_II_WITH_PETSC
     // clear content of matrix since the next calculate_system_matrix-commands add their result
-    system_matrix = 0.0;
+    if(system_matrix.m() > 0)
+      system_matrix = 0.0;
 #endif
 
     calculate_preconditioner();
@@ -169,12 +180,15 @@ public:
   vmult(VectorType & dst, VectorType const & src) const override
   {
 #ifdef DEAL_II_WITH_PETSC
-    apply_petsc_operation(dst,
-                          src,
-                          [&](PETScWrappers::VectorBase &       petsc_dst,
-                              PETScWrappers::VectorBase const & petsc_src) {
-                            amg.vmult(petsc_dst, petsc_src);
-                          });
+    if(system_matrix.m() > 0)
+      apply_petsc_operation(dst,
+                            src,
+                            [&](PETScWrappers::VectorBase &       petsc_dst,
+                                PETScWrappers::VectorBase const & petsc_src) {
+                              amg.vmult(petsc_dst, petsc_src);
+                            },
+                            petsc_vector_dst,
+                            petsc_vector_src);
 #else
     (void)dst;
     (void)src;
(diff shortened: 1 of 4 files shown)


include/exadg/solvers_and_preconditioners/utilities/petsc_operation.h
```diff
@@ -96,9 +89,35 @@ apply_petsc_operation(VectorType &                                           dst
...
+ *  preconditioner) with a dealii::LinearAlgebra::distributed::Vector,
+ *  allocating a PETSc vectors and then calling the other function
+ */
+template<typename VectorType>
+void
+apply_petsc_operation(VectorType &                                           dst,
+                      VectorType const &                                     src,
+                      std::function<void(PETScWrappers::VectorBase &,
+                                         PETScWrappers::VectorBase const &)> petsc_operation,
+                      MPI_Comm const &                                       petsc_mpi_communicator)
```
peterrum, line 108: Is this function still needed?

peterrum (approved):

include/exadg/solvers_and_preconditioners/utilities/petsc_operation.h
kronbichler: We use it for the `Operator::vmult_matrix_based` (used for benchmarking) where we do not want to hold a `Vec` object of PETSc, and also for the case we set up a CG solver with an AMG preconditioner (again because I did not want to pollute the interface with more objects than just PETSc things). Also, that part is probably less performance-critical than the `PreconditionerBoomerAMG` class because the run time will typically be higher, so I opted for the smaller code size there.

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
```diff
@@ -159,7 +169,8 @@ class PreconditionerBoomerAMG : public PreconditionerBase<Number>
   {
 #ifdef DEAL_II_WITH_PETSC
     // clear content of matrix since the next calculate_system_matrix-commands add their result
-    system_matrix = 0.0;
+    if(system_matrix.m() > 0)
```
nfehn, line 172: The meaning of this if is not a 100% clear to me. Did you introduce it because of the sub-communicator or is it a general optimization? Maybe a comment would help the non-expert here.

include/exadg/solvers_and_preconditioners/utilities/petsc_operation.h
```diff
@@ -30,36 +30,29 @@ using namespace dealii;
...
  */
 #ifdef DEAL_II_WITH_PETSC
 template<typename VectorType>
 void
 apply_petsc_operation(VectorType &                                           dst,
                       VectorType const &                                     src,
                       std::function<void(PETScWrappers::VectorBase &,
-                                         PETScWrappers::VectorBase const &)> petsc_operation)
+                                         PETScWrappers::VectorBase const &)> petsc_operation,
+                      Vec &                                                  petsc_vector_dst,
```
nfehn, line 43: I wonder if the wrapping idea of this function would become more clear if the PETSc vectors would come as third and fourth arguments, and finally the lambda?

As a general question: Where does the type `Vec` come from? Did we define this in ExaDG?

nfehn:
Overall, and apart from some minor comments given above, this looks like very clean code to me.

include/exadg/solvers_and_preconditioners/utilities/petsc_operation.h
kronbichler: I agree, I moved the lambda to the last argument. I also put the lambda last in the other `apply_petsc_operation` for consistency (even though it is only a communicator there), see fc0d323d.

The type `Vec` is the name of the PETSc struct, https://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/Vec/Vec.html

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
kronbichler: I added the comment in my fc0d323d commit (sorry I was most effective to work on all comments at the same time and could not easily split it up afterwards).

include/exadg/solvers_and_preconditioners/utilities/petsc_operation.h
nfehn: Ok, I just thought it looks strange that `Vec` is written without any namespace. We could think about `typedef Vec VectorTypePETSc;` with a short comment, and write explicitly `dealii::PETScWrappers::VectorBase` here in order to be more explicit about the types and where they come from (it might get quite complicated for users without dealii background)? It probably does not make sense to wrap the vector into `PETScWrappers::VectorBase` outside of this function (because this is the main content of `apply_petsc_operation`).

include/exadg/solvers_and_preconditioners/utilities/petsc_operation.h
```diff
@@ -96,9 +90,35 @@ apply_petsc_operation(VectorType &                                           dst
...
+ */
+template<typename VectorType>
+void
+apply_petsc_operation(VectorType &                                           dst,
+                      VectorType const &                                     src,
+                      MPI_Comm const &                                       petsc_mpi_communicator,
+                      std::function<void(PETScWrappers::VectorBase &,
+                                         PETScWrappers::VectorBase const &)> petsc_operation)
+{
+  Vec vector_dst, vector_src;
```
nfehn, line 111: here I would write `petsc_vector_dst` for consistency with the other `apply_petsc_operation()` function.

include/exadg/solvers_and_preconditioners/utilities/petsc_operation.h
kronbichler: Very good idea. Done in 0f0e2b69.

include/exadg/solvers_and_preconditioners/utilities/petsc_operation.h
kronbichler: Done in 0f0e2b69.

include/exadg/solvers_and_preconditioners/utilities/petsc_operation.h
```diff
@@ -28,38 +28,37 @@ namespace ExaDG
...
+/*
+ * A typedef to make use of PETSc data structure clearer
+ */
+typedef Vec VectorTypePETSc;
+
 /*
  *  This function wraps the copy of a PETSc object (sparse matrix,
- *  preconditioner) with a dealii::LinearAlgebra::distributed::Vector.
+ *  preconditioner) with a dealii::LinearAlgebra::distributed::Vector, taking
+ *  pre-allocated PETSc vector objects (with struct name `Vec) for the
```
nfehn, line 40: `Vec -> adjust comment and/or incorporate new typedef.

kronbichler: See 644e7104.

nfehn (approved):

PR #69, opened 2021-05-08
Drop MPI processes from coarser grid level in global coarsening

kronbichler:
This was done in close collaboration with @peterrum - the goal of this rather involved code is to drop MPI processes as we pass to coarser levels in multigrid. We should at some point integrate this feature into deal.II (combined with the ability to directly coarsen `parallel::fullydistributed::Triangulation` for the setup of multigrid), but until deal.II catches up on this topic we need to implement a brute-force strategy here in ExaDG to be able to run with 100k+ MPI processes and complicated coarse meshes.

### DIFF
--- a/include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
+++ b/include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
@@ -334,33 +334,35 @@ replicate_distributed_triangulation_on_node(
 
     unsigned int counter = 0;
 
-    for(const auto & cell : distributed_tria.cell_iterators_on_level(0))
+    for(auto const & cell : distributed_tria.cell_iterators_on_level(0))
       cell_data_sorted.emplace_back(cell->id().get_coarse_cell_id(), cell_data[counter++]);
 
-    std::sort(cell_data_sorted.begin(), cell_data_sorted.end(), [](const auto & a, const auto & b) {
+    std::sort(cell_data_sorted.begin(), cell_data_sorted.end(), [](auto const & a, auto const & b) {
       return a.first < b.first;
     });
 
     cell_data.clear();
 
-    for(const auto & i : cell_data_sorted)
+    for(auto const & i : cell_data_sorted)
       cell_data.emplace_back(i.second);
 
     serial_tria.create_triangulation(points, cell_data, sub_cell_data);
   }
 
   // execute refinement on first process of node if there is refinement left
-  if(distributed_tria.n_global_levels() > 1)
+  unsigned int const n_levels = distributed_tria.n_global_levels();
+  if(n_levels > 1)
   {
-    // collect refinement flags from distributed triangulation first on rank 0
+    // collect refinement flags from the complete distributed triangulation on
+    // global rank 0 by an MPI_Gather step
     std::vector<std::vector<std::vector<CellId>>> refinement_flags(
-      distributed_tria.n_global_levels());
+      n_levels - 1);
     {
-      for(unsigned int l = 0; l < distributed_tria.n_global_levels(); ++l)
+      for(unsigned int l = 0; l < n_levels - 1; ++l)
       {
         std::vector<CellId> local_refinement_flags;
 
-        for(const auto & cell : distributed_tria.cell_iterators_on_level(l))
+        for(auto const & cell : distributed_tria.cell_iterators_on_level(l))
           if(cell->has_children())
             local_refinement_flags.push_back(cell->id());
 
@@ -368,8 +370,8 @@ replicate_distributed_triangulation_on_node(
       }
 
       // create new communicator that only involves the first MPI process of
-      // each node, to be able to broadcast the refinement flags of rank 0
-      // to each compute node
+      // each node, to be able to broadcast the refinement flags of global
+      // rank 0 to the first MPI rank on each compute node
       MPI_Comm comm_node;
       MPI_Comm_split(mpi_comm, is_first_process_on_node, my_rank, &comm_node);
 
@@ -382,12 +384,12 @@ replicate_distributed_triangulation_on_node(
     // perform refinement from refine flags
     if(is_first_process_on_node)
     {
-      for(unsigned int l = 0; l < refinement_flags.size(); ++l)
+      for(unsigned int l = 0; l < n_levels - 1; ++l)
       {
         unsigned int counter = 0;
-        for(const auto & flags : refinement_flags[l])
+        for(auto const & refinement_per_process : refinement_flags[l])
         {
-          for(const auto & cell_id : flags)
+          for(auto const & cell_id : refinement_per_process)
           {
             serial_tria.create_cell_iterator(cell_id)->set_refine_flag();
             counter++;
@@ -507,7 +509,7 @@ create_geometric_coarsening_sequence(Triangulation<dim, spacedim> const & fine_t
 
     // Step 3: Continue as above but with the serial triangulation that gets
     // distributed
-    unsigned int n_partitions = Utilities::MPI::n_mpi_processes(MPI_COMM_WORLD);
+    unsigned int n_partitions = Utilities::MPI::n_mpi_processes(mpi_comm);
     for(unsigned int level = tria_copy.n_global_levels(); level > 0; --level)
     {
       // reduce the number of MPI ranks per coarsening step by at most a
@@ -523,7 +525,7 @@ create_geometric_coarsening_sequence(Triangulation<dim, spacedim> const & fine_t
         TriangulationDescription::Utilities::create_description_from_triangulation_in_groups<dim,
                                                                                              dim>(
           [&](auto & tria) { tria.copy_triangulation(serial_tria); },
-          [&](auto & tria, auto const &, const auto) {
+          [&](auto & tria, auto const &, auto const) {
 #ifdef DEAL_II_WITH_METIS
             GridTools::partition_triangulation(n_partitions, tria);
 #else
@@ -553,7 +555,10 @@ create_geometric_coarsening_sequence(Triangulation<dim, spacedim> const & fine_t
 
   for(unsigned int i = 0; i < coarse_grid_triangulations.size(); ++i)
     AssertThrow(i + 1 == coarse_grid_triangulations[i]->n_global_levels(),
-                ExcDimensionMismatch(i + 1, coarse_grid_triangulations[i]->n_global_levels()));
+                ExcMessage("While creating coarser grids, expected a triangulation with " +
+                           std::to_string(i + 1) + " levels, but obtained " +
+                           std::to_string(coarse_grid_triangulations[i]->n_global_levels()) +
+                           " levels."));
 
   return coarse_grid_triangulations;
 }
--- a/include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
+++ b/include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
@@ -307,9 +307,104 @@ MultigridPreconditionerBase<dim, Number>::check_levels(std::vector<MGLevelInfo> 
 }
 
 /**
+ * Helper function for creating a geometric coarsening sequence: Replicate a
+ * parallel::distributed::Triangulation completely in a serial
+ * triangulation. This can potentially be very memory-consuming, so the
+ * triangulation is only filled on the first rank of a compute node, i.e.,
+ * those MPI ranks where `rank % n_processes_per_node` is zero.
+ */
+template<int dim, int spacedim>
+void
+replicate_distributed_triangulation_on_node(
+  parallel::distributed::Triangulation<dim, spacedim> const & distributed_tria,
+  MPI_Comm const &                                            mpi_comm,
+  unsigned int const                                          n_processes_per_node,
+  dealii::Triangulation<dim, spacedim> &                      serial_tria)
+{
+  unsigned int const my_rank                  = Utilities::MPI::this_mpi_process(mpi_comm);
+  bool const         is_first_process_on_node = (my_rank % n_processes_per_node) == 0;
+
+  // copy level 0 of distributed triangulation
+  if(is_first_process_on_node)
+  {
+    auto [points, cell_data, sub_cell_data] =
+      GridTools::get_coarse_mesh_description(distributed_tria);
+
+    std::vector<std::pair<unsigned int, CellData<dim>>> cell_data_sorted;
+
+    unsigned int counter = 0;
+
+    for(const auto & cell : distributed_tria.cell_iterators_on_level(0))
+      cell_data_sorted.emplace_back(cell->id().get_coarse_cell_id(), cell_data[counter++]);
+
+    std::sort(cell_data_sorted.begin(), cell_data_sorted.end(), [](const auto & a, const auto & b) {
+      return a.first < b.first;
+    });
+
+    cell_data.clear();
+
+    for(const auto & i : cell_data_sorted)
+      cell_data.emplace_back(i.second);
+
+    serial_tria.create_triangulation(points, cell_data, sub_cell_data);
+  }
+
+  // execute refinement on first process of node if there is refinement left
+  if(distributed_tria.n_global_levels() > 1)
+  {
+    // collect refinement flags from distributed triangulation first on rank 0
+    std::vector<std::vector<std::vector<CellId>>> refinement_flags(
+      distributed_tria.n_global_levels());
+    {
+      for(unsigned int l = 0; l < distributed_tria.n_global_levels(); ++l)
+      {
+        std::vector<CellId> local_refinement_flags;
+
+        for(const auto & cell : distributed_tria.cell_iterators_on_level(l))
+          if(cell->has_children())
+            local_refinement_flags.push_back(cell->id());
+
+        refinement_flags[l] = Utilities::MPI::gather(mpi_comm, local_refinement_flags, 0);
+      }
+
+      // create new communicator that only involves the first MPI process of
+      // each node, to be able to broadcast the refinement flags of rank 0
+      // to each compute node
+      MPI_Comm comm_node;
+      MPI_Comm_split(mpi_comm, is_first_process_on_node, my_rank, &comm_node);
+
+      if(is_first_process_on_node)
+        refinement_flags = Utilities::MPI::broadcast(comm_node, refinement_flags);
+
+      MPI_Comm_free(&comm_node);
+    }
+
+    // perform refinement from refine flags
+    if(is_first_process_on_node)
+    {
+      for(unsigned int l = 0; l < refinement_flags.size(); ++l)
+      {
+        unsigned int counter = 0;
+        for(const auto & flags : refinement_flags[l])
+        {
+          for(const auto & cell_id : flags)
+          {
+            serial_tria.create_cell_iterator(cell_id)->set_refine_flag();
+            counter++;
+          }
+        }
+
+        if(counter > 0)
+          serial_tria.execute_coarsening_and_refinement();
+      }
+    }
+  }
+}
+
+/**
  * Similar to dealii::MGTransferGlobalCoarseningTools::create_geometric_coarsening_sequence
  * with the difference that the (coarse-grid) p:d:T is converted to a p:f:T
- * right aways.
+ * right away.
  */
 template<int dim, int spacedim>
 std::vector<std::shared_ptr<Triangulation<dim, spacedim> const>>
@@ -332,47 +427,60 @@ create_geometric_coarsening_sequence(Triangulation<dim, spacedim> const & fine_t
 
     Assert(fine_triangulation, ExcNotImplemented());
 
-    // clone distributed triangulation for coarsening
-    parallel::distributed::Triangulation<dim, spacedim> temp_tria(
+    // clone distributed triangulation and start coarsening
+    parallel::distributed::Triangulation<dim, spacedim> tria_copy(
       fine_triangulation->get_communicator(), fine_triangulation->get_mesh_smoothing());
 
-    MPI_Comm mpi_comm = fine_triangulation_in.get_communicator();
-    temp_tria.copy_triangulation(*fine_triangulation);
-    temp_tria.coarsen_global();
+    tria_copy.copy_triangulation(*fine_triangulation);
+    tria_copy.coarsen_global();
 
-    unsigned int min_level = fine_triangulation->n_global_levels() - 1;
+    MPI_Comm mpi_comm = fine_triangulation->get_communicator();
 
     // as long as we have enough cells per process, we can perform regular
-    // coarsening with all MPI processes
-    for(; min_level > 0 &&
-          temp_tria.n_global_active_cells() / Utilities::MPI::n_mpi_processes(mpi_comm) > 100;
-        --min_level)
+    // coarsening with all MPI processes TODO: The number of 400 cells per MPI
+    // process (or 50 if the next refinement were done in 3D) was found to be a
+    // good tradeoff between communication cost and workload size of linear
+    // polynomials, resulting in small run times in preliminary studies. This
+    // could be generalized by a parameter to set in the application files.
+    unsigned int n_cells_per_process = 400;
+    for(unsigned int level = fine_triangulation->n_global_levels() - 1;
+        level > 0 && tria_copy.n_global_active_cells() / Utilities::MPI::n_mpi_processes(mpi_comm) >
+                       n_cells_per_process;
+        --level)
     {
-      // create empty (fully distributed) triangulation
-      auto new_tria = std::make_shared<parallel::fullydistributed::Triangulation<dim, spacedim>>(
-        fine_triangulation->get_communicator());
+      // extract relevant information from distributed triangulation
+      auto const construction_data =
+        TriangulationDescription::Utilities::create_description_from_triangulation(tria_copy,
+                                                                                   mpi_comm);
+
+      // create fully distributed triangulation
+      auto level_tria =
+        std::make_shared<parallel::fullydistributed::Triangulation<dim, spacedim>>(mpi_comm);
 
       for(auto const i : fine_triangulation->get_manifold_ids())
         if(i != numbers::flat_manifold_id)
-          new_tria->set_manifold(i, fine_triangulation->get_manifold(i));
+          level_tria->set_manifold(i, fine_triangulation->get_manifold(i));
 
-      // extract relevant information from distributed triangulation
-      auto const construction_data =
-        TriangulationDescription::Utilities::create_description_from_triangulation(
-          temp_tria, fine_triangulation->get_communicator());
+      level_tria->create_triangulation(construction_data);
 
-      // actually create triangulation
-      new_tria->create_triangulation(construction_data);
+      coarse_grid_triangulations[level - 1] = level_tria;
 
-      // save mesh
-      coarse_grid_triangulations[min_level - 1] = new_tria;
-      temp_tria.coarsen_global();
+      if(level > 1)
+        tria_copy.coarsen_global();
     }
 
-    // find out about the number of MPI processes in shared memory regions,
-    // which will later be used to construct a coarser triangulation only on a
-    // single process per MPI rank
-    unsigned int const group_size = [&]() {
+    // TODO: The following code is a brute-force attempt to create a new
+    // partitioning of the mesh to be fed to a
+    // parallel::fullydistributed::Triangulation with fewer MPI processes. The
+    // main idea is to replicate the complete triangulation obtained in the
+    // previous loop on specific MPI processes and partition it from there
+    // again.
+    //
+    // Step 1: Since the full triangulation with refinements can in general be
+    // many GB of size, we only want to fit a single copy on each node. We
+    // identify the nodes by letting MPI count the number of MPI processes in
+    // a shared memory region.
+    unsigned int const n_processes_per_node = [&]() {
       int rank;
       MPI_Comm_rank(mpi_comm, &rank);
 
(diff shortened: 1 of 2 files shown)


include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -336,14 +336,18 @@ create_geometric_coarsening_sequence(Triangulation<dim, spacedim> const & fine_t
     parallel::distributed::Triangulation<dim, spacedim> temp_tria(
       fine_triangulation->get_communicator(), fine_triangulation->get_mesh_smoothing());
 
+    MPI_Comm mpi_comm = fine_triangulation_in.get_communicator();
```
nfehn, line 339: why don't you work with the communicator of `fine_triangulation`? In my opinion, it would be more natural to work with the tria casted to `p::d::T` at this point of the code, e.g. to make the flow of information more clear or to allow to split this function into smaller sub-units.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -336,14 +336,18 @@ create_geometric_coarsening_sequence(Triangulation<dim, spacedim> const & fine_t
...
     temp_tria.copy_triangulation(*fine_triangulation);
+    temp_tria.coarsen_global();
 
-    // create coarse meshes
-    for(unsigned int l = (fine_triangulation->n_global_levels() - 1); l > 0; --l)
-    {
-      // coarsen mesh
-      temp_tria.coarsen_global();
+    unsigned int min_level = fine_triangulation->n_global_levels() - 1;
```
nfehn, line 343: why is this variable called `min_level` and not just `level`, i.e., it is of type `unsigned int` (no `const`)?

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -336,14 +336,18 @@ create_geometric_coarsening_sequence(Triangulation<dim, spacedim> const & fine_t
     parallel::distributed::Triangulation<dim, spacedim> temp_tria(
       fine_triangulation->get_communicator(), fine_triangulation->get_mesh_smoothing());
 
+    MPI_Comm mpi_comm = fine_triangulation_in.get_communicator();
     temp_tria.copy_triangulation(*fine_triangulation);
+    temp_tria.coarsen_global();
```
nfehn, line 341: Just for my understanding: Is this a necessary change?

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -360,11 +364,153 @@ create_geometric_coarsening_sequence(Triangulation<dim, spacedim> const & fine_t
       // actually create triangulation
       new_tria->create_triangulation(construction_data);
 
+      // save mesh
+      coarse_grid_triangulations[min_level - 1] = new_tria;
+      temp_tria.coarsen_global();
```
nfehn, line 369: Doesn't this mean you coarsen once more in the end where it isn't really needed, or do I overlook something?

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -360,11 +364,153 @@ create_geometric_coarsening_sequence(Triangulation<dim, spacedim> const & fine_t
...
+      int size_shared;
+      MPI_Comm_size(comm_shared, &size_shared);
+      MPI_Comm_free(&comm_shared);
+
+      // determine maximum, since some shared memory communicators
+      // might not be filed completely
+      return Utilities::MPI::max(size_shared, mpi_comm);
+    }();
+
+    const unsigned int my_rank    = Utilities::MPI::this_mpi_process(mpi_comm);
```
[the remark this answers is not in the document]
nfehn: you might want to use our `./contrib/utilities/search-and-replace-all` script for that

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -360,11 +364,153 @@ create_geometric_coarsening_sequence(Triangulation<dim, spacedim> const & fine_t
...
+
+      MPI_Comm comm_shared;
+      MPI_Comm_split_type(mpi_comm, MPI_COMM_TYPE_SHARED, rank, MPI_INFO_NULL, &comm_shared);
+
+      int size_shared;
+      MPI_Comm_size(comm_shared, &size_shared);
+      MPI_Comm_free(&comm_shared);
+
+      // determine maximum, since some shared memory communicators
+      // might not be filed completely
```
nfehn, line 387: filed -> filled. Meaning is unclear to me?

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -360,11 +364,153 @@ create_geometric_coarsening_sequence(Triangulation<dim, spacedim> const & fine_t
...
+      std::sort(cell_data_temp.begin(), cell_data_temp.end(), [](const auto & a, const auto & b) {
+        return a.first < b.first;
+      });
+
+      cell_data.clear();
+
+      for(const auto & i : cell_data_temp)
+        cell_data.emplace_back(i.second);
+
+      tria_serial.create_triangulation(points, cell_data, sub_cell_data);
```
nfehn, line 416: This code only depends on `temp_tria`? -> implement as a free function to render code more readable/understandable?

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -360,11 +364,153 @@ create_geometric_coarsening_sequence(Triangulation<dim, spacedim> const & fine_t
...
+
+          if(counter > 0)
+            tria_serial.execute_coarsening_and_refinement();
+        }
+      }
+    }
+
+    // continue as above but do not work on p:d:T but on the serial
+    // triangulation
+    unsigned int n_partitions_before = Utilities::MPI::n_mpi_processes(MPI_COMM_WORLD);
```
nfehn, line 468: I suggest to rename this variable, `before` is not clear. According to the for loop below, `n_partitions_fine` would be more appropriate. Technically, it seems as if we do not even need two variables `n_partitions_before`, `n_partitions`, but that a single variable would be enough?

Moreover, the naming should be in line with the name chosen for `new_tria`, see also my other comment above.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -360,11 +364,153 @@ create_geometric_coarsening_sequence(Triangulation<dim, spacedim> const & fine_t
...
+      MPI_Comm comm_shared;
+      MPI_Comm_split_type(mpi_comm, MPI_COMM_TYPE_SHARED, rank, MPI_INFO_NULL, &comm_shared);
+
+      int size_shared;
+      MPI_Comm_size(comm_shared, &size_shared);
+      MPI_Comm_free(&comm_shared);
+
+      // determine maximum, since some shared memory communicators
+      // might not be filed completely
+      return Utilities::MPI::max(size_shared, mpi_comm);
```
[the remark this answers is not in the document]
kronbichler: This is for correctness, but we never use this case of different MPI ranks on different compute nodes. I changed this into an assertion now.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
kronbichler: Rephrased.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -360,11 +364,153 @@ create_geometric_coarsening_sequence(Triangulation<dim, spacedim> const & fine_t
...
       new_tria->create_triangulation(construction_data);
 
+      // save mesh
+      coarse_grid_triangulations[min_level - 1] = new_tria;
+      temp_tria.coarsen_global();
+    }
+
+    // find out about the number of MPI processes in shared memory regions,
+    // which will later be used to construct a coarser triangulation only on a
+    // single process per MPI rank
```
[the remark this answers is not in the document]
kronbichler: The first statement was wrong. The second refers to a compute node. I put the code in a free function now and used the term "node" (even though "node" is node specific to MPI). I hope this is clearer now.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -336,14 +336,18 @@ create_geometric_coarsening_sequence(Triangulation<dim, spacedim> const & fine_t
...
+    unsigned int min_level = fine_triangulation->n_global_levels() - 1;
 
+    // as long as we have enough cells per process, we can perform regular
+    // coarsening with all MPI processes
+    for(; min_level > 0 &&
+          temp_tria.n_global_active_cells() / Utilities::MPI::n_mpi_processes(mpi_comm) > 100;
+        --min_level)
+    {
       // create empty (fully distributed) triangulation
       auto new_tria = std::make_shared<parallel::fullydistributed::Triangulation<dim, spacedim>>(
```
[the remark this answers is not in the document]
kronbichler: I agree, I now call it `level_tria` to indicate that it is the triangulation for the level. Regarding the duplication, I did try to streamline other parts of the code. Now we still repeat a few lines (four statements), but in my opinion trying to get that into a separate function would probably be less readable. We might always change it later if we feel for it.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
kronbichler: At this spot we actually still want to coarsen (unless the level is the last one, i.e., 1), because we want to further process the coarsened mesh somewhere else. I tried to rearrange the place where we coarsen but think that this place is the most intuitive for the rest of the function where we want to replicate some of the mesh.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
kronbichler: Good idea, I did this with the next block as well in a new function `replicate_distributed_triangulation_on_node`, to make that part clear from the name.

kronbichler:
I addressed the comments in the new commit d34ad0c - I had to rewrite most parts, so it is better to take a fresh look at the code.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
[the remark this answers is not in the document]
nfehn: replacing "const auto" by "auto const" should resolve the remaining const issues

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -332,39 +427,134 @@ create_geometric_coarsening_sequence(Triangulation<dim, spacedim> const & fine_t
...
+
+      // coarsen mesh
+      if(Utilities::MPI::this_mpi_process(mpi_comm) % n_processes_per_node == 0 && level > 1)
+        serial_tria.coarsen_global();
     }
   }
 
+  for(unsigned int i = 0; i < coarse_grid_triangulations.size(); ++i)
+    AssertThrow(i + 1 == coarse_grid_triangulations[i]->n_global_levels(),
+                ExcDimensionMismatch(i + 1, coarse_grid_triangulations[i]->n_global_levels()));
```
nfehn, line 556: I am not sure, but wouldn't a written message be more helpful?

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -306,10 +306,105 @@ MultigridPreconditionerBase<dim, Number>::check_levels(std::vector<MGLevelInfo>
...
+/**
+ * Helper function for creating a geometric coarsening sequence: Replicate a
+ * parallel::distributed::Triangulation completely in a serial
+ * triangulation. This can potentially be very memory-consuming, so the
+ * triangulation is only filled on the first rank of a compute node, i.e.,
+ * those MPI ranks where `rank % n_processes_per_node` is zero.
+ */
+template<int dim, int spacedim>
+void
+replicate_distributed_triangulation_on_node(
```
nfehn, line 318: I like this refactoring!

I wonder if ordering the parameters according to the `dst`, `src` logic (widely used in dealii or in ExaDG) would be better. Further, the word `serial` in the function name would also help. `serialize_distributed_triangulation_per_node()`? Not sure if this is better?

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -306,10 +306,105 @@ MultigridPreconditionerBase<dim, Number>::check_levels(std::vector<MGLevelInfo>
...
+  parallel::distributed::Triangulation<dim, spacedim> const & distributed_tria,
+  MPI_Comm const &                                            mpi_comm,
+  unsigned int const                                          n_processes_per_node,
+  dealii::Triangulation<dim, spacedim> &                      serial_tria)
+{
+  unsigned int const my_rank                  = Utilities::MPI::this_mpi_process(mpi_comm);
+  bool const         is_first_process_on_node = (my_rank % n_processes_per_node) == 0;
```
nfehn, line 325: The same code is used somewhere below to identify the first process on node. For me, this code is so general that it should become a utility function in ExaDG -> `is_first_process_on_node(mpi_comm, n_processes_per_node)`. Ideally, `n_processes_per_node` would not be part of the interface but happen inside that function, but this design decision probably depends on how expensive it is to call the lambda `n_processes_per_node`.

Regarding this function, I believe that passing `is_first_process_on_node` would be a better design. I don't see what `n_processes_per_node` is used for apart from determining `is_first_process_on_node`. So my preferred option in terms of design would be a call like `is_first_process_on_node(mpi_comm)` outside of this function, and passing the result to this function. Does this make sense?

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -306,10 +306,105 @@ MultigridPreconditionerBase<dim, Number>::check_levels(std::vector<MGLevelInfo>
...
+    for(const auto & i : cell_data_sorted)
+      cell_data.emplace_back(i.second);
+
+    serial_tria.create_triangulation(points, cell_data, sub_cell_data);
+  }
+
+  // execute refinement on first process of node if there is refinement left
+  if(distributed_tria.n_global_levels() > 1)
+  {
+    // collect refinement flags from distributed triangulation first on rank 0
```
nfehn, line 355: The following is not clear to me here: Are we talking about rank 0 (globally), or the first process of a node? If the former is the case, could you briefly explain the background?

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -306,10 +306,105 @@ MultigridPreconditionerBase<dim, Number>::check_levels(std::vector<MGLevelInfo>
...
+    std::vector<std::vector<std::vector<CellId>>> refinement_flags(
+      distributed_tria.n_global_levels());
+    {
+      for(unsigned int l = 0; l < distributed_tria.n_global_levels(); ++l)
```
nfehn, line 359: Introduce variable `n_levels` above instead of multiple calls to `distributed_tria.n_global_levels()`.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -306,10 +306,105 @@ MultigridPreconditionerBase<dim, Number>::check_levels(std::vector<MGLevelInfo>
...
+      if(is_first_process_on_node)
+        refinement_flags = Utilities::MPI::broadcast(comm_node, refinement_flags);
+
+      MPI_Comm_free(&comm_node);
+    }
+
+    // perform refinement from refine flags
+    if(is_first_process_on_node)
+    {
+      for(unsigned int l = 0; l < refinement_flags.size(); ++l)
```
nfehn, line 385: Use of variable `n_levels` would be particularly helpful HERE in my opinion.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -306,10 +306,105 @@ MultigridPreconditionerBase<dim, Number>::check_levels(std::vector<MGLevelInfo>
...
+      cell_data.emplace_back(i.second);
+
+    serial_tria.create_triangulation(points, cell_data, sub_cell_data);
+  }
+
+  // execute refinement on first process of node if there is refinement left
+  if(distributed_tria.n_global_levels() > 1)
+  {
+    // collect refinement flags from distributed triangulation first on rank 0
+    std::vector<std::vector<std::vector<CellId>>> refinement_flags(
```
nfehn, line 356: I wonder why we have three nested vectors, below I only see two???

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -306,10 +306,105 @@ MultigridPreconditionerBase<dim, Number>::check_levels(std::vector<MGLevelInfo>
...
+      MPI_Comm_free(&comm_node);
+    }
+
+    // perform refinement from refine flags
+    if(is_first_process_on_node)
+    {
+      for(unsigned int l = 0; l < refinement_flags.size(); ++l)
+      {
+        unsigned int counter = 0;
+        for(const auto & flags : refinement_flags[l])
```
nfehn, line 388: Maybe my question above (three vs. two vectors) refers to what this loop here is for, and whether a name other than `flags` would help me understanding the code.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -332,39 +427,134 @@ create_geometric_coarsening_sequence(Triangulation<dim, spacedim> const & fine_t
...
+    // Step 2: Create coarse grid on the first MPI process of each node
+    Triangulation<dim, spacedim> serial_tria;
+    replicate_distributed_triangulation_on_node(tria_copy,
+                                                mpi_comm,
+                                                n_processes_per_node,
+                                                serial_tria);
+
+    // Step 3: Continue as above but with the serial triangulation that gets
+    // distributed
+    unsigned int n_partitions = Utilities::MPI::n_mpi_processes(MPI_COMM_WORLD);
```
nfehn, line 510: why `MPI_COMM_WORLD` (which we generally do not want to use in such files in ExaDG as far as I know)?

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
kronbichler: It is the global rank 0 that collects information on all MPI ranks. This follows the `MPI_Gather` concept, but using the `Utilities::MPI::gather` function that can do this for pretty general types. A few lines below, we then broadcast the result again. So in some sense, this is an `MPI_Allgather` function (collect data from all ranks and send it back to all ranks), with the restriction that only the first rank on each node gets the information.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
kronbichler: The vector you probably do not see is what gets created within `Utilities::MPI::gather`, a vector for the vector collected on each MPI process. So the three vectors are level, the MPI process origin, and the cells that should be refined. There is certainly redundant information in the second and third vectors, as the ghosts and owned cells are seen by more than one rank, but that would only increase the code size.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
kronbichler: Let's go with `refinement_per_process` then.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
kronbichler: Silly mistake. Thanks.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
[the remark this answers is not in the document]
kronbichler: Done in e6b6fc41.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
kronbichler: I added a comment in e6b6fc41.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
kronbichler: Added in e6b6fc41.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
kronbichler: Yes. e6b6fc41 - note that I reduced to `n_levels - 1` where appropriate as we won't do refinement on the last level by construction.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
kronbichler: See e6b6fc41.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
kronbichler: Done in e6b6fc41.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
kronbichler: Good idea. I added the function in `exadg/utilities/mpi.h` as `identify_first_process_on_node` - I want to return both the question of whether it is the first process, and the number of processes per node that we also need in this function, as it would be expensive to recompute this. (If I measured correctly, we're talking about more than 10 seconds per call on 3k nodes.)

kronbichler: Have a look at the new commit 02b166b6 that addresses this issue.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
kronbichler: I decided against `serialize_...` because that has a different meaning in terms of storing a stream of bytes so that might potentially confuse the reader. I opted for name `gather_distributed_triangulation_by_node` as the name `gather` bears exactly the meaning of replicating all information in the MPI context.

kronbichler:
Thanks for spotting these, I forgot them. Fixed in fd52e34a.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -332,39 +428,115 @@ create_geometric_coarsening_sequence(Triangulation<dim, spacedim> const & fine_t
...
       for(auto const i : fine_triangulation->get_manifold_ids())
         if(i != numbers::flat_manifold_id)
-          new_tria->set_manifold(i, fine_triangulation->get_manifold(i));
+          level_tria->set_manifold(i, fine_triangulation->get_manifold(i));
+
+      level_tria->create_triangulation(construction_data);
+
+      coarse_grid_triangulations[level - 1] = level_tria;
+
+      if(level > 1)
```
nfehn, line 469: In my opinion, the state of `level` does not fit to the coarsening state of `tria_copy`, which makes the code difficult to understand.

Since `tria_copy` has already been coarsened in line 436, the initialization of `level` in the for-loop should be different. Here in this line, this change would imply `if(level > 0)`. I will add suggestions to the other places where `level` is used.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -332,39 +428,115 @@ create_geometric_coarsening_sequence(Triangulation<dim, spacedim> const & fine_t
...
 
       for(auto const i : fine_triangulation->get_manifold_ids())
         if(i != numbers::flat_manifold_id)
-          new_tria->set_manifold(i, fine_triangulation->get_manifold(i));
+          level_tria->set_manifold(i, fine_triangulation->get_manifold(i));
+
+      level_tria->create_triangulation(construction_data);
+
+      coarse_grid_triangulations[level - 1] = level_tria;
+
```
nfehn, line 468:
```suggestion
coarse_grid_triangulations[level] = level_tria;
```

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -332,39 +428,115 @@ create_geometric_coarsening_sequence(Triangulation<dim, spacedim> const & fine_t
...
+    // coarsening with all MPI processes TODO: The number of 400 cells per MPI
+    // process (or 50 if the next refinement were done in 3D) was found to be a
+    // good tradeoff between communication cost and workload size of linear
+    // polynomials, resulting in small run times in preliminary studies. This
+    // could be generalized by a parameter to set in the application files.
+    unsigned int n_cells_per_process = 400;
+    for(unsigned int level = fine_triangulation->n_global_levels() - 1;
+        level > 0 && tria_copy.n_global_active_cells() / Utilities::MPI::n_mpi_processes(mpi_comm) >
+                       n_cells_per_process;
+        --level)
```
nfehn, line 450:
```suggestion
for(int level = fine_triangulation->n_global_levels() - 2; level >= 0 && ...; --level)
```
Note that the type of the variable then has to be `int` instead of `unsigned int`.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
nfehn, line 469:
```suggestion
if(level > 0)
```

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -332,39 +428,115 @@ create_geometric_coarsening_sequence(Triangulation<dim, spacedim> const & fine_t
...
+    Triangulation<dim, spacedim> serial_tria;
+    gather_distributed_triangulation_by_node(serial_tria,
+                                             tria_copy,
+                                             mpi_comm,
+                                             is_first_process_on_node);
+
+    // Continue as above but with the serial triangulation that gets
+    // distributed
+    unsigned int n_partitions = Utilities::MPI::n_mpi_processes(mpi_comm);
+    for(unsigned int level = tria_copy.n_global_levels(); level > 0; --level)
```
nfehn, line 490:
```suggestion
    for(int level = tria_copy.n_global_levels()-1; level >= 0; --level)
```

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -332,39 +428,115 @@ create_geometric_coarsening_sequence(Triangulation<dim, spacedim> const & fine_t
...
-      // actually create triangulation
-      new_tria->create_triangulation(construction_data);
+      for(auto const i : fine_triangulation->get_manifold_ids())
+        if(i != numbers::flat_manifold_id)
+          level_tria->set_manifold(i, fine_triangulation->get_manifold(i));
+      level_tria->create_triangulation(construction_data);
 
       // save mesh
-      coarse_grid_triangulations[l - 1] = new_tria;
+      coarse_grid_triangulations[level - 1] = level_tria;
```
nfehn, line 525:
```suggestion
      coarse_grid_triangulations[level] = level_tria;
```

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -332,39 +428,115 @@ create_geometric_coarsening_sequence(Triangulation<dim, spacedim> const & fine_t
...
+
+      // coarsen mesh
+      if(is_first_process_on_node && level > 1)
```
nfehn, line 528:
```suggestion
      if(is_first_process_on_node && level > 0)
```

kronbichler:
Addressed in 490f751

nfehn (approved):

nfehn:
Did you run the tests? Then we could merge in my opinion.

kronbichler:
I did run both the lung test case and some of the local tests locally and they pass, so let's merge. I will start a bigger computation later today and see if everything is still sane, but that is easier from the merged master.

PR #70, opened 2021-05-10
Fix a few more occurrences of virtual/override

kronbichler:
With a fresh build, I found a few more of these occurrences that I overlooked in #45. (I see a bunch of additional warnings regarding float-integer conversions due to `std::pow` that I want to tackle in the future, which is why I did not detect them when rebasing #45 and only got now on a closer look).

### DIFF
--- a/applications/poisson/sine/application.h
+++ b/applications/poisson/sine/application.h
@@ -147,7 +147,7 @@ public:
   MeshType    mesh_type        = MeshType::Cartesian;
 
   void
-  set_input_parameters(InputParameters & param)
+  set_input_parameters(InputParameters & param) final
   {
     // MATHEMATICAL MODEL
     param.right_hand_side = true;
@@ -182,7 +182,7 @@ public:
               PeriodicFaces &                     periodic_faces,
               unsigned int const                  n_refine_space,
               std::shared_ptr<Mapping<dim>> &     mapping,
-              unsigned int const                  mapping_degree)
+              unsigned int const                  mapping_degree) final
   {
     (void)periodic_faces;
 
@@ -250,7 +250,8 @@ public:
   }
 
 
-  void set_boundary_conditions(std::shared_ptr<BoundaryDescriptor<0, dim>> boundary_descriptor)
+  void
+    set_boundary_conditions(std::shared_ptr<BoundaryDescriptor<0, dim>> boundary_descriptor) final
   {
     typedef typename std::pair<types::boundary_id, std::shared_ptr<Function<dim>>> pair;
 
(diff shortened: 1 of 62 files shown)


nfehn:
Looks reasonable. I just wonder why you did not get much more warnings, because there should be much more test cases? Related to #45, the `final` syntax would make sense for the test cases, because there will be no further inheritance?

So I think we should either keep the application test cases out of this PR, or roll it out consistently throughout ExaDG.

kronbichler:
I have no idea why the clang compiler only picks up some instances and not all. Maybe it's an `override/final` keyword somewhere that triggers it, I must admit I did not find the reason. I suggest we try to do it consistently also for applications, because I hope to get warning-free with that compiler soon (once I also fix the double-int conversions)

Now I have commit 7124e0f doing that. I'm not entirely sure that I found all instances by my scripts, but at least the ones I got warnings for and I spotted easily. We can always extend on top of them if we see new warnings or if anyone observes things that are inconsistent.

applications/poisson/sine/application.h
```diff
@@ -250,7 +250,8 @@ class Application : public ApplicationBase<dim, Number>
   }
 
 
-  void set_boundary_conditions(std::shared_ptr<BoundaryDescriptor<0, dim>> boundary_descriptor)
+  void
+    set_boundary_conditions(std::shared_ptr<BoundaryDescriptor<0, dim>> boundary_descriptor) final
```
nfehn, line 254: Is the alignment correct here according to clang-format?

nfehn:
For the `incompressible_flow_with_transport` module, it seems as if the functions with ending `_scalar()` have not been detected. 

Also, it seems as if the function `set_input_parameters()` has not been detected in many cases (not just for the `_scalar()` case mentioned above, but also for FSI problems, `set_input_parameters_precursor()` etc., so I think scrolling through all the test cases might be unavoidable (it is an effort of approximately half an hour or an hour in my experience, I did that quite often in the past ;)).

kronbichler:
I now fixed additional appearances I could spot from the base classes and the `set_input_parameters` functions that had different argument names I did not catch with my file. While I did not got through all files individually and compared to all base class members, this is as good as it can get with reasonable effort (and no additional help from the compiler).

nfehn:
There are two more functions for FSI test cases: `set_material_ale()`, `set_material_structure()`

nfehn:
I think the test case  exadg/applications/incompressible_flow_with_transport/template/application.h needs to be adjusted, which is probably best to do manually

nfehn:
There is also a function called `set_mesh_movement_function()`.

With these changes, I think we are done.

kronbichler:
I applied these, and also made sure to check `set-mesh_movement_function`.

nfehn:
I did not find `set_mesh_movement_function()` in the commit? Maybe you had a typo?

nfehn (approved):

kronbichler:
Thank you, I had indeed missed one instance that from visual inspection after running `grep`, see a9fc5af. Now all occurrences in `applications` are `final` (and the rest in `include` is declaration only).

PR #72, opened 2021-05-19
Introduce utility functions create_operator() and create_time_integrator()

nfehn:
The setup routines for incompressible Navier-Stokes applications distinguish between Coupled/DualSplitting/PressureCorrection schemes, which is quite lengthy for all the different drivers and can easily be simplified by introducing utility functions `create_operator()` and `create_time_integrator()`.

This is still work-in-progress and needs to be rolled out to all drivers in ExaDG that involve the IncNS module.


kronbichler (approved):

kronbichler:
(I am fine with the current state, but of course we need to wait for the other modules.)

PR #74, opened 2021-05-19
Avoid implicit float-to-int conversions

kronbichler:
This pull request allows me to finally compile ExaDG without warnings. The compiler complained about the following kind of double-to-float conversions (to give one representative example):
```
In file included from /home/kronbichler/Work/dg/exadg/applications/incompressible_navier_stokes/fda_benchmark/solver.cpp:26:
/home/kronbichler/Work/dg/exadg/applications/incompressible_navier_stokes/fda_benchmark/application.h:196:36: warning: implicit conversion turns floating-point number into integer: 'double' to 'unsigned int' [-Wfloat-conversion]
      n_points = 20 * (degree + 1) * std::pow(2.0, refine_space);
               ~ ~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
What happens is that (at least for clang) `std::pow` takes a floating point argument and produces a floating point result, which is propagated through the multiplication and finally assigned to the integer `n_points`. That implicit conversion is what the compiler complains about. This PR addresses conversions by two approaches:
- Use deal.II's dedicated integer power mechanism `Utilities::pow` for integer powers, e.g. `(degree+1)^dim`, to avoid creating floating point numbers in the first place
- Where floating point numbers are necessary, e.g. for `round(log(n) / log(2))`, I added explicit casts.

### DIFF
--- a/include/exadg/postprocessor/kinetic_energy_spectrum.cpp
+++ b/include/exadg/postprocessor/kinetic_energy_spectrum.cpp
@@ -138,17 +138,17 @@ public:
     std::vector<types::global_dof_index> indices_has, indices_want;
 
     for(auto const & I : local_cells)
-      for(types::global_dof_index i = 0; i < pow_(points_dst, dim); i++)
+      for(types::global_dof_index i = 0; i < dealii::Utilities::pow(points_dst, dim); i++)
         for(types::global_dof_index d = 0; d < dim; d++)
         {
           types::global_dof_index index =
             (I / (n_cells_1D * n_cells_1D) * points_dst + i / (points_dst * points_dst)) *
-              pow_(n_cells_1D * points_dst, 2) +
+              dealii::Utilities::pow(n_cells_1D * points_dst, 2) +
             (((I / n_cells_1D) % n_cells_1D) * points_dst + ((i / points_dst) % points_dst)) *
-              pow_(n_cells_1D * points_dst, 1) +
+              dealii::Utilities::pow(n_cells_1D * points_dst, 1) +
             (I % (n_cells_1D)*points_dst + i % (points_dst));
 
-          indices_has.push_back(d * pow_(points_dst * n_cells_1D, dim) + index);
+          indices_has.push_back(d * dealii::Utilities::pow(points_dst * n_cells_1D, dim) + index);
         }
 
     types::global_dof_index N  = s.cells * s.points_dst;
@@ -161,9 +161,10 @@ public:
         for(types::global_dof_index j = 0; j < N; j++)
           for(types::global_dof_index i = 0; i < Nx; i++, c++)
             if(i < N)
-              indices_want.push_back(d * pow_(points_dst * n_cells_1D, dim) +
-                                     (k + start) * pow_(points_dst * n_cells_1D, 2) +
-                                     j * pow_(points_dst * n_cells_1D, 1) + i);
+              indices_want.push_back(d * dealii::Utilities::pow(points_dst * n_cells_1D, dim) +
+                                     (k + start) *
+                                       dealii::Utilities::pow(points_dst * n_cells_1D, 2) +
+                                     j * dealii::Utilities::pow(points_dst * n_cells_1D, 1) + i);
             else
               indices_want.push_back(numbers::invalid_dof_index); // x-padding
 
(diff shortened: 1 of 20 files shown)


include/exadg/postprocessor/kinetic_energy_spectrum.cpp
```diff
@@ -204,12 +205,16 @@ class DealSpectrumWrapper
...
       timer.start("Permutation");
-      ArrayView<double> dst(
-        fftw.u_real,
-        s.dim * pow_(static_cast<types::global_dof_index>(s.cells * s.points_dst), s.dim) * 2);
+      ArrayView<double>       dst(fftw.u_real,
+                            s.dim *
+                              dealii::Utilities::pow(static_cast<types::global_dof_index>(
+                                                       s.cells * s.points_dst),
+                                                     s.dim) *
+                              2);
```
nfehn, line 213: Is there a way to refactor this code so that the alignment is not that ugly? A first try would be to introduce an empty line between the variables `dst` and `src_` so that there is more space left for the `dst` variable. Or, why not remove the namespace dealii from this file to make the expressions shorter?

include/exadg/postprocessor/kinetic_energy_spectrum.cpp
```diff
@@ -250,31 +255,20 @@ class DealSpectrumWrapper
...
-  pow_(const types::global_dof_index base, const types::global_dof_index exp)
-  {
-    types::global_dof_index result = 1.0;
-
-    for(types::global_dof_index i = 0; i < exp; i++)
-      result *= base;
-
-    return result;
-  };
-
```
nfehn, line 263: This is no longer used?

include/exadg/postprocessor/spectral_analysis/bijection.h
```diff
@@ -206,9 +207,8 @@ class Bijection
         int temp_counter = counter;
         for(int d = 0; d < s.dim; d++)
         {
-          int r = temp_counter % n;
+          Y[d] = temp_counter % n;
           temp_counter /= n;
-          Y[d] = r + 0.5;
```
nfehn, line 211: Can you briefly explain why this change is legal?

include/exadg/utilities/hypercube_resolution_parameters.h
```diff
@@ -77,7 +77,7 @@ fill_resolutions_vector(
   int    refine_level = 0;
   double n_cells      = 1.0;
 
-  while(n_cells <= std::pow(2, dim) * n_cells_max)
+  while(n_cells <= std::pow(2.0, dim) * n_cells_max)
```
nfehn, line 80: why didn't you switch to `int` for this file?

include/exadg/postprocessor/spectral_analysis/bijection.h
kronbichler: I was wondering about this one, too. My reasoning is that `r` is a non-negative integer, and `Y` is an array of integers. So I cannot see that `(int)(r + 0.5)` gives anything else than `r`, so one can simply cast that back. I did not try to really understand the algorithm here. Do we have a good test for this?

nfehn: Let's ask @peterrum ?

include/exadg/postprocessor/kinetic_energy_spectrum.cpp
kronbichler: Indeed, this is the same as `Utilities::pow()` - I made sure that the argument in `pow` is `types::global_dof_index` to also work with large sizes.

include/exadg/postprocessor/kinetic_energy_spectrum.cpp
kronbichler: I introduced a separate variable `size`, see c6b0cb09

include/exadg/utilities/hypercube_resolution_parameters.h
kronbichler: Can you have a look at 3442131d?

include/exadg/utilities/hypercube_resolution_parameters.h
```diff
@@ -71,13 +71,13 @@ fill_resolutions_vector(
 {
   unsigned int l = 0, n_subdivisions_1d = 1;
 
-  double n_cells_min = (double)n_dofs_min / dofs_per_element;
-  double n_cells_max = (double)n_dofs_max / dofs_per_element;
+  types::global_dof_index n_cells_min = n_dofs_min / dofs_per_element;
+  types::global_dof_index n_cells_max = (n_dofs_max + dofs_per_element - 1) / dofs_per_element;
```
nfehn, line 75: I am currently not sure if this is correct. The user expects that `n_dofs_min <= n_dofs <= n_dofs_max` for all `n_dofs` realized in the throughput experiment. So we would have to ceil `n_cells_min` and floor `n_cells_max`?

kronbichler: I agree, my formula did it the wrong way around (floor/ceil), including slightly too much.

include/exadg/postprocessor/spectral_analysis/bijection.h
nfehn: @peterrum would you like to open an issue or pull request for that? I think we should keep this out of the present PR.

nfehn (approved):

peterrum (approved):

PR #75, opened 2021-05-20
refactor MatrixFreeData to avoid code duplication

nfehn:
Two questions came up during refactoring:

- Why do we use `data.tasks_parallel_scheme = MatrixFree<dim, Number>::AdditionalData::partition_partition;` within the drivers, and `data.tasks_parallel_scheme = MatrixFree<dim, MultigridNumber>::AdditionalData::none;` within multigrid? Can we unify this? Is this still up to date with the most recent deal.II version, etc.?
- In the new constructor of `MatrixFreeData` (previously within the `Driver`s): Why do we cast to `parallel::distributed::Triangulation<dim>` if the function `Categorization::do_cell_based_loops()` called afterwards only expects `Triangulation<dim>`?

### DIFF
--- a/include/exadg/matrix_free/matrix_free_data.h
+++ b/include/exadg/matrix_free/matrix_free_data.h
@@ -19,8 +19,8 @@
  *  ______________________________________________________________________
  */
 
-#ifndef INCLUDE_FUNCTIONALITIES_MATRIX_FREE_WRAPPER_H_
-#define INCLUDE_FUNCTIONALITIES_MATRIX_FREE_WRAPPER_H_
+#ifndef INCLUDE_FUNCTIONALITIES_MATRIX_FREE_DATA_H_
+#define INCLUDE_FUNCTIONALITIES_MATRIX_FREE_DATA_H_
 
 // deal.II
 #include <deal.II/distributed/tria.h>
@@ -208,4 +208,4 @@ private:
 };
 } // namespace ExaDG
 
-#endif /* INCLUDE_FUNCTIONALITIES_MATRIX_FREE_WRAPPER_H_ */
+#endif /* INCLUDE_FUNCTIONALITIES_MATRIX_FREE_DATA_H_ */
--- a/include/exadg/matrix_free/matrix_free_data.h
+++ b/include/exadg/matrix_free/matrix_free_data.h
@@ -39,26 +39,11 @@ struct MatrixFreeData
 {
 public:
   /**
-   * Default constructor. Does not care about categorization of cells in case of cell based face
-   * loops.
+   * Default constructor.
    */
   MatrixFreeData()
   {
     data.tasks_parallel_scheme = MatrixFree<dim, Number>::AdditionalData::none;
-  }
-
-  /**
-   * Cnstructor that cares about categorization of cells in case of cell based face loops.
-   */
-  MatrixFreeData(std::shared_ptr<Triangulation<dim> const> const triangulation,
-                 bool const                                      use_cell_based_face_loops)
-  {
-    data.tasks_parallel_scheme = MatrixFree<dim, Number>::AdditionalData::none;
-
-    if(use_cell_based_face_loops)
-    {
-      Categorization::do_cell_based_loops(*triangulation, data);
-    }
   }
 
   /**
--- a/include/exadg/compressible_navier_stokes/driver.cpp
+++ b/include/exadg/compressible_navier_stokes/driver.cpp
@@ -124,7 +124,7 @@ Driver<dim, Number>::setup(std::shared_ptr<ApplicationBase<dim, Number>> app,
                                                mpi_comm));
 
   // initialize matrix_free
-  matrix_free_data.reset(new MatrixFreeData<dim, Number>(triangulation, false));
+  matrix_free_data.reset(new MatrixFreeData<dim, Number>());
   matrix_free_data->append(pde_operator);
 
   matrix_free.reset(new MatrixFree<dim, Number>());
--- a/include/exadg/compressible_navier_stokes/driver.cpp
+++ b/include/exadg/compressible_navier_stokes/driver.cpp
@@ -111,23 +111,21 @@ Driver<dim, Number>::setup(std::shared_ptr<ApplicationBase<dim, Number>> app,
   application->set_field_functions(field_functions);
 
   // initialize compressible Navier-Stokes operator
-  comp_navier_stokes_operator.reset(new Operator<dim, Number>(*triangulation,
-                                                              mapping,
-                                                              degree,
-                                                              boundary_descriptor_density,
-                                                              boundary_descriptor_velocity,
-                                                              boundary_descriptor_pressure,
-                                                              boundary_descriptor_energy,
-                                                              field_functions,
-                                                              param,
-                                                              "fluid",
-                                                              mpi_comm));
+  pde_operator.reset(new Operator<dim, Number>(*triangulation,
+                                               mapping,
+                                               degree,
+                                               boundary_descriptor_density,
+                                               boundary_descriptor_velocity,
+                                               boundary_descriptor_pressure,
+                                               boundary_descriptor_energy,
+                                               field_functions,
+                                               param,
+                                               "fluid",
+                                               mpi_comm));
 
   // initialize matrix_free
-  matrix_free_data.reset(new MatrixFreeData<dim, Number>());
-  matrix_free_data->data.tasks_parallel_scheme =
-    MatrixFree<dim, Number>::AdditionalData::partition_partition;
-  comp_navier_stokes_operator->fill_matrix_free_data(*matrix_free_data);
+  matrix_free_data.reset(new MatrixFreeData<dim, Number>(triangulation, false));
+  matrix_free_data->append(pde_operator);
 
   matrix_free.reset(new MatrixFree<dim, Number>());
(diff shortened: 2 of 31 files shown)


kronbichler:
It should only be `data.tasks_parallel_scheme = MatrixFree<dim, MultigridNumber>::AdditionalData::none`. The setting is for thread-parallel computations, but we disable that component anyway by setting the parameter `1` in `MPI_InitFinalize` the solvers, e.g. https://github.com/exadg/exadg/blob/0040d0912c8dbe92f32f1441ee0e7de1579f3cf9/include/exadg/incompressible_navier_stokes/solver.h#L103
I think this is some leftover from long ago (2013/2014 time frame).

Regarding the second question, I think we do not need the cast any more, it is due to some preliminary interfaces in deal.II I assume.

nfehn:
@kronbichler Can you have a look at the changes once more?

kronbichler (approved):

nfehn:
Are we ready to merge here? @peterrum maybe you would also like to have a look at these changes?

include/exadg/matrix_free/matrix_free_data.h
```diff
@@ -36,6 +38,40 @@ template<int dim, typename Number>
...
+    }
+  }
+
+  /**
+   * Append MatrixFreeData by the needs of (another) pde_operator provided as argument to this
+   * function.
+   */
+  template<typename Operator>
+  void
+  append(std::shared_ptr<Operator> pde_operator)
```
peterrum, line 70: Why not use the name `insert()` here as in the case of the other functions? As far as I see this function does nothing else than delegating the task to the actual operator, which calls the `insert_*()` functions. Or alternatively, one could call the insert functions `append_*`.

include/exadg/matrix_free/matrix_free_data.h
```diff
@@ -36,6 +38,40 @@ template<int dim, typename Number>
...
+  MatrixFreeData(std::shared_ptr<Triangulation<dim> const> const triangulation,
+                 bool const                                      use_cell_based_face_loops)
+  {
+    data.tasks_parallel_scheme = MatrixFree<dim, Number>::AdditionalData::none;
+
+    if(use_cell_based_face_loops)
+    {
+      Categorization::do_cell_based_loops(*triangulation, data);
+    }
+  }
```
peterrum, line 62: Why are you passing a triangulation object to the constructor? One does not store a reference to it and one only needs it if ECL should be enabled. One could actually postpone categorizing till the first time `AdditionalData` is needed for setting up of `MatrixFree` (by that time a `Triangulation` should be definitely be attached, which could be queried for triangulation).

From having a look at the usage of the constructor in the drivers, I think it would be the best to remove the triangulation from the constructor and to have a constructor similar to `MF::AdditionalData` (with only optional parameters).

nfehn: I don't understand this point completely. I made this change because we had the same code (if cell based face loops enabled, do the categorization) in all the drivers (code duplication).

What exactly is your concern? That the pointer to triangulation is unitialized? I understand your last point as moving this part of the code to deal.II in the long run, but then this is something not to be discussed here in ExaDG?

include/exadg/matrix_free/matrix_free_data.h
nfehn: I don't see this contradiction so clearly. The difference is the following: For the insert functions, the argument passed to the function is the one that is inserted into the vector of items with the same data type. This is not the case for `MatrixFreeData`, i.e., the call

```
matrix_free_data->append(pde_operator);
```

does not directly insert a PDE operator into `MatrixFreeData`, but rather fills the `MatrixFreeData` object with information provided by the PDE operator. Note also that this function already has a function called `append_mapping_flags()`, and I consider the new function more in this logic than in the insert logic.

Regarding a renaming of the `insert_` functions, I think one would have to first change the implementation of the insert functions, which currently assert that the element to be inserted does not already exist, i.e. wouldn't these functions look different when really implementing an append-logic?

include/exadg/matrix_free/matrix_free_data.h
nfehn: Maybe let me add another point why this functionality is part of the constructor and not a separate function like the `append()` function or another one. The idea is that `MatrixFree` works on a single triangulation, but potentially for a multitude of fields for coupled problems (each of them requiring different dof-handlers and quadrature rules). Since the categorization part is related to the triangulation (or the functionality that is the same for all the different fields), the idea was to do this in the constructor (done once for one `MatrixFreeData` object for all the different fields).

include/exadg/matrix_free/matrix_free_data.h
peterrum: I am must admit I prefer how it has been done before without the additional indirection via `MatrixFreeData`. But feel free to ignore the comment.

include/exadg/matrix_free/matrix_free_data.h
peterrum: > Since the categorization part is related to the triangulation (or the functionality that is the same for all the different fields), the idea was to do this in the constructor (done once for one MatrixFreeData object for all the different fields).

In the case of AMR, you will need to call it after each mesh refinement so one would need to to call the `Categorization::do_cell_based_loops()` function where `MF::reinit()` is called and at that place you have access to `Triangulation` via a `DoFHandler`.

nfehn: And what does that mean concretely? I can change it back to how it was before. But since we call it in every `Driver`, this is sub-optimal design. Why isn't it part of `MF::reinit()` in deal.II?

nfehn:
@kronbichler Could you briefly check again after the most recent changes? Then, I would like to get this merged to continue with other things.

include/exadg/matrix_free/matrix_free_data.h
kronbichler: The reason why the categorization is not part of deal.II is that this is only one option to group work. The main problem is that categories can express more than these settings, and the library cannot really know how to combine things in case those are conflicting. Therefore, it leaves the decision of grouping to the user code.

kronbichler: Regarding the code duplication: Why not add a function `matrix_free_data->get_additional_data()` that then can call the categorization (if enabled) before it returns the reference to `AdditionalData`? Or, if we like to express it more clearly, call it `construct_additional_data()`? The function would then nicely mirror the `get_dof_handler()` and other functions?

nfehn: But then we are pretty close to the constructor variant that I had before in my opinion, which @peterrum voted against. I think now I don't want to change it again. In particular, from your comments it seems as if we currently don't know whether this is functionality belonging to `AdditionalData` or rather `MatrixFree::reinit()`, independently of how it is currently realized in deal.II. Intuitively, I would say it belongs to `MatrixFree::reinit()` and `AdditionalData` potentially needs additional parameters so that the user can specify the configuration to be used.

peterrum: If I understand Martin's description correctly, this is what he proposes:
```cpp
AdditionalData
construct_additional_data() const
{
  auto data = this->data;

  data.tasks_parallel_scheme = MatrixFree<dim, Number>::AdditionalData::none;

  if(use_cell_based_face_loops)
    Categorization::do_cell_based_loops(dof_handlers.front().get_triangulation(), data);

  return data;
}
```
This function would be called when `MatrixFree` is setup, which is also needed during AMR. So it would solve my concern from above.

nfehn: Ok, but I would prefer `get_additional_data()` for consistency

nfehn: I again decided against this change because this change is in my opinion only allowed if we make `data` private. But: we need `data` as a public member in the multigrid implementation in order to set the h-level. This makes things very complicated since it is then no longer understandable for the user how to access then the `AdditionalData` object (public member vs. get()-function, or a get() vs. a construct()-function????). I am more and more convinced that this needs to be done by `matrix_free->reinit()` to obtain a clean solution.

kronbichler: The reason that it is complicated to find a clean solution here (and I agree with the concerns and difficulties) alone is not the right implication to do it in `MatrixFree::reinit` - the `reinit()` function needs deal with whatever requirements are set there. This means that it is not always possible to provide the optimal interface for a particular downstream code like ExaDG. The main problem is that the categories can serve multiple purposes, and automatically applying one way of setting things up can destroy other use cases. As I really fear that it is challenging to get all possible cases right when automatically trying to detect what the user *intends* to do (rather letting the user say what should be done), it needs to be controlled by an additional parameter. It could make sense to have that parameter in `MF::AdditionalData` in my opinion, but the solution requires to think about the possible use cases in the context of `reinit` and what assertions are needed such that requirements are not in conflict.

From these considerations I would guess that in the short run it is probably 5x as much programmer work as repeating the statements here and maintaining ExaDG with redundancy in the next few years (and we have come a good way along that way already with the extensive discussion we're having), but even that is not the final metric: Since we have spent time on the use cases already, this effort we put in now might as well cover cases in the future where the new parameter comes in handy, so we save future trouble.

nfehn:
I think there are no major concerns any more, so I will merge

Issue #78, opened 2021-05-25
Introduce ExaDG::Grid class?


nfehn:
This issue aims at discussing whether an `ExaDG::Grid` class could be helpful and how this class should look like. In the ExaDG code, we realized that `Triangulation`, `PeriodicFaces`, and `Mapping` are tightly coupled. An idea would be to combine these items to a class `ExaDG::Grid`, with `triangulation`, `periodic_faces`, and `mapping` as public members. 

- This way, the code redundancy found in all `Driver` classes when constructing the `dealii::Triangulation` depending on the triangulation type (distributed, fully-distributed) could be avoided https://github.com/exadg/exadg/blob/c520cdfa6a491535d61b171e29ccad06f9d1832b/include/exadg/poisson/driver.cpp#L75-L89
- The interface of `Application::create_grid()` could be simplified https://github.com/exadg/exadg/blob/c520cdfa6a491535d61b171e29ccad06f9d1832b/include/exadg/poisson/driver.cpp#L93
- certain `Triangulation` functionality such as creating a fully-distributed triangulation currently leads to code redundancy that should actually only be implemented once https://github.com/exadg/exadg/blob/c520cdfa6a491535d61b171e29ccad06f9d1832b/applications/incompressible_navier_stokes/flow_past_cylinder/application.h#L476-L505
- This new class `ExaDG::Grid` could incorporate the functionality to read a grid from an external file (a grid generated, e.g., via gmsh) so that this functionality only needs to be implemented once in ExaDG.

@kronbichler @peterrum @bergbauer Please feel free to share your ideas.

PR #79, opened 2021-05-25
introduce pure virtual MovingMeshInterface class

nfehn:
A preparation for realizing #78 

This PR breaks the inheritance sequence of `MovingMeshBase` (`Mapping` -> ... -> `MappingQCache` -> `MappingDoFVector` -> `MovingMeshBase`) by using composition instead. This way, the time integrators in ExaDG only depend on the pure virtual interface of `MovingMeshInterface` and no longer on the implementation of `MappingDoFVector`. From a different point of view, this PR aims to separate moving mesh functionality from moving mesh data structures. The reason behind is that for an upcoming `ExaDG::Grid` class, the inheritance style is no longer expedient in my opinion as this conflicts with modularity and flexibility. The idea I currently have in mind is that a grid with time dependent mapping will not be derived from `ExaDG::Grid` (say `ExaDG::MovingGrid`), but instead `ExaDG::Grid` has a member function `attach_time_dependent_mapping()` and can internally decide via a getter function `ExaDG::Grid::get_mapping()` which mapping is the correct one to be passed to all other data structures and functionality in ExaDG and deal.II that need a mapping. The moving mesh classes, `MovingMeshBase` and derived ones, would care about the moving mesh functionality, but `ExaDG::Grid` does not need to know about that and how this is realized as long as it has a pointer to a `Mapping` to serve the interfaces of ExaDG and deal.II.

### DIFF
--- a/include/exadg/grid/moving_mesh_function.h
+++ b/include/exadg/grid/moving_mesh_function.h
@@ -81,10 +81,10 @@ public:
     FE_Nothing<dim> dummy_fe;
     FEValues<dim>   fe_values(*this->mapping_undeformed,
                             dummy_fe,
-                            QGaussLobatto<dim>(this->get_degree() + 1),
+                            QGaussLobatto<dim>(this->moving_mapping->get_degree() + 1),
                             update_quadrature_points);
 
-    MappingQCache<dim>::initialize(
+    this->moving_mapping->initialize(
       triangulation,
       [&](typename Triangulation<dim>::cell_iterator const & cell) -> std::vector<Point<dim>> {
         fe_values.reinit(cell);
@@ -94,8 +94,8 @@ public:
         for(unsigned int i = 0; i < fe_values.n_quadrature_points; ++i)
         {
           // need to adjust for hierarchic numbering of MappingQCache
-          Point<dim> const point =
-            fe_values.quadrature_point(this->hierarchic_to_lexicographic_numbering[i]);
+          Point<dim> const point = fe_values.quadrature_point(
+            this->moving_mapping->hierarchic_to_lexicographic_numbering[i]);
           Point<dim> displacement;
           for(unsigned int d = 0; d < dim; ++d)
             displacement[d] = displacement_function->value(point, d);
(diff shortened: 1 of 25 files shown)


nfehn:
@kronbichler Could you have a brief look at this PR? Upcoming PRs would use the changes proposed here (and the changes made here are probably not final).

include/exadg/grid/moving_mesh_function.h
```diff
@@ -81,10 +81,10 @@ class MovingMeshFunction : public MovingMeshBase<dim, Number>
     FE_Nothing<dim> dummy_fe;
```
kronbichler, line 81: I'm wondering if the `initialize` function is still needed. It seems that `update` is only directed here without doing its own work, and I can't find uses of `MovingMeshFunction::initialize()` elsewhere. So I would simply move this whole function into `update()`.

kronbichler:
Apart from that question, I like the design and think this will be useful.

include/exadg/grid/moving_mesh_function.h
nfehn: @kronbichler I made the function private. I would currently like to keep the function itself as it makes the code slightly more structured and provides additional documentation, e.g., that the function is interpreted as a displacement.

kronbichler: I agree with the documentation aspect, so it is fine for me to merge (even though I do not see a real advantage in terms of structure because it merely adds an indirection).

kronbichler (approved):

kronbichler:
We would need to rebase on top of #81 to get the CI test through (change in deal.II, in particular the change in the template argument of `DataOutFaces`) - but since this was passing before and the only change is a private declaration, I think we can also merge immediately.

include/exadg/grid/moving_mesh_function.h
nfehn: I am probably just a bit pedantic about code readability ;). To give an example, the line that sets the time for the function would somehow get lost when merging the two function into one. In the current form, the procedural design is more clear in my opinion. This is something of interest to the user. The implementation details are then part of a private member function.

PR #80, opened 2021-05-26
introduce class ExaDG::Grid.

### LINKED ISSUE #78: Introduce ExaDG::Grid class?

nfehn:
resolves #78 

The goal of this MR is to simplify interfaces, render them more stable (as we detected that we had to change this part quite often in the past), and being prepared for upcoming features such as, e.g., reading in grid files or simplicial/mixed meshes, which should all be handled by the new `ExaDG::Grid` class in order to not always touch many other places in the code (including all the application files).

@kronbichler @peterrum @bergbauer Since this is quite elementary for the design of `ExaDG`, please share your opinion and what we should have in mind already right now e.g. for upcoming features mentioned above.

Upcoming PRs: Many interfaces in `ExaDG` are so far not adapted to the new `Grid` class, which is certainly a major point of discussion, i.e., how deep to inject this new class into `ExaDG` and at which level of the implementation to work with raw (here: `deal.II`) data types. I would plan to at least "inject" `ExaDG::Grid` into the spatial discretization operators.

### DIFF
--- a/include/exadg/grid/grid.h
+++ b/include/exadg/grid/grid.h
@@ -22,10 +22,18 @@
 #ifndef INCLUDE_EXADG_GRID_GRID_H_
 #define INCLUDE_EXADG_GRID_GRID_H_
 
+// deal.II
+#include <deal.II/distributed/fully_distributed_tria.h>
+#include <deal.II/distributed/tria.h>
+#include <deal.II/grid/grid_tools.h>
+
+// ExaDG
 #include <exadg/grid/enum_types.h>
 
 namespace ExaDG
 {
+using namespace dealii;
+
 struct GridData
 {
   GridData()
--- a/include/exadg/grid/grid.h
+++ b/include/exadg/grid/grid.h
@@ -1,0 +1,100 @@
+/*  ______________________________________________________________________
+ *
+ *  ExaDG - High-Order Discontinuous Galerkin for the Exa-Scale
+ *
+ *  Copyright (C) 2021 by the ExaDG authors
+ *
+ *  This program is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *  ______________________________________________________________________
+ */
+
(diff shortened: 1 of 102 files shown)


include/exadg/grid/grid.h
```diff
@@ -0,0 +1,100 @@
...
+
+  /**
+   * dealii::GridTools::PeriodicFacePair's.
+   */
+  PeriodicFaces periodic_faces;
+
+  /**
+   * dealii::Mapping.
+   */
+  std::shared_ptr<Mapping<dim>> mapping;
```
peterrum, line 93: I would make these private.

peterrum (commented):
On the first glace it looks good :+1:

include/exadg/grid/grid.h
nfehn: I think it makes sense to think more about the public variant. For example, the function `create_grid()` needs to access the individual items, e.g.,

```
    grid->get_triangulation().refine_global(data.n_refine_global);
```
which would require a get-function that is not `const`. In terms of software design, this is equivalent to a public member variable in my opinion. The getter-variant is just more complicated.

kronbichler:
This looks good from the general level. (I looked at the interfaces and drivers, as well as some application.)

nfehn:
In principle, I plan more changes, but I wonder if the present state is already one to be merged because it will be difficult to separate additional upcoming changes from previous ones during code review?

nfehn:
> This looks good from the general level. (I looked at the interfaces and drivers, as well as some application.)

@kronbichler Is this an approval?

kronbichler (approved):
Now it is an approval :smile: (as the initial state was marked as draft).

PR #81, opened 2021-05-28
Avoid deprecated functionality of deal.II

kronbichler:
With the 9.3 release of deal.II and switch to version 10.0-pre, some old functionality has been marked deprecated (it was on the early deprecation state before). This PR makes the necessary updates. All interfaces I use now should also have been available in 9.3 for a few months at least, so the code should be stable. There is one additional change that is a slightly different category, namely the `find_active_cell_around_point` function that has changed behavior in case points are not found, as specified in https://github.com/dealii/dealii/pull/11276.

### DIFF
--- a/include/exadg/grid/find_all_active_cells_around_point.h
+++ b/include/exadg/grid/find_all_active_cells_around_point.h
@@ -40,19 +40,16 @@ find_all_active_cells_around_point(Mapping<dim> const &                         
 
   std::vector<Pair> adjacent_cells;
 
-  try
-  {
-    Pair first_cell =
-      GridTools::find_active_cell_around_point(cache, point, cell_hint, marked_vertices, tolerance);
+  Pair first_cell =
+    GridTools::find_active_cell_around_point(cache, point, cell_hint, marked_vertices, tolerance);
 
+  if(first_cell.first != tria.end())
+  {
     // update cell_hint to have a good hint when the function is called next time
     cell_hint = first_cell.first;
 
     adjacent_cells =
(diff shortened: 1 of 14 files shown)


kronbichler:
@nfehn do we have a test I can run for the `find_active_cell...` function?

nfehn:
I think it is best to run some examples manually in `applications/fluid_structure_interaction/` (e.g. `cylinder_with_flag` or `pressure_wave`) for different number of processors (and potentially different mesh refinement levels) to make sure that everything is still working. Someone reported in deal.II that the `find_active_cell_around_point()` function is now much slower, which would be bad as this would render the setup expensive for FSI.

nfehn (approved):

kronbichler:
I ran `cylinder_with_flag` with up to 7 MPI ranks and 0, 1, 2 refinements. Everything seems fine for the first 500 time steps. However, for 1 refinement of the fluid and 2 refinements for the structure I got a crash in the FGMRES solver (either ALE or structure, as far as I can see) between time step 500 and 600, struggling to converge below 1e-5. If I remember correctly you had some issues sometimes @nfehn? How plausible is it that the active point search is responsible for a failure at that stage? The pressure wave also runs for the first 10 time steps with several MPI rank configurations.

kronbichler:
I also verified by checking traces of running the code on the pressure wave that we trigger the `else` case added by the search, so the functionality should be equivalent to before.

nfehn:
Yes, I think running these examples should be enough. I am pretty sure that the simulation would crash much earlier (during setup or the first time step) if a bug in searching neighboring points was introduced.

PR #82, opened 2021-05-29
Actually use subcommunicator for PETSc matrix

kronbichler:
In #68, I did prepare for the PETSc matrix to use sub-communicators but not actually apply it. This is done here. The code is a bit awkward because I need to go into the function that creates the matrix and the function that deletes the matrix again, in order to be able to free the MPI subcommunicator. I tested with valgrind that there are no memory leaks. However, I left a TODO for the case that someone has a better idea than me to handle this case.

### DIFF
--- a/include/exadg/operators/operator_base.cpp
+++ b/include/exadg/operators/operator_base.cpp
@@ -775,7 +775,8 @@ OperatorBase<dim, Number, n_components>::internal_init_system_matrix(
   // actual communication, i.e., which are left out from sub-communication
   types::global_dof_index const sum_of_locally_owned_dofs =
     Utilities::MPI::sum(owned_dofs.n_elements(), mpi_comm);
-  AssertThrow(sum_of_locally_owned_dofs == owned_dofs.size() || owned_dofs.n_elements() == 0,
+  bool const my_rank_is_part_of_subcommunicator = sum_of_locally_owned_dofs == owned_dofs.size();
+  AssertThrow(my_rank_is_part_of_subcommunicator || owned_dofs.n_elements() == 0,
               ExcMessage("The given communicator mpi_comm in init_system_matrix does not span "
                          "all MPI processes needed to cover the index space of all dofs: " +
                          std::to_string(sum_of_locally_owned_dofs) + " vs " +
@@ -800,8 +801,7 @@ OperatorBase<dim, Number, n_components>::internal_init_system_matrix(
   else /* if (!is_dg && !is_mg)*/
     DoFTools::make_sparsity_pattern(dof_handler, dsp, *this->constraint);
 
-  // only work on this aspect if we are part of the communicator with the dofs
-  if(sum_of_locally_owned_dofs == owned_dofs.size())
+  if(my_rank_is_part_of_subcommunicator)
   {
     SparsityTools::distribute_sparsity_pattern(dsp, owned_dofs, mpi_comm, relevant_dofs);
     system_matrix.reinit(owned_dofs, owned_dofs, dsp, mpi_comm);
--- a/include/exadg/operators/operator_base.cpp
+++ b/include/exadg/operators/operator_base.cpp
@@ -795,6 +795,12 @@ OperatorBase<dim, Number, n_components>::internal_init_system_matrix(
     for(const auto & cell : dof_handler.active_cell_iterators())
       if(cell->is_locally_owned())
         ++n_locally_owned_cells;
+
+    // In case some of the MPI ranks do not have cells, we create a
+    // sub-communicator to exclude all those processes from the PETSc
+    // communication and hence speed up those operations. Note that we have to
+    // free the communicator again, which happens in the preconditioner_amg.h
+    // file which owns the matrix passed to the present function.
     MPI_Comm petsc_communicator;
     if(Utilities::MPI::min(n_locally_owned_cells, dof_handler.get_communicator()) == 0)
       MPI_Comm_split(dof_handler.get_communicator(),
--- a/include/exadg/operators/operator_base.cpp
+++ b/include/exadg/operators/operator_base.cpp
@@ -723,9 +723,10 @@ make_sparsity_pattern(const DoFHandler<dim, spacedim> & dof,
 template<int dim, typename Number, int n_components>
 void
 OperatorBase<dim, Number, n_components>::init_system_matrix(
-  TrilinosWrappers::SparseMatrix & system_matrix) const
+  TrilinosWrappers::SparseMatrix & system_matrix,
+  MPI_Comm const &                 mpi_comm) const
 {
-  internal_init_system_matrix(system_matrix);
+  internal_init_system_matrix(system_matrix, mpi_comm);
 }
 
 template<int dim, typename Number, int n_components>
@@ -741,9 +742,10 @@ OperatorBase<dim, Number, n_components>::calculate_system_matrix(
 template<int dim, typename Number, int n_components>
 void
 OperatorBase<dim, Number, n_components>::init_system_matrix(
-  PETScWrappers::MPI::SparseMatrix & system_matrix) const
+  PETScWrappers::MPI::SparseMatrix & system_matrix,
+  MPI_Comm const &                   mpi_comm) const
 {
-  internal_init_system_matrix(system_matrix);
+  internal_init_system_matrix(system_matrix, mpi_comm);
 }
 
 template<int dim, typename Number, int n_components>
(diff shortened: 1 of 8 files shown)


kronbichler:
This needs to be rebased on top of #81 to fix the CI build error.

include/exadg/operators/operator_base.cpp
```diff
@@ -788,7 +788,33 @@ OperatorBase<dim, Number, n_components>::internal_init_system_matrix(
...
+    // file which owns the matrix passed to the present function.
+    MPI_Comm petsc_communicator;
+    if(Utilities::MPI::min(n_locally_owned_cells, dof_handler.get_communicator()) == 0)
+      MPI_Comm_split(dof_handler.get_communicator(),
+                     n_locally_owned_cells > 0,
+                     Utilities::MPI::this_mpi_process(dof_handler.get_communicator()),
+                     &petsc_communicator);
+    else
+      petsc_communicator = dof_handler.get_communicator();
+    if(n_locally_owned_cells > 0)
```
nfehn, line 812: An empty line above this one would ease readability.

include/exadg/operators/operator_base.cpp
```diff
@@ -788,7 +788,33 @@ OperatorBase<dim, Number, n_components>::internal_init_system_matrix(
...
+      system_matrix.reinit(owned_dofs, owned_dofs, dsp, petsc_communicator);
+  }
+  else
+#endif
+    system_matrix.reinit(owned_dofs, owned_dofs, dsp, dof_handler.get_communicator());
```
nfehn, line 817: We should use brackets around this line for readability

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
```diff
@@ -149,10 +149,27 @@ class PreconditionerBoomerAMG : public PreconditionerBase<Number>
   ~PreconditionerBoomerAMG()
   {
 #ifdef DEAL_II_WITH_PETSC
-    PetscErrorCode ierr = VecDestroy(&petsc_vector_dst);
-    AssertThrow(ierr == 0, ExcPETScError(ierr));
-    ierr = VecDestroy(&petsc_vector_src);
-    AssertThrow(ierr == 0, ExcPETScError(ierr));
+    bool const has_subcommunicator =
+      Utilities::MPI::min(system_matrix.m(),
+                          pde_operator.get_matrix_free().get_dof_handler().get_communicator()) == 0;
```
nfehn, line 154: just a few minor comments to make this more readable: i) Introduce a variable for the MPI comm to shorten this expression, ii) use brackets for the expression `(... == 0)` right to the assignment, iii) introduce an empty line before the if-statement below

nfehn:
Regarding the design problem: I think we need to remove the sub-communicator from the `OperatorBase` implementation. Basically, the design in `ExaDG` is that preconditioners hold pointers to the operator, from which they can query what they need. So my question is: why isn't it possible that the preconditioner constructs the sub-communicator, and hands it over to `OperatorBase::init_system_matrix()`. It seems as if all that is needed is a getter function like `OperatorBase::get_dof_handler()`, which can be realized via the existing getters `OperatorBase::get_matrix_free()` and `OperatorBase::get_dof_index()`, can't it? 

In my opinion, this realization would then be in line with the single-responsibility and RAII principles.

kronbichler:
Yes, passing the communicator into the `init_system_matrix` is a clean solution for that, it seems I was mainly try not to touch the interfaces before. (Note that I added the deletion of the sub-communicator only now when I cleaned things up for the PR, the runs were done with only the communicator created.)

kronbichler:
I am trying to get an even cleaner solution in terms of freeing memory, but it is quite some work to pass the communicator through all interfaces. I will need until tomorrow for that.

kronbichler:
I now completely rewrote this and wrapped the communicator into its own deleter. This should hopefully make clear who has the responsibility to free the communicator without dirty tricks.

include/exadg/operators/operator_base.cpp
```diff
@@ -759,12 +761,24 @@ template<int dim, typename Number, int n_components>
...
 {
   DoFHandler<dim> const & dof_handler = this->matrix_free->get_dof_handler(this->data.dof_index);
 
   IndexSet const & owned_dofs =
     is_mg ? dof_handler.locally_owned_mg_dofs(this->level) : dof_handler.locally_owned_dofs();
+
+  // check for a valid subcommunicator - the second check is needed on those
+  // MPI ranks which will not participate in the actual communication, i.e.,
+  // which are left out
+  AssertThrow(Utilities::MPI::sum(owned_dofs.n_elements(), mpi_comm) == owned_dofs.size() ||
```
nfehn, line 775: Question: does `owned_dofs.size()` involve a call to MPI, or why is it more than the dofs of one MPI process? Should be documented more clearly in my opinion.

include/exadg/operators/operator_base.cpp
```diff
@@ -784,11 +798,12 @@ OperatorBase<dim, Number, n_components>::internal_init_system_matrix(
   else /* if (!is_dg && !is_mg)*/
     DoFTools::make_sparsity_pattern(dof_handler, dsp, *this->constraint);
 
-  SparsityTools::distribute_sparsity_pattern(dsp,
-                                             owned_dofs,
-                                             dof_handler.get_communicator(),
-                                             relevant_dofs);
-  system_matrix.reinit(owned_dofs, owned_dofs, dsp, dof_handler.get_communicator());
+  // only work on this aspect if we are part of the communicator with the dofs
+  if(Utilities::MPI::sum(owned_dofs.n_elements(), mpi_comm) == owned_dofs.size())
```
nfehn, line 802: seems to be the same expression as above, so maybe introduce a variable, e.g. `bool my_rank_is_part_of_sub_communicator`?

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
```diff
@@ -37,6 +37,44 @@ namespace ExaDG
 {
 using namespace dealii;
 
+template<typename Operator>
+std::unique_ptr<MPI_Comm, void (*)(MPI_Comm *)>
+create_subcommunicator(Operator const & op)
+{
+  unsigned int n_locally_owned_cells = 0;
+  auto const & dof_handler           = op.get_matrix_free().get_dof_handler();
```
nfehn, line 45: `MatrixFree` can have multiple `DofHandler`s, so I think this is error prone (`op.get_dof_index()`!). I my opinion it would make much more sense to keep the `Operator` out of the interface and use `DofHandler` instead.

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
```diff
@@ -37,6 +37,44 @@ namespace ExaDG
...
+  for(const auto & cell : dof_handler.active_cell_iterators())
+    if(cell->is_locally_owned())
+      ++n_locally_owned_cells;
+
+  // In case some of the MPI ranks do not have cells, we create a
+  // sub-communicator to exclude all those processes from the PETSc
+  // communication and hence speed up those operations. Note that we have to
+  // free the communicator again, which happens in the preconditioner_amg.h
```
nfehn, line 53: we are now in the precondition_amg.h file -> update comment

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
```diff
@@ -37,6 +37,44 @@ namespace ExaDG
...
+    std::unique_ptr<MPI_Comm, void (*)(MPI_Comm *)> subcommunicator(new MPI_Comm,
+                                                                    [](MPI_Comm * comm) {
+                                                                      MPI_Comm_free(comm);
+                                                                      delete comm;
+                                                                    });
+    MPI_Comm_split(dof_handler.get_communicator(),
+                   n_locally_owned_cells > 0,
+                   Utilities::MPI::this_mpi_process(dof_handler.get_communicator()),
+                   subcommunicator.get());
+    return subcommunicator;
```
nfehn, line 66: Please introduce empty lines between the three statements for readability. The same below.

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
```diff
@@ -37,6 +37,44 @@ namespace ExaDG
...
+  auto const & dof_handler           = op.get_matrix_free().get_dof_handler();
+  for(const auto & cell : dof_handler.active_cell_iterators())
+    if(cell->is_locally_owned())
+      ++n_locally_owned_cells;
+
+  // In case some of the MPI ranks do not have cells, we create a
+  // sub-communicator to exclude all those processes from the PETSc
+  // communication and hence speed up those operations. Note that we have to
+  // free the communicator again, which happens in the preconditioner_amg.h
+  // file which owns the matrix passed to the present function.
```
nfehn, line 54: file?

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
nfehn: With this change, this function should only depend on `dealii::DofHandler`, and MPI --> shift to deal.II lateron.

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
```diff
@@ -37,6 +37,44 @@ namespace ExaDG
...
+                   Utilities::MPI::this_mpi_process(dof_handler.get_communicator()),
+                   subcommunicator.get());
+    return subcommunicator;
+  }
+  else
+  {
+    std::unique_ptr<MPI_Comm, void (*)(MPI_Comm *)> communicator(new MPI_Comm, [](MPI_Comm * comm) {
+      delete comm;
+    });
+    *communicator = dof_handler.get_communicator();
```
nfehn, line 73: `dof_handler.get_communicator()` called for the fourth time in this function. Maybe introduce a variable?

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
nfehn: forget this one

nfehn:
Overall, this looks like clean code.

include/exadg/operators/operator_base.cpp
kronbichler: One is the number of dofs on the present process, one is for the global sum. I have the check because a subcommunicator leaves out some ranks and we need to make sure that none of those carries dofs. Fixed in cc655075

include/exadg/operators/operator_base.cpp
kronbichler: Fixed in cc655075

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
kronbichler: Fixed in b643a551

include/exadg/operators/operator_base.cpp
kronbichler: I decided to use an additional bool variable as suggested rather than the initial size variable, because it allows me to skip the comment. See af04232b

nfehn (approved):

PR #83, opened 2021-05-31
inject ExaDG::Grid into spatial discretization classes

nfehn:
This is the next PR related to the introduction of the new `ExaDG::Grid` class, see PR #80.

In the present PR, the `ExaDG::Grid` class is injected into the spatial discretization operators. It could make sense to inject `ExaDG::Grid` into `MultigridPreconditioner` in a next step (?).

The time integrators no longer depend on `MovingMeshInterface` and `MatrixFree`, since these aspects are more closely related to the spatial discretization scheme and should therefore be handled by the `pde_operator` in the time integrator class, i.e., matrix-free evaluation is a detail of the spatial discretization that should not be visible by the time integrator.

Some renaming has been performed due to a change in terminology from "Mesh" to "Grid", related to PR #80.

### DIFF
--- a/include/exadg/grid/grid_motion_interface.h
+++ b/include/exadg/grid/grid_motion_interface.h
@@ -24,18 +24,18 @@
 #include <deal.II/fe/mapping.h>
 #include <deal.II/lac/la_parallel_vector.h>
 
-#ifndef INCLUDE_EXADG_GRID_MOVING_MESH_INTERFACE_H_
-#  define INCLUDE_EXADG_GRID_MOVING_MESH_INTERFACE_H_
+#ifndef INCLUDE_EXADG_GRID_GRID_MOTION_INTERFACE_H_
+#  define INCLUDE_EXADG_GRID_GRID_MOTION_INTERFACE_H_
 
 namespace ExaDG
 {
 using namespace dealii;
 
 /**
- * Pure-virtual interface class for moving mesh functionality.
+ * Pure-virtual interface class for moving grid functionality.
  */
 template<int dim, typename Number>
-class MovingMeshInterface
+class GridMotionInterface
 {
 public:
   typedef LinearAlgebra::distributed::Vector<Number> VectorType;
@@ -43,18 +43,18 @@ public:
   /**
    * Destructor.
    */
-  virtual ~MovingMeshInterface()
+  virtual ~GridMotionInterface()
   {
   }
 
   /**
-   * Updates the mapping, i.e., moves the mesh.
+   * Updates the mapping, i.e., moves the grid.
    */
   virtual void
   update(double const time, bool const print_solver_info) = 0;
 
   /**
-   * Print the number of iterations for PDE type mesh motion problems.
+   * Print the number of iterations for PDE type grid motion problems.
    */
   virtual void
   print_iterations() const
@@ -80,4 +80,4 @@ public:
 } // namespace ExaDG
 
 
-#endif /* INCLUDE_EXADG_GRID_MOVING_MESH_INTERFACE_H_ */
+#endif /* INCLUDE_EXADG_GRID_GRID_MOTION_INTERFACE_H_ */
--- a/applications/compressible_navier_stokes/couette/application.h
+++ b/applications/compressible_navier_stokes/couette/application.h
@@ -179,10 +179,10 @@ public:
     param.use_combined_operator = false;
   }
 
-  std::shared_ptr<Grid<dim>>
+  std::shared_ptr<Grid<dim, Number>>
   create_grid(GridData const & data, MPI_Comm const & mpi_comm) final
   {
-    std::shared_ptr<Grid<dim>> grid = std::make_shared<Grid<dim>>(data, mpi_comm);
+    std::shared_ptr<Grid<dim, Number>> grid = std::make_shared<Grid<dim, Number>>(data, mpi_comm);
 
     std::vector<unsigned int> repetitions({2, 1});
     Point<dim>                point1(0.0, 0.0), point2(L, H);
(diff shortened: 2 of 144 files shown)


include/exadg/grid/grid_motion_interface.h
```diff
@@ -24,37 +24,37 @@
...
 {
 using namespace dealii;
 
 /**
- * Pure-virtual interface class for moving mesh functionality.
+ * Pure-virtual interface class for moving grid functionality.
  */
 template<int dim, typename Number>
-class MovingMeshInterface
+class GridMotionInterface
```
kronbichler, line 38: I am not very happy the template `Number` in this class. On the one hand, this forces all the downstream uses of `Grid` to carry that parameter, which results in a huge list of changes in this patch, but I can't see the advantage because `MappingQCache` then stores the data stores as doubles internally. I see that there could be cases where the `grid_coordinates` vector in the virtual function `fill_grid_coordinates_vector` could be `float` from the caller's site, but I would prefer to convert the `float` vector outside of the grid interface, given that the `Triangulation` and associated data is a `double` description as well.

nfehn: I was also all but happy when realizing this yesterday... Principally, one could write `MovingMeshInterface` with `double` and let `MovingMeshBase` do all the type conversions (derived `GridMotion` classes like Poisson or elasticity-type grid motions involve spatial operators `Operator<dim,Number>` (which introduce `Number`), but for this part of the code I think this would in principle be realizable in an understandable way, even though I think the conversion would have to be done inside the grid classes). Solving the mesh motion problem in double precision would be a peformance penalty? 

However, a main motivation to do this type of refactoring was to get rid of `MovingMeshInterface` and `MatrixFree` in the **time integrators**. The time integrators also have `Number` as template because of the DoF-vectors and call `pde_operator->fill_grid_coordinates_vector()` (so this function would have to be templated with `Number`?), to obtain the mesh velocity (stored as `LA::distributed::Vector<Number>` in the time integrator). When computing in `float` precision, wouldn't this mean that we have a `float`-> `double` conversion within `pde_operator->fill_grid_coordinates_vector()` before calling `MovingMeshInterface<dim>::update()` (no `Number` template), and another conversion `double` -> `float` inside `MovingMeshBase` (or derived classes) before solving the mesh motion problem in `float` precision?

Maybe my point here is that the number of changed files in this patch should not be main argument, as fixing all these type conversions problems would have been much more work according to experiences I made in the past. Rather, the argument should be how we want the library to look like in the future (in terms of design, code readability, and peformance). The time integrator classes and `ApplicationBase` classes are already templated with `Number`.

kronbichler: I agree that the number of lines is not the main argument here, and instead the interface. I think performance is a secondary metric here because double/float conversion is typically much faster than filling data into `MappingQCache` with code like this:
https://github.com/exadg/exadg/blob/c9d148c0176bd710052eb60b49d1582a926d31a6/include/exadg/grid/mapping_dof_vector.h#L245-L255

I agree that for code clarity we should strive not to move back and forth between `double` and `Number` vectors. 

So for the two main options:
- We have all moving mesh base classes (`MovingMeshBase`, `MappingDoFVector`, `MovingMeshFunction`) and the underlying `Grid` classes without template on `Number`. The conversion happens in the derived moving mesh classes, e.g. `MovingMeshPoisson::update()` after solving and before calling https://github.com/exadg/exadg/blob/c9d148c0176bd710052eb60b49d1582a926d31a6/include/exadg/grid/moving_mesh_poisson.h#L89-L91
I think doing the conversion manually would be better than templating the function; this would add another line of code here and in all other derived mesh classes. The setup would stay the same, as far as I can see, because the surrounding classes (like the incomp NS driver) make the setup appropriately. The more annoying part, as you point out, is the `fill_grid_coordinates_vector` function. How is that handled right now, can I solve the mesh in `float` but do the time integration in `double`? Or what kind of cross operations (if any) do we support? (As a side remark, I think at some point we should add the option to solve the moving mesh on a coarser mesh or lower polynomial degree, as transfinite interpolation should be possible and much quicker than the solution of a Poisson/elasticity problem on a high-order mesh; but that is a bigger project.)
- We do as suggested here, which ensures a consistent interface within all mapping classes and across calling sites with the vectors. Instead, the conversion appears in `MappingDoFVector` and the associated free functions. It comes with the expense of the additional template argument.

I don't feel too strong about the way I suggested, so putting everything on the table we might probably still decide for the templated way, but I wanted to understand the full rationale.

nfehn: > How is that handled right now, can I solve the mesh in float but do the time integration in double?

I think this is currently not supported. The classes are templated, but the `Driver` classes use the same `Number` type to instantiate essentially all classes.

Right now, I would strive for the second option, mainly in order to not hinder the progress right now and other related PRs that I would like to realize in the next days/weeks. I am definitely willing to change again 100 files or more in the future if this is necessary or changes the code for the good (as I did countless times in the past).

I consider the first option a viable strategy as well and the information you provided is helpful. Maybe we also have to gain more experience whether we will be able to keep the template `Number` out of the new `ExaDG::Grid` class, which will probably grow in functionality in the future. For now, would it be an option that you shift the description of your first option to a separate issue (describing an alternative to the current option or the option suggested in the present PR), so that we do not forget about this?

kronbichler: Yes, that is a good idea.

kronbichler (approved):

include/exadg/grid/grid_motion_interface.h
kronbichler: See #89.

PR #88, opened 2021-06-02
Make coarser triangulation with deal.II facilities

kronbichler:
This is due to extended functionality inside deal.II, mainly https://github.com/dealii/dealii/pull/12356 and the preparation steps before that. Note that I duplicate the function to coarsen from deal.II because I want to do different things on different levels.

This runs also several times faster than the old serial code.

### DIFF
--- a/include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
+++ b/include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
@@ -323,9 +323,8 @@ template<int dim, int spacedim = dim>
 class BalancedGranularityPartitionPolicy : public RepartitioningPolicyTools::Base<dim, spacedim>
 {
 public:
-  BalancedGranularityPartitionPolicy(types::global_cell_index const n_cells_fine,
-                                     unsigned int const             n_mpi_processes)
-    : n_cells_per_level{n_cells_fine}, n_mpi_processes_per_level{n_mpi_processes}
+  BalancedGranularityPartitionPolicy(unsigned int const n_mpi_processes)
+    : n_mpi_processes_per_level{n_mpi_processes}
   {
   }
 
@@ -336,8 +335,10 @@ public:
 
     // TODO: We hard-code a grain-size limit of 200 cells per processor
     // (assuming linear finite elements and typical behavior of
-    // supercomputers), but we also do not want to coarsen by more than a
-    // factor of 8 to not create too many MPI messages.
+    // supercomputers). In case we have fewer cells on the fine level, we do
+    // not immediately go to 200 cells per rank, but limit the growth by a
+    // factor of 8, which limits makes sure that we do not create too many
+    // messages for individual MPI processes.
     unsigned int const grain_size_limit =
       std::min(200U, 8 * n_cells / n_mpi_processes_per_level.back() + 1);
 
@@ -349,13 +350,11 @@ public:
     // The vector 'partitions' contains the partition numbers. To get the
     // number of partitions, we take the infinity norm.
     n_mpi_processes_per_level.push_back(static_cast<unsigned int>(partitions.linfty_norm()) + 1);
-    n_cells_per_level.push_back(n_cells);
     return partitions;
   }
 
 private:
-  mutable std::vector<types::global_cell_index> n_cells_per_level;
-  mutable std::vector<unsigned int>             n_mpi_processes_per_level;
+  mutable std::vector<unsigned int> n_mpi_processes_per_level;
 };
 
 #else
@@ -470,9 +469,8 @@ create_geometric_coarsening_sequence(Triangulation<dim, spacedim> const & fine_t
 
   return dealii::MGTransferGlobalCoarseningTools::create_geometric_coarsening_sequence(
     fine_triangulation_in,
-    BalancedGranularityPartitionPolicy<dim>(fine_triangulation_in.n_global_active_cells(),
-                                            Utilities::MPI::n_mpi_processes(
-                                              fine_triangulation_in.get_communicator())));
+    BalancedGranularityPartitionPolicy<dim>(
+      Utilities::MPI::n_mpi_processes(fine_triangulation_in.get_communicator())));
 
 #else
 
--- a/include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
+++ b/include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
@@ -21,6 +21,7 @@
 
 // deal.II
 #include <deal.II/distributed/fully_distributed_tria.h>
+#include <deal.II/distributed/repartitioning_policy_tools.h>
 #include <deal.II/fe/fe_dgq.h>
 #include <deal.II/fe/fe_q.h>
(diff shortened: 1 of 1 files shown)


include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -420,124 +327,67 @@ create_geometric_coarsening_sequence(Triangulation<dim, spacedim> const & fine_t
   });
 
   // for a single level nothing has to be done
-  if(fine_triangulation_in.n_global_levels() > 1)
-  {
-    auto const fine_triangulation =
-      dynamic_cast<parallel::distributed::Triangulation<dim, spacedim> const *>(
-        &fine_triangulation_in);
-
-    Assert(fine_triangulation, ExcNotImplemented());
```
kronbichler, line 429: Right now, I unconditionally delete the old code; if we still want to support dealii-9.3 (the last release version), which might be a good idea, we should wrap this into `#ifdef` clauses and switch between the two variants. (Noting that the newer one is faster and takes less memory.)

nfehn: I thought dealii-9.3 is the new release? So did these changes come into dealii shortly after this year's release, or do you mean support of dealii-9.2?

peterrum: We branched off the release branch before so these features are not part of 9.3.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -420,124 +327,67 @@ create_geometric_coarsening_sequence(Triangulation<dim, spacedim> const & fine_t
...
+    // TODO: the grain size in terms of the number of cells is currently
+    // hard-coded to 200 here, plus a factor to account for the fact that we
+    // do not want to reduce the number of participating MPI ranks by more
+    // than about a factor of 8, in order to limit the number of messages a
+    // process on the coarse level might need to handle.
+    unsigned int grain_size_limit =
+      8 * temp_triangulation.n_global_active_cells() / n_processes_last;
+    RepartitioningPolicyTools::MinimalGranularityPolicy<dim> policy(
+      std::min<unsigned int>(200, grain_size_limit));
+    auto const partition = policy.partition(temp_triangulation);
```
nfehn, line 365: I think `partition` is not an intuitive name here (I am also not a fan of using `auto` in such cases as it makes the code difficult to understand; I guess the problem also originates from naming in deal.II since it is not intuitively clear to me what `policy.partition()` is doing; `partition` can be understood as noun or verb here). Here, I think of `partition` as one part of the triangulation, but it is then unclear to me why `partition.linfty_norm()` further below has something to do with the number of processes? Further below, I would guess `partition` takes the role of an mpi comm?

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -420,124 +327,67 @@ create_geometric_coarsening_sequence(Triangulation<dim, spacedim> const & fine_t
...
+    n_processes_last     = static_cast<unsigned int>(partition.linfty_norm() + 1);
+    partition.update_ghost_values();
```
nfehn, line 367: Does it make a difference if changing the order of this line with the one above? If so, this would be a side-effect that needs to be documented. If not, I would place the line with the assignment of `n_processes_last = ...` somewhere else, e.g. at the end of the for-loop body, to indicate more clearly that this is just an update required for the next iteration of the for-loop.

kronbichler:
I now added `#if` guards for the 10.0 version of deal.II and the 9.3 release (and checked both ways locally), and rewrote the code significantly: I now add a new partitioning class that can be passed on to deal.II's function.

nfehn:
You could rebase so that the checks are run for 9.3 and 10.0

kronbichler:
Alright, it seems the CI does fine on both the 10.0 and 9.3 release.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -309,6 +312,54 @@ MultigridPreconditionerBase<dim, Number>::check_levels(std::vector<MGLevelInfo>
...
+
+    // The vector 'partitions' contains the partition numbers. To get the
+    // number of partitions, we take the infinity norm.
+    n_mpi_processes_per_level.push_back(static_cast<unsigned int>(partitions.linfty_norm()) + 1);
+    n_cells_per_level.push_back(n_cells);
+    return partitions;
+  }
+
+private:
+  mutable std::vector<types::global_cell_index> n_cells_per_level;
```
nfehn, line 357: Why do you need this variable? It is private, but only filled in the function `partition()` without being used by this class or another one if I see it correctly.

kronbichler: It is not used in the final version; however, it might be that we use more clever strategies in the future, so I left it in when I derived the algorithm in the present form. If you want, we can remove it now and bring it back in case we do a more refined strategy.

nfehn: It does not help right now in my opinion.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -309,6 +312,54 @@ MultigridPreconditionerBase<dim, Number>::check_levels(std::vector<MGLevelInfo>
...
+
+  virtual LinearAlgebra::distributed::Vector<double>
+  partition(Triangulation<dim, spacedim> const & tria_coarse_in) const override
+  {
+    types::global_cell_index const n_cells = tria_coarse_in.n_global_active_cells();
+
+    // TODO: We hard-code a grain-size limit of 200 cells per processor
+    // (assuming linear finite elements and typical behavior of
+    // supercomputers), but we also do not want to coarsen by more than a
+    // factor of 8 to not create too many MPI messages.
```
nfehn, line 340: I don't understand the comment "coarsen by more than a factor of 8" completely (does it refer to the mesh or the MPI processes?). I would read the code like reducing the number of MPI processes by a factor of 8 for the next coarser level (at least as a target value). Maybe also information is missing here whether the vectors are filled from fine-to-coarse or coarse-to-fine.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
nfehn: I think you already changed this, so please resolve if this is the case.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
kronbichler: Updated comment in 4e6b0ff.

peterrum (approved):
Looks good to me :+1:

PR #94, opened 2021-06-08
Make polynomial degree an input parameter

nfehn:
I think the polynomial degree of the shape functions should become an input parameter rather than being treated as an additional, separate parameter.

 As a motivation, think of more complex elements or codes supporting different types of elements that can not be described by a single int parameter (FE_Q vs. FE_DGQ, etc.). Moreover, the interface for the spatial discretization operator simplifies. Finally, when using solvers from ExaDG in a multiphysics environment with coupling to other problems, there should be one data structure that collects all `ExaDG::Poisson` (or `ExaDG::IncNS`) parameters (the fact that we allow loops over h and k as a simple way of conducting mesh convergence studies is something specific to `ExaDG`).

This PR is work in progress. Currently, the intended design is applied to the `Poisson` module. If we agree on that, I would apply it to the rest of `ExaDG`.

### DIFF
--- a/applications/poisson/gaussian/tests/gaussian.output
+++ b/applications/poisson/gaussian/tests/gaussian.output
@@ -27,6 +27,7 @@ Spatial Discretization:
   Triangulation type:                        Distributed
   Mapping:                                   Isoparametric
   Element type:                              DG
+  Polynomial degree:                         1
   IP factor:                                 1.0000e+00
 
 Solver:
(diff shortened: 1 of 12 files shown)


kronbichler (approved):

nfehn:
@kronbichler I start to continue now ... :)

nfehn:
@kronbichler @peterrum I think all modules are updated now. Ready to review again from my side.

peterrum (approved):
Looks good to me!

kronbichler (approved):

PR #105, opened 2021-12-03
Refactor time step calculation and related interfaces

### LINKED ISSUE #101: Shift parts of time step calculation from TimeIntegrator to SpatialOperator
`ConvDiff::TimeIntBDF::calculate_time_step_size()` queries `pde_operator->get_polynomial_degree()` (the latter function exists only for that purpose)

https://github.com/exadg/exadg/blob/2eabd006f482001b22251632b3c97644a9918706/include/exadg/convection_diffusion/time_integration/time_int_bdf.cpp#L274

However, since this is information is related to the spatial operator, it would make sense to shift parts of the implementation related to the time step calculation to the spatial operator.

Note that this is already done in case of adaptive time stepping (local CFL condition), but not in the case which evaluates the global CFL condition.

This means that the PDE operator needs to implement a method `Operator::calculate_time_step_cfl_global(cfl, max_velocity, param.exponent_fe_degree_convection)` which queries the polynomial degree via `param.degree` and computes the minimal element length via `this->calculate_minimum_element_length();`. This function then calls the free function `calculate_time_step_cfl_global(cfl, max_velocity, h_min, degree, param.exponent_fe_degree_convection)`.

This affects the `ConvDiff`, `IncNS`, and `CompNS` modules.

nfehn:
closes #101

Single-responsibility is now clearer between temporal discretization (`TimeIntegration`) and spatial discretization (`Operator`).

### DIFF
--- a/include/exadg/incompressible_navier_stokes/postprocessor/output_generator.cpp
+++ b/include/exadg/incompressible_navier_stokes/postprocessor/output_generator.cpp
@@ -452,11 +452,9 @@ OutputGenerator<dim, Number>::calculate_additional_fields(VectorType const & vel
 
     if(output_data.write_cfl)
     {
-      auto param = navier_stokes_operator->get_param();
-      auto time_step_size =
-        navier_stokes_operator->calculate_time_step_cfl(velocity,
-                                                        param.cfl,
-                                                        param.cfl_exponent_fe_degree_velocity);
+      auto   param          = navier_stokes_operator->get_param();
+      double time_step_size = navier_stokes_operator->calculate_time_step_cfl(velocity);
+      time_step_size *= param.cfl;
 
       navier_stokes_operator->calculate_cfl_from_time_step(cfl, velocity, time_step_size);
     }
--- a/include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
+++ b/include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
@@ -936,9 +936,9 @@ template<int dim, typename Number>
 double
 SpatialOperatorBase<dim, Number>::calculate_characteristic_element_length() const
 {
-  double minimum_element_length = calculate_minimum_element_length();
+  double const h_min = calculate_minimum_element_length();
 
-  return ExaDG::calculate_characteristic_element_length(minimum_element_length, param.degree_u);
+  return ExaDG::calculate_characteristic_element_length(h_min, param.degree_u);
 }
 
 template<int dim, typename Number>
--- a/include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
+++ b/include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
@@ -740,15 +740,6 @@ SpatialOperatorBase<dim, Number>::get_viscosity_boundary_face(unsigned int const
   return viscosity;
 }
 
-// Polynomial degree required for CFL condition, e.g., CFL_k = CFL / k^{exp}.
-// TODO remove this function
-template<int dim, typename Number>
-unsigned int
-SpatialOperatorBase<dim, Number>::get_polynomial_degree() const
-{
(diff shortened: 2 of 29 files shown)


include/exadg/incompressible_navier_stokes/postprocessor/output_generator.cpp
```diff
@@ -452,11 +452,9 @@ OutputGenerator<dim, Number>::calculate_additional_fields(VectorType const & vel
 
     if(output_data.write_cfl)
     {
-      auto param = navier_stokes_operator->get_param();
-      auto time_step_size =
-        navier_stokes_operator->calculate_time_step_cfl(velocity,
-                                                        param.cfl,
-                                                        param.cfl_exponent_fe_degree_velocity);
+      auto   param          = navier_stokes_operator->get_param();
```
peterrum, line 455: Should this be a reference?

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
```diff
@@ -922,6 +932,14 @@ SpatialOperatorBase<dim, Number>::calculate_cfl_from_time_step(VectorType &
                              param.cfl_exponent_fe_degree_velocity);
 }
 
+template<int dim, typename Number>
+double
+SpatialOperatorBase<dim, Number>::calculate_characteristic_element_length() const
+{
+  double minimum_element_length = calculate_minimum_element_length();
```
peterrum, line 939:
```suggestion
  double const minimum_element_length = calculate_minimum_element_length();
```

peterrum (approved):

include/exadg/incompressible_navier_stokes/postprocessor/output_generator.cpp
nfehn: I think this is ok. 

I see another problem here. The result is not correct if one specifies temporal refinements in the input file. So we should find another solution for this.
@bergbauer  A relative CFL number should be enough/sufficient for this kind of result? So we could calculate the time step size with `cfl=1`, and afterwards compute the dof-vector of cfl-values, which takes `cfl=1` in one cell and smaller values in other cells. I guess we could then remove the getter `navier_stokes_operator->get_garam()`.

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
nfehn: see https://github.com/exadg/exadg/pull/105/commits/ae02c1633268b74cda449071598d088a154ad802

include/exadg/incompressible_navier_stokes/postprocessor/output_generator.cpp
nfehn: ... should be done in a separate PR.

bergbauer: The intention of this output is to see where in the geometry/mesh the CFL condition restricts our time step. So yes, a relative measure is sufficient but should be named appropriately.

peterrum (approved):

include/exadg/incompressible_navier_stokes/postprocessor/output_generator.cpp
kronbichler: Regarding my understanding of
> Should this be a reference?

I think that the copy is expensive (in relative terms, it is "only" several dozen of instructions), compared to the single variable `cfl` that you want to retrieve. I would personally prefer to simply write `time_step_size *= navier_stokes_operator->get_param();` in the use in line 457 below. But I guess this will be refactored anyway, so let's keep it for now.

kronbichler (approved):

include/exadg/incompressible_navier_stokes/postprocessor/output_generator.cpp
nfehn: yes, see #108

PR #106, opened 2021-12-03
refactor design of ApplicationBase

nfehn:
This PR proposes to shift ownership of certain objects from the class `Driver` to the class `ApplicationBase`. With this new design, interfaces and responsibilities should be clearer now. For example, the interface of `create_postprocessor()` is simpler now, 

It is currently done for the `CompNS` module, and within this module for the "couette" test case only (other test cases are currently deactivated). If we agree on this design, I could apply this to the rest of ExaDG.


include/exadg/compressible_navier_stokes/user_interface/application_base.h
```diff
@@ -60,39 +61,76 @@ class ApplicationBase
...
     n_subdivisions_1d_hypercube = n_subdivisions_1d;
   }
 
+  MPI_Comm const & mpi_comm;
+
+  InputParameters                          parameters;
+  std::shared_ptr<BoundaryDescriptor<dim>> boundary_descriptor;
+  std::shared_ptr<FieldFunctions<dim>>     field_functions;
+  GridData                                 grid_data;
+  std::shared_ptr<Grid<dim, Number>>       grid;
```
peterrum, line 130: I would make these protected so that the derived classes can access them directly but not the driver. The driver could get these via getter functions.

peterrum (commented):
What this PR does is to move the classes that are set up by the user/application (i.e., InputParameters, BoundaryDescriptor, FieldFunctions, Grid) into the application base. This implies that the applications are not responsible to fill these classes of the driver, but to provide these classes to the driver. I am fine with this change, although I also liked that the applications were stateless. Let's go with the new approach. I don't see any issues with the concept for now.

nfehn:
I am not sure about how the optimal design looks like. We should have an open discussion about this and think about coupled problems (volume coupled with one triangulation but several fields; or surface coupled like FSI) and how to realize the `Application` class in a modular way (the current design is not optimal in my opinion):
- One application class that aggregates several single-field application classes (e.g. `app.fluid.parameters` and `app.solid.set_boundary_descriptor()`). 
- Or one application class that adds all the low level data structure for the different fields (e.g. `app.parameters_fluid` and `app.set_boundary_descriptor_solid()`).

nfehn:
I refined the changes from last week. I forgot that `Grid` should be a member of `Driver` (due to ALE functionality that needs to attach a moving grid to the `Grid` class). I think we should not change this as compared to the current master branch since this gives us increased flexibility.

With these changes, this PR essentially shifts `InputParameters`, `BoundaryDescriptor`, and `FieldFunctions` from the `Driver` to the `Application`. Related to the changes introduced in #94, the code proposed here is now clearer in terms of which class is responsible for setting input parameters.

kronbichler (approved):

nfehn:
I have now updated all modules and test cases in ExaDG.

nfehn:
Can I mege this one?

include/exadg/compressible_navier_stokes/user_interface/implement_get_application.h
```diff
@@ -0,0 +1,43 @@
...
+{
+template<int dim, typename Number>
+std::shared_ptr<ApplicationBase<dim, Number>>
+get_application(std::string input_file, MPI_Comm const & comm)
+{
+  return std::make_shared<Application<dim, Number>>(input_file, comm);
+}
+
+} // namespace CompNS
+} // namespace ExaDG
```
peterrum, line 39: Could you explain the purpose of these `declare_get_application.h` and `implement_get_application.h`. Is the only purpose so that one does not have to write `std::make_shared<Application<dim, Number>>(input_file, comm)` directly? But one has to include this fill in all applications.

nfehn: Yes, I would have to touch this function for all the applications whenever the interface changes. So I decided to implement it only once and include it in all applications (technically, this is possible since we have separate executables for each application and since each application can therefore have the same name `Application`, which is different from the "ApplicationSelector" concept with different name for each application which we had some time ago). I know the code is only a single line of implementation, but I came to the conclusion that being more strict in terms of "don't repeat yourself" makes sense here.

peterrum: I won't block this change, since the general structure of `get_application()` is not new.

However, I don't see how this helps to reduce code duplication. Currently, there are 7 `declare_get_application.h` and 7 `implement_get_application.h` files and `implement_get_application.h` has to be included in all applications. My question is why does someone has to return a shared pointer of the application. Why isn't it enough to let the user provide an application class and let the solver/drive create a shared pointer of it.

nfehn: 7 is not much if you count - in comparison - the number of all applications (which I would need to change due to the additional mpi_comm parameter introduced in this PR).

But the driver/solver does not know about the concrete application. Principally, one could let the driver instantiate the application, but this would force the user to always name a particular application `Application`. note that `solver.h` (and also `driver.h`) only contains a forward declaration with `ApplicationBase`! My suggestion is much more flexible, as it allows to name the application whatever the user likes best, e.g., `MyFancyApplication`. In this case, the user than just has to instantiate the application manually instead of including the new `implement_get_application()`. I argue that such a user knows what to do. However, what I want is to minimize the effort for unexperienced users. Adding another switch-branch to the `ApplicationSelector` as we had some time ago is already too complex for a new user that just wants to experiment with ExaDG in my opinion.

peterrum: Let's go with the current approach. Maybe there will be cases that one want to have multiple fluid applications and in that case they need different names or have to be in different namespaces.

Issue #104, opened 2021-12-02
Remove GeometryInfo as preparation for simplex support


nfehn:

peterrum:
Remaining instances of `GeometryInfo` in `include/` are:

https://github.com/exadg/exadg/blob/35a791d68c7e05e127302e9c82edc9890506887b/include/exadg/incompressible_navier_stokes/postprocessor/line_plot_calculation_statistics_homogeneous.cpp#L170

https://github.com/exadg/exadg/blob/35a791d68c7e05e127302e9c82edc9890506887b/include/exadg/incompressible_navier_stokes/postprocessor/line_plot_calculation_statistics_homogeneous.cpp#L238

https://github.com/exadg/exadg/blob/35a791d68c7e05e127302e9c82edc9890506887b/include/exadg/incompressible_navier_stokes/postprocessor/line_plot_calculation_statistics_homogeneous.cpp#L290

https://github.com/exadg/exadg/blob/35a791d68c7e05e127302e9c82edc9890506887b/include/exadg/vector_tools/interpolate_solution.h#L67

https://github.com/exadg/exadg/blob/35a791d68c7e05e127302e9c82edc9890506887b/include/exadg/vector_tools/point_value.h#L47

since these two functions are not supported by `ReferenceCell`. The usage of these function calls could be eliminated by the usage of the functions `VectorTools::point_values()` as proposed by https://github.com/exadg/exadg/pull/107#issuecomment-986811792.

nfehn:
deal.II issue https://github.com/dealii/dealii/issues/13308 should be resolved first.

nfehn:
Note that there are other occurrences of `GeometryInfo` in the `applications` directory.

PR #107, opened 2021-12-04
Remove most instance of GeometryInfo within include/

peterrum:
references #104

Remaining instances are:

https://github.com/exadg/exadg/blob/35a791d68c7e05e127302e9c82edc9890506887b/include/exadg/incompressible_navier_stokes/postprocessor/line_plot_calculation_statistics_homogeneous.cpp#L170

https://github.com/exadg/exadg/blob/35a791d68c7e05e127302e9c82edc9890506887b/include/exadg/incompressible_navier_stokes/postprocessor/line_plot_calculation_statistics_homogeneous.cpp#L238

https://github.com/exadg/exadg/blob/35a791d68c7e05e127302e9c82edc9890506887b/include/exadg/incompressible_navier_stokes/postprocessor/line_plot_calculation_statistics_homogeneous.cpp#L290

https://github.com/exadg/exadg/blob/35a791d68c7e05e127302e9c82edc9890506887b/include/exadg/vector_tools/interpolate_solution.h#L67

https://github.com/exadg/exadg/blob/35a791d68c7e05e127302e9c82edc9890506887b/include/exadg/vector_tools/point_value.h#L47

since these two functions are not supported by `ReferenceCell`.

### DIFF
--- a/include/exadg/functions_and_boundary_conditions/interface_coupling.h
+++ b/include/exadg/functions_and_boundary_conditions/interface_coupling.h
@@ -458,13 +458,13 @@ public:
     {
       if(!cell->is_artificial() && cell->at_boundary())
       {
-        for(unsigned int f = 0; f < GeometryInfo<dim>::faces_per_cell; ++f)
+        for(const unsigned int f : cell->face_indices())
         {
           if(cell->face(f)->at_boundary())
           {
             if(map_bc.find(cell->face(f)->boundary_id()) != map_bc.end())
             {
-              for(unsigned int v = 0; v < GeometryInfo<dim - 1>::vertices_per_cell; ++v)
+              for(const unsigned int v : cell->face(f)->vertex_indices())
               {
                 marked_vertices[cell->face(f)->vertex_index(v)] = true;
               }
--- a/include/exadg/functions_and_boundary_conditions/verify_boundary_conditions.h
+++ b/include/exadg/functions_and_boundary_conditions/verify_boundary_conditions.h
@@ -50,7 +50,7 @@ verify_boundary_conditions(BoundaryDescriptor const & boundary_descriptor,
   // Make sure that each boundary face has exactly one boundary type
   for(auto cell : *grid.triangulation)
   {
-    for(unsigned int f = 0; f < GeometryInfo<dim>::faces_per_cell; ++f)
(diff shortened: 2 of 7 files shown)


nfehn (approved):

nfehn:
One aspect is unclear to me: Are the remaining instances a prerequisite for simplex support? I.e., do we have to address them as a preparation for simplex support according to #104?

If so, how we would solve this? Should we add asserts to the code?

kronbichler (approved):

kronbichler:
I don't think they are strictly needed. There is no standard solution for that in deal.II. Some use cases look like we could replace them by https://dealii.org/developer/doxygen/deal.II/namespaceVectorTools.html#a1ad6eceb6cbeaa505baf7f938289bbde or https://dealii.org/developer/doxygen/deal.II/namespaceVectorTools.html#aada37fd09d8be1b9ed9bb12d698f8ee4 (or the variant with the cache), so that might be some opportunity in the future (not for this PR).

nfehn:
But those routines would currently crash when run with a simplex mesh?

I would be happy if you, @peterrum or @kronbichler, could update issue #104 (maybe even change the title of the issue).

peterrum:
> I don't think they are strictly needed. There is no standard solution for that in deal.II. Some use cases look like we could replace them by https://dealii.org/developer/doxygen/deal.II/namespaceVectorTools.html#a1ad6eceb6cbeaa505baf7f938289bbde or https://dealii.org/developer/doxygen/deal.II/namespaceVectorTools.html#aada37fd09d8be1b9ed9bb12d698f8ee4 (or the variant with the cache), so that might be some opportunity in the future (not for this PR).

Indeed the use case of the function that are calling these GeometryInfo functions seem to be very similar to the use case of the VectorTool functions.

> But those routines would currently crash when run with a simplex mesh?

Yes.

nfehn:
Another solution to #104 would be to add asserts to the code that ensure that the mesh has quad/hex elements only? If so, let's add this information also to issue #104 (and we can merge this PR).

peterrum:
> Another solution to #104 would be to add asserts to the code that ensure that the mesh has quad/hex elements only? If so, let's add this information also to issue #104 (and we can merge this PR).

Done.

PR #110, opened 2021-12-07
Add poisson/sine_simplex test case

peterrum:
depends on https://github.com/dealii/dealii/pull/13042


nfehn:
@peterrum Thanks a lot for the initiative!

I think it is not justified to copy one test case with the only difference being the mesh. We should introduce this as a variant of the existing test case.

This brings me to my next point. I think the if-clause checking for quad/hex meshes is not really understandable. Simplex is not the else-branch of a all-hex mesh in my opinion?! ExaDG's philosophy is to introduce a layer between deal.II's low level functionalitites and high-level application code. This means that we need to introduce a new item of `InputParameter` that decides about the type of mesh/elements (which for example can be used in the function `create_grid()` to choose the correct branch for grid generation (hex vs. simplex): e.g.

```
enum class ElementType
{
  Tensor,
  Simplex
};
```

This way, we can easily assert in `InputParameters::check_input_parameters()` which functionality can currently not be used along with simplices, e.g. `triangulation_type == TriangulationType::Serial` or `preconditioner == Preconditioner::None` etc.. This is not possible in the current form since  `InputParameters::check_input_parameters()` does not have information about the triangulation and whether all reference cells are hyper cubes.

@peterrum I am not sure if you want to make the changes described here. So I could offer that I take the code you presented here and that I realize the changes suggested above. Just let me know what you prefer. In any case, your PR is very helpful.

peterrum:
> I think it is not justified to copy one test case with the only difference being the mesh. We should introduce this as a variant of the existing test case.

I did understand your comment on Monday the way that we should have a test case for meshes with simplices and one for external meshes. Apart from what I thought I did understand, I think it would be an option worth considering to have these cases separately so that we can point users to there very simple cases dedicated to these aspects. 

> This brings me to my next point. I think the if-clause checking for quad/hex meshes is not really understandable. Simplex is not the else-branch of a all-hex mesh in my opinion?! ExaDG's philosophy is to introduce a layer between deal.II's low level functionalitites and high-level application code. This means that we need to introduce a new item of InputParameter that decides about the type of mesh/elements (which for example can be used in the function create_grid() to choose the correct branch for grid generation (hex vs. simplex): e.g.
enum class ElementType
{
  Tensor,
  Simplex
};
This way, we can easily assert in InputParameters::check_input_parameters() which functionality can currently not be used along with simplices, e.g. triangulation_type == TriangulationType::Serial or preconditioner == Preconditioner::None etc.. This is not possible in the current form since InputParameters::check_input_parameters() does not have information about the triangulation and whether all reference cells are hyper cubes.

I agree the opposite of hex-mesh is not simplex-mesh. However, I don't completely understand what you suggest. On application level (for applications that should work for hex-and simplex-meshes), one could consider to introduce an application-specific parameter. However, this parameter shouldn't be visible to the core of the library. The core already gets the Grid class (consisting of Triangulation/Mapping), which encodes the information if we have hex-, simplex-, or mixed meshes. Adding an additional parameter is error-prone and could lead to the possibility that Grid and parameter gets out of sync.

I would suggest to adjust the code like:
```cpp
      if(this->grid->triangulation->all_reference_cells_are_hyper_cube())
        fe = std::make_shared<FE_DGQ<dim>>(param.degree);
      else if(this->grid->triangulation->all_reference_cells_are_simplex())
        fe = std::make_shared<FE_SimplexDGP<dim>>(param.degree);
      else
        AssertThrow(false, ExcNotImplemented());
```
See also https://github.com/dealii/dealii/pull/13044.

`check_input_parameters()` is a problem. But what about moving the check after the generation of the `Grid` object?

> I am not sure if you want to make the changes described here. So I could offer that I take the code you presented here and that I realize the changes suggested above. 

I could do smaller adjustments. But if you want bigger changes, I would appreciate if you could do these, since I am not that familiar with the parameters and all the details in ExaDG. But be aware that you need https://github.com/dealii/dealii/pull/13042; hopefully this PR is merged soon!

include/exadg/poisson/spatial_discretization/operator.cpp
```diff
@@ -98,7 +119,8 @@ Operator<dim, Number, n_components>::distribute_dofs()
 
   dof_handler.distribute_dofs(*fe);
 
-  dof_handler.distribute_mg_dofs();
+  if(this->grid->triangulation->all_reference_cells_are_hyper_cube())
+    dof_handler.distribute_mg_dofs();
```
nfehn, line 123: This is very difficult to understand for non-dealii principal developers. What is the background?

include/exadg/grid/enum_types.h
```diff
@@ -39,7 +39,8 @@ enum class TriangulationType
 {
   Undefined,
   Distributed,
-  FullyDistributed
+  FullyDistributed,
+  Serial
```
nfehn, line 43: Why did you introduce this here? Can you please explain this in detail?

nfehn:
> I did understand your comment on Monday the way that we should have a test case for meshes with simplices and one for external meshes.

On Monday I was not sure whether we should have a separate test case or include it in existing ones (I think I mentioned both options). In light of PR #106, I am convinced that we should not produce so much code if it is mostly copied (e.g. you did not remove Cartesian/Curvilinear in this test case but added an assert for the curvilinear case; this is not understandable in my opinion when considering this test case stand alone).

include/exadg/poisson/spatial_discretization/operator.cpp
peterrum: `DoFHandler::distribute_mg_dofs()` is not working for simplex meshes. To be able to use multigrid, one needs to use global coarsening for now.

include/exadg/grid/enum_types.h
peterrum: Because there was enough to think about to get simplex running.

nfehn: What do you mean? Is Serial mandatory for simplex? Or is this change not directly related to this PR dealing with simplex?

include/exadg/poisson/spatial_discretization/operator.cpp
nfehn: Thanks. This implies that we need more asserts for the simplex branch related to which preconditioners can be used (with additional explanations). The reasons behind can not be deduced from the above if statement and the above if would still allow to select „standard dealii multigrid“ for simplex, which would then probably crash at some point.

@kronbichler as mentioned elsewhere in the past (and not directly related to this PR), distribute_mg_dofs() is a problem in terms of design and we need to discuss this in a broader context in my opinion.

include/exadg/grid/enum_types.h
peterrum: > Is Serial mandatory for simplex?

No. It can be extracted into separate PR.

nfehn:
> I agree the opposite of hex-mesh is not simplex-mesh. However, I don't completely understand what you suggest. On application level (for applications that should work for hex-and simplex-meshes), one could consider to introduce an application-specific parameter. However, this parameter shouldn't be visible to the core of the library. The core already gets the Grid class (consisting of Triangulation/Mapping), which encodes the information if we have hex-, simplex-, or mixed meshes. Adding an additional parameter is error-prone and could lead to the possibility that Grid and parameter gets out of sync.

I understand your point with redundancy of parameters. I would probably also accept your new solution with else-if for simplex and an else branch with Assert(false, ...). However, I think it is a useful concept in ExaDG to not try to deduce all the information from dealii data structures. For example, we need more asserts for the Poisson-Simplex case than we have currently, but we cannot expect that the triangulation is available at those places where the assert is needed in order to query such information. So for a final assessment I would need to see whether approach is really suitable to cover all necessary simplex-asserts in ExaDG.

include/exadg/poisson/spatial_discretization/operator.cpp
kronbichler: Regarding `distribute_mg_dofs`: I agree that this is annoying and that it would be good to be able to eliminate it. From an application point of view, one could argue that we already have the ingredients to achieve this, as we might simply switch all solvers to global coarsening. One could achieve similar (or the same) partitioning with that setting, so the parallel scalability of the solver would be the same. Furthermore, one could pull over the optimizations from the local smoothing in terms of data access to global coarsening, eliminating any possible performance penalty still visible for the latter. (I have not measured lately, but maybe @peterrum has.) The setup would remain somewhat more expensive due to the need to create additional triangulations, but for ExaDG this is essentially a one-time cost. Also the increase in memory consumption should be tolerable in general, compared to other data structures we are using. (The scalar Poisson and convection-diffusion modules might be the only really seeing a noticeable effect.)

My main concern at this stage is how well it is possible to hide global coarsening data structures in a complex application scenario: What about the case where multiple fields want to create a multigrid hierarchy. Can we collect all information in the `Grid` class or, more generally, in a central data structure, such that we avoid going through these steps more than once? This is because my reasoning above regarding costs relies on the fact that we might not do this more than once per grid. Think about dynamic adaptivity where we change the mesh more often.

Opinions?

nfehn: For the standard dealii multigrid approach (I mean local smoothing), can we be more conservative in the sense that `distribute_mg_dofs()` is only done in multigrid and that some data structures need to be set up twice in the worst case? What about the performance penalty for this case?

nfehn: > Can we collect all information in the Grid class ...?

The question is whether we then simply shift the problem? We would then have multigrid functionality/data structures in a class (`Grid`) that is used very generally and not just for multigrid.

kronbichler: Thinking about this, I see two alternatives: Either we call `dof_handler.distribute_mg_dofs` inside the multigrid solver (possibly doing `const_cast<DoFHandler&>` on the object, to only call it on demand and at the place where we need it. The other alternative is to set up an additional `DoFHandler` on which we can call `distribute_mg_dofs`, keeping everything isolated to the multigrid solver. For the latter, the perspective of `DoFHandler` has the potential problem that we might have a renumbering on the original `DoFHandler`, for which I do not see a simple way right now to make sure they match. We do not do any renumbering at this point, but we might possibly do in the future and should keep responsibilities clear.

From a performance perspective in the solver classes, there would not be any penalty in this case against what we have now. The second option would pay a bit in terms of storage cost of the additional `DoFHandler`.

nfehn: I would prefer such an approach. Our multigrid implementation has the option `update()`, which updates the multigrid preconditioner in case something has changed (e.g. parameters, mesh motion, etc.). Maybe the problem that you mention would fit into this logic.

kronbichler: > >  Can we collect all information in the Grid class ...?
> 
> The question is whether we then simply shift the problem? We would then have multigrid functionality/data structures in a class (Grid) that is used very generally and not just for multigrid.

I agree, it does put a burden on that class and leads to questions about the responsibility. However, this is in conflict with the desire to expose information created for a particular linear solver (regarding the sharing of say a full multigrid hierarchy created from a single triangulation) to another solver. Hence, I see that we need to decide how to balance the conflict between the level of separation and performance/memory costs associated to repeatedly do the same thing. In consequence, we would need to decide who would keep track of such "hidden states" going from one solver to the other.

kronbichler: > I would prefer such an approach.

Let me have a look at it later today or on Monday, when I've finished some other topics.

nfehn: > I agree, it does put a burden on that class and leads to questions about the responsibility. However, this is in conflict with the desire to expose information created for a particular linear solver (regarding the sharing of say a full multigrid hierarchy created from a single triangulation) to another solver. Hence, I see that we need to decide how to balance the conflict between the level of separation and performance/memory costs associated to repeatedly do the same thing. In consequence, we would need to decide who would keep track of such "hidden states" going from one solver to the other.

It would be acceptable in the grid class if it really works for all cases, but not if we would need if-else-branches depending on other parameters of ExaDG. I don't have the overview of which type of coarsening works for which types of dealii-triagnulations and which types of elements ..

include/exadg/grid/enum_types.h
nfehn: has been done in #116

include/exadg/poisson/spatial_discretization/operator.cpp
nfehn: Regarding `distribute_mg_dofs()`: @kronbichler, the first option you described previously

> Either we call dof_handler.distribute_mg_dofs inside the multigrid solver (possibly doing const_cast<DoFHandler&> on the object), to only call it on demand and at the place where we need it.

seems to be most straight-forward to realize. Could you make a suggestion? I could also help you. However, I now want to realize the simplex-support for Poisson and I currently ask myself how to best solve the `distribute_mg_dofs()` problem (so it would be helpful if we could make progress here).

nfehn: After `distribute_mg_dofs()` is called, the code deals with affine constraints (`DoFTools::make_zero_boundary_constraints()`). In order to shift `distribute_mg_dofs()` to the multigrid preconditioner (which is set up later in the code), we have to make sure that other parts of the code (such as the boundary conditions mentioned above) do not rely on the call `distribute_mg_dofs()`.

nfehn:
@necioglu have we now addressed everything in this PR? If so, we could close this one.

PR #111, opened 2021-12-07
Do not store FEEvaluation objects in OperatorBase

kronbichler:
Related to the simplex example of #110, I think we should work towards avoiding storing a copy of `FEEvalation` and `FEFaceEvaluation` as class members. These classes are designed to operator in a "stateless" scenario, i.e., to be created at the place where `MatrixFree` allows us to enter during `loop/cell_loop`, and released at the end of this loop. For mixed meshes in the future, we would need to add the `cell_range` or `face_range` to the constructor to be able to select the right quadrature formula.

I think we should discuss in which direction we would want to go here. I make a first suggestion with minimal changes to the `OperatorBase` class, where I only move the construction into the various functions. I also adapted the `LaplaceOperator` application to show how the realization for derived classes would like like, but I did not touch the other classes yet. So ExaDG does not compile at this point.

Things get more messy in some of the non-linear operators or for the convective operator, which need a second `FEEvaluation`/`CellIntegrator` for evaluating the linearization point/velocity. The current proposed solution does not offer a good solution for those, so we would need to set up a class member for those, to be able to isolate the `reinit_cell` function from the inner workings (`do_cell_integral`). The longer-term solution for those might be to put all data structures needed for evaluation of the individual classes into `do_****_integral` into a separate data structure that we set up via `virtual OperatorBase::InternalDataBase OperatorBase::create_evaluation_instance() const`. It creates one additional indirection, so I do not want to test it before we reach consensus here.

Another goal we should discuss at this stage of refactoring is that we might want to switch to a point-wise evaluation setup at some point, i.e., to only implement the operation at a point, https://github.com/exadg/exadg/blob/22330b644cab6a2686ddaee5b212873240c85402/include/exadg/poisson/spatial_discretization/laplace_operator.cpp#L124 , and not spell out the the loop over quadrature points any more (so have deal.II run that internally). `FEEvaluation` does not yet support it, and I actually believe that the right class would be something like an `Iterator` class on top of `FEEvaluation`, so there is no clear path yet. Still, this is something we would need for GPU support in the future and might also be beneficial for more performance-optimized versions of the evaluation e.g. in cell-based iterative solvers.

### DIFF
--- a/include/exadg/operators/operator_base.cpp
+++ b/include/exadg/operators/operator_base.cpp
@@ -69,18 +69,6 @@ OperatorBase<dim, Number, n_components>::reinit(MatrixFree<dim, Number> const & 
   // in 3D, and thus necessarily has dofs_per_vertex=0
   is_dg = (this->matrix_free->get_dof_handler(this->data.dof_index).get_fe().dofs_per_vertex == 0);
 
-  integrator   = std::make_shared<IntegratorCell>(*this->matrix_free,
-                                                this->data.dof_index,
-                                                this->data.quad_index);
-  integrator_m = std::make_shared<IntegratorFace>(*this->matrix_free,
-                                                  true,
-                                                  this->data.dof_index,
-                                                  this->data.quad_index);
-  integrator_p = std::make_shared<IntegratorFace>(*this->matrix_free,
-                                                  false,
-                                                  this->data.dof_index,
-                                                  this->data.quad_index);
-
   if(!is_dg)
   {
     constrained_indices.clear();
@@ -423,7 +411,7 @@ OperatorBase<dim, Number, n_components>::add_diagonal(VectorType & diagonal) con
       [&](auto & integrator) -> void {
         // TODO this line is currently needed as bugfix, but should be
         // removed because reinit is now done twice
-        this->reinit_cell(integrator.get_current_cell_index());
+        this->reinit_cell(integrator.get_current_cell_index(), integrator);
 
         integrator.evaluate(integrator_flags.cell_evaluate.value,
                             integrator_flags.cell_evaluate.gradient,
@@ -594,26 +582,29 @@ OperatorBase<dim, Number, n_components>::apply_add_block_diagonal_elementwise(
   VectorizedArray<Number> const * const src,
   unsigned int const                    problem_size) const
 {
+  IntegratorCell integrator(*this->matrix_free, data.dof_index, data.quad_index);
+  IntegratorFace integrator_m(*this->matrix_free, true, data.dof_index, data.quad_index);
+  IntegratorFace integrator_p(*this->matrix_free, false, data.dof_index, data.quad_index);
   (void)problem_size;
 
   AssertThrow(is_dg, ExcMessage("Block Jacobi only implemented for DG!"));
 
-  this->reinit_cell(cell);
+  this->reinit_cell(cell, integrator);
 
-  for(unsigned int i = 0; i < integrator->dofs_per_cell; ++i)
-    integrator->begin_dof_values()[i] = src[i];
+  for(unsigned int i = 0; i < integrator.dofs_per_cell; ++i)
+    integrator.begin_dof_values()[i] = src[i];
 
-  integrator->evaluate(integrator_flags.cell_evaluate.value,
-                       integrator_flags.cell_evaluate.gradient,
-                       integrator_flags.cell_evaluate.hessian);
+  integrator.evaluate(integrator_flags.cell_evaluate.value,
+                      integrator_flags.cell_evaluate.gradient,
+                      integrator_flags.cell_evaluate.hessian);
 
-  this->do_cell_integral(*integrator);
+  this->do_cell_integral(integrator);
 
-  integrator->integrate(integrator_flags.cell_integrate.value,
-                        integrator_flags.cell_integrate.gradient);
+  integrator.integrate(integrator_flags.cell_integrate.value,
+                       integrator_flags.cell_integrate.gradient);
 
-  for(unsigned int i = 0; i < integrator->dofs_per_cell; ++i)
-    dst[i] += integrator->begin_dof_values()[i];
+  for(unsigned int i = 0; i < integrator.dofs_per_cell; ++i)
+    dst[i] += integrator.begin_dof_values()[i];
 
   if(is_dg && evaluate_face_integrals())
   {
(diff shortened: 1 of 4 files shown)


nfehn:
With this PR, the day has come where my wishes have been granted!

The problem is exactly the case with additional `FEEvaluation` objects (like in linearized operators; or convective operators with transport velocity). Those cases would require that the number of parameters in the interfaces like `reinit_cell()` are variable so that we do not have to store the objects as member variables. I would have done it in the past, if I had had a solution to this. 

In the following, I just think out loud ... I think we need an additional software layer (similar to `OperatorBase` but more intelligent, let's call it `MatrixFreeEvaluator` in the following). The question is: Can we provide a vector of `FEEvaluationData/IntegratorData` (collects dof-index, quad-index, and integrator-flags for each item) to `MatrixFreeEvaluator`, so that this class instantiates objects of `FEEvaluation/Integrator` when looping through the mesh according to this list/vector. Indices should be encoded by strings as we currently do this in `ExaDG::MatrixFreeData` in order to avoid hard-coded indices. The concrete implementation of an operator has to follow an interface, so that `MatrixFreeEvaluator` (holding a pointer to the operator) can call functions like `operator->reinit_cell()`, `operator->do_cell_integral()` (without parameters!). Inside the quadrature point or flux computations, the operator accesses the different `FEEvaluation/Integrator` objects e.g. by `matrix_free->get_integrator("temperature").get_gradient(q)` and `matrix_free->get_integrator("velocity").get_value(q)` (instead of `matrix_free->get_integrator(0).get_gradient(q)` and `matrix_free->get_integrator(1).get_value(q)`). Does this make sense? (@kronbichler I do not fully understand your paragraph ending with "one additional indirection").

Developments towards quadrature-point-based evaluations make sense in my opinion, see the "kernel" design in ExaDG. However, this might actually be a different topic? If I see it correctly we are already very close to such a design by our functions `do_cell_integral()`, `do_face_integral()`? We would have to write the for-loops outside the function and just add `q` as another parameter to the function. I think this should be realizable more easily and could be done in ExaDG independently of the more difficult problem of handling the `FEEvaluation/Integrator` objects.

nfehn:
I am not sure if it makes sense to merge a partial solution into ExaDG before we have a concrete plan of how to deal with the most difficult cases.

kronbichler:
The quadrature stuff is mostly independent indeed. The reason I mention it here is that we should think about that case when we redesign interfaces, in order to avoid going down the wrong direction.

> I am not sure if it makes sense to merge a partial solution into ExaDG before we have a concrete plan of how to deal with the most difficult cases.

No, I would think we should aim for a solution before we merge. I see this PR as a first shoot to initiate the discussion.

nfehn:
> However, this class would not store the actual FEEvaluation classes but rather return a wrapper around a vector/map of FEEvaluation object for the current cell batch. The wrapper class could be asked for the actual FEEvaluations via the strings.

So which class would store the `FEEvaluation` objects?

peterrum:
> So which class would store the FEEvaluation objects?

What I wanted to say that there need to be two objects. The object that is owned by the class (e.g. `OperatorBase`) only has the recipe how to create `FEEvaluation` objects. The actual creation of the object that wraps the vector of `FEEvaluation` objects happens inside of the matrix-free loop.

nfehn:
I thought once more about this topic. Maybe we should try to integrate this functionality into existing classes such as `FEEvaluation` and `MatrixFree`. Let me explain this ... In some sense, this PR describes that the interfaces of `MatrixFree` should be improved. When we started to write application code using `MatrixFree`, we realized that we write the same code again and again, but that smaller units such as `reinit_cell()/face()` and `do_cell/face_integral()` (with loop over quadrature points or formulated for a single quadrature point only) should actually be the code that needs to be implemented by applications. From this perspective, this PR suggests to change or complement the interface of `MatrixFree` for more flexibility, in particular operators with additional `FEEvaluation` objects for additional fields.

kronbichler:
Let me start the reply with going back to this one:

> The concrete implementation of an operator has to follow an interface, so that `MatrixFreeEvaluator` (holding a pointer to the operator) can call functions like `operator->reinit_cell()`, `operator->do_cell_integral()` (without parameters!). Inside the quadrature point or flux computations, the operator accesses the different FEEvaluation/Integrator objects e.g. by `matrix_free->get_integrator("temperature").get_gradient(q)` and `matrix_free->get_integrator("velocity").get_value(q)` (instead of `matrix_free->get_integrator(0).get_gradient(q)` and `matrix_free->get_integrator(1).get_value(q))`. Does this make sense?

I assume you mean that `reinit_cell(const unsigned int cell_batch_index)` takes an argument, but `do_cell_integral` does not get any argument? That seems to be reasonable.

Going back for a moment to the question whether a unification is possible, do you mean that we say inside `reinit_cell` (or, more precisely, somewhere in a construction routine) which of the fields we want to call `reinit` on? The reason is that we do not want to spend work on unnecessary `reinit`, as the function is not for free, especially not in a multiple-setup case. We need some way of "registration" because we might have two different "evaluators" attaching to the same field as well, say a velocity variable for the actual variable in a linearized equation, and the linearization point/state coming from a second field of the same type.

Regarding the quadrature point operation, I do not think something like `matrix_free->get_integrator("temperature").get_gradient(q)` is feasible with the regular tools in C++. The problem is that any kind of string operations is involving several dozens of instructions, so looking up things at the quadrature point level is going to make performance tank. (Things are not quite as bad as in the past before the fast-string optimizations of the C++ library, which g++ enabled with C++11 if I remember correctly, as before you would do memory allocations on every such call, implying >100x slowdowns.) So in general one would need at least some intermediate layer that can mediate between a string identifier "temperature" and a variable name `fe_eval_temperature` or at least an integer index into something like `matrix_free_wrapper.get_integrator(temperature_index).get_gradient()`. I am not an expert along those lines, but I do not think it would be infeasible to bring the names into a compile-time context (with run-time resolution of string names from the available fields), as I believe I've seen some of these naming issues been addressed with pre-processor or some boost magic in other contexts.

Now back to the actual question:
> Maybe we should try to integrate this functionality into existing classes such as `FEEvaluation` and `MatrixFree`. Let me explain this ... In some sense, this PR describes that the interfaces of `MatrixFree` should be improved. When we started to write application code using `MatrixFree`, we realized that we write the same code again and again, but that smaller units such as `reinit_cell()/face()` and `do_cell/face_integral()` (with loop over quadrature points or formulated for a single quadrature point only) should actually be the code that needs to be implemented by applications.

I think this is something that represents a new layer, at least compared to the currently existing classes: The `FE(Face)Evaluation` is handling the task of evaluating a single variable in a specific context, whereas `MatrixFree` is more or less a mere data container. I think in several of our discussions among @nfehn @peterrum and me we used a term `MatrixFreeWrapper` or similar, so that might be something to consider. I do not disagree here, actually I agree quite a lot because I think `FEEvaluation` and `FEFaceEvaluation` is not just too low-level for the tasks discussed here, but also holding back performance in other cases where a lower-level approach might help, again considering also the case how things need to look for a GPU code. To give just one example, in https://github.com/dealii/dealii/pull/13056 I spent a lot of effort to reduce the overhead caused by some intermediate layers spurred by work on this one here, so I see both the applicability in ExaDG and having a higher granularity to aggregate some calls for some supported cases with higher performance needs as facets of a question for the future. (Now there might come up a question about code generation as many other packages do.)

nfehn:
@kronbichler If this is meant as a draft and not planned to be integrated into ExaDG in its current form/design, I think we should close this PR now.

PR #113, opened 2021-12-09
CMake: add subdirectories automatically for applications

nfehn:
Scenario: A user wants to create a new application test case, say for the Poisson module

Previous behavior: 
- copy a folder, e.g. "cp poisson/template poisson/my_example"
- in poisson/CMakeLists.txt: add the directory "ADD_SUBDIRECTORY(my_example)"
- in my_example/CMakeLists.txt: adjust the exe names so that they do not conflict with the names of other application examples

New behavior:
- just copy the folder "cp poisson/template poisson/my_example"; everything else should be done automatically

Currently only done for Poisson module. I extend it to the rest of ExaDG once we agreed.

@slfuchs Could be relevant for @partexa


peterrum (approved):
I like the changes. Less writing. However, you need to be aware that it is now not that trivial to disable individual applications (while refactoring in the future).

> just copy the folder "cp poisson/template poisson/my_example"; everything else should be done automatically

One has to add that cmake has to be re-run else the new folder is not found.

Regarding your original intention, I am not sure if we should motivate users to add new folders. Personally, I would regard the folders rather as off-the-shelf applications and examples to present features in ExaDG. Users should be motivated to create applications outside of the ExaDG folder, just like we do in the ExaDG-application repo. I think that would not be more difficult for the user: in deal.II, we also always copy an existing CMakeLists.txt file, which should be also possible here.

Just imagine that a user who is not that familiar with git wants to update the code! In the worst case, we have added a folder with the same name as the user has done.

Also, you were considering spack/versioning in the future. There, ExaDG is installed somewhere and you won't be able to add folders.

nfehn:
Expecting a user to set up a new project and link ExaDG is too much in my opinion. A user of ExaDG should not be confronted with touching CMake. For people new to ExaDG the first steps will be to play around with existing test cases (change parameters, etc.), and then add another test case. This should be realizable completely within ExaDG in my opinion. So I do not want to push users in a certain direction; I am convinced that they want to do it like that anyway. Otherwise many users will just say it's too difficult to work with this software.

@peterrum What you describe above characterizes in my opinion an experienced developer. We are on the same track that this is how we want to build complex applications or multiphysics codes on top of ExaDG. Nevertheless, I believe it would be good if ExaDG provided a template project also for the scenario that you describe, so that users do not have to think about how to create such a project. Do you want to create an issue, or a first draft for such a template project?

nfehn:
I think I have updated everything.

kronbichler (approved):
This looks good.

applications/compressible_navier_stokes/CMakeLists.txt
```diff
@@ -1,10 +1,9 @@
...
-ADD_SUBDIRECTORY(template)
-ADD_SUBDIRECTORY(turbulent_channel)
\ No newline at end of file
+PROJECT(${PROJECT_NAME})
+
+SUBDIRLIST(SUBDIRS ${CMAKE_CURRENT_SOURCE_DIR})
+
+FOREACH(subdir ${SUBDIRS})
+  ADD_SUBDIRECTORY(${subdir})
+ENDFOREACH()
```
kronbichler, line 9: Github complains here and in a number of other files that you do not have a newline at the end of the file. Would you mind making these consistent. (Single new-line at the end of file?)

peterrum (approved):

applications/compressible_navier_stokes/CMakeLists.txt
nfehn: done in https://github.com/exadg/exadg/pull/113/commits/1a8206fe474f2113021abc61506534a9a75003ca

kronbichler (approved):

PR #116, opened 2021-12-10
add TriangulationType::Serial

nfehn:
realizes a part of #110 as a separate PR.

I assume we do not need to add asserts in ExaDG since we all interfaces should be written with `Triangulation` (some time ago we had `parallel::distributed::Triangulation`)?


kronbichler (approved):
Indeed, I think this should be robust as we only should depend on the base class with specializations as necessary. (As long as we do not expect our solvers to be parallel, but I think even that case should work by just running the same job on all ranks by `MPI_COMM_SELF`.)

peterrum (approved):
Thanks!

PR #118, opened 2021-12-20
Add dealii interface for optional petsc

bergbauer:
Fixes build for deal.II without PETSc

### DIFF
--- a/include/exadg/solvers_and_preconditioners/multigrid/multigrid_parameters.h
+++ b/include/exadg/solvers_and_preconditioners/multigrid/multigrid_parameters.h
@@ -88,6 +88,74 @@ struct AdditionalData
 } // namespace dealii
 #endif
 
+#ifndef DEAL_II_WITH_PETSC
+namespace dealii::PETScWrappers::PreconditionBoomerAMG
+{
+// copy of interface from deal.II (lac/petsc_precondition.h)
+struct AdditionalData
+{
+  enum class RelaxationType
+  {
+    Jacobi,
+    sequentialGaussSeidel,
+    seqboundaryGaussSeidel,
+    SORJacobi,
+    backwardSORJacobi,
+    symmetricSORJacobi,
+    l1scaledSORJacobi,
+    GaussianElimination,
+    l1GaussSeidel,
+    backwardl1GaussSeidel,
+    CG,
+    Chebyshev,
+    FCFJacobi,
+    l1scaledJacobi,
+    None
+  };
+
+  AdditionalData(const bool           symmetric_operator               = false,
+                 const double         strong_threshold                 = 0.25,
+                 const double         max_row_sum                      = 0.9,
+                 const unsigned int   aggressive_coarsening_num_levels = 0,
+                 const bool           output_details                   = false,
+                 const RelaxationType relaxation_type_up               = RelaxationType::SORJacobi,
+                 const RelaxationType relaxation_type_down             = RelaxationType::SORJacobi,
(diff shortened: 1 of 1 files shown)


peterrum (approved):

kronbichler (approved):

nfehn:
I added the "do not merge" label since I want to discuss something before merging (since we already have 2 approvals).

I would prefer a solution that guards the occurrences of Petsc or BoomerAMG in the code instead of copying the interface.

peterrum:
This is how we do it also for Trilinos. I think the problem is that we can set the (PETSc/Trilinos) parameters directly in the application.

nfehn:
> This is how we do it also for Trilinos. I think the problem is that we can set the (PETSc/Trilinos) parameters directly in the application.

Yes, my comment also applies to how we do it for Trilinos. But I am convinced that we should not continue like this just because it has been done like this before. There is always the point at which we have to decide to do things in a better or different way ...

Then, the question would be: do applications directly set Trilinos/Petsc parameters  (which ones) and why not using an ifdef also in those functions? If this is too much work or not justified in our opinion, we would need to wrap this data structure in ExaDG in my opinion.

bergbauer:
Let's merge the fix and discuss better solutions in an issue?

nfehn:
Actually, "no" from my side. At least I want to receive an answer to one of my questions. If we can realize it in better way by guarding the code, why not do it immediately?

peterrum:
It seems that non of the applications in  this repo explicitly set `boomer_data` or `ml_data`. I am not sure how it looks like in the application repo. 

> Yes, my comment also applies to how we do it for Trilinos. But I am convinced that we should not continue like this just because it has been done like this before. There is always the point at which we have to decide to do things in a better or different way ...

The question is who should write the ifdefs. The user in the applications. As it is now we provide these dummy structs so that the user does not need to do so. I like it this way.

nfehn:
Yes, but you ignore that code is copied. This way, it is difficult come to a solution. I think we need to discuss where exactly ifdefs would be needed in case of guards.

> As it is now we provide these dummy structs so that the user does not need to do so. I like it this way.

> It seems that non of the applications in this repo explicitly set boomer_data or ml_data.

I see a contraction between the two sentences above.

> The question is who should write the ifdefs. The user in the applications. As it is now we provide these dummy structs so that the user does not need to do so. I like it this way.

From a different perspective, you are saying that these changes should be implemented in deal.II directly. Our "applications" are an application of ExaDG, while the whole ExaDG code is an application of deal.II. So could you please argue why it is not possible to realize this in deal.II?

So far, I have not seen the stringent argument that justifies to copy code from deal.II to ExaDG.

bergbauer:
I also don't like the current solution because it makes changes to exadg necessary when the deal.II interface changes. Anyways, this PR was intended as a bugfix for a failing master build mirroring the approach taken for trilinos. That's why I suggested to merge this and discuss this in an issue which can then be resolved in a follow up PR.

> It seems that non of the applications in this repo explicitly set boomer_data or ml_data. I am not sure how it looks like in the application repo.
We have that case in the application repo but I could be guarded in those cases.

Both other approaches (guarding or upstream change) are viable options in my opinion.

nfehn:
Ok, then we have a plan what to do. We should start with guarding and see how this works. If this becomes too difficult in applications we can adapt the procedure.

Lets merge this one.

kronbichler:
I agree with the procedure to guard code (maybe assisted by useful default values inside the respective solver section and their `ifdef`, as to minimize the settings), and in case we observe difficulties we can wrap that into a class/struct in ExaDG.

PR #119, opened 2021-12-21
Add test case for flow around sphere

kronbichler:
This PR suggests a new test case for the flow around a sphere. It is similar to the flow around a cylinder, but it stresses several different components of our implementation:
- By working on this test case, I observed that the orientation + refinement combination in deal.II was broken (which eventually led to https://github.com/dealii/dealii/pull/13101). This makes it useful to check ExaDG capabilities as well.
- There is still one potential issue (that I do not know how to fix yet), which happens when I have some hanging nodes in the front of the sphere. It could be related to the orientation (but no test in deal.II triggers), but also things like inf-sup or stabilization parameters somewhere.
- The test case demonstrates some more advanced manifold manipulations, doing a boundary-adapted mesh inside a `SphericalManifold`.
- I had initially intended to include `TransfiniteInterpolationManifold` but decided against it for the current implementation in deal.II as it increases costs.

We could think of giving the user more control about the adaptive refinement, which is something we might want to discuss.

### DIFF
--- a/applications/incompressible_navier_stokes/flow_past_sphere/CMakeLists.txt
+++ b/applications/incompressible_navier_stokes/flow_past_sphere/CMakeLists.txt
@@ -1,0 +1,3 @@
+PROJECT(incompressible_flow_past_sphere)
+
+EXADG_PICKUP_EXE(solver.cpp incompressible_flow_past_sphere solver)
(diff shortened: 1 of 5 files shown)


kronbichler:
Regarding the observed instability: This seems to be cured by using a ramping in the start of the simulation, as going to full speed for high Reynolds numbers is obviously too much.

PR #121, opened 2021-12-21
redesign #ifdef's for Trilinos and Petsc

nfehn:
follow up to #118 .

I think the code becomes much simpler by these changes and code duplication from deal.II is avoided.


kronbichler (approved):
Yes, I fully agree with this change, this looks much nicer (and removing 100 lines suggests that this is right).

nfehn:
@bergbauer @peterrum Does one of you also want to review? Then, I think we can merge.

peterrum (approved):

PR #122, opened 2021-12-21
Improve compute penalty parameter

kronbichler:
In some FSI cases, @nfehn and I observed very high cost of initializing `FEFaceValues` (for high-order mappings). This PR avoids this by computing the volume and surface area by the much faster `FEEvaluation` and `FEFaceEvaluation` concepts. The difficulty is that we need to exchange data between different MPI processes, for which this PR suggests to use the global active cell index (avaiable since the 9.3 release of deal.II) and a distributed vector. This should avoid the bottleneck entirely and make the function as fast as a cell loop.


include/exadg/operators/interior_penalty_parameter.h
```diff
@@ -44,48 +44,87 @@ calculate_penalty_parameter(AlignedVector<VectorizedArray<Number>> & array_penal
...
+      for(unsigned int v = 0; v < matrix_free.n_active_entries_per_face_batch(face); ++v)
       {
-        fe_face_values.reinit(cell, f);
-        Number const factor = (cell->at_boundary(f) && !cell->has_periodic_neighbor(f)) ? 1. : 0.5;
-        for(unsigned int q = 0; q < face_quadrature.size(); ++q)
+        // inner face: adjust weight and also write to the "exterior" face
+        if(face < matrix_free.n_inner_face_batches())
         {
-          surface_area += fe_face_values.JxW(q) * factor;
+          face_area[v] *= 0.5;
```
nfehn, line 108: Why don't you separate this code from the code below that adds to `surface_areas` (i.e. shift the `if()` clause to line 102 and briefly explain why `*=0.5`)? The main motivation for this is that multiplication does not have to be done separately for each `v`, so I think this change would improve readability a lot. This would also very naturally resolve the issue that I raised above regarding the comment (that was difficult to understand for me).

nfehn:
I only have (several) minor comments. Maybe also @peterrum could have a brief look at this PR once you have addressed the requested changes.

peterrum:
I'll do. @kronbichler could you ping me once you are ready?

include/exadg/operators/interior_penalty_parameter.h
```diff
@@ -44,48 +44,87 @@ calculate_penalty_parameter(AlignedVector<VectorizedArray<Number>> & array_penal
...
+  unsigned int const n_cells = matrix_free.n_cell_batches() + matrix_free.n_ghost_cell_batches();
   array_penalty_parameter.resize(n_cells);
 
-  Mapping<dim> const &       mapping = *matrix_free.get_mapping_info().mapping;
-  FiniteElement<dim> const & fe      = matrix_free.get_dof_handler(dof_index).get_fe();
-  unsigned int const         degree  = fe.degree;
+  // Compute cell volume with usual FEEvaluation object and store it in
+  // array_penalty_parameter for now
+  {
+    FEEvaluation<dim, -1, 0, 1, Number> fe_eval(matrix_free, dof_index);
```
nfehn, line 53: Is it correct that this uses precompiled code for fast evaluation only up to degree 6 as written in the dealii docu? In ExaDG, we have the parameter `EXADG_DEGREE_MAX`, so why don't you use `CellIntegrator` and/or `FaceIntegrator` as in other parts of ExaDG?

kronbichler: `FEEvaluation<dim, -1, 0, n_components, Number>` **is** `CellIntegrator<dim, n_components, Number`: https://github.com/exadg/exadg/blob/01939cbd3441629dacd624d5cef041cbb9904db6/include/exadg/matrix_free/integrators.h#L31-L35
So we get the same maximal degree in either case, relying on the order implied through the dynamic linking / shared library load (dlopen) to pick up ExaDG's maximal degree rather than deal.II's. It's right that we should switch to `CellIntegrator` here, it's just my habits picking the other :wink:

kronbichler:
In 3449da6 I added better comments and re-organized the code: I now first compute the surface areas and then go through the cells, which allows me to directly store the result back without keeping the intermediate volume in the vector.

include/exadg/operators/interior_penalty_parameter.h
```diff
@@ -25,65 +25,120 @@
...
 
-      array_penalty_parameter[i][v] = surface_area / volume;
+  surface_areas.compress(VectorOperation::add);
+  surface_areas.update_ghost_values();
+
+  // As a second step, we use a CellInterator to compute the volume and then
+  // combine it with the surface areas computed before
+  unsigned int const n_cells = matrix_free.n_cell_batches() + matrix_free.n_ghost_cell_batches();
+  array_penalty_parameter.resize(n_cells);
+  {
```
nfehn, line 130: Is there a reason that you open two scopes in this function? It does not really make sense in my opinion that you define `n_face_batches` inside the scope but `n_cells` outside the scope. I think I would prefer to remove the scopes.

kronbichler: I agree, the scopes are not necessary (I had when I originally had more temporary variables that would have clashed between cell and face loops, but now it's only the evaluator which is unambiguous). Done in 5c06970b

nfehn (approved):

include/exadg/operators/interior_penalty_parameter.h
```diff
@@ -25,67 +25,118 @@
...
   }
+
+  surface_areas.compress(VectorOperation::add);
+  surface_areas.update_ghost_values();
+
+  // As a second step, we use a CellInterator to compute the volume and then
+  // combine it with the surface areas computed before
+  unsigned int const n_cells = matrix_free.n_cell_batches() + matrix_free.n_ghost_cell_batches();
+  array_penalty_parameter.resize(n_cells);
+  CellIntegrator<dim, 1, Number> integrator(matrix_free, dof_index);
```
peterrum, line 128: quad_index does not matter?

peterrum (approved):
This is how the global indices where designed to be used :+1:

include/exadg/operators/interior_penalty_parameter.h
nfehn: I also had this question. I think we should extend the interface of this function (and remove the default value for dof_index). In the old code version, the quadrature rule was constructed inside the function.

peterrum: Let's keep this to a follow-up PR!

nfehn: Why? This PR might introduce a bug!?

nfehn: Apart from matrix-based vs. matrix-free evaluation, the code might do something different now

peterrum: > This PR might introduce a bug!?

You are right. But you might need to help Martin. I quickly take a look where the function is called and at some of those places no quad_index is available (e.g., https://github.com/exadg/exadg/blob/7af16c94168fc0126315d55c038ea55ad9e0fe47/include/exadg/incompressible_navier_stokes/spatial_discretization/operators/viscous_operator.h#L64-L67). Also one needs to decide which quad index to use which might be not unique.

kronbichler: Yes, we can introduce that parameter, but the follow-up question is what quad index we should be choosing. Before, we used `degree + 1`, while the natural choice now is the one in the evaluator context where we're using it (which is often `degree + 1`). We might have minor deviations in behavior between the previous version and current one in those applications that do not use `degree + 1` quadrature points (maybe we do everywhere now, but in the future we might decide differently). Anyway, it shouldn't matter much given that nature of the information we're computing: an approximation of the area/volume. The underlying formula is by no means exact, so a minor quadrature error up or down is irrelevant, and often we either get the exact result or a reasonable estimate already with a single quadrature point. In all our use cases, the number of quadrature points scales with the polynomial degree (i.e., also the mapping degree in the deformed case), so we can expect very small quadrature errors.

kronbichler: > Apart from matrix-based vs. matrix-free evaluation, the code might do something different now

I am confused by the terms *matrix-based* vs *matrix-free*, but I assume you mean `FEValues` vs `FEEvaluation`? In principle, they should be equivalent apart from round-off errors. But the main issue I see is that I could have introduced a bug in what I execute here, as the parameter is hard to verify. I tested several Poisson problems, but it might be good to keep an eye on it.

kronbichler: I now added this with the separate commit ab0ff36. It is not particularly pretty to pull it through all interfaces, but this might be needed at some point anyway (see e.g. #128).

include/exadg/poisson/spatial_discretization/laplace_operator.cpp
```diff
@@ -54,7 +54,7 @@ LaplaceOperator<dim, Number, n_components>::calculate_penalty_parameter(
   MatrixFree<dim, Number> const & matrix_free,
   unsigned int const              dof_index)
```
nfehn, line 55: I think it would be more consistent to also extend this interface by `quad_index`

include/exadg/convection_diffusion/spatial_discretization/operators/combined_operator.cpp
```diff
@@ -60,7 +60,10 @@ CombinedOperator<dim, Number>::initialize(MatrixFree<dim, Number> const &   matr
   if(operator_data.diffusive_problem)
   {
     diffusive_kernel = std::make_shared<Operators::DiffusiveKernel<dim, Number>>();
-    diffusive_kernel->reinit(matrix_free, data.diffusive_kernel_data, data.dof_index);
+    diffusive_kernel->reinit(matrix_free,
+                             data.diffusive_kernel_data,
+                             data.dof_index,
+                             data.quad_index);
```
nfehn, line 66: This is potentially problematic: For the combined operator, `quad_index` might be the quadrature index for over-integration (if `param.use_overintegration=true`). However, the viscous kernel in `ConvDiff::Operator` is always initialized with `get_quad_index()` (= degree+1). Hence, the results might be different depending on whether we apply the operators in one matrix-free loop or separately. However, we want to ensure that this is not the case and that the two variants only differ in terms of computational costs.

The solution to this is the following: Add another parameter `quad_index_standard` to `ConvDiff::CombinedOperatorData` (to be initialized in `ConvDiff::Operator`) and use this parameter `data.quad_index_stanard` to initialize the diffusive kernel (lines 63-66 above). The computation of the integrals will still be done using `data.quad_index` (which might be standard integration or over-integration depending on the input parameters).

Of course, we could also change the logic in the future, but I think the present PR should not change it since it is a refactoring.

include/exadg/convection_diffusion/spatial_discretization/operators/combined_operator.cpp
```diff
@@ -119,7 +122,9 @@ void
 CombinedOperator<dim, Number>::update_after_grid_motion()
 {
   if(operator_data.diffusive_problem)
-    diffusive_kernel->calculate_penalty_parameter(*this->matrix_free, operator_data.dof_index);
+    diffusive_kernel->calculate_penalty_parameter(*this->matrix_free,
+                                                  operator_data.dof_index,
+                                                  operator_data.quad_index);
```
nfehn, line 127: With the above comment, this would then read `operator_data.quad_index_standard`.

include/exadg/incompressible_navier_stokes/spatial_discretization/operators/momentum_operator.cpp
```diff
@@ -42,7 +42,8 @@ MomentumOperator<dim, Number>::initialize(MatrixFree<dim, Number> const &   matr
     this->viscous_kernel = std::make_shared<Operators::ViscousKernel<dim, Number>>();
     this->viscous_kernel->reinit(matrix_free,
                                  operator_data.viscous_kernel_data,
-                                 operator_data.dof_index);
+                                 operator_data.dof_index,
+                                 operator_data.quad_index);
```
nfehn, line 46: Same as above: Add parameter `MomentumOperatorData::quad_index_standard` initialized in `IncNS::SpatialOperatorBase` as `data.quad_index_standard = get_quad_index_velocity_linear();` and use this parameter in the line above, i.e. `operator_data.quad_index_standard`.

include/exadg/incompressible_navier_stokes/spatial_discretization/operators/momentum_operator.cpp
```diff
@@ -133,7 +134,8 @@ MomentumOperator<dim, Number>::update_after_grid_motion()
 {
   if(operator_data.viscous_problem)
     viscous_kernel->calculate_penalty_parameter(this->get_matrix_free(),
-                                                this->get_data().dof_index);
+                                                this->get_data().dof_index,
+                                                this->get_data().quad_index);
```
nfehn, line 138: Then, this should be `quad_index_standard`.

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
```diff
@@ -301,7 +301,10 @@ SpatialOperatorBase<dim, Number>::initialize_operators(std::string const & dof_i
   viscous_kernel_data.viscosity_is_variable        = param.use_turbulence_model;
   viscous_kernel_data.variable_normal_vector       = param.neumann_with_variable_normal_vector;
   viscous_kernel = std::make_shared<Operators::ViscousKernel<dim, Number>>();
-  viscous_kernel->reinit(*matrix_free, viscous_kernel_data, get_dof_index_velocity());
+  viscous_kernel->reinit(*matrix_free,
+                         viscous_kernel_data,
+                         get_dof_index_velocity(),
+                         get_quad_index_velocity_linear());
```
nfehn, line 307: Here, the standard quadrature rule is used which is why it should also be used in case of the momentum operator (operators combined to a single matrix-free loop) for consistency (or in order to avoid misunderstandings).

include/exadg/compressible_navier_stokes/spatial_discretization/kernels_and_operators.h
```diff
@@ -981,7 +981,8 @@ class ViscousOperator
 
     IP::calculate_penalty_parameter<dim, Number>(array_penalty_parameter,
                                                  *matrix_free,
-                                                 data.dof_index);
+                                                 data.dof_index,
+                                                 data.quad_index);
```
nfehn, line 985: Add parameter `ViscousOperatorData::quad_index_standard` (initialized in `CompNS::Operator` with `quad_index_standard`) and use it here. The reason is that the present PR should not change numerics in my opinion. We could redefine this in a follow-up PR.

Note that nothing has to be done for `CompNS::CombinedOperatorData`! The reason is that the `CompNS` module does not use the kernel-design (that is used for the `ConvDiff` and `IncNS` modules). 

Introducing the kernel design for `CompNS` for consistency with other ExaDG modules would have to be done in a follow-up PR.

include/exadg/operators/interior_penalty_parameter.h
nfehn: > I am confused by the terms matrix-based vs matrix-free, but I assume you mean FEValues vs FEEvaluation

Yes. I have written this very quickly yesterday.

nfehn: > It is not particularly pretty to pull it through all interfaces, but this might be needed at some point anyway (see e.g. #128).

We might want to refactor later, but the present PR should make efforts to not change the numerics.

nfehn: > You are right. But you might need to help Martin. I quickly take a look where the function is called and at some of those places no quad_index is available

I have carefully gone through all changes and added comments where adjustments are necessary.

kronbichler:
I can make those changes, but before we spend more time on this side topic, let me raise a question: What would you think if we introduce a special parameter for `quad_index` in FEEvaluation (either the "default" value or a special one we can provide), https://github.com/dealii/dealii/blob/be72dbb5a5f97aee1c8cf330b66664c848176fba/include/deal.II/matrix_free/fe_evaluation.h#L2049, that picks up the quadrature formula which is closest (from above) to `degree + 1`? This might be `numbers::invalid_unsigned_int` (similar to the `hp` parameters) or a more expressive name, e.g. `FEEvaluation::detect_quadrature_formula_from_degree`. I am not trying to hide from this discussion here, but I would like to think about the best solution for the long term and if we can make usability improvements, given this experience.

peterrum:
> I can make those changes, but before we spend more time on this side topic, let me raise a question: What would you think if we introduce a special parameter for quad_index in FEEvaluation (either the "default" value or a special one we can provide), https://github.com/dealii/dealii/blob/be72dbb5a5f97aee1c8cf330b66664c848176fba/include/deal.II/matrix_free/fe_evaluation.h#L2049, that picks up the quadrature formula which is closest (from above) to degree + 1? This might be numbers::invalid_unsigned_int (similar to the hp parameters) or a more expressive name, e.g. FEEvaluation::detect_quadrature_formula_from_degree. I am not trying to hide from this discussion here, but I would like to think about the best solution for the long term and if we can make usability improvements, given this experience.

I was thinking about the same yesterday. We also had something similar at one time in the code (I don't recall in which context): we looped over all quadrature rules and checked the number of quadrature points. The problem I saw were:
- how does the generalization look like: simplex, wedge, pyramid, ...
- let's assume we have two sets of quadrature rules for `degree + 1`: one with Gauss–Legendre quadrature and the other with Gauss–Lobatto quadrature. Which one do we want to take? Furthermore, the result will depend on the order how the quadrature rules have been registered. Probably, choosing any of the two quadrature rules will work but might lead to confusions in the future...

nfehn:
I am not sure whether more automation provides better usability in this context. Personally, I don't like initializing variables with some special numbers or invalid numbers in order to express that a special path should be taken. The code becomes difficult to understand and, most often, it requires a dictionary for translation.

kronbichler:
In the constructor of `FEEvaluation` we can only afford a reasonably cheap operation. A solution actually looking into the quadrature formulas and determining whether it is the most accurate one brings us back to the other discussion we had in terms of the overall evaluator design, allowing a user to collect all information to attach to a particular evaluation context and not just plain integers.

nfehn:
I suggest to complete this PR first and get is merged soon. Only a few additional lines of code should be necessary as described in detail in the code review.

kronbichler:
I rebased this PR and added the `quad_index_standard` to the convection-diffusion and incompressible Navier-Stokes modules. I am not very happy about the fact that we have to pull this parameter through so many interfaces, given that it is only giving a rough estimate. More precisely, I am comparing the state after the last (fourth) commit to the one after the second commit (where we simply pick the zero-th quadrature index, and take the possibility of a small quadrature error into account, modulo some interface cleanup that I addressed in the third commit during rebasing). At least it is clearly better than after the third commit, so we might go on from here.

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
```diff
@@ -522,8 +525,9 @@ SpatialOperatorBase<dim, Number>::initialize_operators(std::string const & dof_i
 
   data.bc = boundary_descriptor->velocity;
 
-  data.dof_index  = get_dof_index_velocity();
-  data.quad_index = get_quad_index_velocity_linearized();
+  data.dof_index           = get_dof_index_velocity();
+  data.quad_index          = get_quad_index_velocity_linearized();
+  data.quad_index_standard = get_quad_index_velocity_linear();
```
nfehn, line 530: Shouldn't this be a parameter of `ViscousKernelData`? As a positive side-effect, the interface of `ViscousKernel::reinit()` would not need to be changed and hopefully also `MomentumOperatorData` does not have to be touched here. It's a long time ago, so maybe I overlook something right now ...

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
```diff
@@ -396,7 +396,10 @@ SpatialOperatorBase<dim, Number>::initialize_operators(std::string const & dof_i
   viscous_kernel_data.viscosity_is_variable        = param.use_turbulence_model;
   viscous_kernel_data.variable_normal_vector       = param.neumann_with_variable_normal_vector;
   viscous_kernel = std::make_shared<Operators::ViscousKernel<dim, Number>>();
-  viscous_kernel->reinit(*matrix_free, viscous_kernel_data, get_dof_index_velocity());
+  viscous_kernel->reinit(*matrix_free,
+                         viscous_kernel_data,
+                         get_dof_index_velocity(),
+                         get_quad_index_velocity_linear());
```
nfehn, line 402: According to this comment https://github.com/exadg/exadg/pull/122/files#r1024055107, we would write a few lines above `viscous_kernel_data.quad_index_penalty_parameter = get_quad_index_velocity_linear();`. The reinit function would only have 3 arguments as before.

include/exadg/compressible_navier_stokes/spatial_discretization/kernels_and_operators.h
```diff
@@ -930,6 +931,7 @@ struct ViscousOperatorData
 
   unsigned int dof_index;
   unsigned int quad_index;
+  unsigned int quad_index_standard;
```
nfehn, line 934: If I remember correctly, we never introduced `Kernel` or `KernelData` for the module `CompNS`. So I agree with the design chosen here, namely adding the parameter to `OperatorData`.  For clarity, I think the name `quad_index_penalty_parameter` (or similar) would be better. It is probably not used for anything else.

nfehn:
I agree with your thoughts on pulling the quad index through so many interfaces, and that we would like to avoid this. I think the solution is adding it to the respective `KernelData`. This way, the combined operators for convection-diffusion and Navier-Stokes should have all the information they need, given the modular design that we have chosen for these combined operators (through composition used in the `Operator` and the `OperatorData` classes).

nfehn:
@kronbichler Is it ok for you that we try to merge #296 first, in order to make progress with simplex?

PR #126, opened 2022-01-06
Refactor InterfaceCoupling

peterrum:
...so that it uses `dealii::Utilities::MPI::RemotePointEvaluation`.

depends on ~https://github.com/dealii/dealii/pull/13173, https://github.com/dealii/dealii/pull/13177, https://github.com/dealii/dealii/pull/13178, https://github.com/dealii/dealii/pull/13187, https://github.com/exadg/exadg/pull/127~


include/exadg/functions_and_boundary_conditions/interface_coupling.h
```diff
@@ -533,175 +140,69 @@ class InterfaceCoupling
...
-          AssertThrow(array_cache_src[q].size() > 0,
-                      ExcMessage("No adjacent points have been found."));
-        }
-      }
-
-
-      /*
-       * 4. Communication: transfer results back to dst-side
-       */
-      communicator.process(mpi_map_solution_src, array_solution_dst, 11 + quadrature);
```
peterrum, line 609: I am not sure why this communication step is performed here.

include/exadg/functions_and_boundary_conditions/interface_coupling.h
```diff
@@ -23,433 +23,53 @@
...
-        unsigned int const                        dof_index_dst_in,
+  setup(std::shared_ptr<MatrixFree<dim, Number>>  matrix_free_dst,
+        unsigned int const                        dof_index_dst,
         std::vector<quad_index> const &           quad_indices_dst_in,
         MapBoundaryCondition const &              map_bc_in,
-        std::shared_ptr<Triangulation<dim>> const triangulation_src_in,
+        std::shared_ptr<Triangulation<dim>> const triangulation_src,
         DoFHandler<dim> const &                   dof_handler_src_in,
-        Mapping<dim> const &                      mapping_src_in,
+        Mapping<dim> const &                      mapping_src,
```
kronbichler, line 66: I think we should use consistent names here in terms of the suffix `_in` or not.

@nfehn a question regarding the argument `triangulation_src`: This seems redudant as we also pass in `dof_handler_src_in`, which, according to the name and its use in FSI `driver.cpp`, must necessarily point to the same `triangulation` as the one that gets passed in. So I would opt for not repeating the same information through different channels and query it via `dof_handler_src_in.get_triangulation()` (not necessarily in this PR).

include/exadg/functions_and_boundary_conditions/interface_coupling.h
```diff
@@ -533,175 +140,69 @@ class InterfaceCoupling
...
-        map_cache_src.find(quadrature)->second;
-      std::map<mpi_rank, ArrayVectorTensor> & mpi_map_solution_src =
-        map_solution_src.find(quadrature)->second;
-
-      for(auto it : mpi_map_cache_src)
-      {
-        mpi_rank const      proc               = it.first;
-        ArrayVectorCache &  array_cache_src    = mpi_map_cache_src.find(proc)->second;
-        ArrayVectorTensor & array_solution_src = mpi_map_solution_src.find(proc)->second;
+      LinearAlgebra::distributed::Vector<float> temp;
```
kronbichler, line 177: Why `float` and not `VectorType`?

include/exadg/functions_and_boundary_conditions/interface_coupling.h
```diff
@@ -533,175 +142,70 @@ class InterfaceCoupling
...
+                                          mapping_src);
     }
 
     // finally, give boundary condition access to the data
     for(auto boundary : map_bc)
     {
       boundary.second->set_data_pointer(map_index_dst, map_solution_dst);
     }
+
+    update_data(dof_vector_src_in); // TODO: needed?
```
kronbichler, line 169: I think the name `setup` suggests we do not exchange any data, but I think @nfehn should decide.

include/exadg/functions_and_boundary_conditions/interface_coupling.h
```diff
@@ -533,175 +142,70 @@ class InterfaceCoupling
...
-        }
-      }
-
-      /*
-       * 2. Communication: transfer results back to dst-side
-       */
-      InterfaceCommunicator<dim, dim> & communicator = map_communicator.find(quadrature)->second;
-
-      communicator.process(mpi_map_solution_src, array_solution_dst, 11 + quadrature);
+      const auto result =
```
kronbichler, line 179: Do you think it would make sense to spell out the type here (I guess it depends whether we have a good typedef otherwise it might get too verbose)? Also, we should do
```suggestion
      auto const result =
```

kronbichler:
This looks like a great simplification, now that we can get the data uniform. I guess we need some backward compatibility setting for deal.II 9.3 (I assume we could choose something similar as I did in #120 with a copy to `double` vectors, in absence of the most recent changes slated for inclusion deal.II regarding the `float` vectors).

kronbichler (approved):
This looks good to me, apart from the questions where we still need to decide.

include/exadg/functions_and_boundary_conditions/interface_coupling.h
peterrum: > I think we should use consistent names here in terms of the suffix _in or not.

I just renamed those variables that are not stored in the class in the style of `variable = variable_in`.

peterrum: > a question regarding the argument triangulation_src

I have removed it.

include/exadg/functions_and_boundary_conditions/interface_coupling.h
```diff
@@ -533,175 +142,89 @@ class InterfaceCoupling
...
-       */
-      InterfaceCommunicator<dim, dim> & communicator = map_communicator.find(quadrature)->second;
-
-      communicator.process(mpi_map_solution_src, array_solution_dst, 11 + quadrature);
+#if DEAL_II_VERSION_GTE(10, 0, 0)
+      auto const result =
+        VectorTools::point_values<n_components>(map_communicator[quadrature],
+                                                *dof_handler_src,
+                                                dof_vector_src,
+                                                VectorTools::EvaluationFlags::avg);
```
peterrum, line 195: @nfehn Averaging is done by deal.II so that the vectors of vectors could be removed.

include/exadg/functions_and_boundary_conditions/function_cached.h
```diff
@@ -38,7 +38,7 @@ class FunctionCached
 private:
   typedef std::tuple<unsigned int /*face*/, unsigned int /*q*/, unsigned int /*v*/> Id;
   typedef std::map<Id, types::global_dof_index>                                     MapVectorIndex;
-  typedef std::vector<std::vector<Tensor<rank, dim, Number>>> ArrayVectorTensor;
+  typedef std::vector<Tensor<rank, dim, Number>>                                    ArrayTensors;
```
nfehn, line 41: The new name (with plural s) is not consistent with the other names that use the singular form, so I would suggest `ArrayTensor` for now.

include/exadg/functions_and_boundary_conditions/interface_coupling.h
```diff
@@ -23,438 +23,59 @@
...
     (n_components == 1) ? 0 : ((n_components == dim) ? 1 : numbers::invalid_unsigned_int);
 
-  typedef InterfaceCoupling<dim, n_components, Number> This;
-
-  typedef LinearAlgebra::distributed::Vector<Number> VectorType;
-  typedef LinearAlgebra::distributed::Vector<double> VectorTypeDouble;
-  typedef FaceIntegrator<dim, n_components, Number>  Integrator;
-  typedef std::pair<unsigned int, unsigned int>      Range;
+  using VectorType       = LinearAlgebra::distributed::Vector<Number>;
+  using VectorTypeDouble = LinearAlgebra::distributed::Vector<double>;
```
nfehn, line 40: I think `DEALII_VERSION_GTE` could/should also be added here?

include/exadg/functions_and_boundary_conditions/interface_coupling.h
nfehn: @peterrum Did you introduce this because you removed the communication step above? Or are these two independent steps?

I would argue that setup prepares everything so that the boundary conditions can be used/evaluated. This would imply that a DoF-vector handed over to the function is indeed evaluated. If the "exchange of data" is done in the old code, we should also do it here in my opinion, in order to not introduce bugs. Right now, I do not remember all the details of the implementation.

include/exadg/functions_and_boundary_conditions/interface_coupling.h
nfehn: I am also not sure and don't remember where this comes from. For example, I wonder about `11`? @peterrum you probably know best what `communicator.process()` was doing.

include/exadg/functions_and_boundary_conditions/interface_coupling.h
```diff
@@ -533,175 +142,89 @@ class InterfaceCoupling
...
-        map_cache_src.find(quadrature)->second;
-      std::map<mpi_rank, ArrayVectorTensor> & mpi_map_solution_src =
-        map_solution_src.find(quadrature)->second;
-
+      array_solution_dst.resize(array_q_points_dst.size(), Tensor<rank, dim, double>());
 
       /*
        * 2. Communication: receive and cache quadrature points of other ranks,
        *    redundantly store own q-points (those that are needed)
        */
```
nfehn, line 150: You removed 3. and 4., so we need to make sure that this comment is up-to-date. Please go through your changes from a "comment" point of view.

include/exadg/functions_and_boundary_conditions/interface_coupling.h
nfehn: Is this comment still relevant? It seems as if you already removed the vector of vectors?

include/exadg/functions_and_boundary_conditions/interface_coupling.h
nfehn: const as a trailing keyword is defined in our ExaDG guidelines in the wiki. So I agree. I comment on the other aspect somewhere else.

include/exadg/functions_and_boundary_conditions/interface_coupling.h
```diff
@@ -533,175 +142,89 @@ class InterfaceCoupling
...
+      auto const result =
+        VectorTools::point_values<n_components>(map_communicator[quadrature],
+                                                *dof_handler_src,
+                                                dof_vector_src_double,
+                                                VectorTools::EvaluationFlags::avg);
+#endif
+
+      for(unsigned int i = 0; i < result.size(); ++i)
+        for(unsigned int c = 0; c < n_components; ++c)
+          map_solution_dst[quadrature][i][c] = result[i][c];
```
nfehn, line 206: Because of this line, I suggest to not use `auto` above, but spell out the type explicitly.

include/exadg/functions_and_boundary_conditions/interface_coupling.h
```diff
@@ -533,175 +142,89 @@ class InterfaceCoupling
...
-  std::vector<quad_index>                  quad_rules_dst;
+  std::vector<quad_index> quad_rules_dst;
 
   mutable std::map<quad_index, MapIndex>              map_index_dst;
   mutable std::map<quad_index, ArrayQuadraturePoints> map_q_points_dst;
-  mutable std::map<quad_index, ArrayVectorTensor>     map_solution_dst;
+  mutable std::map<quad_index, ArrayTensors>          map_solution_dst;
 
-  std::map<quad_index, InterfaceCommunicator<dim, dim>> map_communicator;
+  std::map<quad_index, Utilities::MPI::RemotePointEvaluation<dim>> map_communicator;
```
nfehn, line 220: I think the name `map_communicator` is not longer reasonable. I suggest to use `map_evaluator`.

include/exadg/functions_and_boundary_conditions/interface_coupling.h
nfehn: Here, I would tend to agree with @kronbichler to consistently write `_in` (then also for the last argument `tolerance`).

I think technically you are correct that triangulation is not needed as argument and can be queried via the dof_handler. I think I used this as an argument because the data structures are also used in the sense of mesh = {tria, mapping}, and not only dof_handler = {tria, finite_element}. So the code is somewhat clearer with separate arguments in my opinion. Overall, I am not fully convinced regarding dealii's design choice to query the triangulation via the dof_handler, but this would have to be discussed elsewhere (where my opinion will be of no relevance ... In general, I am of the opinion that we do not have to repeat design choices from dealii just because dealii decided to do it like this.). It appears somewhat arbitrary to me that dealii queries tria via dof_handler, but always uses mapping as a separate parameter. As we currently do not strive for making dealii "replacable" within ExaDG, I can live with removing this parameter.

kronbichler: The main problem I see is that the mesh underlying the `DoFHandler` can be incompatible with the triangulation explicitly handed in, causing the risk for errors in case one accidentally hands in something different. We could of course check that again, but at what cost? I agree that choices do not have to be repeated, but here it would merely make our lives more complicated due to repeating code that needs to be maintained.

The reason why `DoFHandler` keeps a pointer to the triangulation is partly historical as deal.II evolved ~20 years ago. The more important reason is that `DoFHandler` in itself would merely be a storage location for numbers of the unknowns on cells, but those numbers make no sense without the cells (or faces/lines/vertices) provided by the triangulation, so `get_triangulation` is a response to the underlying grid to interpret numbers stored in `DoFHandler`. The same could be said about a `Mesh` or its fundamental ingredient `Mapping`, but here the case is not as straight-forward because
- Non-bilinear geometries were not a central design goal 20 years ago and came around later (even though before I started using deal.II), so it came by as an "addition" to existing data structures.
- The basic mappings do not *store* anything related to the mesh, but rather compute the necessary parts on demand, thus taking a different role than a `DoFHandler`, as they can be fed by an individual `cell` and do not need a `Triangulation` to be useful (the correct term would be a "mesh").
- There is sometimes a desire to create "different" mappings as one performs different steps in a simulation (e.g. two different polynomial degrees of the geometry to adhere to different accuracies of quadrature formulas), and sometimes one might be tempted to even use the same `DoFHandler` with those different mappings (albeit one is likely in the territory of variational crimes).

Either way, I see no *clean* solution that meets the typical requirements and would provide a clear advantage over the pair `Mapping, DoFHandler` (the latter implicitly containing `Triangulation`). The combination is error prone mostly because deal.II also allows users not to provide the mapping argument in several contexts, thus one can miss out these parts. Always using three arguments `Triangulation, Mapping, DoFNumbers` (I intentionally use a different name) or `Mesh, DoFHandler` (redundant triangulation; can be kept in sync with some signal mechanisms) or a super data structure `MeshDoFHandler` (consisting of `Triangulation, Mapping, DoFHandler`; duplicating some of the content, especially in more complicated problems) all have their downsides, too.

nfehn: I agree that we made the most reasonable decision here. But don't you think DoFHandler is also kind of a super data structure, given that DoFHandler itself provides cell iterator functions for example (as another example, we already discussed the MG functionality as part of DoFHandler elsewhere)? Where is "single-responsibility" here? I wonder whether one could have been more strict about what is functionality belonging to triangulation and/or to "dof numbers". To provide just one more example, consider the public function `DoFHandler::get_communicator()`, where the documentation says "Return MPI communicator used by the underlying triangulation. ". In my opinion this is simply inconsequent, independently of the design choices made some time ago and discussed above. To keep interfaces lean, one should enforce `dof_handler.get_triangulation().get_communicator()` instead of such short cuts that blur "single responsibility".

> numbers make no sense without the cells (or faces/lines/vertices) provided by the triangulation, so get_triangulation is a response to the underlying grid to interpret numbers stored in DoFHandler. 

This sounds like a solution that you are describing here, i.e. `triangulation` has to be passed to public functions of `DoFHandler` where information on cells/faces/etc is needed, but no pointer to the triangulation is stored, at least no public getter `get_triangulation()`. In this sense, the dof-handler would be some kind of helper/utility functionality (and if data needs to be stored, it has to be enforced in debug mode that one is using the correct data structures).

nfehn:
@peterrum Many thanks for these changes. This PR is a good example of what we describe in ExaDG's readme under "Philosophy" (https://github.com/exadg/exadg#philosophy).

nfehn:
These changes should be tested "manually" by running FSI examples which use this functionality.

peterrum:
@nfehn Thanks for the review. I'll try to integrate the suggestions later today or tomorrow!

include/exadg/functions_and_boundary_conditions/interface_coupling.h
peterrum: Yes, it has been removed.

peterrum:
@nfehn I have applied the comments. I have run the `cylinder_with_flag` test case. I can run the other benchmarks tomorrow as well.

include/exadg/functions_and_boundary_conditions/interface_coupling.h
peterrum: > Did you introduce this because you removed the communication step above? Or are these two independent steps?

Yes this is the replacement of the old communication step. I am still wondering what the purpose of it is. I guess to set the initial values. But why is it done during initialization of the communication class and not in a separate call?

nfehn: I think this originates from different perspectives. From my point of view, `InterfaceCoupling` is more than the communication class. In other words, the communication class is (one, the main, ...) part of the `InterfaceCoupling`. Hence, `InterfaceCoupling::setup()` wants to make sure that everything is initialized. So for now I believe we should keep this definition, but in principle we could also redefine it.

nfehn:
In principle, one example should be enough because the algorithm in terms of interface coupling is the same for all examples. However, since the cylinder example is a 2d example, it could make sense to run a 3d example as well, e.g. pressure_wave.

nfehn (approved):

PR #127, opened 2022-01-07
Copy vectors in MGTransferC

peterrum:
... else the partitioners of the vectors are incompatible with MatrixFree.

### DIFF
--- a/include/exadg/solvers_and_preconditioners/multigrid/transfers/mg_transfer_c.cpp
+++ b/include/exadg/solvers_and_preconditioners/multigrid/transfers/mg_transfer_c.cpp
@@ -100,10 +100,6 @@ MGTransferC<dim, Number, VectorType, components>::do_restrict_and_add(VectorType
   data_composite.initialize_dof_vector(src_temp, 1);
   src_temp.copy_locally_owned_data_from(src);
 
-  VectorType dst_temp;
-  data_composite.initialize_dof_vector(dst_temp, 0);
-  dst_temp.copy_locally_owned_data_from(dst);
-
   for(unsigned int cell = 0; cell < data_composite.n_cell_batches(); ++cell)
   {
     fe_eval_cg.reinit(cell);
@@ -114,12 +110,10 @@ MGTransferC<dim, Number, VectorType, components>::do_restrict_and_add(VectorType
     for(unsigned int i = 0; i < fe_eval_cg.static_dofs_per_cell; i++)
       fe_eval_cg.begin_dof_values()[i] = fe_eval_dg.begin_dof_values()[i];
 
-    fe_eval_cg.distribute_local_to_global(dst_temp);
+    fe_eval_cg.distribute_local_to_global(dst);
   }
 
-  dst_temp.compress(VectorOperation::add);
-
-  dst.copy_locally_owned_data_from(dst_temp);
+  dst.compress(VectorOperation::add);
 }
 
 template<int dim, typename Number, typename VectorType, int components>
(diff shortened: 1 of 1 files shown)


include/exadg/solvers_and_preconditioners/multigrid/transfers/mg_transfer_c.cpp
```diff
@@ -96,20 +96,30 @@ MGTransferC<dim, Number, VectorType, components>::do_restrict_and_add(VectorType
   FEEvaluation<dim, degree, 1, components, Number> fe_eval_cg(data_composite, 0);
   FEEvaluation<dim, degree, 1, components, Number> fe_eval_dg(data_composite, 1);
 
+  VectorType src_temp;
+  data_composite.initialize_dof_vector(src_temp, 1);
+  src_temp.copy_locally_owned_data_from(src);
```
kronbichler, line 101: Do we need this one? In case we never use anything but an actual DG space here, all DoFs on locally owned cells will be locally owned, so we do not need the copy here. The real difference should be in the CG space where the ghost space might or might not match. (The same is true for the prolongate function below.) One side effect would be that the more expensive vector with more DoFs does not get copied here, saving memory transfer.

include/exadg/solvers_and_preconditioners/multigrid/transfers/mg_transfer_c.cpp
```diff
@@ -96,20 +96,30 @@ MGTransferC<dim, Number, VectorType, components>::do_restrict_and_add(VectorType
...
+
+  VectorType dst_temp;
+  data_composite.initialize_dof_vector(dst_temp, 0);
+  dst_temp.copy_locally_owned_data_from(dst);
```
kronbichler, line 105: Note also that we do not call `update_ghost_values` here, so we've hardcoded the locally owned partitioner already anyway.

include/exadg/solvers_and_preconditioners/multigrid/transfers/mg_transfer_c.cpp
peterrum: OK. I was able to remove a set of temporal vectors.

kronbichler: Now I'm confused, because I would have expected the exact opposite to be possible, i.e., use no temporary vector for `src` in `restrict` (which corresponds to the DG space), but a temporary vector for `dst` (the continuous FEM space), whereas I would expect a temporary vector for `src` in `prolongate`. Can you check whether the current version really works with MPI?

peterrum: Yes, this is working. The problem is that the DG vector has ghost values, but MatrixFree/FEEvaluation is expecting a vector without ghost values. I guess this is related to the `AdditionalData`. Playing around with it could be an option, but also a bit brittle, since ghosting also depends on whether all faces are owned...

kronbichler: I see, thanks for the background. So it is rather the assertion in https://github.com/dealii/dealii/blob/e37add0c8852ca92525961c144fc2df54a146029/include/deal.II/matrix_free/fe_evaluation.h#L3377-L3393 that is too conservative: If all indices on the present cell are indeed locally owned (not ghosted), we would only need to check whether the locally owned range coincides, as the locally owned part is guaranteed to be ok. While trying to get this more precise in deal.II would be useful to some context, I guess we should switch this to the deal.II global coarsening setup at some point anyway, so we can keep this for now and switch later.

kronbichler (approved):

nfehn:
I have seen this only now. Thanks for fixing this! @kronbichler, you added the bug label: It is somewhat difficult for me to see why there is a bug and where/when it occurred. A comment in the code like "we need to introduce a temporary vector because ..." would also be helpful.

peterrum:
As far as I see no simulation using c-transfer should run successfully in parallel and debug mode. The bug was introduced during the performance optimization for the SC paper (https://github.com/peterrum/exadg/commit/cad426c55efc03de24966d590b6f55b9e72e4ddc#diff-8874767505807783152bf447bbadcc9dcaeb050d34119953c5717af3e4edffb9), since the copy is strictly not needed. However, we use `MatrixFree`/`FEEvaluation` here, which have certain requirements regarding partitioners.

kronbichler:
> The bug was introduced during the performance optimization for the SC paper (peterrum@cad426c#diff-8874767505807783152bf447bbadcc9dcaeb050d34119953c5717af3e4edffb9), since the copy is strictly not needed. However, we use MatrixFree/FEEvaluation here, which have certain requirements regarding partitioners.

Indeed, it is not a functional *bug* that got fixed, because all code that we used was correct, but a conservative assertion mechanism in deal.II that makes this to an observed bug.

PR #131, opened 2022-01-12
Simplex support for Poisson operator

### LINKED ISSUE #103: Exemplarily show use of simplex-meshes for Poisson-module

nfehn:
Closes #103 

This PR replaces #110. Compared to #110, I added more Asserts in `Poisson::Operator`. Other parts of #110 have already been realized in other PRs.

I added a TODO in the code, since we still discuss in #110 how to deal with `distribute_mg_dofs()`.

This PR is still work-in-progress, but discussion might be helpful already right now.

### DIFF
--- a/include/exadg/poisson/spatial_discretization/operator.cpp
+++ b/include/exadg/poisson/spatial_discretization/operator.cpp
@@ -22,6 +22,7 @@
 // deal.II
 #include <deal.II/fe/fe_dgq.h>
 #include <deal.II/fe/fe_q.h>
+#include <deal.II/fe/fe_simplex_p.h>
 #include <deal.II/fe/fe_system.h>
 #ifdef DEAL_II_WITH_PETSC
 #  include <deal.II/lac/petsc_vector.h>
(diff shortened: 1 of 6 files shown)


nfehn:
@peterrum Did I use something that was not part of dealii-9.3 (probably all_reference_cells_are_simplex()?)? What should I do best?

include/exadg/poisson/spatial_discretization/operator.cpp
```diff
@@ -76,20 +77,52 @@ Operator<dim, Number, n_components>::distribute_dofs()
...
+        fe = std::make_shared<FESystem<dim>>(FE_SimplexDGP<dim>(param.degree), dim);
+      else
+        AssertThrow(false, ExcNotImplemented());
+    }
     else if(param.spatial_discretization == SpatialDiscretization::CG)
-      fe = std::make_shared<FESystem<dim>>(FE_Q<dim>(param.degree), dim);
+    {
+      if(this->grid->triangulation->all_reference_cells_are_hyper_cube())
+        fe = std::make_shared<FESystem<dim>>(FE_Q<dim>(param.degree), dim);
+      else if(this->grid->triangulation->all_reference_cells_are_simplex())
```
peterrum, line 117: Just move the content of https://github.com/dealii/dealii/blob/be72dbb5a5f97aee1c8cf330b66664c848176fba/source/grid/tria.cc#L14668-L14669 into a free function: `bool all_reference_cells_are_simplex(Triangulation const & tria);`. You get `reference_cells` by calling `Triangulation::get_reference_cells()`.

nfehn:
See PR #147 for aspects related to triangulation/mapping with simplex meshes

nfehn:
This line, https://github.com/exadg/exadg/blob/b36a6b64704c3fe5ebdefdb631b1e4d306682689/include/exadg/grid/grid.h#L76, is currently a problem for the simplex case.

We need to find a design how to set the type of mesh during construction of `Grid`. Previously, the mapping was set up in the application, but this line of code was then essentially copied to every application, which is redundant code.

nfehn:
I think I will merge this PR so that @necioglu can continue working. To solve the above problem, my suggestion would be to equip `GridData`, see https://github.com/exadg/exadg/blob/b36a6b64704c3fe5ebdefdb631b1e4d306682689/include/exadg/grid/grid_data.h#L30,
 with an enum

```C++
enum class ElementType
{
Hypercube,
Simplex
};
```
so that the grid class can correctly set up the mapping depending on the element type.

To some extent, this is redundant information, but I would simply try out whether this leads to problems.

PR #132, opened 2022-01-12
Remove distribute_mg_dofs from field solvers

kronbichler:
Following what we discussed in https://github.com/exadg/exadg/pull/110#discussion_r783097083, I realized that we always create a `DoFHandler` copy inside the multigrid solver for which we call `distribute_mg_dofs`, https://github.com/exadg/exadg/blob/f2ff9d113b0898c8eb5d089b2ee5f0d3112c9297/include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp#L799-L808 , because we do https://github.com/exadg/exadg/blob/f2ff9d113b0898c8eb5d089b2ee5f0d3112c9297/include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp#L157-L168 .

Also, we only fill the constraints on the levels on this `DoFHandler`, which is called in the right order, https://github.com/exadg/exadg/blob/master/include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp#L812

I ran tests in the convection-diffusion, structure and Poisson module, which all work for me. @nfehn do you agree or do you know what to check in a broader sense?

### DIFF
--- a/include/exadg/convection_diffusion/spatial_discretization/operator.cpp
+++ b/include/exadg/convection_diffusion/spatial_discretization/operator.cpp
@@ -289,12 +289,10 @@ Operator<dim, Number>::distribute_dofs()
 {
   // enumerate degrees of freedom
   dof_handler.distribute_dofs(fe);
-  dof_handler.distribute_mg_dofs();
 
   if(needs_own_dof_handler_velocity())
   {
     dof_handler_velocity->distribute_dofs(*fe_velocity);
-    dof_handler_velocity->distribute_mg_dofs();
   }
 
   unsigned int const ndofs_per_cell = Utilities::pow(param.degree + 1, dim);
--- a/include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
+++ b/include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
@@ -235,11 +235,8 @@ SpatialOperatorBase<dim, Number>::distribute_dofs()
 {
   // enumerate degrees of freedom
   dof_handler_u.distribute_dofs(*fe_u);
-  dof_handler_u.distribute_mg_dofs();
   dof_handler_p.distribute_dofs(fe_p);
-  dof_handler_p.distribute_mg_dofs();
   dof_handler_u_scalar.distribute_dofs(fe_u_scalar);
-  dof_handler_u_scalar.distribute_mg_dofs(); // probably, we don't need this
 
   unsigned int const ndofs_per_cell_velocity = Utilities::pow(param.degree_u + 1, dim) * dim;
   unsigned int const ndofs_per_cell_pressure =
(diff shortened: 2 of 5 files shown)


peterrum (approved):
Looks reasonable. The question is what to do with:

https://github.com/exadg/exadg/blob/2e767595001f2bf63b5be992a3fbbae219744201/include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp#L808

But I guess we can discuss this later, once we try to make multgrid work with simplices.

kronbichler:
Yes, but that is the relevant one in the sense that all multigrid classes rely on it. I think for now we need to disable multigrid in #131 in the case we have simplices.

nfehn:
So we just did unnecessary work in the `SpatialOperator`s?

> I ran tests in the convection-diffusion, structure and Poisson module, which all work for me. @nfehn do you agree or do you know what to check in a broader sense?

We should think once about what happens for operators with additional velocity transport field in combination with multigrid preconditioners. In these cases, the velocity vector needs to be restricted to coarser meshes, and we set up corresponding data structures, see https://github.com/exadg/exadg/blob/f2ff9d113b0898c8eb5d089b2ee5f0d3112c9297/include/exadg/convection_diffusion/preconditioners/multigrid_preconditioner.cpp#L211-L226. Hence, the same function as highlighted above should be called in the end (which calls `distribute_mg_dofs()` for this multigrid-variant).

kronbichler:
Yes, it seems that we fixed this in some of our refactorings. Regarding the convection-diffusion case: Indeed, this is what I guessed as well. But I could not find a test case with that code path - do we have one that I just didn't see?

nfehn:
> Yes, but that is the relevant one in the sense that all multigrid classes rely on it. I think for now we need to disable multigrid in #131 in the case we have simplices.

Actually, we should only have to deactivate "standard multigrid": Global coarsening multigrid should be possible?

@peterrum has previously written 

> To be able to use multigrid [for simplices], one needs to use global coarsening for now.

We already have such an if-else-branch in our multigrid implementation. So the else-branch (which already has an assert that no hanging nodes occur) should add the assert

```AssertThrow(triangulation->all_reference_cells_are_hyper_cube(), ExcMessage("This multigrid implementation is currently only available for hyper cube elements."));```

kronbichler:
Right, then let me add a second commit that adds this assertion.

nfehn (approved):

nfehn:
> But I could not find a test case with that code path - do we have one that I just didn't see?

I think you would have to test this manually for one of the `incompressible_flow_with_transport` applications, formulating the convective term implicitly and using multigrid as preconditioner (with multigrid applied to both convective and diffusive terms; I mention this because we have the option to let multigrid operate only on symmetric terms (given that smoothers might be more involved in the non-symmetric case)).

nfehn:
@kronbichler Please let me know if this PR is ready to merge from your side (I think last week we had the open point of velocity vectors in the multigrid algorithm).

kronbichler:
I wanted to have a look at the example you mentioned, but did not yet find the time to do so. But it should be fine to be merged in my opinion.

PR #134, opened 2022-01-12
shift n_subdivisions_1d_hypercube from ApplicationBase to GridData

nfehn:
Responsibilities should be clearer with this change. Shifting this parameter to the grid class has the advantage that the grid class knows how to deal with hex vs. simplex meshes for example.


nfehn:
@peterrum @kronbichler could one of you have a brief look at the changes, because I would like to continue on this basis?

peterrum (approved):

PR #138, opened 2022-01-20
a minor simplification in FTI::Driver related to ALE

nfehn:
If `ale_formulation == true` it is ensured automatically that

https://github.com/exadg/exadg/blob/683595d85527a036d36bc800e30f4f5261ac91a5/include/exadg/incompressible_navier_stokes/user_interface/parameters.cpp#L264-L267


nfehn:
I realized that the `ConvDiff` module didn't do this check in `Parameters::check()`, so I added it.

peterrum (approved):

PR #141, opened 2022-01-20
ALE: check for restarted simulation in postprocessing()

nfehn:
Strictly speaking, `time_step_number == 1` is not identical to `time == start_time`. The reason behind is that in our implementation the time step number always start with 1 after a restart (which has, of course, reasons). 

Add `if(... && !restarted_simulation)` to correctly identify the initial time.

### DIFF
--- a/include/exadg/convection_diffusion/time_integration/time_int_bdf.cpp
+++ b/include/exadg/convection_diffusion/time_integration/time_int_bdf.cpp
@@ -714,9 +714,10 @@ TimeIntBDF<dim, Number>::postprocessing() const
   Timer timer;
   timer.restart();
 
-  // the mesh has to be at the correct position to allow a computation of
-  // errors at start_time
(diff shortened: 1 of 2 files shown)


peterrum:
> with 1 after a restart (which has, of course, reasons).

Why don't you restart with the same time step number?

peterrum (approved):

nfehn:
> Out of curiosity, why don't you restart with the same time step number?

I think one could realize both variants. The present one is maybe simpler in the sense that the time integrator does not need an extra member variable `unsigned int time_step_number_start` and that no time step number has to be written to the restart files. A technical issue might be that this member variable `time_step_number_start` should be defined `const`, but that restart related content is currently done in `TimeIntBDFBase<Number>::setup()` and not in the constructor. Again one might change this, but this requires probably more refactoring of the time integrator design. I am undecided about which variant is better, and I also see benefits to continue with the same time step number after the restart as you suggested.

PR #145, opened 2022-01-27
remove "using namespace dealii"

nfehn:
I think it is advantageous to see dependencies on dealii explicitly in ExaDG. We do not want to expect ExaDG-developers to also be dealii-developers.

### DIFF
--- a/applications/compressible_navier_stokes/couette/application.h
+++ b/applications/compressible_navier_stokes/couette/application.h
@@ -26,8 +26,6 @@ namespace ExaDG
 {
 namespace CompNS
 {
-using namespace dealii;
-
 // problem specific parameters
 double const DYN_VISCOSITY = 1.0e-2;
 double const GAMMA         = 1.4;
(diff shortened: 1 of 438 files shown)


nfehn:
I updated the wiki-page "coding conventions"

nfehn:
I updated the wiki-page "coding conventions"

kronbichler (approved):
I agree on this change from an encapsulation perspective. Some of the places get a bit tedious to read, such as `dealii::Point` or `dealii::Tensor` that are rather unambiguous. We can always decide to pull certain data structures into ExaDG's namespace, such as we do for `CellIntegrator = dealii::FEEvaluation` (or decide not do it, depending on how our preferences develop). I did not check all occurrences but I assume that the compiler will signal any opportunity missed out (and that you grepped for `using namespace dealii` to eliminate all).

nfehn:
One should check the files so that comments and std::cout output is not affected, as I did the changes somewhat automatically

PR #146, opened 2022-01-31
new Grid functionality: perform local refinements

nfehn:
I shifted this functionality from our application project exadg-applications to ExaDG, since the functionality is actually more general.

I did not change the implementation, so that this PR is only a refactoring in combination with the related PR in exadg-applications. (The label "refactoring" is to be understood in combination with the app project exadg-applications).

As indicated by the documentation of the new function, the design should be changed from `std::vector` to `std::map` in the future.


bergbauer (approved):
LGTM!

PR #147, opened 2022-02-01
Generalize and simplify create grid functionality

nfehn:
The function `create_grid()` in the various applications often contains the same lines of code. In particular, in case of `parallel::fullydistributed::Tria`, the "complicated" call to `dealii::TriangulationDescription::???` with many lambdas is "duplicated code" in many applications, where we essentially always use the same setting.

This PR suggests to generalize this functionality and equip the `Grid` class with a function `create_triangulation()`, which takes a lambda function (constructing the coarse triangulation) as an argument. This new function distinguishes between `Serial/Distributed/FullyDistributed` triangulation types according to parameters specified in `GridData`. Moreover, local and global refinements are done by this new function. 

This PR therefore generalizes and simplifies the `UserInterface`.

<del> Currently, this PR only contains the implementation of the new function, but it is not used so far in ExaDG applications. </del>

### DIFF
--- a/include/exadg/grid/enum_types.cpp
+++ b/include/exadg/grid/enum_types.cpp
@@ -52,6 +52,27 @@ enum_to_string(TriangulationType const enum_type)
 }
 
 std::string
+enum_to_string(PartitioningType const enum_type)
+{
+  std::string string_type;
+
+  switch(enum_type)
+  {
+    case PartitioningType::Metis:
+      string_type = "Metis";
+      break;
+    case PartitioningType::p4est:
+      string_type = "p4est";
+      break;
+    default:
+      AssertThrow(false, dealii::ExcMessage("Not implemented."));
+      break;
+  }
+
+  return string_type;
+}
+
+std::string
 enum_to_string(MappingType const enum_type)
 {
   std::string string_type;
--- a/include/exadg/grid/enum_types.cpp
+++ b/include/exadg/grid/enum_types.cpp
@@ -61,8 +61,8 @@ enum_to_string(PartitioningType const enum_type)
     case PartitioningType::Metis:
       string_type = "Metis";
       break;
-    case PartitioningType::p4est:
-      string_type = "p4est";
+    case PartitioningType::z_order:
+      string_type = "z_order";
       break;
     default:
       AssertThrow(false, dealii::ExcMessage("Not implemented."));
--- a/include/exadg/grid/enum_types.h
+++ b/include/exadg/grid/enum_types.h
@@ -46,6 +46,18 @@ std::string
 enum_to_string(TriangulationType const enum_type);
 
 /*
+ * Partitioning type (relevant for fully-distributed triangulation)
+ */
+enum class PartitioningType
+{
+  Metis,
+  p4est
+};
+
+std::string
+enum_to_string(PartitioningType const enum_type);
+
+/*
  *  Mapping type (polynomial degree)
  */
 enum class MappingType
--- a/include/exadg/grid/enum_types.h
+++ b/include/exadg/grid/enum_types.h
@@ -51,7 +51,7 @@ enum_to_string(TriangulationType const enum_type);
 enum class PartitioningType
 {
   Metis,
-  p4est
+  z_order
 };
 
 std::string
--- a/include/exadg/grid/grid.h
+++ b/include/exadg/grid/grid.h
@@ -71,6 +71,7 @@ public:
     }
 
     // mapping
+    // TODO SIMPLEX: this will not work in case of simplex meshes (-> use MappingFE)
     mapping = std::make_shared<dealii::MappingQGeneric<dim>>(data.mapping_degree);
   }
 
@@ -166,6 +167,7 @@ private:
 
       unsigned int const group_size = 1;
 
+      // TODO SIMPLEX: this will not work in case of simplex meshes
       auto const description = dealii::TriangulationDescription::Utilities::
         create_description_from_triangulation_in_groups<dim, dim>(
           serial_grid_generator,
--- a/include/exadg/grid/grid.h
+++ b/include/exadg/grid/grid.h
@@ -102,8 +102,8 @@ public:
       };
 
       auto const serial_grid_partitioner = [&](dealii::Triangulation<dim, dim> & tria_serial,
-                                              MPI_Comm const                    comm,
-                                              unsigned int const                group_size) {
+                                               MPI_Comm const                    comm,
+                                               unsigned int const                group_size) {
         (void)group_size;
         if(data.partitioning_type == PartitioningType::Metis)
         {
--- a/include/exadg/grid/grid.h
+++ b/include/exadg/grid/grid.h
@@ -140,6 +140,58 @@ public:
     }
   }
 
+  void
+  create_triangulation(
+    GridData const &                                          data,
+    std::function<void(dealii::Triangulation<dim> &)> const & create_fine_triangulation)
+  {
+    if(data.triangulation_type == TriangulationType::Serial or
+       data.triangulation_type == TriangulationType::Distributed)
+    {
+      create_fine_triangulation(*triangulation);
+    }
+    else if(data.triangulation_type == TriangulationType::FullyDistributed)
+    {
+      auto const serial_grid_partitioner = [&](dealii::Triangulation<dim, dim> & tria_serial,
+                                               MPI_Comm const                    comm,
+                                               unsigned int const                group_size) {
+        (void)group_size;
+        if(data.partitioning_type == PartitioningType::Metis)
+        {
+          dealii::GridTools::partition_triangulation(dealii::Utilities::MPI::n_mpi_processes(comm),
+                                                     tria_serial);
+        }
+        else if(data.partitioning_type == PartitioningType::p4est)
+        {
+          dealii::GridTools::partition_triangulation_zorder(
+            dealii::Utilities::MPI::n_mpi_processes(comm), tria_serial);
+        }
+        else
+        {
+          AssertThrow(false, dealii::ExcNotImplemented());
+        }
+      };
+
+      unsigned int const group_size = 1;
+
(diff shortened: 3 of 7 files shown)


include/exadg/grid/enum_types.cpp
```diff
@@ -51,6 +51,27 @@ enum_to_string(TriangulationType const enum_type)
...
+{
+  std::string string_type;
+
+  switch(enum_type)
+  {
+    case PartitioningType::Metis:
+      string_type = "Metis";
+      break;
+    case PartitioningType::p4est:
+      string_type = "p4est";
```
peterrum, line 65:
```suggestion
    case PartitioningType::z_order:
      string_type = "z_order";
```

include/exadg/grid/enum_types.h
```diff
@@ -45,6 +45,18 @@ enum class TriangulationType
 std::string
 enum_to_string(TriangulationType const enum_type);
 
+/*
+ * Partitioning type (relevant for fully-distributed triangulation)
+ */
+enum class PartitioningType
+{
+  Metis,
+  p4est
```
peterrum, line 54:
```suggestion
  z_order
```

include/exadg/grid/grid.h
```diff
@@ -73,6 +74,124 @@ class Grid
     mapping = std::make_shared<dealii::MappingQGeneric<dim>>(data.mapping_degree);
   }
 
+  void
+  create_and_refine_triangulation(
+    GridData const &                                          data,
+    std::vector<int> const &                                  vector_local_refinements,
+    std::function<void(dealii::Triangulation<dim> &)> const & create_coarse_triangulation)
```
peterrum, line 81: What is the purpose to have two versions of `create_and_refine_triangulation()`? I would switch the parameters and only have one version..

include/exadg/grid/grid.h
```diff
@@ -73,6 +74,124 @@ class Grid
...
+  {
+    if(data.triangulation_type == TriangulationType::Serial or
+       data.triangulation_type == TriangulationType::Distributed)
+    {
+      create_coarse_triangulation(*triangulation);
+
+      if(vector_local_refinements.size() > 0)
```
peterrum, line 88: Why not do the check within `refine_local`?

include/exadg/grid/grid.h
```diff
@@ -73,6 +74,124 @@ class Grid
     mapping = std::make_shared<dealii::MappingQGeneric<dim>>(data.mapping_degree);
   }
 
+  void
+  create_and_refine_triangulation(
+    GridData const &                                          data,
+    std::vector<int> const &                                  vector_local_refinements,
```
peterrum, line 80:
```suggestion
    std::vector<unsigned int> const &                                  vector_local_refinements,
```

include/exadg/grid/grid.h
```diff
@@ -73,6 +74,124 @@ class Grid
...
+      unsigned int const group_size = 1;
+
+      auto const description = dealii::TriangulationDescription::Utilities::
+        create_description_from_triangulation_in_groups<dim, dim>(
+          serial_grid_generator,
+          serial_grid_partitioner,
+          triangulation->get_communicator(),
+          group_size,
+          dealii::Triangulation<dim>::none,
+          dealii::TriangulationDescription::construct_multigrid_hierarchy);
```
peterrum, line 133: This will not not work with simplex/mixed meshes.

include/exadg/grid/grid.h
```diff
@@ -73,6 +74,124 @@ class Grid
...
+          dealii::GridTools::partition_triangulation_zorder(
+            dealii::Utilities::MPI::n_mpi_processes(comm), tria_serial);
+        }
+        else
+        {
+          AssertThrow(false, dealii::ExcNotImplemented());
+        }
+      };
+
+      unsigned int const group_size = 1;
```
peterrum, line 124: I guess we could turn this into an input argument.

peterrum (approved):
Looks good.

include/exadg/grid/grid.h
nfehn: It is described here https://gitlab.lrz.de/exadg/exadg-applications/-/merge_requests/62#note_1620973.

Ideally, refinement would always be done after creating the coarse/serial triangulation. However, this is not possible for certain dealii grid-manipulation tools (working with manifolds). Therefore, I suggest to support both variants in our code.

include/exadg/grid/grid.h
nfehn: what exactly? the whole function call or just certain parameters?

include/exadg/grid/grid.h
nfehn: Yes. The question is whether we should do it now or later when we need it. In the past, I only introduced parameters when we really want to make use of them.

include/exadg/grid/grid.h
nfehn: I think this is a matter of taste. I thought for a user (who might look into the `Grid` class, but not deaper), it is good to see when local refinement is invoked, since `vector_local_refinements` is a data structure that needs to be provided by the user. Otherwise, this would be somewhat hidden from the user, and in my opinion the code would become more difficult to understand.

But we can also realize the other variant if a majority of developers prefers that variant.

include/exadg/grid/grid.h
nfehn: Technically, we could combine both variants to a single function (a boolean parameter decides whether refinement is invoked or not), but I have the fear that it becomes increasingly difficult for the user to understand how to use the code. Note that the main purpose of this PR is to simplify the use of p:f:T for "unexperienced" users.

In principle I agree with you. For now, my suggestion would be that we do not consider the interface of `Grid` fixed once and for all, but get some experience with the new functionality and maybe improve it in the future. Note that there are other functionalities that we want to include such as simplex meshes and reading grids from files. Hence, I expect the `Grid` class to be changed and refactored in the near future.

I have the following idea: We keep two functions in the public interface, but combine the two functions to a single one in the implementation (i.e. we create a private member function).

include/exadg/grid/enum_types.cpp
nfehn: done

include/exadg/grid/grid.h
nfehn: Let me also mention that there is an upcoming PR that changes the design from std::vector to std::map (at least this is my suggestion).

nfehn: The function `refine_local()` should already have such an if-statement. So the if-statement here is somehow just for clarity or readability.

include/exadg/grid/grid.h
nfehn: done here https://github.com/exadg/exadg/pull/147/commits/1b8e4c42909bd5d6485c65ad8995f081e7ddfdbe

include/exadg/grid/grid.h
nfehn: done here https://github.com/exadg/exadg/pull/147/commits/1b8e4c42909bd5d6485c65ad8995f081e7ddfdbe

include/exadg/grid/grid.h
nfehn: let's do this later

include/exadg/grid/grid.h
nfehn: I still think it might be good to have an input parameter that decides whether the mesh is hypercube/simplex instead of deducing that information from the triangulation.

Assume we add such an `enum class` parameter to `GridData`, e.g. `enum class ElementType {Hypercube, Simplex};`. Then, the function `create_triangulation()` could check in the end that the created triangulation is consistent with the input parameter. The rest of the code would then work with the parameter `ElementType` and not with `Triangulation` (and `all_cells_are_hypercube/simplex()`).

nfehn:
@peterrum: A question related to this comment https://github.com/exadg/exadg/pull/147#discussion_r797327134 on simplex meshes:

In case of simplex, what about the mapping initialized in the constructor of this class? Which mapping should be used for simplex instead? When calling the constructor, we cannot ask the triangulation whether it is hypercube or simplex since the triangulation has not been created at this point, suggesting that we introduce an input parameter {hypercube,simplex}.

peterrum:
> Which mapping should be used for simplex instead? 

`MappingFE` (https://www.dealii.org/developer/doxygen/deal.II/classMappingFE.html) is the one to select. It works for all types of cells, including hypercube.

> Assume we add such an enum class parameter to GridData, e.g. enum class ElementType {Hypercube, Simplex};. Then, the function create_triangulation() could check in the end that the created triangulation is consistent with the input parameter. The rest of the code would then work with the parameter ElementType and not with Triangulation (and all_cells_are_hypercube/simplex()).

I personally have bad experience with such a parameter. I have introduced such a logic in another project(s) and did not turn out to be very beneficial and safe, since the input might contradict itself.

> suggesting that we introduce an input parameter {hypercube,simplex}.

One would need to postpone the creating of the mapping as much as possible.

peterrum (approved):

include/exadg/grid/grid.h
nfehn: @peterrum, what would I have to change? Maybe you could write this to PR #131 dealing with simplex (see comment below, where I suggest to postpone simplex-content in this PR and address it in PR #131).

nfehn:
I introduced "TODO SIMPLEX" comments, but suggest to address this elsewhere, see e.g. PR #131.

include/exadg/grid/grid.h
nfehn: @peterrum Do you agree with the new design (which we can still change in the future)?

nfehn:
@kronbichler do you agree?

kronbichler (approved):
Yes, this looks good. It's a pity that we can't reasonably use `MappingFEField` also in the hex mesh case (it is many times slower and consumes too much memory for higher orders, as there is no tensor production optimization possible for a general `FiniteElement` structure).

PR #152, opened 2022-02-03
rename Mortar -> Cached

nfehn:
The name "mortar" currently used for boundary conditions in surface-coupled problems (e.g. FSI) with partitioned solvers is irritating. I suggest to perform the renaming "Mortar -> Cached" in analogy to the `FunctionCached` class that is used to evaluate this type of boundary conditions.


kronbichler (approved):
This sounds good. The name `Mortar` might have been confusing also in the sense that it sometimes implies a more accurate treatment than just the exchange of node values, be it the support points of the Lagrange polynomials or the Gauss quadrature points. (Say segment-based quadrature, where we split up the integration domain into parts that we integrate over true polynomials on both sides of an interface.)

PR #156, opened 2022-02-04
Remove option IncNS::Parameters::store_previous_boundary_values = false

### LINKED ISSUE #154: Always enforce store_previous_boundary_values = true for IncNS projection methods
Currently, `store_previous_boundary_values = true` is enforced in case of ALE.

In the Eulerian case, also `store_previous_boundary_values = false` can be used in principle. However, this formulation is not generic and requires boundary conditions to be an analytical function of time (precursor simulations and the lung application are examples where this is not the case, since boundary conditions are updated by other routines and are only valid for time t_{n+1} in the current time step, but can not be evaluated correctly for previous times t_{n-i} in the current time step ... One has to keep track of previous boundary values explicitly in that case, which is done by `store_previous_boundary_values = true`). 
Moreover, `store_previous_boundary_values = false` requires the velocity time derivative du/dt to be prescribed analytically as pressure Neumann BC in case of the dual splitting scheme (often termed `PressureBC_dudt` in the ExaDG code). When enforcing `store_previous_boundary_values = true`, the user interface would be unified for all IncNS solvers without special treaments for projection methods.

See Section 2.3.3. (equations (2.41) and (2.42)) in https://mediatum.ub.tum.de/doc/1601025/1601025.pdf for theoretical background (Eulerian formulation).

Therefore, I suggest to only support `store_previous_boundary_values = true` from now on.

nfehn:
closes #154

### DIFF
--- a/include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
+++ b/include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
@@ -153,15 +153,13 @@ SpatialOperatorBase<dim, Number>::fill_matrix_free_data(
                                      field + quad_index_p);
   matrix_free_data.insert_quadrature(dealii::QGauss<1>(param.degree_u + (param.degree_u + 2) / 2),
                                      field + quad_index_u_nonlinear);
-  if(param.store_previous_boundary_values)
-  {
-    // TODO create those quadrature rules only when needed
-    matrix_free_data.insert_quadrature(dealii::QGaussLobatto<1>(param.degree_u + 1),
-                                       field + quad_index_u_gauss_lobatto);
-    matrix_free_data.insert_quadrature(dealii::QGaussLobatto<1>(param.get_degree_p(param.degree_u) +
-                                                                1),
-                                       field + quad_index_p_gauss_lobatto);
-  }
+
+  // TODO create those quadrature rules only when needed
+  matrix_free_data.insert_quadrature(dealii::QGaussLobatto<1>(param.degree_u + 1),
+                                     field + quad_index_u_gauss_lobatto);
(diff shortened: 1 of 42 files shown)


nfehn:
In terms of `UserInterface`, this commit https://github.com/exadg/exadg/pull/156/commits/1ceda040a3e1ecc4fcd2f41fa4620af31f699c00 is the most essential change

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
```diff
@@ -153,14 +153,13 @@ SpatialOperatorBase<dim, Number>::fill_matrix_free_data(
...
-                                                                1),
-                                       field + quad_index_p_gauss_lobatto);
-  }
+
+  // TODO create those quadrature rules only when needed
+  matrix_free_data.insert_quadrature(dealii::QGaussLobatto<1>(param.degree_u + 1),
+                                     field + quad_index_u_gauss_lobatto);
+  matrix_free_data.insert_quadrature(dealii::QGaussLobatto<1>(param.get_degree_p(param.degree_u) +
+                                                              1),
+                                     field + quad_index_p_gauss_lobatto);
```
kronbichler, line 162: To enable an additional quadrature formula implies a considerable increase in memory consumption on general deformed meshes. In the longer run, we need to equip `FEEvaluation` with an entry point where we can compute the mapping information on the fly. Here I am not talking about the possible performance benefit (for which the current interfaces of deal.II are likely inappropriate as they force us into too much overhead), but really just the memory overhead for rarely used functionality.

kronbichler: For sure, this is nothing we need to adjust in this PR, but we need to work towards a solution avoiding excessive memory consumption of `MatrixFree` when we need new evaluation points.

kronbichler (approved):

PR #157, opened 2022-02-04
fix bug introduced in PR #147

nfehn:
bug was not detected since function `create_but_do_not_refine_triangulation()` is not called in ExaDG (only in derived projects).


kronbichler (approved):
Just so I get it right: This would not compile when trying to use the function?

peterrum (approved):

nfehn:
Yes, the private function `do_create_triangulation()` has four parameters (without default arguments), but the function was called with three parameters. Since the public function is called in the project exadg-applications, the ExaDG code compiled without errors.

Issue #140, opened 2022-01-20
Refactor FSI::Driver::setup()


nfehn:
In `FSI::Driver::setup()`, we currently have the following order of initialization steps:

1. single field problems: initialize spatial operators
2. initialize interface coupling (F->S, S->F, S->ALE)
3. single field problems: initialize time integrators and solvers

From the perspective of modular code design, it would make sense to switch steps 2.) and 3.), to obtain:

1. single field problems: initialize spatial operators, time integrators, and solvers
2. initialize interface coupling (F->S, S->F, S->ALE)

I think it should be possible to find a solution, since initialization of time integrators and solvers should not depend on the interface coupling (at least at first sight). Obviously, the setup of time integrators and solvers does not depend directly on the interface coupling objects, but indirectly given that the interface coupling affects ("manipulates") the boundary conditions (and initializes the boundary values of the FSI-interface with zeros). Then, the following aspects have to be discussed in order to ensure that the code remains valid after performing the refactoring suggested above:
- some solvers or preconditioners perform some Krylov iterations during setup, but they should only use the homogeneous parts of the operators without boundary values. Hence, this part should be unproblematic, given that the boundary IDs as well as the type of boundary condition (Dirichlet vs Neumann vs. DirichletCached vs. NeumannCached) are set up in the very beginning of `FSI::Driver::setup()`. @kronbichler: Does this hold for all types of spatial discretization techniques (CG, DG, Hdiv, ...)?
- the time integrator for the structural problem computes an initial acceleration during setup, where the momentum operator is evaluated (including all the boundary conditions such as the BCs initialized by the interface coupling, I think) and subsequently solving a mass matrix system. Hence, this part would be affected by the refactoring suggested above. I currently assume the code would crash since the data arrays for the boundary conditions at the FSI-interface would not have been initialized at this point when changing the order of initialization as suggested above. @peterrum I guess this is the reason why the interface coupling initializes the data structures (so that `compute_initial_acceleration()` can be done in the `setup()` function of the time integrator), a question that came up during #126 (see https://github.com/exadg/exadg/pull/126#discussion_r779610238).

The question is then how to address this problem?

nfehn:
From my perspective, the most reasonable solution would be to perform `compute_initial_acceleration()` at the place where it is needed/used first, in the function `solve_timestep()` 

[...]

PR #160, opened 2022-02-07
split FSI::ApplicationBase::setup() into two parts (structure and fluid/ALE)

nfehn:
@DavidSCN probably relevant/helpful for you


peterrum (approved):
Looks good! Could we do something similar also in the driver? It really felt bad to have to copy all the intialization routines and to split them up and "only" replace the communicator.

nfehn:
@peterrum yes, but to do this I first have to address issue #140

PR #148, opened 2022-02-01
Add preCICE coupled FSI Solid and Fluid solvers

DavidSCN:
Hello,

@peterrum and I spent some time to set up FSI solvers using preCICE instead of the ExaDG-own FSI coupling.

The current implementation is very much preliminary and there are certainly some follow-up tasks, but before going now further into detail, I would like to ask if there is interest in getting these and further changes into the code or not. The implementation doesn't fully follow the code guide lines right now (I guess), because I took some of the code from other places with other guide lines. If there is no interest in such these preCICE coupled solvers, I will probably craft something in my fork.

For now, I set up our very basic [perpendicular-flap](https://precice.org/tutorials-perpendicular-flap.html) tutorial and compared the results of an ExaDG coupling with a deal.II - OpenFOAM coupling. The results cannot directly be compared with the other results on the tutorial website, as all of the reference results there use linear-elasticity as opposed to the non-linear elastic models applied here. The ExaDG solid solver uses the St. Venant-Kirchhoff material model whereas the deal.II solver uses a neo-Hookean material model, so that apparent small differences are reasonable. Another important difference is that the Fluid solver of ExaDG receives `Velocity` and `Displacement` data, whereas the OpenFOAM solver computes the velocity data based on the `Displacement` data.  Note also that the results shown here are generated using loose coupling (no sub-iterations), implicit/strong coupling has been tested as well and works as expected with the current implementation.

comparison

Regarding the run-time of the considered test case, I have the following results:

On the fluid side, using 2.096 cells and (velocity + pressure) 11.528 DoFs

```bash
FSI                           1.10e+02 s    100.00 %
  ALE                         3.87e+01 s     35.15 %
  preCICE                     5.25e+01 s     47.73 %
  Fluid                       1.86e+01 s     16.96 %
  Other                       1.82e-01 s      0.17 %
```

on the solid side with 6.050 DoFs 

```bash
FSI                 1.11e+02 s    100.00 %
  preCICE           7.83e+01 s     70.80 %
  Structure         3.14e+01 s     28.34 %
  Other             9.53e-01 s      0.86 %
```

Coupling OpenFOAM using 2.790 cells and 11.160 (velocity + pressure) DoFs and our own deal.II solid mechanics solver (6.050 as above DoFs) results in a total time of 53.4s with around  ~20sec of time spent on the OpenFOAM side and 33sec of the time spent in the deal.II solver. 

[...]

PR #162, opened 2022-02-09
Refactor FSI setup() function

### LINKED ISSUE #140: Refactor FSI::Driver::setup()

nfehn:
closes #140

currently WIP


nfehn:
commit https://github.com/exadg/exadg/pull/162/commits/4e0806942d1c7ac753dff2b71fa1844dce8c0226 addresses comment https://github.com/exadg/exadg/issues/140#issuecomment-1033732202

nfehn:
please review the code

nfehn:
Commit https://github.com/exadg/exadg/pull/162/commits/5f0d85c44cf7e2ac66f9bc6ded0d6d0579d9f28e allows to reduce the public interface of the spatial operators and shows that the previous design (passing dof-indices and quad-indices from `SpatialOperator` to `InterfaceCoupling` through the `Driver`) was sub-optimal, with unclear responsibilities between `SpatialOperator` and `InterfaceCoupling`.

nfehn:
This PR became more work than expected, but I think it is ready for review now. Due to PR #148, we should make progress and merge the present PR soon.

nfehn:
If there will not appear review comments, I think I will merge during the course of the day.

nfehn:
@peterrum @kronbichler This PR also addresses https://github.com/exadg/exadg/pull/126#discussion_r779610238

include/exadg/functions_and_boundary_conditions/function_cached.h
```diff
@@ -71,16 +73,16 @@ class FunctionCached
...
-    map_solution            = &map_solution_in;
+    map_map_vector_index = &map_map_vector_index_;
+    map_array_solution   = &map_array_solution_;
   }
 
 private:
-  std::map<unsigned int, MapVectorIndex> const * global_map_vector_index;
-  std::map<unsigned int, ArrayTensor> const *    map_solution;
+  std::map<unsigned int, MapVectorIndex> const *      map_map_vector_index;
+  std::map<unsigned int, ArraySolutionValues> const * map_array_solution;
```
peterrum, line 85: These variable names really don't help to understand the meaning. What about given a name that is more descriptive or alternatively add a comment!?

include/exadg/functions_and_boundary_conditions/function_cached.h
```diff
@@ -71,16 +73,16 @@ class FunctionCached
   }
 
   void
-  set_data_pointer(std::map<unsigned int, MapVectorIndex> const & global_map_vector_index_in,
-                   std::map<unsigned int, ArrayTensor> const &    map_solution_in)
+  set_data_pointer(std::map<unsigned int, MapVectorIndex> const &      map_map_vector_index_,
```
peterrum, line 76: Is `_` instead of `_in` the new way to indicate that the parameter is an input?

include/exadg/functions_and_boundary_conditions/interface_coupling.h
```diff
@@ -145,13 +109,101 @@ class InterfaceCoupling
...
+
+template<int dim, int n_components, typename Number>
+class InterfaceCoupling
+{
+private:
+  static unsigned int const rank =
+    (n_components == 1) ? 0 : ((n_components == dim) ? 1 : dealii::numbers::invalid_unsigned_int);
+
+  using MapBoundaryCondition =
+    std::map<dealii::types::boundary_id, std::shared_ptr<FunctionCached<rank, dim, double>>>;
```
peterrum, line 155: Since we are here. I think we should not compute `rank` ourselves and instead use:

https://github.com/dealii/dealii/blob/2105cad54b490002bf27ec6cc26d326c8eb057d0/include/deal.II/numerics/vector_tools_evaluate.h#L116-L119

Using `Tensor` here will not work if we use these classes for scalar valued solutions.

include/exadg/functions_and_boundary_conditions/interface_coupling.h
```diff
@@ -145,13 +109,101 @@ class InterfaceCoupling
...
+    }
+
+    // finally, give boundary condition access to the data
+    for(auto boundary : map_bc_)
+    {
+      boundary.second->set_data_pointer(map_vector_index, map_solution);
+    }
+  }
+
+  std::vector<quad_index> const &
```
peterrum, line 121:
```suggestion
  std::vector<quad_index_type> const &
```

include/exadg/poisson/spatial_discretization/operator.cpp
```diff
@@ -226,6 +226,24 @@ Operator<dim, Number, n_components>::setup(
...
+  {
+    interface_data_dirichlet_cached =
+      std::make_shared<ContainerInterfaceData<dim, n_components, Number>>();
+    std::vector<unsigned int> quad_indices;
+    if(param.spatial_discretization == SpatialDiscretization::DG)
+      quad_indices.emplace_back(get_quad_index());
+    else if(param.spatial_discretization == SpatialDiscretization::CG)
+      quad_indices.emplace_back(get_quad_index_gauss_lobatto());
+    else
+      AssertThrow(false, dealii::ExcMessage("not implemented."));
```
peterrum, line 239: What about:
```suggestion
      AssertThrow(false, dealii::ExcNotImplemented());
```

include/exadg/poisson/spatial_discretization/operator.cpp
```diff
@@ -226,6 +226,24 @@ Operator<dim, Number, n_components>::setup(
   matrix_free      = matrix_free_in;
   matrix_free_data = matrix_free_data_in;
 
+  if(not(boundary_descriptor->dirichlet_cached_bc.empty()))
```
peterrum, line 229: Where is the Poisson coupled to other fields?

include/exadg/functions_and_boundary_conditions/function_cached.h
nfehn: I agree that it is not easy to find good names. However, I think that the names have at least improved compared to the previous version (For example, I was irritated by the name `global_map_...` in the previous version, which initiated the renaming performed here). I don't know if there are general C++ recommendations how to name maps ... if there are suggestion, we can of course further improve the names.

include/exadg/functions_and_boundary_conditions/function_cached.h
nfehn: We have both versions in the code (the version with `_` only was a recommendation given in the PRACE course "Advanced C++ with focus on software engineering").
We could think about fixing the style and enforce it uniformly within ExaDG in the future.

include/exadg/poisson/spatial_discretization/operator.cpp
nfehn: Yes, would probably make sense. As for other comments, I would prefer to do this consistently within ExaDG due to many occurrences.
As a very general comment or question: I am not sure if application projects such as ExaDG should use  `Assert`/`AssertThrow` (from dealii) everywhere, or instead (re-)define assert-functionality within the library? One problem that I see is for example that `Assert`/`AssertThrow` does not have the `dealii` namespace, see PR #145.

include/exadg/poisson/spatial_discretization/operator.cpp
nfehn: Poisson is used as mesh motion algorithm / PDE model for the ALE-field in FSI. The displacement of the mesh at the FSI interface is prescribed by the structure (and is prescribed via a `DirichletCached` BC).

include/exadg/functions_and_boundary_conditions/interface_coupling.h
nfehn: I consider this also a follow-up PR. I currently don't know how to deal with the value_type/double mixture that we have in ExaDG, due to compatibility with dealii-9.3 on the one hand and due to the fact that we do not want to introduce a second template parameter `Number` for `BoundaryDescriptor`s and `Function`s  on the other hand (even though I am open to changing this in ExaDG in the future ... but I think we should discuss this in a broader context, e.g. as separate issues in ExaDG).

include/exadg/functions_and_boundary_conditions/function_cached.h
peterrum: The name only indicates the type of the variable. For that one has an IDE. If someone takes a look at the code, he/she would like to know what is mapped without the need to look into the setup routines.

include/exadg/functions_and_boundary_conditions/interface_coupling.h
peterrum: But please indicate somewhere/somehow that it is a type. In the same file the name is also used as variable name:

https://github.com/nfehn/exadg/blob/0d2cb65d0267db4318ecdcd4aba83986d1424c47/include/exadg/functions_and_boundary_conditions/interface_coupling.h#L201

include/exadg/poisson/spatial_discretization/operator.cpp
peterrum: We can create our own exceptions. That shouldn't be a problem. Let me create PR tomorrow for that!

kronbichler: Regarding `Assert`: That is a macro and thus gets resolved by the preprocessor. One can avoid the use of those, see e.g. the comment https://github.com/dealii/dealii/blob/09bae93a06eca37f12e27be378647160377a63a1/include/deal.II/base/exceptions.h#L1434-L1492 . When it comes to defining our own macros instead, the driving force is likely the added value we get from having a new definition, as the current variant is pretty self-explaining and I don't see fundamental problems with the current usage.

nfehn: Yes, I had this comment in mind. Might this raise problems in the future? Why using macros and not something more modern, or do you think the macro-style is still the best solution nowadays?

The question is whether we should introduce dependencies on deal.II in files/modules, that actually do not depend on deal.II (apart from exceptions). Even though ExaDG will probably always have deal.II as a mandatory dependency, I think it would still make sense to not copy such functionality from deal.II.

kronbichler: I don't expect problems in the regular use case by the macro, but it of course depends on if there is a conflicting definition of the macro in another package, or if one wants to redefine it to something else (this includes the `DeclExceptionMsg` and friends). One can disable those flags and keep separate.

The question for us in ExaDG is whether it is worth the effort to design something new, i.e., we have the competence to design something substantially better and if we want to spend the developer time on this topic rather than some other topic.

> Why using macros and not something more modern, or do you think the macro-style is still the best solution nowadays?

I am not an expert on this topic and do not intend to become one, so I can't really say much beyond a personal opinion: For now I see little downsides with this approach, it is robust, makes it pretty simple to assert things or create new exceptions and without an unreasonably high learning curve. I would act if it does not fit our needs any more to the extent that creating a solution for us consumes less aggregate time than fighting with undesired behavior. I can't see that now. Any way, I am happy to learn about alternatives that are as easy or easier to use, and you would consider "more modern".

PR #171, opened 2022-02-14
Add exceptions.h

peterrum:


nfehn (approved):

peterrum:
@nfehn Feel free to add additional exceptions in the future. For how to use the macros, you can take a look at https://github.com/dealii/dealii/blob/master/include/deal.II/base/exceptions.h!

include/exadg/utilities/exceptions.h
```diff
@@ -0,0 +1,45 @@
...
+                 "You are trying to use functionality in ExaDG that is "
+                 "currently not implemented. In many cases, this indicates "
+                 "that there simply didn't appear much of a need for it, or "
+                 "that the author of the original code did not have the "
+                 "time to implement a particular case. If you hit this "
+                 "exception, it is therefore worth the time to look into "
+                 "the code to find out whether you may be able to "
+                 "implement the missing functionality. If you do, please "
+                 "consider providing a patch to the ExaDG development "
+                 "sources (see the ExaDG Wiki page on how to contribute).");
```
kronbichler, line 39: Just a comment, but this smells very slightly like plagiarism when we use a verbatim copy of the text in deal.II (with only the package names exchanged) without explicitly saying that this is a copy of the respective `ExcNotImplemented` in deal.II.

peterrum: Is it with the following comment better?

https://github.com/exadg/exadg/blob/e6c4599a1a41e0eb10891fc0cd18b899a12546c1/include/exadg/utilities/exceptions.h#L30

I really don't care what the message is. The PR only shows how to introduce new exceptions.

kronbichler (approved):
I agree, it's just that I want to avoid having this kind of style of language (coming from a particular person not active in ExaDG) without proper reference to it.

include/exadg/utilities/exceptions.h
```diff
@@ -0,0 +1,48 @@
...
+
+#ifndef INCLUDE_EXADG_UTILITIES_EXCEPTIONS_H_
+#define INCLUDE_EXADG_UTILITIES_EXCEPTIONS_H_
+
+#include <deal.II/base/exceptions.h>
+
+namespace ExaDG
+{
+/**
+ * The following exception is based on and motivated by dealii::ExcNotImplemented.
```
nfehn, line 30: Why do you write `DeclExceptionMsg` without `dealii::` namespace? This is exactly the kind of dependency that I criticized or questioned here https://github.com/exadg/exadg/pull/162#discussion_r806131679. I more and more come to the opinion that this should only be used internally in deal.II

include/exadg/utilities/exceptions.h
nfehn: Yes, no need to adapt it word-by-word, given that the message is also very wordy ...

Let's collect ideas regarding which information such a message should provide in order to help un-experienced users ...

nfehn: Some ideas from my side: The deal.ii-message is somewhat written from the perspective of developers. In my opinion, the message should mention the possibility that the user has chosen a combination of parameters that is actually not intended. Note that it might be quite difficult to explicitly assert all invalid parameter combinations (even though we try this in `Parameters::check()`). Hence, the exception message could provide the suggestion to carefully check the input parameters again. So in some cases such a message is not a hint to implement additional functionality.  I believe `ExcNotImplemented` is still the correct type of exception, since such problems can not be deduced from the implementation in a single file, but need to take into account the interaction of modules.

kronbichler: I agree, we should have the first sentence indicating that some combination of features either has not been implemented, or does not make sense in combination.

include/exadg/utilities/exceptions.h
kronbichler: It's again a `#define`: https://github.com/dealii/dealii/blob/18a3861abce9e60d2a82e8ba52b8ed1a7c6774ed/include/deal.II/base/exceptions.h#L213-L230

PR #175, opened 2022-02-16
TimerTree: only print results if level <= max_level

nfehn:
This might fix the comment described here https://github.com/exadg/exadg/pull/148#discussion_r808047907

The corresponding test needs to be adjusted, but I currently can not run tests due to ubuntu issues ...

### DIFF
--- a/include/exadg/utilities/timer_tree.h
+++ b/include/exadg/utilities/timer_tree.h
@@ -210,9 +210,39 @@ public:
   {
     unsigned int const length = get_length();
 
-    pcout << std::endl;
+    unsigned int const max_level = get_max_level();
 
-    do_print_level(pcout, level, 0, length);
+    if(level <= max_level)
+    {
+      pcout << std::endl;
+
+      do_print_level(pcout, level, 0, length);
+    }
+    else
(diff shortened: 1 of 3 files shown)


include/exadg/utilities/timer_tree.h
```diff
@@ -210,9 +210,39 @@ class TimerTree
...
+      pcout << std::endl;
+
+      do_print_level(pcout, level, 0, length);
+    }
+    else
+    {
+      pcout << std::endl
+            << "Timings can not be printed for level = " << level << "," << std::endl
+            << "since the maximum level of the timer tree is max_level = " << max_level << "."
+            << std::endl;
```
peterrum, line 226: I don't see how writing something to pcout is beneficial? Why not assert?

nfehn: In fact I believe it would not be the correct behavior to assert here. The user requests information that can not be provided. But the timing information is nothing vital to the simulation, only "additional information". So the simulation should not crash in my opinion, but terminate "normally" with the message that a particular information can not be given.

PR #181, opened 2022-02-28
explain how we treat constrained degrees of freedom in OperatorBase

nfehn:
After a discussion with @sebproell, we found that the code is quite hard to read regarding how we treat constrained degrees of freedom in the context of iterative solvers.

This PR suggests additional explanations in the form of comments, that hopefully improve understanding.


kronbichler (approved):
I agree. I was thinking about whether it might help to mention the concept of homogeneous (linear) vs inhomogeneous (affine) operators, but came to the conclusion that it might more confusing than helpful.

Issue #180, opened 2022-02-28
slurm script error with exadg


fab6:
Hi,

the deal-9.3.3 and exadg compilation worked (did not include petsc and no trilinos yet, no fftw). In addition to  p4est I needed numdiff as local installation (at least for the second reinstallation using a slurm script). After first similar error message  (see below) I recompiled deal/p4est and exadg on the nodes via slurm script. `numdiff` was installed on the frontend.

I used for installation some available modules:
```
module load 2021a GCC/10.3.0 OpenMPI/4.1.1 CMake/3.20.1 libarchive METIS/5.1.0
```
A local quick run on the frontend works:
![image](https://user-images.githubusercontent.com/3047695/155963481-9c21adfe-841c-48bf-9193-eb4148cd13bd.png)

The slurm script looks like:
![image](https://user-images.githubusercontent.com/3047695/155963678-0e677551-b60b-45b0-bb81-ee0b3b5299f3.png)

Now when trying to run e.g. the TGV incompressible application I get an error when running using `slurm`:
```
srun: error: wn21205: tasks 0-5,7-15: Illegal instruction
srun: error: wn21205: task 6: Illegal instruction
```

I get similar error when using the complete path for the executable or e.g. `mpirun -np 16` or solely `srun`.

E.g. for `mpirun` I get
![image](https://user-images.githubusercontent.com/3047695/155966148-25e5573d-06bc-4c90-b9fa-0b11c4b6a119.png)

Do you have an advice how to get the examples running via slurm?
Thank you in advance!
Fab

fab6:
The processor on the frontend is a intel xeon whereas the nodes have
![image](https://user-images.githubusercontent.com/3047695/155969514-9926a08d-60a5-4b67-8b4f-8e1cc197e7cf.png)

So I assume the comilation on the nodes (which I did) is important; though the error confuses me.

kronbichler:
[...]

PR #182, opened 2022-02-28
add comment regarding compiler flags to configure scripts

nfehn:
see discussion in issue #180.

@fab6 FYI.

### DIFF
--- a/scripts/config_dealii.sh
+++ b/scripts/config_dealii.sh
@@ -30,6 +30,12 @@ DEAL=$WORKING_DIRECTORY/sw/dealii
 
 rm -rf CMakeFiles/ CMakeCache.txt
 
+# Note on compiler flags: Note that "-march=native" requires that the hardware 
+# on which you compile the code is consistent with the hardware on which you 
+# execute the code. If this is not the case, consider to specify the target 
+# hardware for compilation, e.g. "-march=haswell" or "-march=skylake-avx512", 
+# where "-march=haswell" can also be used for recent AMD x86-64 hardware. 
+
 cmake \
     -D CMAKE_BUILD_TYPE="DebugRelease" \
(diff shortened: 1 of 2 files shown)


scripts/config_dealii.sh
```diff
@@ -30,6 +30,12 @@ DEAL=$WORKING_DIRECTORY/sw/dealii
 
 rm -rf CMakeFiles/ CMakeCache.txt
 
+# Note on compiler flags: Note that "-march=native" requires that the hardware 
+# on which you compile the code is consistent with the hardware on which you 
+# execute the code. If this is not the case, consider to specify the target 
+# hardware for compilation, e.g. "-march=haswell" or "-march=skylake-avx512", 
+# where "-march=haswell" can also be used for recent AMD x86-64 hardware.
```
bergbauer, line 37: There are also options for amd processors like `-march=znver2` for Zen 2 architecture. 

https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html

nfehn: thanks for this hint. I adjusted the comment with a reference to the link you have provided.

kronbichler: That's indeed an excellent link. Just as a background information, in my experience the exact microarchitecture (e.g., znver2 versus haswell) has only a minor influence on the performance - it affects some minor scheduling issues or preferring one instruction over another, but all of that are handled relatively well in hardware, too. The big influence is which extensions get available, such as FMA, AVX/AVX2, AVX-512, etc., that can map the code e.g. to wider SIMD registers or more powerful instructions, which are not specific to single CPU models like Haswell.

kronbichler (approved):

scripts/config_dealii.sh
nfehn: thanks for the additional information.

nfehn:
I will merge once the checks have passed.

PR #183, opened 2022-03-02
do not impose Dirichlet constraints after Newton solve

nfehn:
It should not be necessary to apply Dirichlet constraints in the nonlinear `Structure` solver after the Newton solve. The constraints have already been set before the solve and no contributions to the constrained degrees of freedom should be added in the Newton solver by the linearized solver.

I added a check that the residual vector is really zero for the constrained degrees of freedom. One might argue that this is not necessary since the Newton solver would otherwise not converge, on the other hand the problem might not be detected by the Newton solver depending on the chosen solver tolerances. The costs of this additional check should be irrelevant compared to the rest.

@sebproell FYI

### DIFF
--- a/include/exadg/operators/operator_base.cpp
+++ b/include/exadg/operators/operator_base.cpp
@@ -243,6 +243,20 @@ OperatorBase<dim, Number, n_components>::set_constrained_values_to_zero(VectorTy
 }
 
 template<int dim, typename Number, int n_components>
+bool
+OperatorBase<dim, Number, n_components>::check_constrained_values_are_zero(
+  VectorType const & vector) const
+{
+  for(unsigned int i = 0; i < constrained_indices.size(); ++i)
+  {
+    if(vector.local_element(constrained_indices[i]) > std::numeric_limits<Number>::min())
+      return false;
+  }
+
+  return true;
+}
+
+template<int dim, typename Number, int n_components>
 void
 OperatorBase<dim, Number, n_components>::calculate_inverse_diagonal(VectorType & diagonal) const
 {
--- a/include/exadg/structure/spatial_discretization/operator.cpp
+++ b/include/exadg/structure/spatial_discretization/operator.cpp
@@ -662,6 +662,11 @@ Operator<dim, Number>::evaluate_nonlinear_residual(VectorType &       dst,
     body_force_operator.evaluate_add(body_forces, src, time);
     dst -= body_forces;
   }
+
+  // check that constrained degrees of freedom are really zero
+  AssertThrow(check_constrained_values_are_zero(dst),
+              dealii::ExcMessage(
+                "Expected that constrained degrees of freedom are zero. Aborting."));
 }
 
 template<int dim, typename Number>
@@ -718,6 +723,16 @@ Operator<dim, Number>::set_constrained_values_to_zero(VectorType & vector) const
 }
 
 template<int dim, typename Number>
+bool
+Operator<dim, Number>::check_constrained_values_are_zero(VectorType const & vector) const
+{
+  if(param.large_deformation)
+    return elasticity_operator_nonlinear.check_constrained_values_are_zero(vector);
+  else
+    return elasticity_operator_linear.check_constrained_values_are_zero(vector);
+}
+
+template<int dim, typename Number>
 std::tuple<unsigned int, unsigned int>
 Operator<dim, Number>::solve_nonlinear(VectorType &       sol,
                                        VectorType const & rhs,
@@ -739,9 +754,6 @@ Operator<dim, Number>::solve_nonlinear(VectorType &       sol,
 
   // solve nonlinear problem
   auto const iter = newton_solver->solve(sol, update);
-
-  // set inhomogeneous Dirichlet values
-  elasticity_operator_nonlinear.set_constrained_values(sol, time);
 
   return iter;
 }
--- a/include/exadg/operators/operator_base.h
+++ b/include/exadg/operators/operator_base.h
@@ -172,6 +172,9 @@ public:
   void
   set_constrained_values_to_zero(VectorType & vector) const;
 
+  bool
+  check_constrained_values_are_zero(VectorType const & vector) const;
+
   void
   calculate_inverse_diagonal(VectorType & diagonal) const;
 
--- a/include/exadg/structure/spatial_discretization/operator.h
+++ b/include/exadg/structure/spatial_discretization/operator.h
@@ -263,6 +263,9 @@ public:
   void
   set_constrained_values_to_zero(VectorType & vector) const;
 
+  bool
+  check_constrained_values_are_zero(VectorType const & vector) const;
+
   /*
    * This function solves the (non-)linear system of equations.
    */


include/exadg/operators/operator_base.cpp
```diff
@@ -242,6 +242,20 @@ OperatorBase<dim, Number, n_components>::set_constrained_values_to_zero(VectorTy
...
 }
 
+template<int dim, typename Number, int n_components>
+bool
+OperatorBase<dim, Number, n_components>::check_constrained_values_are_zero(
+  VectorType const & vector) const
+{
+  for(unsigned int i = 0; i < constrained_indices.size(); ++i)
+  {
+    if(vector.local_element(constrained_indices[i]) > std::numeric_limits<Number>::min())
```
sebproell, line 252: I think this doesn't work if you have a negative entry.

nfehn:
@kronbichler please also check whether you agree with the comment introduced in commit https://github.com/exadg/exadg/pull/183/commits/d6c836e884a219b8ea104d73e803e3ad44622ceb

include/exadg/operators/operator_base.cpp
nfehn: yes, good point

include/exadg/operators/operator_base.cpp
```diff
@@ -242,6 +242,20 @@ OperatorBase<dim, Number, n_components>::set_constrained_values_to_zero(VectorTy
...
 }
 
+template<int dim, typename Number, int n_components>
+bool
+OperatorBase<dim, Number, n_components>::check_constrained_values_are_zero(
+  VectorType const & vector) const
+{
+  for(unsigned int i = 0; i < constrained_indices.size(); ++i)
+  {
+    if(std::abs(vector.local_element(constrained_indices[i])) > std::numeric_limits<Number>::min())
```
peterrum, line 252: I would argue that if `std::abs(vector.local_element(constrained_indices[i]))` in not `0.0` actual there is already an inconsistency in the code.

include/exadg/structure/spatial_discretization/operator.cpp
```diff
@@ -662,6 +662,11 @@ Operator<dim, Number>::evaluate_nonlinear_residual(VectorType &       dst,
     body_force_operator.evaluate_add(body_forces, src, time);
     dst -= body_forces;
   }
+
+  // check that constrained degrees of freedom are really zero
+  AssertThrow(check_constrained_values_are_zero(dst),
```
peterrum, line 667: Do we really need this in release mode?

include/exadg/operators/operator_base.cpp
nfehn: What do you mean? write `== 0.0`? where is the difference?

include/exadg/structure/spatial_discretization/operator.cpp
nfehn: One could probably think about both options. The reason why I originally decided for AssertThrow is that the costs should be small (note that vmult() of the linear operator touches the constraints before and after matrix-vector product), so I thought these costs are negligible since this is only done in every nonlinear residual evaluation (see description of PR). So I would change to Assert only if you and @kronbichler have concerns in terms of computational costs.

include/exadg/operators/operator_base.cpp
kronbichler: I would also say that we should write `== 0.0` here. There are subnormal numbers between the `min` numeric limit and 0 (so some random content could end up being the same here).

include/exadg/structure/spatial_discretization/operator.cpp
kronbichler: The costs are likely small (<1%) in all cases we are using right now, but I would still opt to only use it in debug mode (i.e., use `Assert`):
- The check is a pure verification check to be used during algorithm development and nothing that can should ever go wrong in an application context.
- In case algorithms get extended in the future that would increase the frequency of the evaluation of residual (I think of nonlinear Krylov solvers), the cost could get visible.
- The cost might be low here, but I think the example could indicate to a less experienced programmer that "checks on individual vector elements" in the inner workings of an algorithm are the recommended way also for release builds, which is not true in general.

Note that I am a big believer of the `Assert` category of checks as there are many conditions one would like to make extra checks on during development, but then would be too expensive / superfluous to be checked in a working code with release qualification afterwards. This seems to be a prime example of it, even if the cost is negligible right now.

include/exadg/operators/operator_base.cpp
nfehn: `== 0.0` is what we want to assert. So should I read the exceptional case with `std::abs(value) > 0.0`, or `value != 0.0`?

kronbichler: != 0

include/exadg/structure/time_integration/time_int_gen_alpha.cpp
```diff
@@ -126,16 +126,18 @@ TimeIntGenAlpha<dim, Number>::do_timestep_solve()
...
     pde_operator->compute_rhs_linear(rhs, this->get_mid_time());
+
+    // Constrained degrees of freedom have already been set
+    // correctly in the function compute_rhs_vector() for the rhs vector.
+    // Hence, the following line ensures that adding const_vector to rhs
+    // does not alter the constrained degrees of freedom of rhs.
+    pde_operator->set_constrained_values_to_zero(const_vector);
+
     // shift const_vector to right-hand side
     rhs.add(-1.0, const_vector);
```
nfehn, line 142: Linear problems: This should be refactored/redesigned in a follow-up PR. I think we can improve the logic for imposing inhomogeneous Dirichlet BCs also in case of linear problems.

include/exadg/structure/time_integration/driver_quasi_static_problems.cpp
```diff
@@ -226,17 +226,19 @@ DriverQuasiStatic<dim, Number>::solve_step(double const load_factor)
...
     this->param.update_preconditioner &&
     ((this->step_number - 1) % this->param.update_preconditioner_every_time_steps == 0);
 
-  auto const iter = pde_operator->solve_nonlinear(
+  VectorType const const_vector; // will not be used
+  auto const       iter = pde_operator->solve_nonlinear(
     solution, const_vector, 0.0 /*no mass term*/, load_factor /* = time */, update_preconditioner);
 
+  unsigned int const N_iter_nonlinear = std::get<0>(iter);
+  unsigned int const N_iter_linear    = std::get<1>(iter);
```
sebproell, line 238:
```suggestion
  auto const    [N_iter_nonlinear, N_iter_linear]    = pde_operator->solve_nonlinear(
    solution, const_vector, 0.0 /*no mass term*/, load_factor /* = time */, update_preconditioner);
```

nfehn: Nice! I did this here for consistency with another occurrence in ExaDG. Hence, let's address this in a separate PR.

nfehn:
Please check again and then let's get it merged.

include/exadg/structure/time_integration/driver_quasi_static_problems.cpp
kronbichler: Do we require C++17 (which is needed for this feature)?

nfehn: I think we already have C++17 in ExaDG (I think we needed it for filesystem related functionality). But as said before, let's discuss this elsewhere. I would suggest that @sebproell opens an issue with the "suggested changes" from above. Then, I can go through the code and replace several occurrences.

PR #184, opened 2022-03-03
Constraints for continuous Galerkin discretizations

nfehn:
<del> This is a follow-up PR to #183.</del> This supersedes PR #183. 

Currently, the code contains calls like `set_constrained_values_to_zero()` in the time integrator class. However, this is a detail of the spatial discretization operator and should not occur in the time integrator. With these changes, I hope that the code is clearer in terms of single responsibility and easier to understand.

Currently, the function of `OperatorBase` that shifts contributions of inhomogeneous BCs to the right-hand side vector also sets the constrained degrees of freedom of the rhs vector to the inhomogeneous boundary conditions. However, we can not ensure that other operators overwrite or add values to these vector entries and thereby invalidate what `OperatorBase` has done before. Hence, I believe the code is clearer if we remove these lines of code and instead apply constraints (if desired) immediately before the linear solver is called. For `OperatorBase`, it is only relevant that (i) `vmult()` is computed correctly in case of constrained DoFs, and (ii) that contributions of constrained DoFs are taken into account in unconstrained vector entries of the rhs vector (by accumulating into the rhs-vector!).

<del> It should not be necessary to apply Dirichlet constraints in the nonlinear Structure solver after the Newton solve. The constraints have already been set before the solve and no contributions to the constrained degrees of freedom should be added in the Newton solver by the linearized solver.</del>

<del> This PR currently contains changes from #183. Hence, I have to rebase once #183 is merged. Alternatively, we could just close PR #183 and consider the present one.</del>

@sebproell FYI

### DIFF
--- a/include/exadg/structure/spatial_discretization/operator.cpp
+++ b/include/exadg/structure/spatial_discretization/operator.cpp
@@ -643,10 +643,6 @@ Operator<dim, Number>::evaluate_nonlinear_residual(VectorType &       dst,
                                                    double const       factor,
                                                    double const       time) const
 {
-  // Note that constrained degrees of freedom have to be zero for dst and const_vector
-  // in order to ensure convergence of the Newton solver. This is checked at the
-  // end of this function.
-
   // elasticity operator
   elasticity_operator_nonlinear.set_scaling_factor_mass_operator(factor);
   elasticity_operator_nonlinear.set_time(time);
@@ -667,9 +663,11 @@ Operator<dim, Number>::evaluate_nonlinear_residual(VectorType &       dst,
     dst -= body_forces;
   }
 
-  // check that constrained degrees of freedom are really zero
-  Assert(check_constrained_values_are_zero(dst),
-         dealii::ExcMessage("Expected that constrained degrees of freedom are zero. Aborting."));
+  // To ensure convergence of the Newton solver, the residual has to be zero
+  // for constrained degrees of freedom as well, which might not be the case
+  // in general, e.g. due to const_vector. Hence, we set the constrained
+  // degrees of freedom explicitly to zero.
+  set_constrained_values_to_zero(dst);
 }
 
 template<int dim, typename Number>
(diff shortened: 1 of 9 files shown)


nfehn:
Iteration counts can be reduced considerably by choosing the initial solution for the iterative solver consistent with the rhs-vector entries for constrained degrees of freedom.
The test case that I consider is the "FSI -  cylinder with flag" example. The iteration counts reduced from around 280 to 120 for the linear elasticity mesh motion solver by applying the changes from commit https://github.com/exadg/exadg/pull/184/commits/c6037129463d112cb936a0065b075eefee252190

nfehn:
I just realized that iteration counts are around 60 with the current master branch, so I must have introduced changes (but I currently do not see where this should originate from?).

nfehn:
I made once more changes to ensure that this PR does not introduce changes, but is only a refactoring of the current master branch. Now, I get the same iterations counts for the FSI example as for the current master branch.

The question whether to apply constraints again after the solution of linear/nonlinear systems of equations or whether this is actually superfluous should be addressed separately.

@kronbichler Could you have a look at the changes?

include/exadg/structure/spatial_discretization/operator.cpp
```diff
@@ -740,7 +736,13 @@ Operator<dim, Number>::solve_nonlinear(VectorType &       sol,
...
   auto const iter = newton_solver->solve(sol, update);
 
-  // set inhomogeneous Dirichlet values
+  // This step should actually be optional: The constraints have already been set
+  // before the nonlinear solver is called and no contributions to the constrained
+  // degrees of freedom should be added in the Newton solver by the linearized solver
+  // (because the residual vector forming the rhs of the linearized problem is zero
+  // for constrained degrees of freedom, the initial solution of the linearized
+  // solver is also zero, and the linearized operator contains values of 1 on the
+  // diagonal for constrained degrees of freedom.
```
kronbichler, line 745: Tiny comment: I think you need to close the braces
```suggestion
  // diagonal for constrained degrees of freedom).
```

kronbichler (approved):
I think this is OK, even though it is a bit confusing to modify the right hand side vector (rhs_mutable) inside the solver, rather than doing it on a higher level. But I see no better solution right now, so it is good to proceed here.

kronbichler (approved):

PR #186, opened 2022-03-07
new feature 'overset grids'

nfehn:
This PR introduces the feature of overset grids by the example of the `Poisson` module. Two domains overlap and ExaDG's interface coupling functionality is used to couple the two domains via Dirichlet BCs. A very simple outer iteration is currently used to obtain convergence for the coupled problem.

<del> This PR is currently a draft, since several aspects have been encountered which first need to be generalized in ExaDG. Here is a list of main points: </del>

- `InterfaceCoupling`/`FunctionCached` does currently not work for scalar quantities. -> see commit https://github.com/exadg/exadg/pull/186/commits/69a89d8106c59a82988b768b0740d22be6b30a35
- <del> Hence, I set up a vectorial Poisson problem. However, I had to deactivate the error calculation in `Poisson::Postprocessor` since this seems to work only for scalar problems. <\del>
- The setup of multigrid (map of dirichlet boundary IDs passed to `MultigridPreconditionerBase`) does not take into account Dirichlet cached BCs. In case all boundary conditions are of type `DirichletCached`, multigrid even crashes. It is remarkable that multigrid seems to work if we have both `Dirichlet` and `DirichletCached` BCs (as in some FSI examples). This is a bug. -> See PR #188.
- `InterfaceCoupling` needs to be generalized in case that the two domains do not touch each other along a boundary, but intersect each other (-> `marked_vertices` and setup of `InterfaceCoupling` needs to be generalized). -> See PR #187 as a first remedy.
- The design of the related base application class and the driver class needs to be improved to avoid code duplication. -> first part resolved in commit https://github.com/exadg/exadg/pull/186/commits/f46941264f5c3535a0cd4ab4e073f9bb007469d5, second part in commit https://github.com/exadg/exadg/pull/186/commits/7cb2daae995688ee092511c1c68f6f4555de355

Example: Scalar Poisson problem with constant right-hand side, homogeneous DBC's, DirichletCachedBC's at the overlapping boundaries: Domain1 = [0,1]^2; Domain2 = [0.5, 1.5] x [0, 1]. -> overlap from x_1 = 0.5 to 1.0.

Solution after initialization with zero
fig0
and after 1, 2, 3 iterations (visually converged after 3 iterations).
fig1
fig2
fig3


include/exadg/functions_and_boundary_conditions/interface_coupling.h
```diff
@@ -196,6 +196,17 @@ class InterfaceCoupling
...
+    // Make sure that not all entries of marked_vertices are false
+    bool all_false = true;
+    for(auto iter : marked_vertices)
+    {
+      if(iter == true)
+        all_false = false;
+    }
+
+    if(all_false)
+      marked_vertices = std::vector<bool>(dof_handler_src_.get_triangulation().n_vertices(), true);
```
nfehn, line 209: addressed separately in PR #187

include/exadg/poisson/spatial_discretization/operator.cpp
```diff
@@ -277,13 +277,23 @@ Operator<dim, Number, n_components>::setup_solver()
...
+    typedef typename std::pair<dealii::types::boundary_id, std::shared_ptr<dealii::Function<dim>>>
+      pair;
+
+    std::map<dealii::types::boundary_id, std::shared_ptr<dealii::Function<dim>>>
+      dirichlet_boundary_conditions = laplace_operator.get_data().bc->dirichlet_bc;
+
+    for(auto iter : laplace_operator.get_data().bc->dirichlet_cached_bc)
+      dirichlet_boundary_conditions.insert(
+        pair(iter.first, new dealii::Functions::ZeroFunction<dim>(dim)));
+
```
nfehn, line 289: Addressed separately in PR #188.

peterrum:
Nice, some domain-decomposition method! The next step would be to compute the traditional Schwarz setup: union of rectangle and circle. That was one of the first tests of RPE. I'll try to take a look at the PR later today.

nfehn:
> Nice, some domain-decomposition method! The next step would be to compute the traditional Schwarz setup: union of rectangle and circle. That was one of the first tests of RPE. I'll try to take a look at the PR later today.

Yes, I already thought of this setup, did not realize it so far since some functionality is missing. Can you point me to that test case? Is it part of deal.II? I need functionality like `tria_1->contains(face)` (in the bangerth-favorite naming ;) and interpreted purely geometrically; `face` belonging to another triangulation `tria_2`) to detect automatically on which boundary faces to set `DirichletCached` boundary conditions.

peterrum:
This is the test: https://github.com/dealii/dealii/blob/master/tests/remote_point_evaluation/vector_tools_evaluate_at_points_02.cc

Not very nice. What I do is to only check if the quadrature points are inside of the domain or not.

nfehn:
> This is the test: https://github.com/dealii/dealii/blob/master/tests/remote_point_evaluation/vector_tools_evaluate_at_points_02.cc
>
> Not very nice. What I do is to only check if the quadrature points are inside of the domain or not.

I do not find the place where you check whether the q-points are inside the other triangulation. Does this happen inside RemotePointEvaluation? For me the code looks like you are collecting all points on the boundary, do the evaluation, and set constraints for all the points afterwards?

peterrum:
> Does this happen inside RemotePointEvaluation? For me the code looks like you are collecting all points on the boundary, do the evaluation, and set constraints for all the points afterwards?

Yes. If the point is not found in the other domain, the value has its default value (0.0).

applications/poisson/overset_grids/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+TARGETNAME(TARGET_NAME ${CMAKE_CURRENT_SOURCE_DIR})
```
peterrum, line 1: Why is here no header?

applications/poisson/overset_grids/input.json
```diff
@@ -0,0 +1,18 @@
...
+    "General": {
+        "Precision": "double",
+        "Dim": "2",
+        "IsTest": "false"
+    },
+    "SpatialResolution": {
+        "DegreeMin": "3",
+        "DegreeMax": "3",
+        "RefineSpaceMin": "2",
+        "RefineSpaceMax": "2"
```
peterrum, line 11: Why not specify these independently for the domains?

include/exadg/functions_and_boundary_conditions/interface_coupling.h
```diff
@@ -232,20 +249,24 @@ class InterfaceCoupling
     for(auto quadrature : interface_data_dst->get_quad_indices())
     {
 #if DEAL_II_VERSION_GTE(10, 0, 0)
-      std::vector<dealii::Tensor<rank, dim, Number>> const result =
+      auto const result =
```
peterrum, line 252: @kronbichler Wanted me to be explicit here due to the possible double/Number confusion.

include/exadg/poisson/overset_grids/driver.h
```diff
@@ -0,0 +1,70 @@
...
+
+  void
+  setup();
+
+  void
+  solve();
+
+private:
+  static unsigned int const rank =
+    (n_components == 1) ? 0 : ((n_components == dim) ? 1 : dealii::numbers::invalid_unsigned_int);
```
peterrum, line 49: I don't think it is a good idea to introduce `rank` here. I would rather use the traits from deal.II.

include/exadg/poisson/overset_grids/driver.h
```diff
@@ -0,0 +1,70 @@
...
+  // output to std::cout
+  dealii::ConditionalOStream pcout;
+
+  std::shared_ptr<ApplicationOversetGridsBase<dim, n_components, Number>> application;
+
+  // Poisson solvers
+  std::shared_ptr<SolverPoisson<dim, n_components, Number>> poisson1, poisson2;
+
+  // interface coupling
+  std::shared_ptr<InterfaceCoupling<dim, n_components, Number>> first_to_second, second_to_first;
```
peterrum, line 63: What about making these to vectors? The code is generalizable to n domains.

peterrum (commented):
Looks good to me!

applications/poisson/overset_grids/CMakeLists.txt
nfehn: I think we generally do not have them for these files. I could include them consistently in a separate PR.

include/exadg/functions_and_boundary_conditions/interface_coupling.h
nfehn: this is currently the strategy to deal with https://github.com/dealii/dealii/issues/13517 in the case of scalar quantities. Of course, we still need the copying some lines below.

include/exadg/poisson/overset_grids/driver.h
nfehn: we can talk about this again once https://github.com/dealii/dealii/issues/13517 is resolved.

include/exadg/poisson/overset_grids/driver.h
nfehn: Principally yes. For me this could be a future extension. Right now, there is no clear application behind that serves as a motivation. The main purpose is currently to show the capabilities/functionalitites of the code. For this, I believe a code without vectors is easier to read.

nfehn:
> Yes. If the point is not found in the other domain, the value has its default value (0.0).

Does this mean that you impose a constraint with a value of 0.0 for faces that are not found? This would be restrictive since one might want to impose other inhomogeneous boundary conditions for those faces that are outside the other tria.

applications/poisson/overset_grids/input.json
nfehn: this is something I could change already for the present PR

nfehn: done here https://github.com/exadg/exadg/pull/186/commits/fa5eb24dfa3c6633fbf487f7b7d8e8954bd32165

nfehn:
i would merge once the checks completed

PR #187, opened 2022-03-07
improve robustness of InterfaceCoupling regarding marked_vertices

nfehn:
Independently of this PR, dealii should provide additional measures to improve robustness/correctness of RemotePointEvaluation in my opinion, see dealii issue https://github.com/dealii/dealii/issues/13501.

### DIFF
--- a/include/exadg/functions_and_boundary_conditions/interface_coupling.h
+++ b/include/exadg/functions_and_boundary_conditions/interface_coupling.h
@@ -207,15 +207,8 @@ public:
     // To improve robustness, make sure that not all entries of marked_vertices are false.
     // Otherwise, points will simply not be found by RemotePointEvaluation and results will
     // probably be wrong.
-    bool all_false = true;
-    for(auto iter : marked_vertices)
-    {
-      if(iter == true)
-        all_false = false;
-    }
-
-    if(all_false)
-      marked_vertices = std::vector<bool>(dof_handler_src_.get_triangulation().n_vertices(), true);
+    if (std::all_of(marked_vertices.begin(), marked_vertices.end(), [](bool marked){return marked == false;}))
+      std::fill(marked_vertices.begin(), marked_vertices.end(), true);
 #endif
 
     for(auto quad_index : interface_data_dst->get_quad_indices())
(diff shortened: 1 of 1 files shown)


include/exadg/functions_and_boundary_conditions/interface_coupling.h
```diff
@@ -196,6 +203,19 @@ class InterfaceCoupling
...
+    // probably be wrong.
+    bool all_false = true;
+    for(auto iter : marked_vertices)
+    {
+      if(iter == true)
+        all_false = false;
+    }
+
+    if(all_false)
+      marked_vertices = std::vector<bool>(dof_handler_src_.get_triangulation().n_vertices(), true);
```
kronbichler, line 218: I am not sure if it is preferable, but I would write this as:
```suggestion
    if (std::all_off(marked_vertices.begin(), marked_vertices.end(), false))
      std::fill(marked_vertices.begin(), marked_vertices.end(), true);
```
I am not entirely sure if the `bool` vectors support these operations as they are somewhat special compared to other types like `int`, but I hope they do.

nfehn: I think one has to write is slightly differently using a lambda.

Please check again.

nfehn:
@kronbichler could you review again?

include/exadg/functions_and_boundary_conditions/interface_coupling.h
```diff
@@ -196,6 +203,16 @@ class InterfaceCoupling
...
     }
+
+    // To improve robustness, make sure that not all entries of marked_vertices are false.
+    // Otherwise, points will simply not be found by RemotePointEvaluation and results will
+    // probably be wrong.
+    if(std::all_of(marked_vertices.begin(), marked_vertices.end(), [](bool vertex_is_marked) {
+         return vertex_is_marked == false;
+       }))
+    {
+      std::fill(marked_vertices.begin(), marked_vertices.end(), true);
```
peterrum, line 214:
```suggestion
      marked_vertices.clear();
```

nfehn: Could you explain why?

nfehn:
@peterrum could this change (or something similar) be included into deal.II?

peterrum:
No, I don't think so. The way to indicate that no marked vertices are given, is to pass in an empty vector. RPE just passes the array to the right functions.

PR #188, opened 2022-03-07
take DirichletCached BCs into account for multigrid setup

nfehn:
Currently, the code ignores DirichletCached BCs for the setup of multigrid (which requires a map of Dirichlet BCs). From the perspective of multigrid, there is no difference between both types of Dirichlet BCs. Hence, the present PR is a bugfix.

Note that the code currently crashed in case of pure `DirichletCached` BCs, as mentioned in PR #186.

One might argue that the current design is not optimal if it is possible to produce such bugs without noticing this for a while. However, it is also not clear how to incorporate DirichletCached BCs into standard Dirichlet BCs using dealii::Function. One could derive FunctionCached from dealii::Function (maybe even introduce this in dealii later ...), and then use dynamic casts to identify whether a boundary is of type DirichletCached (which is probably also not nice but maybe less error prone?). What do you think?

### DIFF
--- a/include/exadg/poisson/spatial_discretization/operator.cpp
+++ b/include/exadg/poisson/spatial_discretization/operator.cpp
@@ -285,7 +285,9 @@ Operator<dim, Number, n_components>::setup_solver()
 
     // We also need to add DirichletCached boundary conditions. From the
     // perspective of multigrid, there is no difference between standard
-    // and cached Dirichlet BCs.
+    // and cached Dirichlet BCs. Since multigrid does not need information
+    // about inhomogeneous boundary data, we simply fill the map with
+    // dealii::Functions::ZeroFunction for DirichletCached BCs.
     for(auto iter : laplace_operator.get_data().bc->dirichlet_cached_bc)
       dirichlet_boundary_conditions.insert(
         pair(iter.first, new dealii::Functions::ZeroFunction<dim>(dim)));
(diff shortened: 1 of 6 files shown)


nfehn:
should be ready for review now from my side.

We might want to discuss whether we want to change the interface of `MultigridPreconditionerBase`, but this should be the content of follow-up issues/PRs. The present PR should just fix the bug.

include/exadg/poisson/spatial_discretization/operator.cpp
```diff
@@ -277,13 +277,28 @@ Operator<dim, Number, n_components>::setup_solver()
...
+      dirichlet_boundary_conditions = laplace_operator.get_data().bc->dirichlet_bc;
+
+    // We also need to add DirichletCached boundary conditions. From the
+    // perspective of multigrid, there is no difference between standard
+    // and cached Dirichlet BCs. Since multigrid does not need information
+    // about inhomogeneous boundary data, we simply fill the map with
+    // dealii::Functions::ZeroFunction for DirichletCached BCs.
+    for(auto iter : laplace_operator.get_data().bc->dirichlet_cached_bc)
+      dirichlet_boundary_conditions.insert(
+        pair(iter.first, new dealii::Functions::ZeroFunction<dim>(dim)));
```
kronbichler, line 293: Should this be
```suggestion
        pair(iter.first, new dealii::Functions::ZeroFunction<dim>(1)));
```

kronbichler (approved):
This looks good.

include/exadg/poisson/spatial_discretization/operator.cpp
nfehn: should be n_components, has been fixed

PR #191, opened 2022-03-10
Adjust mpi_type_id to new form

kwic:


peterrum (approved):

include/exadg/postprocessor/kinetic_energy_spectrum.cpp
```diff
@@ -110,7 +110,7 @@ class DealSpectrumWrapper
     MPI_Exscan(&n_local_cells,
                &global_offset,
                1,
-               dealii::Utilities::MPI::mpi_type_id(&global_offset),
+               dealii::Utilities::MPI::mpi_type_id_for_type<decltype(global_offset)>,
```
peterrum, line 113: Thanks for the fix. Since, we also support deal.II 9.3, we need to versions here. It can be accomplished via ifdefs like here: https://github.com/exadg/exadg/blob/fd4290a3e5cc593bfc41b99f85511b20547eaec6/include/exadg/functions_and_boundary_conditions/interface_coupling.h#L176

kwic: Hm. I don't quite get the versioning here. Will 9.3.0 be followed by 10.0.0?

I'd like to do a DEAL_II_VERSION_LTE(9,3,0) to ensure a possible 9.4.0 is also covered, but there is no such macro.

nfehn: we already know that there will be no dealii 9.4 :), so you can do it as suggested by peter

kwic: Done

nfehn (approved):

peterrum:
Now there is also `DEAL_II_VERSION_LTE` (https://github.com/dealii/dealii/pull/13530).

include/exadg/postprocessor/kinetic_energy_spectrum.cpp
nfehn: my statement above turned out incorrect, see https://github.com/dealii/dealii/pull/13553

PR #194, opened 2022-03-11
update and refactor configure scripts

nfehn:
I collected all the directories at the top of the file. Moreover, we should not assume the existence of a directory "sw" for TPLs (this is a rather specific setup). The default setting should be to have all libraries in the WORKING_DIRECTORY.

The Wiki has already been adjusted.


nfehn:
With the setting from commit https://github.com/exadg/exadg/pull/194/commits/e57b9b0073884662784b902da0d942b7ca6beb50, and when installing deal.II to "sw/dealii-install" and p4est to "sw/p4est-install", I get an error when compiling exadg (using dealii-install):

/home/fehn/sw/dealii-install/include/deal.II/distributed/p4est_wrappers.h:24:12: fatal error: p4est_bits.h: No such file or directory
   24 | #  include <p4est_bits.h>

It seems to be fine when using "sw/dealii-build" for compiling exadg, or when not using a separate "p4est-install".

kronbichler:
This is strange: cmake tries to find `p4est` in the directory `-isystem /home/fehn/sw/FAST/include` (this is what is included in the compile line), which is not where you indicate you installed `p4est`, which is rather `-isystem /home/fehn/sw/p4est-install/FAST/include`. I do not often work with the installed version, but I believe you should have the file `/home/fehn/sw/dealii-install/detailed.log`. Can you check which path it defines in `P4EST_INCLUDE_DIRS`?

PR #198, opened 2022-03-15
clarify logic for setting manifold IDs in application examples

nfehn:
I did not observe problems, but I think it makes sense to correct/clarify the implementation nevertheless

So far done for FSI examples. I will address all other occurrences as well.

### DIFF
--- a/applications/fluid_structure_interaction/cylinder_with_flag/application.h
+++ b/applications/fluid_structure_interaction/cylinder_with_flag/application.h
@@ -421,11 +421,13 @@ private:
 
     for(auto cell : this->grid->triangulation->active_cell_iterators())
     {
+      double const TOL = 1.e-12;
+
+      // boundary IDs
       for(unsigned int f = 0; f < dealii::GeometryInfo<dim>::faces_per_cell; ++f)
       {
-        double const x   = cell->face(f)->center()(0);
-        double const y   = cell->face(f)->center()(1);
-        double const TOL = 1.e-12;
+        double const x = cell->face(f)->center()(0);
+        double const y = cell->face(f)->center()(1);
 
         if(std::fabs(y - Y_0) < TOL || std::fabs(y - H) < TOL)
         {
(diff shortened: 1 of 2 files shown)


applications/incompressible_navier_stokes/vortex/application.h
```diff
@@ -455,96 +456,90 @@ class Application : public ApplicationBase<dim, Number>
...
+      else
+      {
+        AssertThrow(false, dealii::ExcMessage("Should not arrive here."));
+      }
 
-      // refine globally due to boundary conditions for vortex problem
-      this->grid->triangulation->refine_global(1);
+      tria_coarse.refine_global(1);
+      // make sure that the triangulation does not have refinements at this point
+      dealii::GridGenerator::flatten_triangulation(tria_coarse, *this->grid->triangulation);
```
nfehn, line 542: @kronbichler, @peterrum: I encountered a problem for this test case if `mesh_type == MeshType::ComplexVolumeManifold` or `mesh_type == MeshType::ComplexSurfaceManifold`. The setup of the Chebyshev coarse grid solver fails (which is not related to the changes regarding manifolds IDs proposed in this PR, the problem also occurs with the previous code version). Due to the boundary conditions for this "vortex problem" (we have both Dirichlet and Neumann on each of the four sides of the rectangular domain), we have  to refine once before setting the boundary IDs. However, this refinement is considered an additional level in multigrid. I assume the boundary conditions are then somewhat undefined for the coarse grid problem (???, what does the deal.II code do in such a case?), so that the Chebyshev coarse grid solver can not converge ...

So I tried to solve this by removing this refinement level by calling `flatten_triangulation()`, considering the resulting (already refined) triangulation as the coarse grid problem. The solver now indeed converges.

However, with `flatten_triangulation()`, the manifolds have also disappeared, but the manifolds are main purpose of this code!? Applying the manifolds after `flatten_triangulation()` for the coarse grid (which is already a "refined" one in terms of number of cells) would be more complicated since manifolds would also have to be applied to cells not located at the spherical internal boundary to achieve the same behavior as before. 

@kronbichler, @peterrum: If you also do not know a solution, I would suggest to remove this code dealing with `mesh_type == MeshType::ComplexVolumeManifold` and `mesh_type == MeshType::ComplexSurfaceManifold` completely from this test case, implying that we would loose some functionality.

nfehn: > Not sure. But cannot you attach the manifolds again after the flattening?

The problem is not the flattening, but that the grid is now finer. Since we apply the manifolds manually for all the relevant cells, this is much more complicated if the grid is refined. Using the coarsest possible grid, the criterion whether the cell touches the spherical internal boundary is sufficient as a criterion to apply the manifolds. This does no longer work when using the refined grid.

nfehn: @kronbichler what do you think?

PR #201, opened 2022-03-16
Generalizations for overset grids functionality

nfehn:
This PR implements generalizations of the overset grids functionality. The overlap region is determined automatically and boundary IDs are set accordingly for DirichletCached BCs.

The function is written generically with a vector of points (which could be a single point (e.g. vertex), or represent a cell/face). Regarding the nomenclature, this vector (or cloud) of points is called an "object" -> `locate_object_relative_to_triangulation()`.

Limitations: 

- only works in serial. For the parallel case, something like `RemotePointEvaluation` could be helpful. @peterrum are you interested to discuss how we could use/extend `RemotePointEvaluation` for this purpose?
- Only takes into account the vertices of a face when classifying a face in terms of `Location::Inside/Outside/Intersected`. For a correct identification of `Location::Inside/Outside/Intersected` all the mapping support points of the (potentially high-order) mapping should be taken into account.

Here are examples for coarse and fine meshes for domain1:

automatic_coarse
automatic_fine

### DIFF
--- a/applications/poisson/overset_grids/application.h
+++ b/applications/poisson/overset_grids/application.h
@@ -77,21 +77,10 @@ public:
     double const       right = 1.0;
     dealii::Point<dim> p1, p2;
     p1[0] = 0.0;
-    p1[1] = 0.0;
+    p1[1] = 0.5;
     p2[0] = right;
-    p2[1] = 1.0;
+    p2[1] = 1.5;
     dealii::GridGenerator::subdivided_hyper_rectangle(*this->grid->triangulation, {3, 3}, p1, p2);
-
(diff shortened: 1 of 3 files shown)


nfehn:
@peterrum is this mergeable from your side? Note that we have the checks with number of mpi processes == 1 in the code.

Another generalization using RemotePointEvaluation could then be done in a follow up.

nfehn:
@peterrum is this mergeable from your side? Note that we have the checks with number of mpi processes == 1 in the code.

Another generalization using RemotePointEvaluation could then be done in a follow up.

peterrum (approved):
Looks good.

PR #208, opened 2022-03-17
use const as trailing keyword

### LINKED ISSUE #203: preCICE-based FSI code: follow coding convention "double const" (and similarily for all other types)
@DavidSCN: a follow up of #148

nfehn:
closes #203

### DIFF
--- a/include/exadg/fluid_structure_interaction/precice/coupling_base.h
+++ b/include/exadg/fluid_structure_interaction/precice/coupling_base.h
@@ -63,10 +63,10 @@ template<int dim, int data_dim, typename VectorizedArrayType>
 class CouplingBase
 {
 public:
-  CouplingBase(std::shared_ptr<const dealii::MatrixFree<dim, double, VectorizedArrayType>> data,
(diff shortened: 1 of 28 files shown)


DavidSCN:
Thank you!

nfehn:
@DavidSCN did you go through the changes? It would be good if someone else has briefly checked the changes before merging.

peterrum (approved):
A short question: have you done this manually or have used a script. In the latter case, it would be nice to make it part of the indentation check of CI.

nfehn:
Unfortunately, only semi-manual. Run `contrib/utilities/search-and-replace-all` for all the data types that occur ;)

nfehn:
If we collect all the data types that we know of in a file/vector, we could automatize this process and we would at least cover most of the cases automatically.

nfehn:
For full automatization, we would need to detect data types automatically, which is not trivial since an empty character is not necessarily the end of the name `Type<int, bool>` for example. Also, how do we make sure that `double const value` is not changed to `double value const` by such a generic script?

PR #213, opened 2022-03-24
remove GeometryInfo

nfehn:


applications/fluid_structure_interaction/perpendicular_flap/application.h
```diff
@@ -336,7 +336,7 @@ class Application : public FluidFSI::ApplicationBase<dim, Number>
 
     for(auto cell : this->grid->triangulation->active_cell_iterators())
     {
-      for(unsigned int f = 0; f < dealii::GeometryInfo<dim>::faces_per_cell; ++f)
+      for(auto const & f : cell.face_indices())
```
peterrum, line 339:
```suggestion
      for(auto const & f : cell->face_indices())
```

applications/fluid_structure_interaction/perpendicular_flap/application.h
```diff
@@ -682,7 +682,7 @@ class Application : public StructureFSI::ApplicationBase<dim, Number>
...
```
peterrum, line 685:
```suggestion
      for(auto const & f : cell->face_indices())
```

peterrum (approved):
Looks good. I am wondering why you write `for(auto const & face : cell.face_indices())` instead of `for(auto const face : cell.face_indices())` (it is just an int).

applications/fluid_structure_interaction/perpendicular_flap/application.h
nfehn: @DavidSCN It seems as if the perpendicular_flap example is not compiled if one does not use precice. However, this test case could also be used for the ExaDG native FSI implementation. Hence, could you change/shift the "precice"-guards from the CMakeLists.txt file for this application to the precice-based FSI solver implementation?

To avoid problems as the one encountered here, we should give it a try to only guard the precice instance in the precice-based FSI driver, but always compile the rest of the driver. I am undecided whether we should guard the whole driver (assume for example that a class in ExaDG is just renamed, so that all places using this class need to be adapted.).

DavidSCN: >  It seems as if the perpendicular_flap example is not compiled if one does not use precice.

That's the case, yes.

I never tried to run the case using the native ExaDG coupling. We can put into the TODO list.

Issue #209, opened 2022-03-18
preCICE-based FSI code: follow-up tasks


DavidSCN:
The preCICE coupled solvers were added in #148. Although the implementation there ended up being much more mature then initially intended, there are some follow up tasks I would like to document. In particular, I would like to work on

- [ ] support for adaptive time-stepping of the coupled simulation
- [ ] compatibility to other coupled codes, which operate solely on `Displacement` and `Stress` as coupling data. This would require some simple time derivative on the Fluid side in order to recover the interface velocity for the NS solver (and probably destroy higher-order time-stepping schemes) from the displacement data.
- [ ] stress computation based on the weak form as described [in this paper](http://dx.doi.org/https://doi.org/10.1115/1.4005187) (not yet completely sure whether this pays off, but I would like to give it a trial)
- [ ] support for multiple boundary IDs forming the coupling interface
- [ ] (maybe) adding an in-solver mapping using `RemotePointEvaluation`, similar to what ExaDG is doing. Hence, only relying on the communication and coupling schemes of preCICE.

PR #214, opened 2022-03-26
[preCICE FSI] Support adaptive-time stepping in serial coupling schemes

DavidSCN:
As promised in #209. While doing this, I realized that subcycling is somewhat challenging and I guess the effort doesn't pay off right now. The current implementation would essentially do the same as ExaDG's implementation in terms of the time-step size, i.e., an adaptive time stepping for staggered coupling schemes and a constant time-window size for parallel coupling schemes. I tried to document capabilities in a more comprehensive way in the code.

### DIFF
--- a/applications/fluid_structure_interaction/perpendicular_flap/application.h
+++ b/applications/fluid_structure_interaction/perpendicular_flap/application.h
@@ -151,12 +151,13 @@ private:
     param.treatment_of_convective_term    = TreatmentOfConvectiveTerm::Explicit;
     param.order_time_integrator           = 1;
     param.start_with_low_order            = true;
-    param.adaptive_time_stepping          = false;
-    param.calculation_of_time_step_size   = TimeStepCalculation::UserSpecified;
+    param.adaptive_time_stepping          = true;
+    param.calculation_of_time_step_size   = TimeStepCalculation::CFL;
     param.time_step_size                  = DELTA_T;
+    param.time_step_size_max              = DELTA_T;
     param.max_velocity                    = U_MEAN;
-    param.cfl                             = 0.5;
-    param.cfl_exponent_fe_degree_velocity = 1.5;
+    param.cfl                             = 0.75;
+    param.cfl_exponent_fe_degree_velocity = 1.2;
 
     // output of solver information
     param.solver_info_data.interval_time_steps = OUTPUT_SOLVER_INFO_EVERY_TIME_STEPS;
@@ -619,7 +620,7 @@ public:
     param.problem_type         = ProblemType::Unsteady;
     param.body_force           = false;
     param.pull_back_body_force = false;
-    param.large_deformation    = true;
+    param.large_deformation    = false;
     param.pull_back_traction   = true;
 
     param.density = DENSITY_STRUCTURE;
--- a/applications/fluid_structure_interaction/perpendicular_flap/application.h
+++ b/applications/fluid_structure_interaction/perpendicular_flap/application.h
@@ -156,8 +156,8 @@ private:
     param.time_step_size                  = DELTA_T;
     param.time_step_size_max              = DELTA_T;
     param.max_velocity                    = U_MEAN;
-    param.cfl                             = 0.75;
-    param.cfl_exponent_fe_degree_velocity = 1.2;
+    param.cfl                             = 0.9;
+    param.cfl_exponent_fe_degree_velocity = 1.5;
 
     // output of solver information
     param.solver_info_data.interval_time_steps = OUTPUT_SOLVER_INFO_EVERY_TIME_STEPS;
--- a/include/exadg/fluid_structure_interaction/precice/driver.h
+++ b/include/exadg/fluid_structure_interaction/precice/driver.h
@@ -83,6 +83,9 @@ protected:
   std::shared_ptr<ExaDG::preCICE::Adapter<dim, dim, VectorType>> precice;
   ExaDG::preCICE::ConfigurationParameters                        precice_parameters;
 
+  // maximum allowed time-step size until we reach a new coupling time window
+  mutable double allowed_time_step_size = 0;
+
   // do not print wall times if is_test
   bool const is_test;
 
--- a/include/exadg/fluid_structure_interaction/precice/driver.h
+++ b/include/exadg/fluid_structure_interaction/precice/driver.h
@@ -83,6 +83,23 @@ protected:
   std::shared_ptr<ExaDG::preCICE::Adapter<dim, dim, VectorType>> precice;
   ExaDG::preCICE::ConfigurationParameters                        precice_parameters;
 
+  // Nomenclature
+  // time window: denotes time interval after which FSi coupling takes place
+  // subcycling: a single field solver performs several time steps until the next FSI coupling time
+  // window is reached
+  // time step size: refers to single field solvers. In case of subcycling, one
+  // time window consists of several time step sizes.
+
+  // The time-window size is determined through the preCICE configuration by the user:
+  // The <coupling-scheme.. tag has options for
+  // <time-window-size method="fixed" ... /> which refers to a constant time window size
+  // <time-window-size method="first-participant" ... /> which refers to an adaptive time window
+  // size prescribed by the first (which is usually the Fluid) participant. The latter is only
+  // applicable in serial coupling schemes.
+  // Both solvers can exploit subcycling, but they need to synchronize at the end of a certain
+  // time-window size. Therefore, the allowed time-step size until we reach the next synchronization
+  // point needs to be determined.
+
   // maximum allowed time-step size until we reach a new coupling time window
   mutable double allowed_time_step_size = 0;
 
--- a/include/exadg/fluid_structure_interaction/precice/driver.h
+++ b/include/exadg/fluid_structure_interaction/precice/driver.h
@@ -89,9 +89,13 @@ protected:
   // window is reached
   // time step size: refers to single field solvers. In case of subcycling, one
   // time window consists of several time step sizes.
+  // serial coupling scheme: the single-field solvers are called sequentially/consecutively with
+  // data transfer from one field to another after in single-field solver.
+  // parallel coupling scheme: the single-field solvers are called simultaneously with data transfer
+  // in both directions afterwards.
 
   // The time-window size is determined through the preCICE configuration by the user:
-  // The <coupling-scheme.. tag has options for
+  // The <coupling-scheme.. /> tag has options for
   // <time-window-size method="fixed" ... /> which refers to a constant time window size
   // <time-window-size method="first-participant" ... /> which refers to an adaptive time window
   // size prescribed by the first (which is usually the Fluid) participant. The latter is only
@@ -101,7 +105,7 @@ protected:
   // point needs to be determined.
 
   // maximum allowed time-step size until we reach a new coupling time window
-  mutable double allowed_time_step_size = 0;
+  mutable double time_until_next_coupling_window = 0;
 
   // do not print wall times if is_test
   bool const is_test;
--- a/include/exadg/fluid_structure_interaction/precice/driver_fluid.h
+++ b/include/exadg/fluid_structure_interaction/precice/driver_fluid.h
@@ -219,19 +219,18 @@ public:
       {
         // computes new time-step size
         fluid->time_integrator->advance_one_timestep_post_solve();
-        // next, we synchronize the time-step sizes. Subcycling would in theory be compatible in
-        // explicit coupling schemes here. In implicit coupling schemes, we need matching
-        // time-window sizes (either constant (serial or parallel schemes) or adaptively (serial
-        // scheme)) as the time-step size push back happens in the
-        // 'advance_one_timestep_post_solve()' and we cannot change two subsequent time-step sizes
-        // without a push back operation, otherwise we falsify the time integrator. However, the
-        // treatment of the boundary conditions is not handled correctly when using subcycling, as
-        // we apply the 'whole displacement' within the first time-step. In order to cope with
-        // subcycling, either the boundary condition needs to be adopted or we need to wait for a
-        // newer preCICE version (currently v2.3.0) which can handle this.
-        // Hence, we do not adjust the time-step-size of the fluid solver, but rather Assert that it
-        // is still valid. Note that we would run into a preCICE error otherwise reporting the same
-        // issue otherwise.
+        // next, we synchronize the time-step sizes. Have a look in the base class for more
+        // explanations. Subcycling would in theory be compatible in explicit coupling schemes here.
+        // In implicit coupling schemes, we need matching time-window sizes as the time-step size
+        // push back happens in the 'advance_one_timestep_post_solve()' and we cannot change two
+        // subsequent time-step sizes without a push back operation, otherwise we falsify the time
+        // integrator. However, the treatment of the boundary conditions is not handled correctly
+        // when using subcycling, as we would apply the 'whole displacement' within the first
+        // time-step (being potentially only a fractio of the whole time-windo). In order to cope
+        // with subcycling, either the boundary condition needs to be adapted or we need to wait for
+        // a newer preCICE version (currently v2.3.0) which can handle this. Hence, we do not adjust
+        // the time-step-size of the fluid solver, but rather Assert that it is still valid. Note
+        // that we would otherwise run into a preCICE assert reporting the same issue.
         // fluid->time_integrator->set_current_time_step_size(
         //   std::min(this->allowed_time_step_size, fluid->time_integrator->get_time_step_size()));
         Assert(
@@ -240,6 +239,7 @@ public:
           dealii::ExcMessage(
             "The solver time-step size exceeded the maximum admissible time-step size allowed by preCICE. "
             "If you select adaptive time-stepping, make sure to let the Fluid participant steer the time-window size. "
+            "(precice config: <coupling-scheme:serial... > <time-window-size method=\"first-participant\" />) "
             "In any other case, please disable adaptive time-stepping."));
       }
     }
--- a/include/exadg/fluid_structure_interaction/precice/driver_fluid.h
+++ b/include/exadg/fluid_structure_interaction/precice/driver_fluid.h
@@ -219,15 +219,28 @@ public:
       {
         // computes new time-step size
         fluid->time_integrator->advance_one_timestep_post_solve();
-        // next, we synchronize the time-step sizes. Subcycling would be possible in explicit
-        // coupling schemes. In implicit coupling schemes, we need matching time-window sizes
-        // (either constant (serial or parallel schemes) or adaptively (serial scheme)) as the
-        // time-step size push back happens in the 'advance_one_timestep_post_solve()' and we cannot
-        // change two subsequent time-step sizes without a push back operation, otherwise we falsify
-        // the time integrator. In case one selects an adaptive time-step scheme here and a constant
-        // time-window size in preCICE, preCICE will throw an error
-        fluid->time_integrator->set_current_time_step_size(
-          std::min(this->allowed_time_step_size, fluid->time_integrator->get_time_step_size()));
+        // next, we synchronize the time-step sizes. Subcycling would in theory be compatible in
+        // explicit coupling schemes here. In implicit coupling schemes, we need matching
+        // time-window sizes (either constant (serial or parallel schemes) or adaptively (serial
+        // scheme)) as the time-step size push back happens in the
+        // 'advance_one_timestep_post_solve()' and we cannot change two subsequent time-step sizes
+        // without a push back operation, otherwise we falsify the time integrator. However, the
+        // treatment of the boundary conditions is not handled correctly when using subcycling, as
+        // we apply the 'whole displacement' within the first time-step. In order to cope with
+        // subcycling, either the boundary condition needs to be adopted or we need to wait for a
+        // newer preCICE version (currently v2.3.0) which can handle this.
+        // Hence, we do not adjust the time-step-size of the fluid solver, but rather Assert that it
+        // is still valid. Note that we would run into a preCICE error otherwise reporting the same
+        // issue otherwise.
+        // fluid->time_integrator->set_current_time_step_size(
+        //   std::min(this->allowed_time_step_size, fluid->time_integrator->get_time_step_size()));
+        Assert(
+          fluid->time_integrator->get_time_step_size() <
+            this->allowed_time_step_size + std::numeric_limits<double>::min(),
+          ExcMessage(
+            "The solver time-step size exceeded the maximum admissible time-step size allowed by preCICE. "
+            "If you select adaptive time-stepping, make sure to let the Fluid participant steer the time-window size. "
+            "In any other case, please disable adaptive time-stepping."));
       }
     }
   }
--- a/include/exadg/fluid_structure_interaction/precice/driver_fluid.h
+++ b/include/exadg/fluid_structure_interaction/precice/driver_fluid.h
@@ -151,7 +151,7 @@ public:
     // initialize preCICE with initial stress data
     VectorType initial_stress;
     fluid->pde_operator->initialize_vector_velocity(initial_stress);
-    this->precice->initialize_precice(initial_stress);
+    this->allowed_time_step_size = this->precice->initialize_precice(initial_stress);
   }
 
 
@@ -178,10 +178,9 @@ public:
   void
   solve() const final
   {
-    Assert(this->application->fluid->get_parameters().adaptive_time_stepping == false,
-           dealii::ExcNotImplemented());
-
     bool is_new_time_window = true;
+    fluid->time_integrator->set_current_time_step_size(
+      std::min(this->allowed_time_step_size, fluid->time_integrator->get_time_step_size()));
     // preCICE dictates when the time loop is finished
     while(this->precice->is_coupling_ongoing())
     {
@@ -205,10 +204,9 @@ public:
         // compute and send stress to solid
         coupling_fluid_to_structure();
 
-        // TODO: Add synchronization for the time-step size here. For now, we only allow a constant
-        // time-step size
         dealii::Timer precice_timer;
-        this->precice->advance(fluid->time_integrator->get_time_step_size());
+        this->allowed_time_step_size =
+          this->precice->advance(fluid->time_integrator->get_time_step_size());
         is_new_time_window = this->precice->is_time_window_complete();
         this->timer_tree.insert({"FSI", "preCICE"}, precice_timer.wall_time());
       }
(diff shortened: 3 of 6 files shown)


applications/fluid_structure_interaction/perpendicular_flap/application.h
```diff
@@ -619,7 +620,7 @@ class Application : public StructureFSI::ApplicationBase<dim, Number>
     param.problem_type         = ProblemType::Unsteady;
     param.body_force           = false;
     param.pull_back_body_force = false;
-    param.large_deformation    = true;
+    param.large_deformation    = false;
```
nfehn, line 623: this changes the mathematical model (linear elasticity is now used). Is this intended?

applications/fluid_structure_interaction/perpendicular_flap/application.h
```diff
@@ -151,12 +151,13 @@ class Application : public FluidFSI::ApplicationBase<dim, Number>
...
-    param.calculation_of_time_step_size   = TimeStepCalculation::UserSpecified;
+    param.adaptive_time_stepping          = true;
+    param.calculation_of_time_step_size   = TimeStepCalculation::CFL;
     param.time_step_size                  = DELTA_T;
+    param.time_step_size_max              = DELTA_T;
     param.max_velocity                    = U_MEAN;
-    param.cfl                             = 0.5;
-    param.cfl_exponent_fe_degree_velocity = 1.5;
+    param.cfl                             = 0.75;
+    param.cfl_exponent_fe_degree_velocity = 1.2;
```
nfehn, line 160: Out of curiosity, where does this change (1.5 -> 1.2) come from? Did you make parameter studies for different polynomial degrees?

include/exadg/fluid_structure_interaction/precice/driver_solid.h
```diff
@@ -181,6 +185,12 @@ class DriverSolid : public Driver<dim, Number>
 
       if(is_new_time_window)
         structure->time_integrator->advance_one_timestep_post_solve();
+
+      // We need to adjust the time-step size in case we reach a coupling time-wimdow
```
nfehn, line 189: .

include/exadg/fluid_structure_interaction/precice/driver_solid.h
```diff
@@ -181,6 +185,12 @@ class DriverSolid : public Driver<dim, Number>
...
+      // We don't take the time-step size of the solver here into account, but rather apply the
+      // available time-step size reported by preCICE, as the GenAlpha time integrator doesn't
+      // compute a time-step size on its own
```
nfehn, line 192: add "." at end of sentence.

include/exadg/fluid_structure_interaction/precice/driver_fluid.h
```diff
@@ -218,7 +216,32 @@ class DriverFluid : public Driver<dim, Number>
...
+        // is still valid. Note that we would run into a preCICE error otherwise reporting the same
+        // issue otherwise.
+        // fluid->time_integrator->set_current_time_step_size(
+        //   std::min(this->allowed_time_step_size, fluid->time_integrator->get_time_step_size()));
+        Assert(
+          fluid->time_integrator->get_time_step_size() <
+            this->allowed_time_step_size + std::numeric_limits<double>::min(),
+          dealii::ExcMessage(
+            "The solver time-step size exceeded the maximum admissible time-step size allowed by preCICE. "
+            "If you select adaptive time-stepping, make sure to let the Fluid participant steer the time-window size. "
```
nfehn, line 242: How do I make this decision ("let the Fluid participant steer the time-window size")? Is it a precice parameter? Please refer to the parameter that makes this decision.

include/exadg/fluid_structure_interaction/precice/driver_fluid.h
```diff
@@ -218,7 +216,32 @@ class DriverFluid : public Driver<dim, Number>
...
+        // scheme)) as the time-step size push back happens in the
+        // 'advance_one_timestep_post_solve()' and we cannot change two subsequent time-step sizes
+        // without a push back operation, otherwise we falsify the time integrator. However, the
+        // treatment of the boundary conditions is not handled correctly when using subcycling, as
+        // we apply the 'whole displacement' within the first time-step. In order to cope with
+        // subcycling, either the boundary condition needs to be adopted or we need to wait for a
+        // newer preCICE version (currently v2.3.0) which can handle this.
+        // Hence, we do not adjust the time-step-size of the fluid solver, but rather Assert that it
+        // is still valid. Note that we would run into a preCICE error otherwise reporting the same
+        // issue otherwise.
```
nfehn, line 234: This text is hard to understand in my opinion. The documentation of `Driver` (either the base class or both derived classes, probably the base class) has to describe certain nomenclature in my opinion, e.g.

time window: denotes time interval after which FSi coupling takes place
subcycling: a single field solver performs several time steps until the next FSI coupling time window is reached
time step size: refers to single field solvers. In case of subcycling, one time window consists of several time step sizes.

I don't know if this is a precise description (I only want to show which type of information is necessary in my opinion).

Next, describe how the time window size is determined, since this probably comes from pecice. Next, also describe whether the single field solvers have to hit the begin/end of a time window exactly, or whether the time step sizes between fluid and the precice time-window (as well as solid and the precice time-window) may be staggered. Then, assuming matching time windows, may the solid and fluid time step sizes differ in case of subcycling?

I don't know whether it make sense to discuss here constant vs. adaptive schemes (in terms of precice and the window, if I understand it correctly?)? The implementation updates `this->allowed_time_step_size`, so it looks like the implementation is capable of adaptive time window sizes?

I do not understand explicit vs. implicit coupling schemes. What do you use currently? I would have expected that an implicit coupling scheme is used.

I think this is not the right place to discuss serial vs. parallel schemes (again, this would need precise definitions of serial/parallel so that it is understandable from an ExaDG perspective).

nfehn:
I don't understand "serial" in the title of this PR? I think to remember that you wanted to concentrate on parallel coupling schemes here in ExaDG (in the discussion how to design the driver class, base class with two derived classes, or one single class) since the serial coupling would be very similar to the ExaDG-native FSI implementation. Also, I don't know why serial/parallel makes a difference at all?

nfehn:
From the current implementation, I do not see which time step sizes are adjusted adaptively (fluid, but there is an assert? also solid?), and which time step sizes / time windows have to be matching, or may be staggered?

nfehn:
> As promised in https://github.com/exadg/exadg/issues/209.

Please describe in more detail what this PR enables exactly as compared to the previous code version. Nomenclature such as "subcyling"/"time window" is not described in #209, so it is not clear why these terms appear here.

applications/fluid_structure_interaction/perpendicular_flap/application.h
DavidSCN: It's actually not intended as part of this workpackage (it has nothing to do with time), but I would like to change it for consistency reasons as the corresponding tutorial also relies on linear elasticity models on the structural side.

applications/fluid_structure_interaction/perpendicular_flap/application.h
DavidSCN: Short answer: no. Before the current changes, I used a fixed time-step size so that this parameters had no influence on the final result. I tried (and ran) the setup now with adaptive time-stepping, but the computed time-step size is very small (compared to the fixed one). The perpendicular flap case is somewhat artificial in the sense that the setup has a very high viscosity (low Re) so that bigger time-step sizes are acceptable and do not lead to unstable behavior. Hence, I just tried to loosen up the time-step restrictions.

include/exadg/fluid_structure_interaction/precice/driver_fluid.h
DavidSCN: Ok, I can clear up things, but it might become lengthy. Should I still put it in this place as a source code comment?

DavidSCN:
> since the serial coupling would be very similar to the ExaDG-native FSI implementation. Also, I don't know why serial/parallel makes a difference at all?

To make it short: this PR implements indeed the same time-step adaptivity as ExaDG, i.e., a serial (explicit or implicit doesn't matter, this is a configuration thing in preCICE) coupling scheme where the Fluid participant prescribes the time-step size. The main difference between parallel and serial coupling schemes is that the time-window size can be adjusted in serial schemes (one participant prescribes, the other one follows). In parallel schemes, each solver could still do its own time-step size, but the coupling time-window will remain the same. This however, requires you currently to adjust your boundary conditions. Imagine the case where the solid solver computes a large displacement and you start subcycling on the fluid side: the mesh motion solver would solve for the 'whole' mesh motion within the first time-step which is obviously wrong. There is a remedy to this problem on its way which would do this adjustments automatically ([Waveforms](https://onlinelibrary.wiley.com/doi/full/10.1002/nme.6443)), so I would not adjust this right now.
Yes, it would be similar to what ExaDG does, but only as long as I use ExaDG for both, Fluid and Solid. With this PR, we can do adaptive in time for serial coupling and constant in time for parallel coupling. In principle, an adaptive time-step size for parallel coupling could be applied, but it doesn't make sense in my opinion.

applications/fluid_structure_interaction/perpendicular_flap/application.h
nfehn: Then, I would use the standard value 1.5 for the exponent and adjust the time step size via larger cfl values.

include/exadg/fluid_structure_interaction/precice/driver_fluid.h
nfehn: One could also limit the scope, so that one does not have to explain everything and that it does not become overly lengthy. For example, if explicit vs. implicit is a detail / configuration option of precice, no need to discuss this here.

include/exadg/fluid_structure_interaction/precice/driver_fluid.h
DavidSCN: I added the respective line in the config.

include/exadg/fluid_structure_interaction/precice/driver_fluid.h
```diff
@@ -218,7 +216,32 @@ class DriverFluid : public Driver<dim, Number>
 
       // post-solve
       if(is_new_time_window)
+      {
+        // computes new time-step size
         fluid->time_integrator->advance_one_timestep_post_solve();
+        // next, we synchronize the time-step sizes. Have a look in the base class for more
```
nfehn, line 222: Where does a synchronization happen?

include/exadg/fluid_structure_interaction/precice/driver_fluid.h
```diff
@@ -218,7 +216,32 @@ class DriverFluid : public Driver<dim, Number>
...
+        // explanations. Subcycling would in theory be compatible in explicit coupling schemes here.
+        // In implicit coupling schemes, we need matching time-window sizes as the time-step size
+        // push back happens in the 'advance_one_timestep_post_solve()' and we cannot change two
+        // subsequent time-step sizes without a push back operation, otherwise we falsify the time
+        // integrator. However, the treatment of the boundary conditions is not handled correctly
```
nfehn, line 227: I do not understand this sentence.

include/exadg/fluid_structure_interaction/precice/driver_fluid.h
```diff
@@ -218,7 +216,32 @@ class DriverFluid : public Driver<dim, Number>
...
+        // when using subcycling, as we would apply the 'whole displacement' within the first
+        // time-step (being potentially only a fractio of the whole time-windo). In order to cope
```
nfehn, line 229: why "potentially"? Doesn't "subcycling" mean that one time-window consists of several time steps? 

There are some typos, but I would remove what is written in the brackets since this is redundant with the "nomenclature" comment.

include/exadg/fluid_structure_interaction/precice/driver.h
```diff
@@ -83,6 +83,26 @@ class Driver
...
+  // <time-window-size method="fixed" ... /> which refers to a constant time window size
+  // <time-window-size method="first-participant" ... /> which refers to an adaptive time window
+  // size prescribed by the first (which is usually the Fluid) participant. The latter is only
+  // applicable in serial coupling schemes.
+  // Both solvers can exploit subcycling, but they need to synchronize at the end of a certain
+  // time-window size. Therefore, the allowed time-step size until we reach the next synchronization
+  // point needs to be determined.
+
+  // maximum allowed time-step size until we reach a new coupling time window
+  mutable double allowed_time_step_size = 0;
```
nfehn, line 104: I am not sure whether I understand the meaning of this variable. I wonder whether "allowed" is intuitive to understand.

In case of subcycling with multiple time-steps per time window: Is the variable `allowed_time_step_size` adjusted from one time step to the next within one time window? If so, why not use instead `time_until_next_coupling_window`?

include/exadg/fluid_structure_interaction/precice/driver.h
```diff
@@ -83,6 +83,26 @@ class Driver
...
 
+  // Nomenclature
+  // time window: denotes time interval after which FSi coupling takes place
+  // subcycling: a single field solver performs several time steps until the next FSI coupling time
+  // window is reached
+  // time step size: refers to single field solvers. In case of subcycling, one
+  // time window consists of several time step sizes.
+
+  // The time-window size is determined through the preCICE configuration by the user:
+  // The <coupling-scheme.. tag has options for
```
nfehn, line 94: Is the syntax `<coupling-scheme..` correct?

include/exadg/fluid_structure_interaction/precice/driver.h
```diff
@@ -83,6 +83,26 @@ class Driver
...
+  // <time-window-size method="fixed" ... /> which refers to a constant time window size
+  // <time-window-size method="first-participant" ... /> which refers to an adaptive time window
+  // size prescribed by the first (which is usually the Fluid) participant. The latter is only
+  // applicable in serial coupling schemes.
```
nfehn, line 98: I think ExaDG developers/users need an explanation of serial/parallel coupling, because one otherwise thinks of one vs. multiple MPI ranks.

serial coupling = the single-field solvers are called sequentially/consecutively with data transfer from one field to another after in single-field solver.
parallel coupling = the single-field solvers are called simultaneously with data transfer in both directions afterwards.

I would also like taking about "sequential" and "simultaneous" coupling schemes (avoiding the serial/parallel notation).

include/exadg/fluid_structure_interaction/precice/driver_fluid.h
nfehn, line 222: If I understand the implementation correctly, then this comment should start with:

"The current implementation does not allow to synchronize the time step size between the fluid field and precice. Instead, the fluid field dictates the time step size. The reasons behind are that ..."

or similar. My point is to start the comment with the capabilities of the current implementation (and subsequent (optional) explanations). In the current form it is hard to understand the differences between (i) what the current implementation can do and (ii) what one would like to have/support in the future.

include/exadg/fluid_structure_interaction/precice/precice_adapter.h
```diff
@@ -115,9 +117,13 @@ class Adapter
                    std::vector<std::string> const &                               read_data_name);
 
   /**
-   * @brief      Advances preCICE after every timestep
+   * @brief Advances preCICE after every timestep
+   *
+   * @param computed_timestep_length the time-step size computed by the solver
+   *
+   * @return double the allowed time-step size until the next coupling time-window is reached
```
nfehn, line 124: related to another comment above: What is the difference between "the allowed time-step size until ..." or simply "the time until ..."?

include/exadg/fluid_structure_interaction/precice/driver_fluid.h
DavidSCN: I'm referring to this line 
https://github.com/exadg/exadg/blob/6e647f226111b184e441eee8d7b8326e48bc757f/include/exadg/time_integration/time_int_bdf_base.cpp#L242  and this nice comment https://github.com/exadg/exadg/blob/6e647f226111b184e441eee8d7b8326e48bc757f/include/exadg/time_integration/time_int_bdf_base.h#L90-L94

Does it make it more clear?

include/exadg/fluid_structure_interaction/precice/driver_fluid.h
DavidSCN: > why "potentially"? Doesn't "subcycling" mean that one time-window consists of several time steps?

Sure. With time-step adaptivity you don't know in advance weather the time-step size is equal to the time-window size, though. I deleted the comment in brackets now.

include/exadg/fluid_structure_interaction/precice/driver.h
DavidSCN: >  Is the variable allowed_time_step_size adjusted from one time step to the next within one time window?

Exactly

I changed it.

include/exadg/fluid_structure_interaction/precice/driver.h
DavidSCN: I understand your motivation for changing the terminology here. However, for someone dealing with preCICE (or configuring preCICE) the terminology serial/parallel coupling scheme will be inevitably. I copied over your explanation comments.

include/exadg/fluid_structure_interaction/precice/driver_fluid.h
DavidSCN: I tried to improve on that. My goal is to still document current thoughts so that I understand limitations when I look into this later on.

include/exadg/fluid_structure_interaction/precice/driver_fluid.h
```diff
@@ -218,7 +216,32 @@ class DriverFluid : public Driver<dim, Number>
...
+        // with subcycling, either the boundary condition needs to be adapted or we need to wait for
+        // a newer preCICE version (currently v2.3.0) which can handle this. Hence, we do not adjust
+        // the time-step-size of the fluid solver, but rather Assert that it is still valid. Note
+        // that we would otherwise run into a preCICE assert reporting the same issue.
+        // fluid->time_integrator->set_current_time_step_size(
+        //   std::min(this->allowed_time_step_size, fluid->time_integrator->get_time_step_size()));
+        Assert(
+          fluid->time_integrator->get_time_step_size() <
+            this->allowed_time_step_size + std::numeric_limits<double>::min(),
```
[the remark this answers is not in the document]
DavidSCN: That's equivalent, I put in your suggestion.

include/exadg/fluid_structure_interaction/precice/precice_adapter.h
DavidSCN: They are the same, I changed the terminology.

include/exadg/fluid_structure_interaction/precice/driver_fluid.h
DavidSCN: The `advance` call return the time remaining until the next time-window. The comment was rephrased anyway, let's see if it is more clear now.

include/exadg/fluid_structure_interaction/precice/driver.h
```diff
@@ -83,6 +83,30 @@ class Driver
...
   ExaDG::preCICE::ConfigurationParameters                        precice_parameters;
 
+  // Nomenclature
+  // time window: denotes time interval after which FSi coupling takes place
+  // subcycling: a single field solver performs several time steps until the next FSI coupling time
+  // window is reached
+  // time step size: refers to single field solvers. In case of subcycling, one
+  // time window consists of several time step sizes.
+  // serial coupling scheme: the single-field solvers are called sequentially/consecutively with
+  // data transfer from one field to another after in single-field solver.
```
nfehn, line 93: I think this should be "after each ..."

include/exadg/fluid_structure_interaction/precice/driver_fluid.h
```diff
@@ -218,7 +217,36 @@ class DriverFluid : public Driver<dim, Number>
...
+        // conditions is not handled correctly when using subcycling, as we would apply the 'whole
+        // displacement' within the first time-step. In order to cope with subcycling, either the
+        // boundary condition needs to be adapted or we need to wait for a newer preCICE version
+        // (currently v2.3.0) which can handle this. Hence, we do not adjust the time-step-size of
+        // the fluid solver, but rather Assert that it is still valid. Note that we would otherwise
+        // run into a preCICE assert reporting the same issue.
+        // fluid->time_integrator->set_current_time_step_size(
+        //   std::min(this->time_until_next_coupling_window,
+        //   fluid->time_integrator->get_time_step_size()));
+        // Have a look in the base class for more explanations.
```
nfehn, line 239: This comment on base class is not clear to me? Is it needed? I would remove the code above.

include/exadg/fluid_structure_interaction/precice/driver_solid.h
```diff
@@ -169,18 +176,23 @@ class DriverSolid : public Driver<dim, Number>
...
       this->timer_tree.insert({"FSI", "preCICE"}, precice_timer.wall_time());
 
-      // Needs to be called before the swaps in post_solve
+      // Needs to be called before the swaps in post_solve.
       this->precice->reload_old_state_if_required([&]() {});
 
       if(is_new_time_window)
         structure->time_integrator->advance_one_timestep_post_solve();
+
+      // We need to adjust the time-step size in case we reach a coupling time-wimdow.
```
nfehn, line 191: window

include/exadg/fluid_structure_interaction/precice/driver_solid.h
```diff
@@ -169,18 +176,23 @@ class DriverSolid : public Driver<dim, Number>
...
+      // We don't take the time-step size of the solver here into account, but rather apply the
+      // available time-step size reported by preCICE, as the GenAlpha time integrator doesn't
```
nfehn, line 193: I would remove the second part "as the GenAlpha ..." (because we should not have references to a particular time integrator GenAlpha here; GenAlpha is not the reason. e.g. this functionality may be added to GenAlpha later; the point here is that the fluid dictates the time step size).

include/exadg/fluid_structure_interaction/precice/driver_solid.h
```diff
@@ -148,6 +149,12 @@ class DriverSolid : public Driver<dim, Number>
   solve() const final
   {
     bool is_new_time_window = true;
+
+    // preCICE prescribes the time-step size
+    structure->time_integrator->set_current_time_step_size(
+      std::min(this->time_until_next_coupling_window,
+               structure->time_integrator->get_time_step_size()));
```
nfehn, line 156: I do not see how the comment fits to the three lines of code below?

include/exadg/fluid_structure_interaction/precice/driver.h
```diff
@@ -83,6 +83,30 @@ class Driver
   std::shared_ptr<ExaDG::preCICE::Adapter<dim, dim, VectorType>> precice;
   ExaDG::preCICE::ConfigurationParameters                        precice_parameters;
 
+  // Nomenclature
+  // time window: denotes time interval after which FSi coupling takes place
```
nfehn, line 87: FSI (capital I)

nfehn: If correct, please add "Each time window size must be the same for the fluid field and the solid field."

include/exadg/fluid_structure_interaction/precice/driver.h
```diff
@@ -83,6 +83,30 @@ class Driver
...
+  // subcycling: a single field solver performs several time steps until the next FSI coupling time
+  // window is reached
+  // time step size: refers to single field solvers. In case of subcycling, one
+  // time window consists of several time step sizes.
```
nfehn, line 91: If correct, please add the relative clause "... of several time steps, which may have different size for the fluid field and the solid field."

include/exadg/fluid_structure_interaction/precice/driver.h
```diff
@@ -83,6 +83,30 @@ class Driver
...
+
+  // The time-window size is determined through the preCICE configuration by the user:
+  // The <coupling-scheme.. /> tag has options for
+  // <time-window-size method="fixed" ... /> which refers to a constant time window size
+  // <time-window-size method="first-participant" ... /> which refers to an adaptive time window
+  // size prescribed by the first (which is usually the Fluid) participant. The latter is only
+  // applicable in serial coupling schemes.
+  // Both solvers can exploit subcycling, but they need to synchronize at the end of a certain
+  // time-window size. Therefore, the allowed time-step size until we reach the next synchronization
+  // point needs to be determined.
```
nfehn, line 105: Is sub-cycling only relevant/possible in case of an adaptive time window size, or also for a constant time window size throughout the simulation? I think this information is still missing. Please add/describe.

From a user perspective, how to I decide to use sub-cycling or no sub-cycling? I think this also needs to be explained.

nfehn:
Sorry for the late review (I was busy with other things) ... I think we can get this merged in the next days. The descriptions are clearer now. Some more questions occurred when going through the code today, and we should try to clarify them before merging.

@DavidSCN could you resolve those conversations you have already resolved in the code (to keep the overview of what is missing)? I don't remember the details and where something still needs to be done or not. Thanks.

include/exadg/fluid_structure_interaction/precice/driver_solid.h
DavidSCN: Actually no. The window refers to the step size of the coupled system, time-step size to the size of the solver.

nfehn: I mean the typo here. What do you mean?

DavidSCN: Ah ok sorry

include/exadg/fluid_structure_interaction/precice/driver.h
nfehn: I thought that the single-field solvers must hit the time window exactly? I will write another comment below ...

nfehn:
I think the difficulty for me is to see from the comments in the code which functionality is supported in the ExaDG implementation and what is general preCICE functionality. As I probably mentioned in previous comments of this PR, I am of the opinion that it does not help to describe the most general preCICE case in this PR, since the implementation currently happens in ExaDG and has some limitations. Maybe we need another opinion on this topic?

A topic I often thought about in the last weeks is also whether we should think/discuss more about where to implement which code. E.g. Does it belong to ExaDG, does it belong to preCICE directly? Maybe the discussions here in this PR are a result of the circumstance that the "separation of concerns" might be unclear right now? In the previous PR, I think @peterrum mentioned that he sees parts of the code in the preCICE repo in the long run. If you think it should be in ExaDG, I am convinced that the code/comments should be written from an ExaDG perspective (with preCICE as a TPL).

DavidSCN:
> Maybe we need another opinion on this topic?

Might be, yes. I think the main problem is that we face different limitations from ExaDG and preCICE which make the whole topic a bit hard to understand. In the end, there are mostly two useful configurations which are both supported in the current implementation: namely a staggered coupling scheme with time adaptivity and a simultaneous coupling scheme without time adaptivity.

DavidSCN:
I had some discussion today about this and we concluded that the actual issue here is that the current implementation doesn't allow subcycling on the fluid side. If the implementation would allow subcycling, all combinations can be configured without further restrictions. I will try to investigate how we could support this on the fluid side and come back to the time adaptivity later on.

PR #216, opened 2022-04-05
switch from active_cell_iterators() to cell_iterators() where reasonable

### LINKED ISSUE #215: use cell_iterators() or active_cell_iterators()?
We currently do not consistently use `cell_iterators()` or `active_cell_iterators()`. This might be irritating for users who do not have detailed background knowledge about dealii, users who just want to "use" ExaDG and, for example, implement their own examples taking existing ones as a template.

The typical use case is (i) to generate a grid (potentially composed of multiple elements, e.g. subdivided hyper cube/rectangle), but without any global refinements. (ii) Next, we loop over the cells (e.g. apply boundary IDs). (iii) Finally, global refinements are performed.

The question is which type of cell iterators should be used? I would argue that it does not make a difference for the use case above. However, we should take into account that `Grid::create_triangulation()` applies local and global refinements in the current implementation. @kronbichler @peterrum What do you consider more general? Is there a disadvantage if we use `cell_iterators()` always for the above use case? Assume unexperienced users switch steps (ii) and (iii) in their code. Using `active_cell_iterators()` might then lead to wrong results / problems, because coarse levels in multigrid algorithms also want to know the correct type of boundary conditions. Is this correct?

nfehn:
closes #215

### DIFF
--- a/include/exadg/postprocessor/kinetic_energy_spectrum.cpp
+++ b/include/exadg/postprocessor/kinetic_energy_spectrum.cpp
@@ -94,9 +94,9 @@ public:
     s.init(dim, n_cells_1D, points_src, points_dst);
 
     std::vector<dealii::types::global_dof_index> local_cells;
-    for(auto const & cell : tria.cell_iterators())
+    for(auto const & cell : tria.active_cell_iterators())
     {
-      if(cell->is_active() && cell->is_locally_owned())
+      if(cell->is_locally_owned())
       {
         auto c = cell->center();
         for(dealii::types::global_dof_index i = 0; i < dim; i++)
--- a/include/exadg/postprocessor/kinetic_energy_spectrum.cpp
+++ b/include/exadg/postprocessor/kinetic_energy_spectrum.cpp
@@ -94,7 +94,8 @@ public:
     s.init(dim, n_cells_1D, points_src, points_dst);
 
     std::vector<dealii::types::global_dof_index> local_cells;
-    for(auto const & cell : tria.active_cell_iterators())
+    for(auto const & cell : tria.cell_iterators())
+    {
       if(cell->is_active() && cell->is_locally_owned())
       {
         auto c = cell->center();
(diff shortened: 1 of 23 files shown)


include/exadg/postprocessor/kinetic_energy_spectrum.cpp
```diff
@@ -94,7 +94,8 @@ class DealSpectrumWrapper
     s.init(dim, n_cells_1D, points_src, points_dst);
 
     std::vector<dealii::types::global_dof_index> local_cells;
-    for(auto const & cell : tria.active_cell_iterators())
+    for(auto const & cell : tria.cell_iterators())
```
nfehn, line 97: @peterrum please check this change. I switched to `cell_iterators()` since the following line writes `if(cell->is_active() ...)`

kronbichler: This is correct, but the previous code could have skipped `cell->is_active()` instead. But it then boils down to what we prefer: Have the explicit check `is_active` in any case, or use a more compact notation with a `for` loop over `active_cell_iterators()`. As for the mental load, I do not see a difference since one needs to be familiar with a similar vocabulary. But since things will horribly go wrong without `active_cell_iterators` in the algorithm, I would tend to keep the version proposed here.

kronbichler (approved):

include/exadg/postprocessor/kinetic_energy_spectrum.cpp
peterrum: Personally, I would keep `active_cell_iterators()` and remove `cell->is_active()`.

nfehn: I don't have a clear preference here. Since this code was contributed by @peterrum, I think I will follow your suggestion.

nfehn: see https://github.com/exadg/exadg/pull/216/commits/e6b491128b7b60fc4f1a34db7a4d5e48977db518

peterrum (approved):
Thanks.

PR #207, opened 2022-03-17
Some refactoring of CMakeLists.txt

nfehn:
related to comment https://github.com/exadg/exadg/pull/148#discussion_r828259812

PR #217, opened 2022-04-05
use EXADG_WITH_LIKWID instead of LIKWID_PERFMON in ExaDG code

nfehn:
related to comment https://github.com/exadg/exadg/pull/207#discussion_r842460458.

I think we should not use `LIKWID_PERFMON` in the ExaDG code, but only in the top-level CMakeLists.txt file (in the section dealing with likwid).

@kronbichler what do you think?


peterrum (approved):

kronbichler (approved):
Yes, this makes sense. Thank you.

PR #218, opened 2022-04-05
Make InterfaceCoupling more robust

### LINKED ISSUE #205: FSI solver does not throw error when points for interface coupling are not found
In case the solid and fluid domains do not touch each other at a fluid-structure-interface, the code does not throw any error, even in debug mode! In my opinion, it should also not happen in release mode ... In any case, such a behavior is not tolerable. See also deal.II PR https://github.com/dealii/dealii/issues/13501.

I assume this behavior comes from PR #126 (using `RemotePointEvaluation` and `VectorTools::point_values()` from deal.II instead of own implementations in ExaDG), because I remember we checked that all the points at the interface are really found when we implemented the `InterfaceCoupling` functionality for FSI (e.g. I remember that at that time, there were also some problems in the deal.II library with find_active_cells_around_point() functions, so we made efforts to ensure that all the points are really found). This original code was a predecessor version of today's `RemotePointEvaluation` in deal.II.

To give an example: This is the pressure wave example:
fsi_pressure_wave_correct

Now, replacing the parameter `R_INNER` by `R_INNER*0.8` in the function `create_grid()` for the fluid https://github.com/exadg/exadg/blob/e858f3ccc19555fe4f00c461d8b134e0202850e5/applications/fluid_structure_interaction/pressure_wave/application.h#L269

results in:
fsi_pressure_wave_incorrect

i.e., the points for the interface coupling are simply not found, but nothing happens (even in debug mode!) and the FSI code does not get informed about this problem.

nfehn:
closes #205

uses functionality introduced by @peterrum in deal.II: see https://github.com/dealii/dealii/pull/13547

### DIFF
--- a/applications/fluid_structure_interaction/pressure_wave/application.h
+++ b/applications/fluid_structure_interaction/pressure_wave/application.h
@@ -268,8 +268,9 @@ private:
   void
   create_grid() final
   {
+    double R_INNER_FLUID = R_INNER*0.8;
     dealii::Triangulation<2> tria_2d;
-    dealii::GridGenerator::hyper_ball(tria_2d, dealii::Point<2>(), R_INNER);
+    dealii::GridGenerator::hyper_ball(tria_2d, dealii::Point<2>(), R_INNER_FLUID);
     dealii::GridGenerator::extrude_triangulation(tria_2d,
                                                  N_CELLS_AXIAL / 4 + 1,
                                                  L,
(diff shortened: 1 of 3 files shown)


nfehn:
we probably need `DEAL_II_VERSION_GTE(9, 4, 0)` ...

peterrum (approved):

kronbichler (approved):

PR #219, opened 2022-04-06
Restructure FSI application to build preCICE and native ExaDG solver

DavidSCN:
As pointed out in https://github.com/exadg/exadg/pull/213#discussion_r841490023 the ExaDG FSI code was not compiled in the current setup. Since the CMake macro `EXADG_PICKUP_EXE` doesn't allow multiple targets, I moved now the precice stuff into a subdirectory in order to comply with the `include` directory. The structure looks as follows

```bash
$ tree -L 2
.
├── application.h
├── CMakeLists.txt
├── precice
│   ├── CMakeLists.txt
│   ├── fluid-exadg
│   ├── precice-config.xml
│   ├── solid-exadg
│   └── solver_precice.cpp
└── solver.cpp
```

Any comments?


nfehn:
My intention in comment https://github.com/exadg/exadg/pull/213#discussion_r841490023 was different. My goal is that the precice driver is compiled when compiling ExaDG even in case preCICE is not installed on a local machine (otherwise ExaDG would actually not be the right repository for this code in my opinion). When making changes to ExaDG (related to ExaDG-native implementations used in the preCICE-based FSI implementation), I want to detect compilation errors on my local machine, not indirectly via github CI. (For clarity, the alternative would be realizing this code in another repository and linking against a fixed (not yet existing) release version of ExaDG.)

In detail, this means that I would prefer a variant where the ifdef's for preCICE appear in the source code and not in the CMakeLists.txt file. Take https://github.com/exadg/exadg/blob/master/include/exadg/structure/preconditioners/amg_preconditioner.h as an example how we typically deal with this in ExaDG. The question whether the preCICE executable for the perpendicular flag example should exist in case preCICE is not installed is - in my opinion - an independent and subsequent question.

DavidSCN:
Compiling the driver into exadg is principle possible, although it might result in slight `ifdef` hell. 

The proposed changes (discussion?) in this PR remain valid, though. I have an application for the `cylinder_with_flag` and due to the intense refactoring, I can reuse the `application.h` and only need to change the solver include

```diff
-<exadg/fluid_structure_interaction/solver.h>
+<exadg/fluid_structure_interaction/precice/solver.h>
```

in order to build the preCICE variant. Should we now create a subdirectory and build both FSI codes or introduce a switch?

nfehn:
Yes, finding a good way how to optimally deal with the ifdef's is not trivial, see e.g. https://github.com/exadg/exadg/pull/121. My point is we have to give this a try and then maybe take the opinions from different developers to find the best solution for this case. This should probably be dealt with in a separate PR to be considered first. Do you want to do this or do you want me to do this?

Including both solvers is probably not possible because you then include two main-functions, don't you? I would consider the changes from the present PR subsequently. What would be missing with respect to the changes from the first commit of this PR is the .json input file for the ExaDG-native FSI solver. A subdirectory for preCICE sound reasonable in my opinion, given that this is also the structure we have in other places implementing the preCICE-variant. Then, just `solver.h` instead of `solver_precice.h` should be enough.

PR #223, opened 2022-04-09
Drop support for deal.II 9.3

peterrum:
With https://github.com/dealii/dealii/pull/13603, we can now drop support of deal.II 9.3.


nfehn (approved):

nfehn:
I agree to no longer support this version of dealii in ExaDG, given that we realized that support for dealii-9.3 is quite some maintenance effort in ExaDG.

kronbichler (approved):

PR #233, opened 2022-06-04
H(div)-conforming incompressible Navier-Stokes

NiklasWik:
Adds support for H(div)-conforming finite element (Raviart-Thomas) in the incompressible N-S setting. 

~~This PR needs to wait for https://github.com/dealii/dealii/pull/13907 to get merged.~~ ✔️


include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
```diff
@@ -249,20 +249,87 @@ template<int dim, typename Number>
...
+        face.cell[1]->face(face.face_idx[1])->boundary_id(),
+        face.face_idx[0] / 2,
+        constraint_u);
+
+
+    // Set boundaries constraints
+    if(boundary_descriptor->velocity->symmetry_bc.empty() == false)
+    {
+      for(auto bc : boundary_descriptor->velocity->symmetry_bc)
+        dealii::VectorTools::project_boundary_values_div_conforming(
```
NiklasWik, line 310: Alternatively, these rows could be commented out for now.

peterrum (commented):
Looks very nice and clean 👍

NiklasWik:
Is this not using the master branch of dealii?

peterrum:
It does. But it seems that there is an issue with the docker deployment pipeline: https://github.com/dealii/dealii/actions/workflows/docker.yml

nfehn:
I was on vacation this week. So I plan to have a detailed look at this PR next week.

NiklasWik:
I pushed another commit which makes so that Hdiv also uses `VectorTools::interpolate` since [#13976](https://github.com/dealii/dealii/pull/13976) fixes it.

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
```diff
@@ -249,20 +249,87 @@ template<int dim, typename Number>
...
+    pcout << std::endl
+          << "Discontinuous Galerkin finite element discretization:" << std::endl
+          << std::endl
+          << std::flush;
+  }
+  else if(param.spatial_discretization == SpatialDiscretization::HDIV)
+  {
+    // In-parameter for RT is the degree in tangent direction, i.e the degree in normal direction
+    // will be degree+1. The FE also saves fe_u.degree = degree+1.
+    // degree - 1 is used so that the same Gauss quadrature can be used for exact integration.
```
nfehn, line 265: I would not say this is the reason for degree-1. We choose "degree-1" in tangential direction and "degree" in normal direction so that the method is inf-sup stable when choosing degree-1 for the pressure.

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
```diff
@@ -249,20 +249,87 @@ template<int dim, typename Number>
...
+    fe_u = std::make_shared<dealii::FE_RaviartThomasNodal<dim>>(param.degree_u - 1);
+
+    pcout << std::endl
+          << "Hdiv-conforming Raviart-Thomas finite element discretization:" << std::endl
+          << "(Note that the given degree for the velocity is the degree in normal direction."
+          << std::endl
+          << " The degree in tangent direction is degree - 1. Note also that the number of dofs"
```
nfehn, line 272: This is hard to understand: With the wording "Note that the given degree", do you refer to the input parameter ``param.degree_u`` or to the constructor argument of the finite element? In my opinion, this does not become clear.

To be more precise, print explicitly the degree in normal direction and the degree in tangential direction to the screen:

```C++
std::cout << "Degree in normal direction: " << param.degree_u << std::endl;
std::cout << "Degree in tangential direction: " << param.degree_u - 1 << std::endl;
```

In my opinion, this is the relevant information for the user. A user does not need to know how the constructor is defined in deal.II. The user wants to know the degrees in normal/tangential direction.

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
```diff
@@ -249,20 +249,87 @@ template<int dim, typename Number>
...
+
+    pcout << std::endl
+          << "Discontinuous Galerkin finite element discretization:" << std::endl
+          << std::endl
+          << std::flush;
+  }
+  else if(param.spatial_discretization == SpatialDiscretization::HDIV)
+  {
+    // In-parameter for RT is the degree in tangent direction, i.e the degree in normal direction
+    // will be degree+1. The FE also saves fe_u.degree = degree+1.
```
nfehn, line 264: This comment should only explain how the constructor is defined in deal.II, i.e. "The constructor of ... takes the degree in tangential direction as an argument. The degree in normal direction is one degree higher."

Whether ``param.degree_u`` is the degree in normal direction or in tangential direction is in my opinion a convention in ExaDG. We could also define it differently, but then we would need to adjust the number of quadrature points, in order to integrate polynomials of degree ``param.degree_u +1`` in normal direction exactly. However, this information should be given at the relevant places in my opinion, i.e. in the class ``Parameters`` and where we define the quadrature rule.

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
```diff
@@ -249,20 +249,87 @@ template<int dim, typename Number>
...
+
   // enumerate degrees of freedom
   dof_handler_u.distribute_dofs(*fe_u);
   dof_handler_p.distribute_dofs(fe_p);
   dof_handler_u_scalar.distribute_dofs(fe_u_scalar);
 
+  if(param.spatial_discretization == SpatialDiscretization::HDIV)
+  {
+    // We need to make sure the normal dofs are shared between cells on the periodic boundaries,
+    // since these are contiguous for HDIV.
```
nfehn, line 291: I think you mean "continuous"? If so, please adjust.

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
```diff
@@ -1030,7 +1135,9 @@ SpatialOperatorBase<dim, Number>::compute_vorticity(VectorType & dst, VectorType
 {
   vorticity_calculator.compute_vorticity(dst, src);
 
-  inverse_mass_velocity.apply(dst, dst);
+  // If HDIV is used, an iterative solver is used,
+  // but here we ignore the number of iterations
```
nfehn, line 1139: I would not comment on HDIV here. Then, we would have to do this whenever we call the function.

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
```diff
@@ -249,20 +249,87 @@ template<int dim, typename Number>
...
+      // dealii::VectorTools::project_boundary_values
+      // dealii::DoFTools::make_zero_boundary_constraints
+    }
+  }
+
   unsigned int const ndofs_per_cell_velocity =
-    dealii::Utilities::pow(param.degree_u + 1, dim) * dim;
+    (get_spatial_discretization() == SpatialDiscretization::L2) ?
+      dealii::Utilities::pow(param.degree_u + 1, dim) * dim :
+      dealii::Utilities::pow(param.degree_u, dim - 1) * (param.degree_u + 1) * dim;
```
nfehn, line 329: I am not sure whether the ``... ? ... : ...`` syntax is suitable here. I think we could be more explicit using `if(L2)`, `else if (Hdiv)`, `else {AssertThrow(false, ExcNotImplemented);}`.

For the Hdiv-case, an additional comment like "continuity constraints in normal direction are not taken into account regarding the number of DoFs per element.".

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.h
```diff
@@ -241,12 +242,18 @@ class SpatialOperatorBase : public dealii::Subscriptor
...
+  dealii::AffineConstraints<Number> const &
+  get_constraint_u() const;
+
   dealii::types::global_dof_index
   get_number_of_dofs() const;
 
   double
   get_viscosity() const;
 
+  SpatialDiscretization const &
```
nfehn, line 254: Could you explain why this member function is needed?

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.cpp
```diff
@@ -610,6 +610,13 @@ TimeIntBDF<dim, Number>::postprocessing() const
     operator_base->move_grid_and_update_dependent_data_structures(this->get_time());
   }
 
+  // We need to distribute the dofs for HDIV before computing the error since
+  // dealii::VectorTools::integrate_difference() does not take the periodic boundary
+  // constraints into account like MatrixFree does, hence reading the wrong value.
+  // distribute() updates the otherwise unused values.
```
nfehn, line 616: Nice comment! 

I do not understand "otherwise unused" in the last line. -> "distribute() updates those values. Note that MatrixFree does this automatically."?

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_dual_splitting.cpp
```diff
@@ -644,7 +653,10 @@ TimeIntBDFDualSplitting<dim, Number>::projection_step()
     if(this->print_solver_info() and not(this->is_test))
     {
       this->pcout << std::endl << "Explicit projection step:";
-      print_wall_time(this->pcout, timer.wall_time());
+      if(this->param.spatial_discretization == SpatialDiscretization::HDIV)
+        print_wall_time_with_mass_iter(this->pcout, n_iter_mass, timer.wall_time());
```
nfehn, line 657: Why a new function? Can't one use one of the existing functions? See other steps of e.g. the dual splitting scheme where we print both wall times and iteration counts. Such a function should be re-usable here I think.

include/exadg/incompressible_navier_stokes/user_interface/enum_types.h
```diff
@@ -638,6 +652,43 @@ std::string
...
+ *  the system
+ *
+ */
+enum class SolverMassHdiv
+{
+  CG
+};
+
+std::string
+enum_to_string(SolverMassHdiv const enum_type);
```
nfehn, line 675: I think we do not need this as long as there is not decision to be made. Let's try to minimize the parameters seen by a user.

include/exadg/incompressible_navier_stokes/user_interface/parameters.cpp
```diff
@@ -403,9 +415,33 @@ Parameters::check(dealii::ConditionalOStream const & pcout) const
     }
   }
 
+  if(spatial_discretization == SpatialDiscretization::HDIV)
+  {
+    AssertThrow(
+      use_divergence_penalty == false || use_continuity_penalty == false,
+      dealii::ExcMessage(
+        "Penalty terms can not be used in case of Hdiv function spaces. Think about using L2 instead, or deactivate the penatly terms."));
```
nfehn, line 423: There is a typo in "penatly". However, I think one does not need the second part of the explanation. I think one will not switch to L2 just to be able to use the penalty terms.

What is the reason that they cannot be used? Why does the code not work with penalty terms?

A maybe better message could be: "The use of penalty terms does not make sense for SpatialDiscretization::HDIV. You have to deactivate them."

include/exadg/incompressible_navier_stokes/user_interface/parameters.cpp
```diff
@@ -403,9 +415,33 @@ Parameters::check(dealii::ConditionalOStream const & pcout) const
...
+    if(temporal_discretization != TemporalDiscretization::BDFCoupledSolution)
+      AssertThrow(
+        preconditioner_velocity_block != MomentumPreconditioner::InverseMassMatrix &&
+          preconditioner_velocity_block != MomentumPreconditioner::BlockJacobi &&
+          preconditioner_velocity_block != MomentumPreconditioner::Multigrid,
+        dealii::ExcMessage(
+          "Multigrid, BlockJacobi, and InverseMassMatrix as preconditioner are not availiable for HDIV. Use either PointJacobi or None instead, or change to L2 discretization. "));
```
nfehn, line 430: Here, I would also not refer to L2. I think this type of commentes is hard to maintain and does not really provide additional information. Typically, one does not switch to a different discretization because a preconditioner is not available for Hdiv.

include/exadg/incompressible_navier_stokes/user_interface/parameters.cpp
```diff
@@ -433,6 +469,11 @@ Parameters::check(dealii::ConditionalOStream const & pcout) const
   // PRESSURE-CORRECTION SCHEME
   if(temporal_discretization == TemporalDiscretization::BDFPressureCorrection)
   {
+    AssertThrow(
+      spatial_discretization == SpatialDiscretization::L2,
+      dealii::ExcMessage(
+        "Pressure correction only works for L2. If you want to use Hdiv conforming elements you have to use the coupled solution"));
```
nfehn, line 475: I would do this differently. I would make sure by the Assert that the spatial discretization is not Hdiv. If it is Hdiv, the ExcMessage should be of type "not implemented", because this reflects better the current situation, right? Hdiv for pressure-correction has just not been implemented? Or do you think there are other special issues in case of Hdiv with pressure-correction?

include/exadg/incompressible_navier_stokes/user_interface/parameters.cpp
```diff
@@ -500,6 +541,10 @@ Parameters::check(dealii::ConditionalOStream const & pcout) const
       use_cell_based_face_loops == true,
       dealii::ExcMessage(
         "Cell based face loops have to be used for matrix-free implementation of block diagonal preconditioner."));
+
+    AssertThrow(spatial_discretization == SpatialDiscretization::L2,
```
nfehn, line 545: Here, I would also assert that the type of discretization is not Hdiv, with the exception message "not implemented.".

include/exadg/incompressible_navier_stokes/user_interface/parameters.h
```diff
@@ -654,6 +657,25 @@ class Parameters
...
+  /**************************************************************************************/
+  /*                                                                                    */
+  /*                            SOLVE MASS SYSTEM HDIV                                  */
+  /*                                                                                    */
+  /**************************************************************************************/
+  // We cannot apply the inverse mass operator while using HDIV. Instead we have to solve
+  // the system
+
+  // description: see enum declaration
+  SolverMassHdiv solver_mass_hdiv;
```
nfehn, line 672: I think I would remove ``..._hdiv`` from the names.

include/exadg/time_integration/time_step_calculation.h
```diff
@@ -253,6 +253,8 @@ calculate_time_step_cfl_local(
   CFLConditionType const                                         cfl_condition_type,
   MPI_Comm const &                                               mpi_comm)
 {
+  // Need to update ghost values for hdiv ...
```
nfehn, line 256: Can you explain this in more detail here.

include/exadg/utilities/print_solver_results.h
```diff
@@ -191,6 +191,20 @@ print_wall_time(dealii::ConditionalOStream const & pcout, double const wall_time
   // clang-format on
 }
 
+inline void
+print_wall_time_with_mass_iter(dealii::ConditionalOStream const & pcout,
```
nfehn, line 195: according to another comment, think about whether this function can be removed and whether an existing function can be used.

nfehn:
As a general comment, it would be good if you include the periodic 2d vortex example in the present PR, so that we have a test case for which we can verify the rsults obtained with HDIV.

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
nfehn: I think the comment that `degree_u` is the degree in normal direction in case of Hdiv + RT should be written here
https://github.com/exadg/exadg/blob/8382b8bacb87b907111ffe60e9cc48d7e86c2a4c/include/exadg/incompressible_navier_stokes/user_interface/parameters.h#L283

The output to cout could/should also be done in `Parameters::print()` (e.g. introduce an if-else for the output depending on L2 vs. HDIV. Then,  a comment in the spatial discretization operator, namely that the constructor argument of dealii::FE takes the tangential degree, should be enough.

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.h
NiklasWik: I need it here
https://github.com/exadg/exadg/blob/b82815944e73824ce05b1623c36fc89785958891/include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.cpp#L617-L618
I also started using it in `spatial_operator_base_cpp` after I introduced it, but I'll make sure it's consistent and not a mixture of `param.spatial_discretization` and `get_spatial_discretization()`.

nfehn: Yes, you could directly ask param (we generally rely on that in the time integration classes in order to avoid writing many getter functions in the spatial discretization classes).

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_dual_splitting.cpp
NiklasWik: I can't really remember the reason.. :)

include/exadg/incompressible_navier_stokes/user_interface/enum_types.h
NiklasWik: I wasn't even checking it when initiating the solver :D

include/exadg/incompressible_navier_stokes/user_interface/parameters.cpp
NiklasWik: I'm pretty sure it was you who wrote this message :D 

The penalty terms are zero in the case of hdiv!

include/exadg/incompressible_navier_stokes/user_interface/parameters.cpp
NiklasWik: Is it maybe better to check for the ones that work?

include/exadg/incompressible_navier_stokes/user_interface/parameters.cpp
NiklasWik: I just haven't tried it. It might have been the inverse mass that wasn't working, so it might work now. I'll still add the assert though.

nfehn:
In the very beginning, you mention dealii PR that needs to be merged. Could you update that information?

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
```diff
@@ -126,7 +125,8 @@ SpatialOperatorBase<dim, Number>::fill_matrix_free_data(
     matrix_free_data.append_mapping_flags(
       Operators::ViscousKernel<dim, Number>::get_mapping_flags(true, true));
 
-  if(param.right_hand_side)
+  // RHSOperator is used to initialize the velocity field for HDIV!
+  if(param.right_hand_side || param.spatial_discretization == SpatialDiscretization::HDIV)
```
nfehn, line 129: why is this still necessary? I thought that the initialization of the velocity field is now done via `dealii::VectorTools`?

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
```diff
@@ -249,25 +249,101 @@ template<int dim, typename Number>
...
+      for(auto bc : boundary_descriptor->velocity->symmetry_bc)
+        dealii::VectorTools::project_boundary_values_div_conforming(
+          dof_handler_u, 0, *(bc.second), bc.first, constraint_u, *get_mapping());
+    }
+    if(boundary_descriptor->velocity->dirichlet_bc.empty() == false)
+    {
+      AssertThrow(
+        false,
+        dealii::ExcMessage(
+          "Dirichlet BC are not properly implemented for HDIV. The BC needs to be strongly applied in order to obtain an exactly divergence-free solution."));
```
nfehn, line 299: Dirichlet BC are -> Dirichlet BCs are

Is it the whole boundary condition that is applied strongly, or just the normal component that is applied strongly?

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
```diff
@@ -249,25 +249,101 @@ template<int dim, typename Number>
...
+    for(auto const & face : periodic_face_pair)
+      dealii::DoFTools::make_periodicity_constraints(
+        dof_handler_u,
+        face.cell[0]->face(face.face_idx[0])->boundary_id(),
+        face.cell[1]->face(face.face_idx[1])->boundary_id(),
+        face.face_idx[0] / 2,
+        constraint_u);
+
+
+    // Set boundaries constraints
```
nfehn, line 287: boundaries -> boundary

I think I would remove this comment, and instead introduce a comment above in line 268, something like "Strong imposition of boundary conditions in normal direction in case of HDIV".

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
```diff
@@ -249,25 +249,101 @@ template<int dim, typename Number>
...
+    if(boundary_descriptor->velocity->symmetry_bc.empty() == false)
+    {
+      for(auto bc : boundary_descriptor->velocity->symmetry_bc)
+        dealii::VectorTools::project_boundary_values_div_conforming(
+          dof_handler_u, 0, *(bc.second), bc.first, constraint_u, *get_mapping());
+    }
+    if(boundary_descriptor->velocity->dirichlet_bc.empty() == false)
```
nfehn, line 294: Above there are two empty lines, here is no empty line (which reads like "if-else if" at first sight. So I would structure the code a bit which comments like "periodic boundaries", "symmetry boundaries", "Dirichlet boundaries", given that `.empty()==false` is also hard to read (using instead `not(....empty())` would be easier in my opinion).

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
```diff
@@ -324,7 +400,31 @@ SpatialOperatorBase<dim, Number>::initialize_operators(std::string const & dof_i
...
+    if(param.preconditioner_mass == PreconditionerMass::PointJacobi)
+    {
+      mass_preconditioner =
+        std::make_shared<JacobiPreconditioner<MassOperator<dim, dim, Number>>>(this->mass_operator);
+      solver_data.use_preconditioner = true;
+    }
+    else if(param.preconditioner_mass == PreconditionerMass::None)
+    {
+      solver_data.use_preconditioner = false;
+    }
```
nfehn, line 422: I think we normally start with None, then the different preconditioners that are implemented, and finally an else branch with AssertThrow(false).

Imagine someone adds `Multigrid` as option for this preconditioner in `Parameters`. How should this developer find all places in the code that need to be adjusted in case you do not add an Assert now.

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
```diff
@@ -249,25 +249,101 @@ template<int dim, typename Number>
...
   pcout << "Velocity:" << std::endl;
-  print_parameter(pcout, "degree of 1D polynomials", param.degree_u);
+  if(get_spatial_discretization() == SpatialDiscretization::L2)
+  {
+    print_parameter(pcout, "degree of 1D polynomials", param.degree_u);
+  }
+  else if(get_spatial_discretization() == SpatialDiscretization::HDIV)
+  {
+    print_parameter(pcout, "degree of 1D polynomials (normal direction)", param.degree_u);
+    print_parameter(pcout, "degree of 1D polynomials (tangential direction)", (param.degree_u - 1));
```
nfehn, line 334: Looks like this needs more characters than the maximum number of characters "reserved" here:
https://github.com/exadg/exadg/blob/8382b8bacb87b907111ffe60e9cc48d7e86c2a4c/include/exadg/utilities/print_functions.h#L70
As a result, the alignment of parameters and their values when printing them to screen will probably no longer work. In my comment here (https://github.com/exadg/exadg/pull/233#discussion_r905978234) you can find another suggestion, or just remove "direction" above.

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
nfehn: I saw that you added the sentence in the first paragraph of the above comment. Please comment on the second paragraph so that it is a bit easier for me to keep the overview and so that I can mark comments as resolved.

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
nfehn: The second paragraph is optional, so I mark this comment as resolved.

nfehn: resolved

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
```diff
@@ -732,13 +832,27 @@ SpatialOperatorBase<dim, Number>::get_constraint_p() const
...
   return param.viscosity;
 }
 
+template<int dim, typename Number>
+SpatialDiscretization const &
+SpatialOperatorBase<dim, Number>::get_spatial_discretization() const
+{
+  return param.spatial_discretization;
+}
+
```
nfehn, line 855: This function should be removed.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_dual_splitting.cpp
```diff
@@ -376,7 +380,13 @@ TimeIntBDFDualSplitting<dim, Number>::convective_step()
   if(this->print_solver_info() and not(this->is_test))
   {
     this->pcout << std::endl << "Explicit convective step:";
-    print_wall_time(this->pcout, timer.wall_time());
+    if(this->param.spatial_discretization == SpatialDiscretization::HDIV)
+    {
+      this->pcout << std::endl << "(mass system solved)";
```
nfehn, line 385: Why do we need this comment? Isn't this already reflected in the print function that prints iteration counts to screen? Instead, the above comment "Explicit convective step" has to be adjusted, because it is no longer explicit in case of HDIV (just "Convective step"). Note that this will cause tests to fail, because the output for the L2 case will change.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_dual_splitting.cpp
```diff
@@ -622,6 +634,7 @@ TimeIntBDFDualSplitting<dim, Number>::projection_step()
 
     if(this->print_solver_info() and not(this->is_test))
     {
+      // No need to print the number of mass iterations here since we should be using L2
```
nfehn, line 637: "should be using L2" is unclear in my opinion. The question is: is it possible to run this part of the code (penalty terms applied!) in combination with HDIV? Whether we allow this parameter combination is to be commented in `Parameters` in my opinion. Given that you did not obtain exactly divergence-free results for the dual-splitting scheme, one might indeed want to experiment with this parameter combination in order to gain understanding.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_dual_splitting.cpp
```diff
@@ -631,7 +644,13 @@ TimeIntBDFDualSplitting<dim, Number>::projection_step()
     if(this->print_solver_info() and not(this->is_test))
     {
       this->pcout << std::endl << "Explicit projection step:";
```
nfehn, line 646: Here, the same applies as above. The step is no longer "explicit" in case of HDIV.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_dual_splitting.cpp
```diff
@@ -631,7 +644,13 @@ TimeIntBDFDualSplitting<dim, Number>::projection_step()
...
-      print_wall_time(this->pcout, timer.wall_time());
+      if(this->param.spatial_discretization == SpatialDiscretization::HDIV)
+      {
+        this->pcout << std::endl << "(mass system solved)";
+        print_solver_info_linear(this->pcout, n_iter_mass, timer.wall_time());
+      }
+      else
+        print_wall_time(this->pcout, timer.wall_time());
```
nfehn, line 653: Here, we should also use something like 

if(L2) - else if(HDIV) - else(AssertThrow(false))

Currently the code reads like we do not have to solve a mass matrix system in case of continuous finite elements, which is not the case and should therefore be avoided.

include/exadg/incompressible_navier_stokes/user_interface/parameters.cpp
```diff
@@ -385,9 +396,32 @@ Parameters::check(dealii::ConditionalOStream const & pcout) const
     }
   }
 
+  if(spatial_discretization == SpatialDiscretization::HDIV)
+  {
+    AssertThrow(
+      use_divergence_penalty == false || use_continuity_penalty == false,
```
nfehn, line 402: why "or" and not "and"?

include/exadg/incompressible_navier_stokes/user_interface/parameters.cpp
```diff
@@ -385,9 +396,32 @@ Parameters::check(dealii::ConditionalOStream const & pcout) const
...
+      dealii::ExcMessage(
+        "The use of penalty terms does not make sense in the case of HDIV since they are evaluated to zero. They should therefore be deactivated."));
+    if(temporal_discretization != TemporalDiscretization::BDFCoupledSolution)
```
nfehn, line 405: Why "!=" and not "=="?

Please introduce brackets for this if branch (it is formally not needed but a coding convention in ExaDG).

include/exadg/incompressible_navier_stokes/user_interface/parameters.cpp
```diff
@@ -385,9 +396,32 @@ Parameters::check(dealii::ConditionalOStream const & pcout) const
...
+      AssertThrow(
+        preconditioner_velocity_block == MomentumPreconditioner::None ||
+          preconditioner_velocity_block == MomentumPreconditioner::PointJacobi,
+        dealii::ExcMessage(
+          "Use either PointJacobi or None as preconditioner for the momentum block in the case of HDIV - Raviart-Thomas."));
```
nfehn, line 410: I think we need the same check for the viscous step of the dual splitting scheme (i.e. only PointJacobi allowed as preconditioner).

include/exadg/incompressible_navier_stokes/user_interface/parameters.cpp
```diff
@@ -697,7 +737,22 @@ Parameters::print_parameters_spatial_discretization(dealii::ConditionalOStream c
 
   grid.print(pcout);
 
-  print_parameter(pcout, "Polynomial degree velocity", degree_u);
+  print_parameter(pcout, "Element type", enum_to_string(spatial_discretization));
```
nfehn, line 740: This comment will cause tests to fail. Hence, you need to adjust the output of tests so that we can merge this PR.

include/exadg/incompressible_navier_stokes/user_interface/parameters.cpp
```diff
@@ -697,7 +737,22 @@ Parameters::print_parameters_spatial_discretization(dealii::ConditionalOStream c
...
+
+  if(spatial_discretization == SpatialDiscretization::L2)
+  {
+    print_parameter(pcout, "Polynomial degree velocity: ", degree_u);
+  }
+  else if(spatial_discretization == SpatialDiscretization::HDIV)
+  {
+    print_parameter(pcout, "Polynomial degree velocity in normal direction: ", degree_u);
+    print_parameter(pcout, "Polynomial degree velocity in tangential direction: ", degree_u);
```
nfehn, line 749: See other comment.

include/exadg/incompressible_navier_stokes/user_interface/parameters.cpp
```diff
@@ -468,6 +505,9 @@ Parameters::check(dealii::ConditionalOStream const & pcout) const
       use_cell_based_face_loops == true,
       dealii::ExcMessage(
         "Cell based face loops have to be used for matrix-free implementation of block diagonal preconditioner."));
+
+    AssertThrow(spatial_discretization != SpatialDiscretization::HDIV,
```
nfehn, line 509: Here, we should also write "==L2" similar to other asserts.

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
nfehn: I have not seen places in the code using the right-hand side operator for HDIV. So let's make sure that this part of the code is correct.

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
nfehn: do symmetry boundary conditions work now?

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
NiklasWik: Yes, I forgot to remove it when I changed back to `VT::interpolate`

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
NiklasWik: I'm not sure about this. I think that one still needs to take the piola transform into account. Maybe the use of `VectorTools::project_boundary_values_div_conforming` is enough here?

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.cpp
```diff
@@ -463,6 +463,13 @@ TimeIntBDF<dim, Number>::postprocessing() const
     operator_base->move_grid_and_update_dependent_data_structures(this->get_time());
   }
 
+  // We need to distribute the dofs for HDIV before computing the error since
+  // dealii::VectorTools::integrate_difference() does not take the periodic boundary
+  // constraints into account like MatrixFree does, hence reading the wrong value.
+  // distribute() updates the constrained values.
+  if(operator_base->get_spatial_discretization() == SpatialDiscretization::HDIV)
+    operator_base->get_constraint_u().distribute(const_cast<VectorType &>(get_velocity(0)));
+
```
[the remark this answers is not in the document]
NiklasWik: Yes, I agree that it doesn't really fit in here, but I think there will be problems below in the else branch when calculating the error, i.e lines 491-492. So if `standard` is ever changed to false... I can add a comment.

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
NiklasWik: But I will need it still in `postprocessor.cpp` since `Parameters param` is protected in `spatial_operator_base.h`, or am I missing something?

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
nfehn: @kronbichler what do you think?

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
nfehn: ah ok. Then, I think it is best to implement a function `SpatialOperatorBase::distribute_constraints()` (or similar) so that the other two functions are not needed. This function is public and can either be used in the `TimeIntegrator` or the `PostProcessor`.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.cpp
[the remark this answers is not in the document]
NiklasWik: I think you would need to call distribute before this line:
https://github.com/exadg/exadg/blob/d405d108cc1231ae7333a1fe2e631b7a8a58dcfb/include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.cpp#L491

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
NiklasWik: Yes!

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
NiklasWik: Ah, missed this. I can move the two lines again from `PostProcessor` to `SpatialOperatorBase`. And then call the funtion from the same place.

nfehn:
A few comments are still open, but we are approaching the finish line. We also need input from e.g. @kronbichler . Let's try to complete this asap.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.cpp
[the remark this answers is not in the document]
kronbichler: How do we handle inhomogeneous Dirichlet boundaries (or periodic boundaries) in the structure module? I think we should try to apply similar techniques, as those are needed for all forms of constrained unknowns. In particular, we should also anticipate the case of adaptive mesh refinement, where the velocity field will be subject to hanging node constraints. I did have a short look, but could not find it immediately (but I remember we discussed it at some point, as there are several solutions to the homogeneous/inhomogeneous representation needed at various stages). This does not necessarily have to happen in this PR, depending on how precisely we are going to solve it.

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
kronbichler: The project function in deal.II will only affect the normal component of the velocity, tangential components need to be handled weakly just as is the case at internal faces. I would think that `project_boundary_values_div_conforming` picks up the right normal direction (including Piola), but I guess that is a topic to be investigated more in the future.

NiklasWik:
Sorry it took a while!  

I've now added a function `SpatialDisretizationBase::distribute_constraint_u()` and checked that Taylor Green Vortex gives correct output.

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
nfehn: For now, we should adjust the comment and write that the normal component of the velocity needs to be applied strongly.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.cpp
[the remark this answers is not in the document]
nfehn: We should remove the else branch anyway from the time integrator, since this is not implemented in a nice way. It has previously been used to test certain things but should better be realized in the postprocessor. This is something I can address in the future.

@kronbichler What I did not understand is whether the aspects you mention are important for obtaining the correct solution, or just for postprocessing? I understood @NiklasWik in the sense that the solution is computed correctly and this discussion is only about doing the postprocessing correctly. Since Niklas has finished his project and will not implemented everything related to inhomogeneous BCs, we should find a solution that allows to complete this PR soon. If you address more general aspects here, could I ask you to summarize them in a separate issue such that they do not get lost and can be addressed later?

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.h
nfehn: @NiklasWik also here (and for other comments) it would be great if you could point me directly to your most recent changes.

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
NiklasWik: This was lastly changed here https://github.com/exadg/exadg/pull/233/commits/9b5780da19d7f31d6c490316d1e10f5297ab78bb

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.h
NiklasWik: This was removed here https://github.com/exadg/exadg/pull/233/commits/dbfbaf9f1489c35dff9d9de8c34338011780e4ff#diff-0cd2dc6d304bbc5350a7fb4cf865ac674a9b35b07d2b3463fc7a887f0943c2e6

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
NiklasWik: Here! https://github.com/exadg/exadg/pull/233/commits/dbfbaf9f1489c35dff9d9de8c34338011780e4ff#diff-b6700465d73837b3d857d6cb48936b37beff4706033534f21a6f4403039eb02c

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
NiklasWik: Done here https://github.com/exadg/exadg/pull/233/commits/54b505e18c5f6c9e4fa220627c351cbab07330f3 !

NiklasWik:
Since all current PRs seem to fail on the same lines during compiling, and since it compiles fine on my computer I'll leave it to you guys who know the problem better than me! 

Let me know if I can provide or change anything else!

include/exadg/incompressible_navier_stokes/user_interface/enum_types.h
```diff
@@ -612,6 +626,28 @@ std::string
...
+/**************************************************************************************/
+/*                                                                                    */
+/*                            SOLVE MASS SYSTEM (projection)                          */
+/*                                                                                    */
+/**************************************************************************************/
+
+/*
+ *  Preconditioner type for solution of mass system, i.e a projection:
+ *
+ *  InverseMassMatrix and BlockJacobi are not availiable..
```
nfehn, line 639: Here are two dots at the end

include/exadg/incompressible_navier_stokes/user_interface/enum_types.h
```diff
@@ -612,6 +626,28 @@ std::string
...
 
+
+/**************************************************************************************/
+/*                                                                                    */
+/*                            SOLVE MASS SYSTEM (projection)                          */
+/*                                                                                    */
+/**************************************************************************************/
+
+/*
+ *  Preconditioner type for solution of mass system, i.e a projection:
```
nfehn, line 637: I think we should avoid "projection" here (and above), since it is used in a different context in case of the Navier-Stokes solvers. There should somewhere also be a comment explaining that these parameters and the available preconditioners are relevant only for the HDIV case.

include/exadg/incompressible_navier_stokes/user_interface/parameters.cpp
```diff
@@ -223,7 +226,11 @@ Parameters::Parameters()
     preconditioner_pressure_block(SchurComplementPreconditioner::PressureConvectionDiffusion),
     multigrid_data_pressure_block(MultigridData()),
     exact_inversion_of_laplace_operator(false),
-    solver_data_pressure_block(SolverData(1e4, 1.e-12, 1.e-6, 100))
+    solver_data_pressure_block(SolverData(1e4, 1.e-12, 1.e-6, 100)),
+
+    // Solving mass system (projection) if Matrix-free inverse mass operator is not avaliable.
```
nfehn, line 231: a comment like this should appear where the parameters are defined (PreconditionerMass for example). Here in the parameter file, such a detailed explanation is not needed in my opinion. I would rather write e.g. "only relevant for HDIV case" here

include/exadg/time_integration/time_step_calculation.h
```diff
@@ -253,6 +253,8 @@ calculate_time_step_cfl_local(
   CFLConditionType const                                         cfl_condition_type,
   MPI_Comm const &                                               mpi_comm)
 {
+  // Need to update ghost values in the case of continuity constraints.
+  velocity.update_ghost_values();
```
nfehn, line 257: This is currently done always (L2 and HDIV). But we could also call this function in 

`IncNS::SpatialOperatorBase<dim, Number>::calculate_time_step_cfl()`

where we could easily distinguish between L2 and HDIV, because the spatial operator knows these parameters. Would such a design be clearer in terms of single responsibility?

nfehn:
@NiklasWik you could rebase and push again to see whether the checks are passing now.

nfehn (approved):

nfehn:
@NiklasWik are you ready for merging?

NiklasWik:
> @NiklasWik are you ready for merging?

Actually no. The 2D problem this adds is no longer working for HDIV and I can't figure out why. 3D tgv is still correct though. Not sure if this has to do with deal.ii or exadg since it works fine for L2.

peterrum:
> Actually no. The 2D problem this adds is no longer working for HDIV and I can't figure out why. 3D tgv is still correct though. Not sure if this has to do with deal.ii or exadg since it works fine for L2.

@NiklasWik Do you happen to have a combination of versions of ExaDG/deal.II that used to work; have you stored somewhere the code from your Master's thesis? From that it might be possible to determine the commit causing the issue. I would be surprised that the issue is coming from deal.II, since we - as far as I remember - didn't do any modifications there regarding RT. But to identify the root might be quite useful! Let's find the problem soon so that we can merge and finalize your Master's thesis!

NiklasWik:
> @NiklasWik Do you happen to have a combination of versions of ExaDG/deal.II that used to work; have you stored somewhere the code from your Master's thesis? 

Unfortunately not! I'm not sure when it stopped working either as when I've done these small changes I've only ever tested the three dimensional tgv.

NiklasWik:
It works using the dual splitting scheme, so that's something.

NiklasWik:
It was just the cfl! I'm pretty sure it was the first thing I tried last weekend... Must have been tired and done something weird. I don't either understand how I ended up pushing with an unstable cfl since I ran this problem quite a few times back in May/June. 

@nfehn should be all good now!

nfehn:
@NiklasWik Thanks for figuring this out! I am a bit irritated because you changed the CFL number to a larger value (0.1->0.4). Normally, we would expect that a smaller CFL number cannot be unstable if CFL=0.4 is stable? So did you see instabilities/blowup for CFL=0.1?

NiklasWik:
> @NiklasWik Thanks for figuring this out! I am a bit irritated because you changed the CFL number to a larger value (0.1->0.4). Normally, we would expect that a smaller CFL number cannot be unstable if CFL=0.4 is stable? So did you see instabilities/blowup for CFL=0.1?

Ah, yes! This is why I was so confused since the first thing I tried was taking a smaller CFL which didn't work. The week after I just compared with the settings of tgv as I thought it might have been something else. But yea, I did see blowup for CFL=0.1.

nfehn:
I might have an idea where this might come from but I am not sure. Could you figure out for which solution strategies (coupled, dual splitting, pressure correction) you observe this problem? For all of them?

nfehn:
Another question: Do you see numerical blowup in the sense that the simulation crashes, or "just" large errors (i.e. a seemingly wrong solution) but a simulation that completes until final time? The reason behind this question is that - from a mathematical/physical perspective - the given problem with its analytical solution is not necessarily "stable" in the sense that a "chaotic" solution might develop, e.g. triggered by small perturbations. These perturbations might be damped more strongly by a (dissipative) time integration scheme with larger time step size (and higher time integration error).

nfehn:
I am just interested whether see this problem also in case of the coupled solver. (I would hope not.)

NiklasWik:
> I am just interested whether see this problem also in case of the coupled solver. (I would hope not.)

Definitely for the coupled solver I'm afraid..

PR #235, opened 2022-06-28
Implement pure c multigrid option

bergbauer:


include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -215,6 +216,11 @@ MultigridPreconditionerBase<dim, Number>::initialize_levels(dealii::Triangulatio
     for(unsigned int h = 0; h < h_levels.size(); h++)
       level_info.push_back({h_levels[h], p_levels.front()});
   }
+  else if(mg_type == MultigridType::cMG)
+  {
+    level_info.push_back({h_levels.back(), p_levels.front()});
+    level_info.push_back({h_levels.back(), p_levels.back()});
```
nfehn, line 222: why front/back() for the p_levels? Probably, I just don't remember ... but let's be sure that this is intended/correct

bergbauer: As far as I understand it, yes!

nfehn:
looks good. I assume you ran a test with c-transfer. Did you run the ExaDG-tests manually to make sure that nothing has changed?

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
peterrum: Not sure but isn't the result of this that you jump from high-order DG to low-order CG? If yes, I guess this is not what you want. Have you checked the number of DoFs on the levels?

Also I wondering if the current implementation works. The deal.II transfer operators can handle the combination of p- and c-transfer; but not the ExaDG version, does it!?

nfehn: but only two p-levels (both at high degree) are created, right? so this would be a pure c-transfer at high degree, isn't it?

bergbauer: Yes only two levels at same degree are created (DG and CG) in the code below the changed lines above this comment.

nfehn:
@bergbauer reported a failure of tests.

Today, I updated to the recent dealii/master branch version. Before that, all tests were passing on my machine. Now, the following tests fail:

```
The following tests FAILED:
	  2 - tests/euler_vortex.release (Failed)
	  7 - tests/convective_explicit_time_int.release (Failed)
	  9 - tests/explicit_time_int.release (Failed)
```

So I guess that must have been introduced by recent changes in deal.II.

nfehn:
@bergbauer Could you finally run the tests on your machine before we merge?

bergbauer:
All tests are passing!

PR #238, opened 2022-07-04
fix Rayleigh-Benard setup

### LINKED ISSUE #237: No convergence with Rayleigh-Benard application
I tried to run the Rayleigh-Benard application with the default [input file](https://github.com/exadg/exadg/blob/master/applications/incompressible_flow_with_transport/rayleigh_benard/input.json). However, it seems like the CG solver for the `IncNS::ProjectionOperator` does not converge. Can you repeat the problem?

I am using deal.II 9.4.0. You can find the console output attached: [default.log](https://github.com/exadg/exadg/files/9039734/default.log)
 
I did not try other applications yet, so I can't say if other applications are affected as well.

nfehn:
closes #237

In my experience, the problem is a bit sensitive w.r.t. the initial perturbations of the temperature field. @marcfehling Can you check the setup proposed here (where I increased the initial perturbations)?

### DIFF
--- a/applications/incompressible_flow_with_transport/rayleigh_benard/application.h
+++ b/applications/incompressible_flow_with_transport/rayleigh_benard/application.h
@@ -50,7 +50,7 @@ public:
 
     double perturbation = time_factor;
     perturbation *=
-      0.25 * delta_T * std::pow(std::sin(dealii::numbers::PI * p[0] / (length / 4.)), 2.0);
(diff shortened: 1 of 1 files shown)


marcfehling:
> In my experience, the problem is a bit sensitive w.r.t. the initial perturbations of the temperature field. @marcfehling Can you check the setup proposed here (where I increased the initial perturbations)?

With your patch, I no longer get the no convergence error. The new configuration runs without problems in Debug mode.

Thank you! I appreciate your quick response :)

PR #242, opened 2022-07-06
Add includes

jh66637:
Some includes are missing, such that ExaDG builds out of the box on our cluster.

### DIFF
--- a/include/exadg/convection_diffusion/user_interface/boundary_descriptor.h
+++ b/include/exadg/convection_diffusion/user_interface/boundary_descriptor.h
@@ -32,6 +32,8 @@
 // ExaDG
 #include <exadg/functions_and_boundary_conditions/function_cached.h>
 
+#include <memory>
+
 namespace ExaDG
 {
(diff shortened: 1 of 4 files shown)


peterrum (approved):

include/exadg/convection_diffusion/user_interface/boundary_descriptor.h
```diff
@@ -32,6 +32,8 @@
 // ExaDG
 #include <exadg/functions_and_boundary_conditions/function_cached.h>
 
+#include <memory>
```
nfehn, line 35: for example here: Why is memory needed here? We should make sure to insert the includes where necessary.

peterrum: I guess due to the shared pointers.

jh66637: @peterrum exactly

PR #245, opened 2022-07-11
Only evaluate quantities if necessary

jh66637:
Add possibility to preprocess data before computing the error. With the current code design, such preprocessing is done in do_postprocessing() functions, which are called each time step. This PR helps to reduce the work per time step by preprocessing data only if necessary. Similar functionality should be added in some places, e.g. for the compressible NSE. If we merge this, I can tackle this in a follow-up PR when I find some time.

### DIFF
--- a/applications/compressible_navier_stokes/turbulent_channel/application.h
+++ b/applications/compressible_navier_stokes/turbulent_channel/application.h
@@ -252,8 +252,7 @@ public:
     if(statistics_turb_ch->time_control_statistics.time_control.needs_evaluation(time,
                                                                                  time_step_number))
     {
-      statistics_turb_ch->evaluate(evaluate_get(this->velocity,
-                                                Utilities::is_unsteady_timestep(time_step_number)),
+      statistics_turb_ch->evaluate(evaluate_get(this->velocity),
                                    Utilities::is_unsteady_timestep(time_step_number));
     }
 
--- a/applications/compressible_navier_stokes/turbulent_channel/application.h
+++ b/applications/compressible_navier_stokes/turbulent_channel/application.h
@@ -252,7 +252,8 @@ public:
     if(statistics_turb_ch->time_control_statistics.time_control.needs_evaluation(time,
                                                                                  time_step_number))
     {
-      statistics_turb_ch->evaluate(this->velocity,
+      statistics_turb_ch->evaluate(evaluate_get(this->velocity,
+                                                Utilities::is_unsteady_timestep(time_step_number)),
                                    Utilities::is_unsteady_timestep(time_step_number));
     }
 
@@ -412,9 +413,8 @@ private:
     pp_data.output_data.time_control_data.is_active        = this->output_parameters.write;
     pp_data.output_data.time_control_data.start_time       = START_TIME;
     pp_data.output_data.time_control_data.trigger_interval = 1.0;
-    pp_data.output_data.directory = this->output_parameters.directory + "vtu/";
-    pp_data.output_data.filename  = this->output_parameters.filename;
-    pp_data.calculate_velocity = true; // activate this for kinetic energy calculations (see below)
+    pp_data.output_data.directory          = this->output_parameters.directory + "vtu/";
+    pp_data.output_data.filename           = this->output_parameters.filename;
     pp_data.output_data.write_pressure     = true;
     pp_data.output_data.write_velocity     = true;
     pp_data.output_data.write_temperature  = true;
--- a/include/exadg/incompressible_navier_stokes/postprocessor/postprocessor.cpp
+++ b/include/exadg/incompressible_navier_stokes/postprocessor/postprocessor.cpp
@@ -116,8 +116,11 @@ PostProcessor<dim, Number>::do_postprocessing(VectorType const &     velocity,
    */
   if(time_control_mean_velocity.needs_evaluation(time, time_step_number))
   {
-    if(time_control_mean_velocity.needs_evaluation(time, time_step_number))
-      mean_velocity.evaluate(Utilities::is_unsteady_timestep(time_step_number));
+    AssertThrow(Utilities::is_unsteady_timestep(time_step_number),
+                dealii::ExcMessage(
+                  "Calculating mean velocity does not make sense for steady problems."));
+
+    mean_velocity.evaluate();
   }
 
 
@@ -128,8 +131,7 @@ PostProcessor<dim, Number>::do_postprocessing(VectorType const &     velocity,
   {
     output_generator.evaluate(velocity,
                               pressure,
-                              evaluate_get(additional_fields_vtu,
-                                           Utilities::is_unsteady_timestep(time_step_number)),
+                              evaluate_get(additional_fields_vtu),
                               time,
                               Utilities::is_unsteady_timestep(time_step_number));
   }
@@ -203,7 +205,7 @@ PostProcessor<dim, Number>::initialize_additional_vectors()
     vorticity.name        = "vorticity";
     vorticity.dof_handler = &navier_stokes_operator->get_dof_handler_u();
     navier_stokes_operator->initialize_vector_velocity(vorticity.get_vector_reference());
-    vorticity.recompute_solution_field = [&](VectorType & dst, VectorType const & src, bool const) {
+    vorticity.recompute_solution_field = [&](VectorType & dst, VectorType const & src) {
       navier_stokes_operator->compute_vorticity(dst, src);
     };
 
@@ -217,10 +219,9 @@ PostProcessor<dim, Number>::initialize_additional_vectors()
     divergence.name        = "div_u";
     divergence.dof_handler = &navier_stokes_operator->get_dof_handler_u_scalar();
     navier_stokes_operator->initialize_vector_velocity_scalar(divergence.get_vector_reference());
-    divergence.recompute_solution_field =
-      [&](VectorType & dst, VectorType const & src, bool const) {
-        navier_stokes_operator->compute_divergence(dst, src);
-      };
+    divergence.recompute_solution_field = [&](VectorType & dst, VectorType const & src) {
+      navier_stokes_operator->compute_divergence(dst, src);
+    };
 
     this->additional_fields_vtu.push_back(&divergence);
   }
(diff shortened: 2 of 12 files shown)


peterrum:
@jh66637 I must admit I don't see the use case. Do you have an example?

jh66637:
https://github.com/exadg/exadg/blob/905814367356a66dedfc152013dd751d9015d180/include/exadg/compressible_navier_stokes/postprocessor/postprocessor.cpp#L87-L105.
do_postprocessing() is called every time step. Subsequently, the calculation of additional vectors is performed each time step, even though it might not be needed. Now, imagine I want to calculate the error of the pressure only. Then I could write e.g.
```
error_calculator.evaluate(pressure, time, time_step_number,[&](){navier_stokes_operator->compute_pressure(pressure, solution);});
```
Now the computation of the additional vector is only triggered if needed. Obviously, in the code above calculate_additional_vectors() would have to be deleted and all evaluate() functions should be equipped with the same concept.

jh66637:
I would preferre to pass a lambda to keep everything local without introducing a lot of additional code. If you like it better, this can probably also be done during construction. 

Regarding, computing quantities twice per time step: I thought about putting the quantities in a pair<timestep_number,Vector> and recompute only if timestep_number differs. This has the advantage that not all quantities are computed if one calculator is triggered. If I understood your idea with ```is_available``` correct, all quantities have to be computed if one calculator is triggered, otherwise you need one ```is_available``` per solution field.

Even though I like the lambda I can also live with different options which get the job done.

nfehn:
> Regarding, computing quantities twice per time step: I thought about putting the quantities in a pair<timestep_number,Vector> and recompute only if timestep_number differs. This has the advantage that not all quantities are computed if one calculator is triggered. If I understood your idea with is_available correct, all quantities have to be computed if one calculator is triggered, otherwise you need one is_available per solution field.

No, I want to make the current implementation more modular by using a variable `is_available` for each field separately. I expect that we need some lines of code for a good design, since the current implementation has just been a "quick solution".

jh66637:
@nfehn I went with Design C. Currently, I applied the changes only to the CNSE. It would be great if you could review the changes and leave your thoughts before I apply the changes in all postprocessors :)

jh66637:
@nfehn I know tests are currently failing. This is because I did not apply the changes to the whole codebase. Is there a chance to review the design before I adapt the rest of the code ? :)

applications/compressible_navier_stokes/turbulent_channel/application.h
```diff
@@ -246,7 +246,7 @@ class MyPostProcessor : public PostProcessor<dim, Number>
   {
     Base::do_postprocessing(solution, time, time_step_number);
 
-    statistics_turb_ch->evaluate(this->velocity, time, time_step_number);
+    statistics_turb_ch->evaluate(this->velocity.get(), time, time_step_number);
```
nfehn, line 249: I think this line is the only weakness of the current design. It always computes the velocity even though this might only be necessary every 10th/100th time step. Passing `SolutionField` to the statistics manager would be natural, but this is difficult since the same implementation is used for incompressible Navier-Stokes, where velocity is already a primary solution variable and introducing solution field there would mean additional implementation effort and somewhat needless complexity. So I can live with the current solution.

nfehn:
@jh66637 You realized quite a good solution! Looking at the code, there is - in my opinion - one obvious deficiency in the code design that we need to fix in the future. This deficiency is due to what I implemented in the past. I think we need to make the check whether a certain postprocessing routine is done in a given time step outside of the function `postprocessing_routine.evaluate()`, i.e.

```c++
if(postprocessing_routine.evaluate_now(time, time_step_number)
  postprocessing_routine.evaluate(solution_field.get());
```
This way, we can keep the postprocessing routines independent of `SolutionField`, thereby minimizing dependencies. We then no longer need two functions `evaluate()`, one with argument of type `VectorType` and one with argument `SolutionField`.

A main advantage of this design is also that we avoid code duplication. I do not want to count how often we implement the same functionality in terms of figuring out whether we arrived at a new time window for postprocessing for a particular postprocessing routine ... For example, PR #255 needs to implement the same function (with counter, reset counter, EPSILON, etc.) that many other postprocessing routines have already implemented. We need to introduce a class `TimeControl` (or similar) that takes care of exactly this (single responsibility principle), which is then used inside each postprocessing routine. Hence, the above if-statement could also look like this

```c++
if(postprocessing_routine.time_control.needs_evaluation(time, time_step_number)
  postprocessing_routine.evaluate(solution_field.get());
```
This way, we avoid implementing the same type of function for every postprocessing routine.

The question is where we should make these changes. If we realize these changes later, we should not forget to revert many changes introduced in the present PR, i.e. we then only need a single `evaluate()` function and no longer need the second `evaluate()` function introduced in the present PR.

jh66637:
I think we should introduce ```time_control``` first. I can do it once https://github.com/exadg/exadg/pull/255 is merged.

include/exadg/incompressible_navier_stokes/postprocessor/postprocessor.cpp
```diff
@@ -182,105 +191,123 @@ void
...
   // cfl
   if(pp_data.output_data.write_cfl)
   {
-    SolutionField<dim, Number> sol;
-    sol.type   = SolutionFieldType::cellwise;
-    sol.name   = "cfl_relative";
-    sol.vector = &cfl_vector;
-    this->additional_fields.push_back(sol);
+    // TODO: IS THIS A BUG OR DO CELL VECTORS NEED NOT INTIALIZATION?
+    cfl_vector.type                     = SolutionFieldType::cellwise;
```
jh66637, line 300: There is no initialization of the Vector. Do Cell vectors need no intialization?

jh66637:
I rebased this PR to https://github.com/exadg/exadg/pull/255 and applied the necessary changes. For review, it is probably best to check the last commit (The first commit is PR https://github.com/exadg/exadg/pull/262).
Except above question, I think I am ready here.
@nfehn Any comments?

include/exadg/incompressible_navier_stokes/postprocessor/postprocessor.cpp
```diff
@@ -193,103 +199,127 @@ PostProcessor<dim, Number>::initialize_additional_vectors()
...
   // cfl
   if(pp_data.output_data.write_cfl)
   {
-    SolutionField<dim, Number> sol;
-    sol.type   = SolutionFieldType::cellwise;
-    sol.name   = "cfl_relative";
-    sol.vector = &cfl_vector;
-    this->additional_fields.push_back(sol);
+    // TODO: IS THIS A BUG OR DO CELL VECTORS NEED NOT INTIALIZATION?
+    cfl_vector.type = SolutionFieldType::cellwise;
```
jh66637, line 311: Do cell vectors need no initialization?

jh66637:
I rebased again onto the current time control version

jh66637:
From my side this can be merged directly after https://github.com/exadg/exadg/pull/262, if you can confirm https://github.com/exadg/exadg/pull/245#discussion_r996444168.

include/exadg/postprocessor/solution_field.h
```diff
@@ -35,23 +35,103 @@ enum class SolutionFieldType
...
+
+  void
+  reinit(VectorType const & solution_in)
+  {
+    is_available = false;
+    solution     = &solution_in;
+  }
+
+  void
+  evaluate(bool const unsteady)
```
nfehn, line 61: I would like to get rid of the parameter `unsteady`, which appears as function argument in four functions in this PR, but it seems as if it is only required by the assert in `compute_mean_velocity()`. I think we should find a better solution for this. What happens if we only create and access the mean velocity object in case of unsteady problems? i.e. `Postprocessor` has the notion of steady/unsteady problems and can care about this, but a solution field should not care about this.

include/exadg/incompressible_navier_stokes/postprocessor/postprocessor.cpp
```diff
@@ -193,103 +199,126 @@ PostProcessor<dim, Number>::initialize_additional_vectors()
...
+    mean_velocity.type        = SolutionFieldType::vector;
+    mean_velocity.name        = "mean_velocity";
+    mean_velocity.dof_handler = &navier_stokes_operator->get_dof_handler_u();
+    navier_stokes_operator->initialize_vector_velocity(mean_velocity.get_vector_reference());
+    mean_velocity.recompute_solution_field =
+      [&](VectorType & dst, VectorType const & src, bool const unsteady) {
+        compute_mean_velocity(dst, src, unsteady);
+      };
+
+    this->additional_fields_vtu.push_back(&mean_velocity);
```
nfehn, line 304: why not doing this only for unsteady problems so that we can get rid of the parameter `unsteady` in the above lambda?

include/exadg/postprocessor/solution_field.h
jh66637: Only creating in the unsteady case is not as easy as it seems, since inside the relevant function it is not known if the simulation is unsteady. 
Instead, I moved the Assert to the point before mean_velocity is evaluated. This simplifies the code a lot, but comes at the price that the user has to think about checking if the simulation is unsteady before evaluating mean_velocity.

nfehn: I think this is exactly how it should look like. I like the current version.

include/exadg/incompressible_navier_stokes/postprocessor/postprocessor.cpp
```diff
@@ -193,190 +201,146 @@ PostProcessor<dim, Number>::initialize_additional_vectors()
...
   {
-    navier_stokes_operator->initialize_vector_velocity_scalar(divergence);
-
-    SolutionField<dim, Number> sol;
-    sol.type        = SolutionFieldType::scalar;
-    sol.name        = "div_u";
-    sol.dof_handler = &navier_stokes_operator->get_dof_handler_u_scalar();
-    sol.vector      = &divergence;
-    this->additional_fields.push_back(sol);
+    divergence.type        = SolutionFieldType::vector;
```
nfehn, line 218: I think this line introduced a bug that has not been detected so far.

PR #248, opened 2022-07-12
Verify the calculation of the diagonal in parallel

bergbauer:
Enables the verification of the diagonal in parallel

Checked correctness with poisson/sine testcase with PointJacobi preconditioner on 12 procs locally.


nfehn (approved):
looks good from my side. I assume `vmult()` does all the necessary synchronisation steps so that each process uses most recent data?

peterrum (approved):

PR #250, opened 2022-07-18
Avoid Copy

jh66637:
Currently, the grid velocity is copied and not passed by const reference.

Also, set_grid_velocity_ptr() should not be const in my opinion since a member is set. This might be up to discussion, since the member is mutable. However, the mutable keyword is only used to be able to access the member in the matrix free loops and has nothing to do with setting the velocity.

### DIFF
--- a/include/exadg/incompressible_navier_stokes/spatial_discretization/operators/convective_operator.h
+++ b/include/exadg/incompressible_navier_stokes/spatial_discretization/operators/convective_operator.h
@@ -93,7 +93,7 @@ public:
 
     if(data.ale)
     {
-      matrix_free.initialize_dof_vector(grid_velocity, dof_index);
+      matrix_free.initialize_dof_vector(grid_velocity.own(), dof_index);
 
       AssertThrow(data.formulation == FormulationConvectiveTerm::ConvectiveFormulation,
                   dealii::ExcMessage(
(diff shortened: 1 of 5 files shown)


nfehn:
makes sense in my opinion

nfehn:
The name `set_grid_velocity_ptr()` (especially "pointer") is irritating/misleading in my opinion. The DoF-vector is copied. So shouldn't we change the name to `set_grid_velocity()`?

Do you get an error when you try to remove the `mutable` keyword?

jh66637:
Removing the mutable doesn't compile because it is accessed inside the matrix free loops (which are const). 

I think we should const the lazy_pointer access (see 2076424e7df07f3c4c41ac66cf55b81207cc4d90). Putting variables there should make it possible to remove the mutable keywords, I am not sure in how many places of the code this can be done.

jh66637:
Not sure why the pipeline fails. It complains about missing template arguments, which is probably not related to this PR. Also I ran clang-format on the files.

nfehn:
@jh66637 is this ready-to-merge from your side? Any comments/concerns by @bergbauer @peterrum @kronbichler (@peterrum has written the lazy_ptr)?

jh66637:
Yes, I would be ready :)

PR #251, opened 2022-07-19
[WIP] Fix Compilation Errors

jh66637:
I think there are compilation errors related to explicitly using double and https://github.com/dealii/dealii/pull/14142

### DIFF
--- a/include/exadg/functions_and_boundary_conditions/interface_coupling.cpp
+++ b/include/exadg/functions_and_boundary_conditions/interface_coupling.cpp
@@ -196,10 +196,10 @@ template class ContainerInterfaceData<2, 2, double>;
 template class ContainerInterfaceData<3, 1, double>;
 template class ContainerInterfaceData<3, 3, double>;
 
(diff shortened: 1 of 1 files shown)


nfehn:
hm, it is actually a design goal in ExaDG to keep  `Number` out of data structures such as `BoundaryDescriptor`. I am not sure if the changes here are the way how we should approach this.

jh66637:
I am currently just checking where the error comes from. I am not intending to keep all the changes

jh66637:
@nfehn The problem is, that deal checks internally for double vs float. So instantiating InterfaceCoupling<float> will fail. So you should not be allowed to call InterfaceCoupling and InterfaceCouplingData with Number==float in the first place if only doubles are supported inside.

jh66637:
To be able to keep Number out, we need something like a copy of float data to double data. Or a major code refracturing. I might give it a try again towards the end of the week and create an Issue in the mean time.

PR #253, opened 2022-07-22
Use MatrixFree::get_affine_constraints() function

bergbauer:
Use the new function get_affine_constraints() introduced in https://github.com/dealii/dealii/pull/14132 in the multigrid preconditioners


nfehn:
looks good from my side. @kronbichler do you also agree? I think then we could merge this.

kronbichler (approved):

PR #255, opened 2022-08-08
Record physical quantities over time

jh66637:
This PR adds functionality to record quantities at certain points in space over time.

To show an example, I started to add this to the compressible Navier Stokes postprocessor.


include/exadg/compressible_navier_stokes/postprocessor/hdf5_generator.cpp
```diff
@@ -0,0 +1,118 @@
...
+{
+  if(hdf5_data.write_rho || hdf5_data.write_rho_u || hdf5_data.write_rho_E)
+  {
+    auto const values = dealii::VectorTools::point_values<dim + 2>(this->get_remote_evaluator(),
+                                                                   *dof_handler,
+                                                                   solution);
+    if(hdf5_data.write_rho)
+    {
+      this->extract_component(hyperslab_data, values, 0);
+      this->write_scalar_quantity("Rho", hyperslab_data);
```
nfehn, line 93: These two functions seem to appear only in combination. I think it would make sense to combine them into a single function.

Also, I do not understand why `hyperslab_data` is an argument of this function? I think of the following interface:
```
this->write_scalar_quantity(values, 0, "Rho");
```

include/exadg/compressible_navier_stokes/postprocessor/hdf5_generator.h
```diff
@@ -0,0 +1,70 @@
...
+        dealii::Mapping<dim> const &    mapping_in,
+        HDF5Data<dim> const &           hdf5_data_in);
+
+private:
+  void
+  do_evaluate(VectorType const & solution) final;
+
+  HDF5Data<dim>                                       hdf5_data;
+  dealii::SmartPointer<dealii::DoFHandler<dim> const> dof_handler;
+  dealii::Vector<Number>                              hyperslab_data;
```
nfehn, line 63: Related to the above comment (https://github.com/exadg/exadg/pull/255/files#r964020314), I believe this variable should not be a member variable of this class.

nfehn:
As a general question, what is the role of HDF5 here? I think HDF5 should not be that present in all the class names. The characteristic aspect of this postprocessing routine seems to be the ability to write results for a set of specified points in space, similarly as we have postprocessing routines for the data in the whole volume or e.g. for data along lines (`LinePlotCalculator`). Against this background, the naming should be driven by the type of postprocessing routine (e.g. `PointwiseOutputGenerator`?).

include/exadg/postprocessor/hdf5_generator_base.h
```diff
@@ -0,0 +1,149 @@
...
+  write_time(double time);
+
+  MPI_Comm const                                                      mpi_comm;
+  dealii::SmartPointer<dealii::Triangulation<dim> const>              tria;
+  dealii::SmartPointer<dealii::Mapping<dim> const>                    mapping;
+  std::shared_ptr<dealii::Utilities::MPI::RemotePointEvaluation<dim>> remote_evaluator;
+  HDF5DataBase<dim>                                                   hdf5_data;
+  unsigned int                                                        counter = 0;
+  unsigned int                                                        n_out_samples;
+  bool                                                                reset_counter = true;
```
nfehn, line 142: initialization should be done in constructor in my opinion (is done like this almost everywhere in ExaDG).

include/exadg/compressible_navier_stokes/postprocessor/hdf5_generator.cpp
jh66637: resolved with [7647d7b](https://github.com/exadg/exadg/pull/255/commits/7647d7b3beaddfbed1cba903da224c0bbb5f09a5)

include/exadg/compressible_navier_stokes/postprocessor/hdf5_generator.h
jh66637: resolved with [7647d7b](https://github.com/exadg/exadg/pull/255/commits/7647d7b3beaddfbed1cba903da224c0bbb5f09a5)

jh66637:
I renamed to ```PointwiseOutput...``` as you suggested

include/exadg/postprocessor/hdf5_generator_base.h
jh66637: resolved in [5f34cc0](https://github.com/exadg/exadg/pull/255/commits/5f34cc07efbf9106a7441fbd8460f044066469ed)

jh66637:
What is left to do in the end is to test again and replace decltype(auto).

applications/compressible_navier_stokes/turbulent_channel/application.h
```diff
@@ -409,6 +409,25 @@ class Application : public ApplicationBase<dim, Number>
     pp_data.output_data.degree             = this->param.degree;
     pp_data.output_data.write_higher_order = false;
 
+    // write data to hdf5
+    pp_data.pointwise_output_data.write_output = true;
```
nfehn, line 413: in the end, this should be set to `false` (and activated by the user) in my opinion

include/exadg/postprocessor/pointwise_output_generator_base.cpp
```diff
@@ -0,0 +1,291 @@
...
+  PointwiseOutputDataBase<dim> const & pointwise_output_data_in)
+{
+  pointwise_output_data = pointwise_output_data_in;
+
+  if(pointwise_output_data.write_output && pointwise_output_data.evaluation_points.size() > 0)
+  {
+    dof_handler = &dof_handler_in;
+    mapping     = &mapping_in;
+
+    // allocate memory for hyperslap
```
nfehn, line 85: spelling: hyperslap vs. hyperslab?

jh66637: resolved in [6e0c9b3](https://github.com/exadg/exadg/pull/255/commits/6e0c9b3db6fa171da3b797e3347ce1c0720106a2)

applications/compressible_navier_stokes/turbulent_channel/application.h
jh66637: resolved in [a674dae](https://github.com/exadg/exadg/pull/255/commits/a674dae71e792a298cc4cef22fc65d0f5b3598f7)

include/exadg/postprocessor/pointwise_output_generator_base.h
```diff
@@ -0,0 +1,198 @@
...
+
+  void
+  setup_base(dealii::DoFHandler<dim> const &      dof_handler_in,
+             dealii::Mapping<dim> const &         mapping_in,
+             PointwiseOutputDataBase<dim> const & pointwise_output_data_in);
+
+  void
+  evaluate(VectorType const & solution, double const & time, int const & time_step_number);
+
+  void
```
nfehn, line 82: These functions down to `compute_point_values()` do not have to be public. The public interface should be minimal.

include/exadg/postprocessor/pointwise_output_generator_base.h
```diff
@@ -0,0 +1,198 @@
...
+  fill_evaluation_points(std::string const & name);
+
+  void
+  add_time_dataset(dealii::HDF5::Group & group, std::string const & name);
+
+  void
+  write_time(double time);
+
+  void
+  write_hyperslab(std::string const & name, dealii::Vector<Number> const & hyperslab_data);
```
nfehn, line 164: The naming `hyperslab` is very prominent in this MR. Does this come from HDF5? What is the meaning? I wonder whether we could find a more expressive name ...

include/exadg/postprocessor/pointwise_output_generator_base.h
```diff
@@ -0,0 +1,198 @@
...
+    AssertThrow(first_component == 0,
+                dealii::ExcMessage("This function can only be called with first_component==0."));
+
+    hyperslab_data = std::move(dealii::Vector<Number>(values.begin(), values.end()));
+    write_hyperslab(name, hyperslab_data);
+  }
+
+  template<int n_components>
+  void
+  write_scalar_quantity(std::string const &                                          name,
```
nfehn, line 108: I don't understand this function name in combination with `Tensor<1, n_components, Number>`? This raises the question whether we should distinguish at all between scalar/vectorial regarding the function names.

include/exadg/postprocessor/pointwise_output_generator_base.h
```diff
@@ -0,0 +1,198 @@
...
+  void
+  write_vectorial_quantity(std::string const & name, dealii::Vector<Number> const & hyperslab_data);
+
+  void
+  write_scalar_quantity(std::string const &         name,
+                        std::vector<Number> const & values,
+                        unsigned int const          first_component = 0)
+  {
+    AssertThrow(first_component == 0,
+                dealii::ExcMessage("This function can only be called with first_component==0."));
```
nfehn, line 100: also this looks a bit strange to me. Why is it needed in addition to the other function `write_scalar_quantity()` above?

scripts/config_dealii.sh
```diff
@@ -56,6 +56,7 @@ cmake \
     -D DEAL_II_CXX_FLAGS_RELEASE="-O3" \
     -D DEAL_II_CXX_FLAGS_DEBUG="-Og" \
     -D DEAL_II_WITH_MPI:BOOL="ON" \
+    -D DEAL_II_WITH_HDF5:BOOL="ON" \
```
nfehn, line 59: ExaDG should compile without errors if this variable is set `OFF`. Is this guaranteed?

nfehn: When trying to use the new functionality of this MR in ExaDG, an error message when running the executable should tell the user that one has to activate HDF5 in the configuration files to use this postprocessing functionality.

include/exadg/postprocessor/pointwise_output_generator_base.h
jh66637: It comes from HDF5. A hyperslab is basically the place where to write data. So hyperslab data would be the data that has to be written to the hyperslab. In this case, it would be the values of a component at all points at a certain time. So ```componentwise_result``` might be a name we could use if you don't want to have the HDF5 specific names here.

nfehn: Yes, I would prefer such a name.

include/exadg/postprocessor/pointwise_output_generator_base.h
jh66637: Say we have a solution vector containing pressure and velocity, then we have to pick out the pressure or velocity before we write to the file. So basically we give the functions all results and leave it up to the function to extract the correct data and write it to the file. I would like to keep it this way if you don't have any concerns.

include/exadg/postprocessor/pointwise_output_generator_base.h
jh66637: Remote Point evaluation gives us different types regarding the number of components that are present in the solution. Regarding the type, we have to extract values or not. With this function overload, the interface stays the same and the user does not have to care about it.

scripts/config_dealii.sh
jh66637: I will check this, but it will take some time to compile deal.ii without hdf5 to test it.

include/exadg/postprocessor/pointwise_output_generator_base.h
jh66637: resolved in [bddce88](https://github.com/exadg/exadg/pull/255/commits/bddce8854304c2194e17aaa84d9d06e5fef4d23f)

jh66637:
When merging this PR, the wiki should also be adapted to show how to install hdf5. I didn't find an option to do that myself. It boils down to:
```sudo apt install libhdf5-dev```

**Comment by jh66637** [comment]
This would be ready from my side :)

File: include/exadg/compressible_navier_stokes/postprocessor/pointwise_output_generator.cpp [hunk 11]
```diff
@@ -0,0 +1,104 @@
...
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *  ______________________________________________________________________
+ */
+
+#include <exadg/compressible_navier_stokes/postprocessor/pointwise_output_generator.h>
+
+#include <deal.II/numerics/vector_tools.h>
+
+#include <exadg/utilities/print_functions.h>
```
nfehn, line 26: as a minor comment: we typically follow the convention in ExaDG to list header files in a certain order, beginning with C/C++, followed by deal.II, and finally ExaDG. Would you like to adjust this?

include/exadg/postprocessor/pointwise_output_generator_base.cpp
```diff
@@ -0,0 +1,325 @@
...
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *  ______________________________________________________________________
+ */
+
+#include <exadg/postprocessor/pointwise_output_generator_base.h>
+
+#include <deal.II/lac/full_matrix.h>
```
nfehn, line 24: same as other comment

include/exadg/postprocessor/pointwise_output_generator_base.h
nfehn: Here, I am not convinced that this is the right design. We have a function argument that needs to be fixed to a certain value but is not used by the function.

Why not giving up scalar and vectorial completely? Why not using a single function with arguments `first_component` and `last_component` (maybe with the default value `last_component=first_component`) for the general case with `Tensor` as function argument. For the special case of a scalar quantity and `std::vector<Number>` as argument, wouldn't it make sense to write directly `write_component(name, std::move(dealii::Vector<Number>(values.begin(), values.end())))` instead of calling `write_scalar_quantity`?

include/exadg/postprocessor/pointwise_output_generator_base.h
nfehn: see discussion in other thread.

include/exadg/postprocessor/pointwise_output_generator_base.h
```diff
@@ -0,0 +1,202 @@
...
+  {
+    AssertIndexRange(comp, n_components);
+
+    auto iter = dst.begin();
+    for(auto const & val : values)
+    {
+      *iter = val[comp];
+      ++iter;
+    }
+  }
```
nfehn, line 183: This functionality is so general that it should appear as a free function at a central place in ExaDG (in directory utilities) in my opinion.

include/exadg/postprocessor/pointwise_output_generator_base.h
```diff
@@ -0,0 +1,202 @@
...
+
+#ifdef DEAL_II_WITH_HDF5
+  void
+  add_evaluation_points_dataset(dealii::HDF5::Group & group, std::string const & name);
+#endif
+
+  void
+  fill_evaluation_points(std::string const & name);
+
+#ifdef DEAL_II_WITH_HDF5
```
nfehn, line 158: I would group together the two functions guarded by `DEAL_II_WITH_HDF5`.

include/exadg/compressible_navier_stokes/postprocessor/pointwise_output_generator.cpp
jh66637: Shouldn't the header file corresponding to the .cpp file be the first include to ensure it is self-contained? I can change it for consistency, though.

include/exadg/postprocessor/pointwise_output_generator_base.h
jh66637: resolved with [93ec7fd](https://github.com/exadg/exadg/pull/255/commits/93ec7fd87353d984a30b73b260daaa2d6da69407)

include/exadg/postprocessor/pointwise_output_generator_base.h
```diff
@@ -0,0 +1,202 @@
...
+  void
+  create_hdf5_file();
+
+#ifdef DEAL_II_WITH_HDF5
+  void
+  add_evaluation_points_dataset(dealii::HDF5::Group & group, std::string const & name);
+#endif
+
+  void
+  fill_evaluation_points(std::string const & name);
```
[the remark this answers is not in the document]
jh66637: The decision to split the functions was motivated as follows:
In hdf5, memory is allocated while adding datasets. Filling datasets writes to the allocated memory. This is kind of a logical splitting from my point of view. 
This is also how we proceed in the other cases. We add a dataset and fill it, e.g. in each time step. Therefore, I think it is most intuitive to split the functionality everywhere.

You can decide if you want it fused. I think both points are valid, and I don't prefer one over the other.

include/exadg/postprocessor/pointwise_output_generator_base.h
jh66637: [ce6a59b](https://github.com/exadg/exadg/pull/255/commits/ce6a59b5edabc2a6164a6c62e803c748b3fdce23)

include/exadg/compressible_navier_stokes/postprocessor/pointwise_output_generator.cpp
nfehn: ok

include/exadg/postprocessor/pointwise_output_generator_base.h
```diff
@@ -22,6 +22,8 @@
 #ifndef INCLUDE_COMPRESSIBLE_NAVIER_STOKES_POSTPROCESSOR_POINTWISE_OUTPUT_GENERATOR_BASE_H_
 #define INCLUDE_COMPRESSIBLE_NAVIER_STOKES_POSTPROCESSOR_POINTWISE_OUTPUT_GENERATOR_BASE_H_
 
+#include <exadg/utilities/extract_component_from_tensors.h>
```
nfehn, line 25: Still, I am not sure whether we should break here with the current ExaDG convention. I am open to change this in the future if other solutions are more intelligent, but for now I would like to keep the convention we currently have.

include/exadg/utilities/extract_component_from_tensors.h
```diff
@@ -0,0 +1,72 @@
...
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *  ______________________________________________________________________
+ */
+
+#ifndef INCLUDE_EXADG_UTILITIES_EXTRACT_COMPONENT_FROM_TENSORS_H_
+#define INCLUDE_EXADG_UTILITIES_EXTRACT_COMPONENT_FROM_TENSORS_H_
+
```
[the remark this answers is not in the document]
jh66637: I simply forgot. 
[8b39851](https://github.com/exadg/exadg/pull/255/commits/8b398515c3cac518c8731375d453c7fd5b20c41f)

include/exadg/postprocessor/pointwise_output_generator_base.h
jh66637: I will change all includes to the exadg convention :)

jh66637: [376c941](https://github.com/exadg/exadg/pull/255/commits/376c941dacec7e145658a1b4dabfd7d77604f9a0)

include/exadg/postprocessor/pointwise_output_generator_base.cpp
jh66637: [376c941](https://github.com/exadg/exadg/pull/255/commits/376c941dacec7e145658a1b4dabfd7d77604f9a0)

include/exadg/postprocessor/pointwise_output_generator_base.h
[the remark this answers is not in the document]
jh66637: The function is called here:
https://github.com/exadg/exadg/pull/255/files#diff-8377e4ef672706838605d3cd1455f00bd46206f876a578310af110b0b3989e3aR147-R149

include/exadg/postprocessor/pointwise_output_generator_base.h
jh66637: We need extract_component also in the scalar case since RPE gives a Tensor in which a scalar and a vector are combined.
Using first and last component would work

include/exadg/postprocessor/pointwise_output_generator_base.h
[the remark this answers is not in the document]
jh66637: OK :)

include/exadg/postprocessor/pointwise_output_generator_base.h
jh66637: To be consistent this would also require:
```add_quantity_dataset(std::string const&name,unsigned int const n_components=1)```
instead of ```add_scalar_quantity_dataset``` ```and add_vectorial_dataset```
Do you agree on this?

Comment (nfehn) [reply include/exadg/postprocessor/pointwise_output_generator_base.h:103]: I am not sure why there is `n_components` as argument. I would like to have only a single function like `add_dataset(name, values, first_component, last_component)`.

Comment (nfehn) [reply include/exadg/postprocessor/pointwise_output_generator_base.h:103]: Ah, I misunderstood it. You are right, here `n_components` is correct.

Comment (jh66637) [reply include/exadg/postprocessor/pointwise_output_generator_base.h]: Slightly different though. Hope you like it:
[2b20809](https://github.com/exadg/exadg/pull/255/commits/2b20809683371a883b4520bdc131134068aae0fd)

Full disclosure: I didn't test it. Before merging, I would like to ensure I didn't break anything by a stupid mistake.

File: include/exadg/postprocessor/pointwise_output_generator_base.cpp [hunk 18]
```diff
@@ -164,26 +164,25 @@ PointwiseOutputGeneratorBase<dim, Number>::setup_base(
 
 template<int dim, typename Number>
 void
-PointwiseOutputGeneratorBase<dim, Number>::add_scalar_quantity_dataset(std::string const & name)
+PointwiseOutputGeneratorBase<dim, Number>::add_quantity(std::string const & name,
+                                                        unsigned int const  n_components = 1)
```
nfehn: I think I would remove the default argument here (just following the standard C++ recommendation to avoid default arguments).

jh66637: [c697ee0](https://github.com/exadg/exadg/pull/255/commits/c697ee086e8bab1c98563b4d9219bf911538c4f9)

PR #256, opened 2022-08-23
Transfer solutions between arbitrary triangulations

jh66637:
With this class, it is possible to transfer solutions between a source and a target triangulation. The triangulations do not have to depend on each other, as it is the case in deal.IIs SolutionTransfer. 
The class uses RemotePointEvaluation and VectorTools::point_values to evaluate solutions in the support points of the elements of the target mesh. Eventually, obtained values are set in the dof vector corresponding to the destination triangulation.

### DIFF
--- a/include/exadg/vector_tools/solution_interpolation_between_triangulations.h
+++ b/include/exadg/vector_tools/solution_interpolation_between_triangulations.h
@@ -1,0 +1,165 @@
+/*  ______________________________________________________________________
+ *
+ *  ExaDG - High-Order Discontinuous Galerkin for the Exa-Scale
+ *
+ *  Copyright (C) 2021 by the ExaDG authors
+ *
+ *  This program is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *  ______________________________________________________________________
+ */
+
+#ifndef INCLUDE_VECTOR_TOOLS_SOLUTION_INTERPOLATION_BETWEEN_TRIANGULATIONS_H
+#define INCLUDE_VECTOR_TOOLS_SOLUTION_INTERPOLATION_BETWEEN_TRIANGULATIONS_H
+
+#include <deal.II/base/mpi_remote_point_evaluation.h>
+#include <deal.II/dofs/dof_handler.h>
+#include <deal.II/fe/mapping.h>
+#include <deal.II/numerics/vector_tools.h>
+
+namespace ExaDG
+{
+/**
+ * Class to transfer solutions between different DoFHandlers via interpolation. This class requires,
+ * that the destination DofHandler has generalized support points.
+ */
+template<int dim>
+class SolutionInterpolationBetweenTriangulations
+{
+public:
+  /**
+   * Set source and target triangulations on which solution transfer is performed.
+   *
+   * @param[in] dof_handler_dst_in Target DofHandler.
+   * @param[in] mapping_dst_in Target Mapping.
+   * @param[in] dof_handler_src_in Source DofHandler.
+   * @param[in] mapping_src_in Source Mapping.
+   */
+  void
+  reinit(dealii::DoFHandler<dim> const & dof_handler_dst_in,
+         dealii::Mapping<dim> const &    mapping_dst_in,
+         dealii::DoFHandler<dim> const & dof_handler_src_in,
+         dealii::Mapping<dim> const &    mapping_src_in)
+  {
+    AssertThrow(
+      dof_handler_dst_in.get_fe().has_generalized_support_points(),
+      dealii::ExcMessage(
+        "Solution can only be interpolated to finite elements that have support points."));
+
+    dof_handler_dst = &dof_handler_dst_in;
+    dof_handler_src = &dof_handler_src_in;
+
+    rpe.reinit(collect_mapped_support_points(dof_handler_dst_in, mapping_dst_in),
+               dof_handler_src_in.get_triangulation(),
+               mapping_src_in);
+  }
+
+  /**
+   * Interpolate the solution from a source to a target triangulation, that were set with reinit.
+   * At support points which are not overlapping the source triangulation, corresponding dof entries
+   * are set to 0.
+   *
+   * @param[in] dst Target Dof Vector.
+   * @param[in] src Source Dof Vector.
+   */
+  template<int n_components, typename VectorType1, typename VectorType2>
+  void
+  interpolate_solution(VectorType1 &                                               dst,
+                       VectorType2 const &                                         src,
+                       dealii::VectorTools::EvaluationFlags::EvaluationFlags const flags =
+                         dealii::VectorTools::EvaluationFlags::avg) const
+  {
+    AssertThrow(src.size() == dof_handler_src->n_dofs(),
+                dealii::ExcMessage("Dimensions do not fit."));
+    AssertThrow(dst.size() == dof_handler_dst->n_dofs(),
+                dealii::ExcMessage("Dimensions do not fit."));
+
+    src.update_ghost_values();
+    auto const values =
+      dealii::VectorTools::point_values<n_components>(rpe, *dof_handler_src, src, flags);
+    src.zero_out_ghost_values();
+
+    fill_dof_vector_with_values<n_components>(dst, *dof_handler_dst, values);
+  }
+
+private:
+  std::vector<dealii::Point<dim>>
+  collect_mapped_support_points(dealii::DoFHandler<dim> const & dof_handler,
+                                dealii::Mapping<dim> const &    mapping) const
+  {
+    std::vector<dealii::Point<dim>> support_points;
+
+    for(auto const & cell : dof_handler.active_cell_iterators())
+    {
+      if(cell->is_locally_owned() == false)
+        continue;
+
+      auto cellwise_support_points = cell->get_fe().get_generalized_support_points();
+
+      for(auto & csp : cellwise_support_points)
+        csp = mapping.transform_unit_to_real_cell(cell, csp);
+
+      support_points.insert(support_points.end(),
+                            cellwise_support_points.begin(),
+                            cellwise_support_points.end());
+    }
+
+    return support_points;
+  }
+
+  template<int n_components, typename VectorType, typename value_type>
+  void
+  fill_dof_vector_with_values(VectorType &                    dst,
+                              dealii::DoFHandler<dim> const & dof_handler,
+                              std::vector<value_type> const & values) const
+  {
+    auto ptr = values.begin();
+    for(auto const & cell : dof_handler.active_cell_iterators())
+    {
+      if(cell->is_locally_owned() == false)
+        continue;
+
+      auto const & fe = cell->get_fe();
+
+      std::vector<double> dof_values(fe.n_dofs_per_cell());
+      unsigned int const  n_support_points = fe.get_generalized_support_points().size();
+      std::vector<dealii::Vector<double>> component_dof_values(
+        n_support_points, dealii::Vector<double>(n_components));
+
+      for(unsigned int i = 0; i < n_support_points; ++i)
+      {
+        if constexpr(n_components == 1)
+          component_dof_values[i] = *ptr;
+        else
+          component_dof_values[i] =
+            std::move(dealii::Vector<double>(ptr->begin_raw(), ptr->end_raw()));
+
+        ++ptr;
+      }
+
+      fe.convert_generalized_support_point_values_to_dof_values(component_dof_values, dof_values);
+
+      cell->set_dof_values(dealii::Vector<typename VectorType::value_type>(dof_values.begin(),
+                                                                           dof_values.end()),
+                           dst);
+    }
+  }
+
+  dealii::SmartPointer<dealii::DoFHandler<dim> const> dof_handler_dst;
+  dealii::SmartPointer<dealii::DoFHandler<dim> const> dof_handler_src;
+  dealii::Utilities::MPI::RemotePointEvaluation<dim>  rpe;
+};
+
+} // namespace ExaDG
+
+#endif /* INCLUDE_VECTOR_TOOLS_SOLUTION_INTERPOLATION_BETWEEN_TRIANGULATIONS_H */


include/exadg/vector_tools/arbitrary_solution_transfer.h
```diff
@@ -0,0 +1,167 @@
...
+   */
+  template<int n_components, typename VectorType1, typename VectorType2>
+  void
+  transfer_solution(VectorType1 &                                               dst,
+                    VectorType2 const &                                         src,
+                    dealii::VectorTools::EvaluationFlags::EvaluationFlags const flags =
+                      dealii::VectorTools::EvaluationFlags::avg) const
+  {
+    src.update_ghost_values();
+    const auto values =
```
nfehn, line 74: should be `auto const` in ExaDG

include/exadg/vector_tools/arbitrary_solution_transfer.h
```diff
@@ -0,0 +1,167 @@
...
+    fill_dof_vector_with_values<n_components>(dst, *dof_handler_dst, values);
+  }
+
+private:
+  std::vector<dealii::Point<dim>>
+  collect_mapped_support_points(dealii::DoFHandler<dim> const & dof_handler,
+                                dealii::Mapping<dim> const &    mapping) const
+  {
+    std::vector<dealii::Point<dim>> global_mapped_support_points;
+    std::vector<dealii::Point<dim>> mapped_support_points;
```
nfehn, line 87: I think the term "nodes" would be more appropriate here? Then, one could distinguish between "global_nodes" and "elementwise/cellwise_nodes".

include/exadg/vector_tools/arbitrary_solution_transfer.h
```diff
@@ -0,0 +1,167 @@
...
+
+    for(auto const & cell : dof_handler.active_cell_iterators())
+    {
+      if(cell->is_locally_owned() == false)
+        continue;
+
+      mapped_support_points.resize(cell->get_fe().get_generalized_support_points().size());
```
nfehn, line 94: The call `cell->get_fe().get_generalized_support_points()` appears twice, so one could introduce a variable, e.g. `nodes_on_unit_cell` (if I understand the code correctly), which is then reused below.

include/exadg/vector_tools/arbitrary_solution_transfer.h
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  template<int n_components, typename VectorType, typename ValueType>
+  void
+  fill_dof_vector_with_values(VectorType &                    dst,
+                              dealii::DoFHandler<dim> const & dof_handler,
+                              std::vector<ValueType> const &  values) const
+  {
+    auto ptr = values.begin();
+    for(const auto & cell : dof_handler.active_cell_iterators())
```
nfehn, line 115: auto const

include/exadg/vector_tools/arbitrary_solution_transfer.h
```diff
@@ -0,0 +1,167 @@
...
+    {
+      if(cell->is_locally_owned() == false)
+        continue;
+
+      unsigned int n_dofs_per_component = cell->get_fe().get_generalized_support_points().size();
```
nfehn, line 120: unsigned int const?

nfehn: `n_nodes` could be a more intuitive name

include/exadg/vector_tools/arbitrary_solution_transfer.h
```diff
@@ -0,0 +1,167 @@
...
+ *  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *  ______________________________________________________________________
+ */
+
+#ifndef INCLUDE_VECTOR_TOOLS_ARBITRARY_SOLUTION_TRANSFER_H
+#define INCLUDE_VECTOR_TOOLS_ARBITRARY_SOLUTION_TRANSFER_H
+
+#include <deal.II/dofs/dof_handler.h>
+#include <deal.II/fe/mapping.h>
+#include <deal.II/matrix_free/fe_point_evaluation.h>
```
nfehn, line 27: why do you include fe_point_evaluation and not mpi_remote_point_evaluation?

include/exadg/vector_tools/arbitrary_solution_transfer.h
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  template<typename Number, typename InputIterator>
+  dealii::Vector<Number>
+  multiple_components_to_dof_values(dealii::FiniteElement<dim> const & fe,
+                                    unsigned int const                 n_dofs_per_comp,
+                                    InputIterator const                first) const
+  {
+    std::vector<dealii::Vector<double>> vec;
+    vec.reserve(fe.n_dofs_per_cell());
```
nfehn, line 145: why n_dofs_per_cell and not n_dofs_per_component?

include/exadg/vector_tools/arbitrary_solution_transfer.h
jh66637: What do you think about "support_points" and "cellwise_support_points"? I would like to reflect where points are evaluated in the name, and nodes is in my opinion not as clear.

include/exadg/vector_tools/arbitrary_solution_transfer.h
jh66637: You are right. Introducing the variable locally makes the code more readable. Due to the change, I restructured the for loop a bit.

include/exadg/vector_tools/arbitrary_solution_transfer.h
jh66637: The naming should be chosen as above. So if we go with "nodes" I am fine with this, if we go with "support_points" I will change it to n_support_points

include/exadg/vector_tools/arbitrary_solution_transfer.h
jh66637: It is possible that there are multiple components at the same support point. Using generalized support points gives us a unique representation of the used support points. However, we might evaluate multiple components at the same support point and thus have to write each component to the dof vector, i.e. n_dofs_per_cell.

jh66637:
@nfehn: I updated everything as discussed above. For the naming, I went with support_points instead of nodes, but we can further discuss this.

include/exadg/vector_tools/arbitrary_solution_transfer.h
nfehn: Not sure if I understood it correctly. My question is why the loop below is over `n_dofs_per_comp` while you reserve for `n_dofs_per_cell`?

include/exadg/vector_tools/arbitrary_solution_transfer.h
```diff
@@ -0,0 +1,164 @@
...
+                             dst);
+      }
+
+      ptr += n_support_points;
+    }
+  }
+
+  template<typename Number, typename InputIterator>
+  dealii::Vector<Number>
+  multiple_components_to_dof_values(dealii::FiniteElement<dim> const & fe,
```
nfehn, line 138: I am undecided whether it makes sense to introduce this function. As a stand-alone function, it is quite hard to read in terms of `InputIterator` in my opinion. I think I would shift the content of this function to the else-branch above (where also the incrementing of the pointer is done very naturally).

include/exadg/vector_tools/arbitrary_solution_transfer.h
```diff
@@ -0,0 +1,164 @@
...
+          dealii::Vector<typename VectorType::value_type>(ptr, ptr + n_support_points), dst);
+      }
+      else
+      {
+        cell->set_dof_values(multiple_components_to_dof_values<typename VectorType::value_type>(
+                               cell->get_fe(), n_support_points, ptr),
+                             dst);
+      }
+
+      ptr += n_support_points;
```
nfehn, line 132: According to comment https://github.com/exadg/exadg/pull/256/files#r960506978, this line would have to be shifted to the if-branch.

include/exadg/vector_tools/arbitrary_solution_transfer.h
```diff
@@ -0,0 +1,164 @@
...
+      {
+        cell->set_dof_values(
+          dealii::Vector<typename VectorType::value_type>(ptr, ptr + n_support_points), dst);
+      }
+      else
+      {
+        cell->set_dof_values(multiple_components_to_dof_values<typename VectorType::value_type>(
+                               cell->get_fe(), n_support_points, ptr),
+                             dst);
+      }
```
nfehn, line 130: It looks like the same deal.II function (`set_dof_values()`) is called in both cases? If so, I would prefer that the vector `dof_values` is filled depending on the parameter `n_components`, and that the function call appears only once after the vector with `dof_values` has been filled.

include/exadg/vector_tools/arbitrary_solution_transfer.h
```diff
@@ -0,0 +1,164 @@
...
+
+      support_points.insert(support_points.end(),
+                            cellwise_support_points.begin(),
+                            cellwise_support_points.end());
+    }
+
+    return support_points;
+  }
+
+  template<int n_components, typename VectorType, typename ValueType>
```
nfehn, line 106: for consistency with ExaDG, we should write `value_type` here. Is this template parameter really necessary, i.e. does this function internally perform float<->double conversions?

include/exadg/vector_tools/arbitrary_solution_transfer.h
```diff
@@ -0,0 +1,164 @@
...
+
+#include <deal.II/base/mpi_remote_point_evaluation.h>
+#include <deal.II/dofs/dof_handler.h>
+#include <deal.II/fe/mapping.h>
+#include <deal.II/numerics/vector_tools.h>
+
+namespace ExaDG
+{
+template<int dim>
+class ArbitrarySolutionTransfer
```
nfehn, line 33: My last question for now concerns the name of the class. `Arbitrary` might be too general or not specific enough. The name should reflect that the solution is transferred between different grids/triangulations. Hence, I would suggest something like `SolutionTransferBetweenGrids`. 

For clarity, the documentation should mention that the implementation assumes a nodal basis on the dst-side, i.e. the solution evaluated at certain points in space (currently called support points) equals entries of the DoF-vector. The src-side appears to be flexible in terms of nodal vs. modal basis functions (all we use is RPE, which interpolates the src-solution correctly into the dst-points).

include/exadg/vector_tools/arbitrary_solution_transfer.h
nfehn: I can live with support points as well.

include/exadg/vector_tools/arbitrary_solution_transfer.h
nfehn: agree

nfehn:
Is the implementation restricted to certain types of finite elements, i.e. does the implementation assume an "isotropic" element in case of vectorial problems, where the different components all have the same number and the same location of the support points? More concretely, I am thinking about "anistropic" Hdiv-conforming spaces with Raviart-Thomas elements which we currently introduce in ExaDG. If the implementation has such a restriction, we should make sure that we insert appropriate Asserts in the code. @kronbichler @peterrum I would be happy if you could briefly comment on this aspect.

include/exadg/vector_tools/arbitrary_solution_transfer.h
nfehn: `SolutionTransferBetweenTriangulations` (as you write in the title) would also be good in my opinion

include/exadg/vector_tools/arbitrary_solution_transfer.h
```diff
@@ -0,0 +1,164 @@
...
+                    VectorType2 const &                                         src,
+                    dealii::VectorTools::EvaluationFlags::EvaluationFlags const flags =
+                      dealii::VectorTools::EvaluationFlags::avg) const
+  {
+    src.update_ghost_values();
+    auto const values =
+      dealii::VectorTools::point_values<n_components>(rpe, *dof_handler_src, src, flags);
+    src.zero_out_ghost_values();
+
+    fill_dof_vector_with_values<n_components>(dst, *dof_handler_dst, values);
```
nfehn, line 78: I thought I have written another comment, but I can not find it anymore ... so I write it again: For me, it looks like the two functions `collect_mapped_support_points()` and `fill_dof_vector_with_values()` assume that we go through the mesh/cells in the same order. Is this implementation safe enough? Put differently: Who takes care of the correct indexing so that it is guaranteed that the `values` returned by `VectorTools::point_values()` are written correctly to the dof-vector? It looks like we are just looping through the vector from begin to end, so there must be some implicit convention behind, right? Shouldn't we explicitly use maps with cell_ids/dof_indices as we use in `InterfaceCoupling`?

jh66637:
@nfehn I think with the latest update I clarified all points you mentioned above (besides the last point). 
Since we collect the support points in the same order, we write to the dof vector, I don't think it is necessary to use maps here, but we can further discuss this.

include/exadg/vector_tools/arbitrary_solution_transfer.h
kronbichler: The convention of `RemotePointEvaluation` and `VectorTools::point_values` is that the order of the returned values be the same as the one in points passed in. I did look at the code and it seems that this is fulfilled in terms of the order of loops. I think the usage should be fine. Clearly, there is still some small ambiguity in terms of the loop done in the two functions `collect_mapped_support_points` and `fill_dof_vector_with_values`, but I do not see an immediate benefit of the functions we have in `InterfaceCoupling`. The argument is that lookup via `std::map` is (probably) rather slow for a volume quantity, and we still rely on the convention that we look up points previously inserted, so I would primarily see this as beneficial during code development, not for catching run-time errors.

kronbichler:
> Is the implementation restricted to certain types of finite elements, i.e. does the implementation assume an "isotropic" element in case of vectorial problems, where the different components all have the same number and the same location of the support points? More concretely, I am thinking about "anistropic" Hdiv-conforming spaces with Raviart-Thomas elements which we currently introduce in ExaDG. If the implementation has such a restriction, we should make sure that we insert appropriate Asserts in the code. @kronbichler @peterrum I would be happy if you could briefly comment on this aspect.

I have no deep insight into the functions dealing with generalized support points, so I can only offer a partial explanation without carefully looking into the code. On the one hand, I agree with the assessment that Raviart-Thomas with the underlying Piola transform might be problematic. The reason is that the place where I've recently convinced myself that things work in a general way, https://github.com/dealii/dealii/blob/982a1c21ac600160a5be1f49a87fdc564c9d10a9/include/deal.II/numerics/vector_tools_interpolate.templates.h#L365-L385 , does not only pick up the solution values, but also applies a `transform` method that makes sure possible transformations in the solution space are taken into account. So this might be a place to check.

Regarding the different numbers of points in different components, my understanding is that the the `convert_generalized_support_point_values_to_dof_values` would take care of it, possibly picking up/leaving out certain components on certain points, or doing a pseudo inverse, I do not recall the details.

My final concern with the function in `FiniteElement` is that it does not utilize the tensor product in shape functions, so it could get slow for higher degrees in 3D. I am not sure if that would mean we need to fix the transfer by our own implementation here in ExaDG (say a very generic 'project' function), or if we should instead attempt to improve the functions in deal.II.

include/exadg/vector_tools/arbitrary_solution_transfer.h
nfehn: ok, so let's use the current approach. I have one short question, though: There is probably no generic deal.II functionality to loop only over the locally owned cells so that we do not have to write the if-statement with `continue` in the code here? I wonder how often these lines of code are written here and there.

include/exadg/vector_tools/arbitrary_solution_transfer.h
nfehn: > For clarity, the documentation should mention that the implementation assumes a nodal basis on the dst-side, i.e. the solution evaluated at certain points in space (currently called support points) equals entries of the DoF-vector. The src-side appears to be flexible in terms of nodal vs. modal basis functions (all we use is RPE, which interpolates the src-solution correctly into the dst-points).

If you agree, you could somewhere place this information in the documentation for clarity.

nfehn:
@kronbichler, regarding the comment

> My final concern with the function in FiniteElement is that it does not utilize the tensor product in shape functions, so it could get slow for higher degrees in 3D. I am not sure if that would mean we need to fix the transfer by our own implementation here in ExaDG (say a very generic 'project' function), or if we should instead attempt to improve the functions in deal.II.

I did not understand which functionality of `FiniteElement` you are referring to? Or do you mean functionality hidden by `RemotePointEvaluation` or `point_values()`?

nfehn:
Overall,  I think you did a very good job @jh66637. Many thanks!

nfehn (approved):

include/exadg/vector_tools/arbitrary_solution_transfer.h
jh66637: As far as I understood the concept of support points, only nodal bases have them. Am I right?

include/exadg/vector_tools/arbitrary_solution_transfer.h
peterrum: @nfehn You can use filter, like here: https://github.com/dealii/dealii/blob/dd99ecfe16df9287195726e4444bb448a5b78fe5/examples/step-85/step-85.cc#L363-L365 (personally, I don't like).

kronbichler: Indeed, I also personally prefer writing `if` statements over these iterator filters, but I am not sure if it more a habit or real preference. A more direct use (from `mg_global_coarsening.templates.h`) is:
```
      for (const auto &cell : dof_handler_fine.active_cell_iterators() |
                                IteratorFilters::LocallyOwnedCell())
        is_locally_owned_fine.add_index(cell_id_translator.translate(cell));
```

kronbichler:
> I did not understand which functionality of FiniteElement you are referring to? Or do you mean functionality hidden by RemotePointEvaluation or point_values()?

I was thinking about `FiniteElement::convert_generalized_support_point_values_to_dof_values` in the assumption that it builds a dense matrix. However, that is not the case, it simply picks indices for our nodal RT element at least: https://github.com/dealii/dealii/blob/dd99ecfe16df9287195726e4444bb448a5b78fe5/source/fe/fe_raviart_thomas_nodal.cc#L254-L282
However, other functions do involve quadratic complexity. I think for the present use case the approach by @jh66637 is fine, as long as we manage to get the interpolation/Piola transform in. The alternative would be to use our infrastructure and do a projection instead, i.e., rather than directly injecting the values at interpolation points, we inject the values to quadrature points, and then we invert a mass matrix.

include/exadg/vector_tools/arbitrary_solution_transfer.h
kronbichler: Support points are indeed only there for nodal bases. However, other element types have something called "generalized support points", which are some representative points that can be used to convert from values at these points to the values of the weights in the polynomial expansion, be it a modal basis or some mixture (like "moments on the edges"). You can see what is done e.g. in the non-nodal version of the Raviart-Thomas finite element: https://github.com/dealii/dealii/blob/dd99ecfe16df9287195726e4444bb448a5b78fe5/source/fe/fe_raviart_thomas.cc#L679-L707
What you can see here is that there is a "weights" matrix that contains the relevant expansions. For a modal basis (and ignoring the unknowns on faces for Raviart-Thomas at the moment), you can think of this as being the inverse of a Vandermonde matrix.

jh66637:
> The alternative would be to use our infrastructure and do a projection instead, i.e., rather than directly injecting the values at interpolation points, we inject the values to quadrature points, and then we invert a mass matrix.

I think this should be the way to go. This way, the class can be used with any FE_*. I will give it a shot when I find the time.

include/exadg/vector_tools/arbitrary_solution_transfer.h
jh66637: Thanks for clarification :)

nfehn: Ah, so I think I was not aware of "generalized". Against the background provided by @kronbichler , I am now unsure whether the class is really restricted to nodal bases on the dst-side? It looks like we are only working with generalized support points e.g. when collecting the points, and finally we convert the generalized support point values to dof values. If I understand the comment by @kronbichler correctly, the class would/could also work with a modal basis on the dst-side (potentially involving a dense interpolation step with quadratic complexity if this is required by a certain non-nodal basis)?

nfehn:
> The alternative would be to use our infrastructure and do a projection instead, i.e., rather than directly injecting the values at interpolation points, we inject the values to quadrature points, and then we invert a mass matrix.

This is very important functionality that we also need for multi-physics coupling in general. I think I would not insert this functionality into the present class. The present class is independent of `MatrixFree` or `VectorizedArray`, which is a good design. In addition to the present class, we should introduce another class that interpolates into the quadrature points of the dst-side, in analogy to `InterfaceCoupling`, but now for volume-coupled problems (or the transfer between two triangulations). This class then provides very general functionality that can be used to evaluate weak forms for coupled problems, or just do a projection (as described above).

include/exadg/vector_tools/arbitrary_solution_transfer.h
jh66637: Yes it could work, but not with the current implementation. I have to adapt something for this. Should not take me long.

jh66637:
So we would have different classes aiming to achieve the same result?

nfehn:
I would say the result is different. In one case a dof-vector is filled, in the other case a vector of q-point data, or "interpolation" vs. "projection". I believe for the projection case one needs much more than we currently have in the class so that I believe it is not justified to include it in this class. This is no problem because the present class is based on interpolation, so I do not expect at all that it also includes the case of projection, which is numerically something different (see e.g. `VectorTools::interpolate` and `VectorTools::project`; one solution would be to rename the present class `Transfer...` into `Interpolate...` to be more precise, which could be done already now or later). My point is that filling a q-point vector is functionality more general than the projection problem. I see the class filling the q-point vector as a sub-problem of the projection problem. Once we have the q-point vector, no knowledge is required about a second triangulation (the src-triangulation) and the dof-vectors living on that triangulation to compute the projection problem. So my goal is to minimize dependencies and to not implement functionality twice. But, of course, in the end we always need to try it out and have a look at concrete implementations.

jh66637:
I see. What is left to do?

nfehn:
For this MR, I think everything has been done.

PR #261, opened 2022-10-12
manage arbitrary multistep coeffs instead of BDF coeffs only

jh66637:
I am aiming to change ```TimeIntBDFBase```. 

My implementations are motivated as follows:

* To introduce Adams-Bashforth and Adams-Moulton time stepping, the same functionality is needed that is currently in ```TimeIntBDFBase```

This is why I moved the coefficients from the base class to child classes. This way, the base class can also be used for Adams methods without introducing further coefficients in the base class.

@nfehn As requested, here is the current inheritance diagram. I also sketched an alternative, which might resolve the comments of the current review.

image


jh66637:
@nfehn @peterrum Could someone review these changes? I would like to add an acoustic module to ExaDG that uses Adams-Bashforth-Moulton for time integration, thus requires this to be merged.

nfehn:
I can look at this PR

nfehn:
In the description of this PR, could you sketch the inheritance diagram that you plan to realize? 

To which extent are base classes required for BDF, Adams-Bashfort, Adams-Moulton or not?

include/exadg/time_integration/time_int_multistep_base.h
```diff
@@ -176,20 +169,20 @@ class TimeIntBDFBase : public TimeIntBase
...
-  initialize_current_solution() = 0;
+  initialize_multistep_dof_vectors() = 0;
 
   /*
-   * Initializes the solution vectors at time t - dt[1], t - dt[1] - dt[2], etc.
+   * Initializes the initial solution
    */
   virtual void
-  initialize_former_solutions() = 0;
+  initialize_current_solution() = 0;
```
nfehn, line 185: Please undo the change regarding switching the order of function definitions in this PR.

include/exadg/convection_diffusion/time_integration/time_int_bdf.h
```diff
@@ -64,10 +66,13 @@ class TimeIntBDF : public TimeIntBDFBase<Number>
...
   void
-  ale_update();
+  ale_update() final;
 
   void
-  print_iterations() const;
+  print_iterations() const final;
+
+  double
+  get_scaling_factor_time_derivative_term() const;
```
nfehn, line 75: calls for a separate BDF base class.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.cpp
```diff
@@ -376,6 +391,13 @@ TimeIntBDF<dim, Number>::print_solver_info() const
                                       this->time_step_number);
 }
 
+template<int dim, typename Number>
+double
+TimeIntBDF<dim, Number>::get_scaling_factor_time_derivative_term() const
+{
+  return bdf.get_gamma0() / this->get_time_step_size();
+}
```
nfehn, line 399: same comment as here https://github.com/exadg/exadg/pull/261/files#r1135864108

include/exadg/time_integration/time_int_multistep_base.h
```diff
@@ -176,20 +169,20 @@ class TimeIntBDFBase : public TimeIntBase
...
+  void
+  initialize_vectors_and_time_step_size(bool do_restart);
 
   /*
-   * Initializes the solution vectors at time t
+   * Initializes vectors needed for the multistep scheme
    */
   virtual void
-  initialize_current_solution() = 0;
+  initialize_multistep_dof_vectors() = 0;
```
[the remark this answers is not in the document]
jh66637: The problem here is, that in Adams methods, we do not need to keep track of former solutions. Instead, we have to keep track of evaluated operators. I still think all multistep methods should be derived from the same base class, since the general principle of those methods are the same.

nfehn:
Can you explain why there are currently two separate functions `initialize_{current|former}_solution()` for BDF (e.g. in the description of this PR)? Could one solution to this problem be that we unify both functions in a single function instead of renaming `initialize_former_solutions()`? What are the pros and cons of this decision, or does this make no sense at all? Are there maybe further options that are easy to understand and that lead to a good inheritance logic?

jh66637:
Both functions have been there before. One is only called if a high order start is required. Having a function that is always called (```init_current_sol()```) and a function that is only called in the case of high order start (```init_former_sol()```) should be kept seperate in my opinion. Since there is no such thing as former solutions in the case of Adams methods, this could be moved to an itermediate class ```BDFBase```. In the case of Adams method we would have something like ```init_former_evaluated_operators()``` instead.

nfehn:
Thanks for the diagram @jh66637.

I see some conflicts in design for both variants:

- current variant: certain BDF functionality is re-implemented for all modules using this type of time integration (ConvDiff, IncNS). see comment https://github.com/exadg/exadg/pull/261#discussion_r1135864108
- alternative variant: The alternative variant does not resolve this since it stores all coefficients (bdf, extra, am) in the base class for multistep methods. Then, we no longer have true "is-a" relationships.

Hence, I believe it needs a base class for BDF as in the current master branch. Similarily, I believe it needs a base class for AB, AM, so that more than one module can use this type of time integration without reimplementing content. If there is overlap between `TimeIntBDFBase` and `TimeIntAdamsBase`, it makes sense to introduce a base class `TimeIntMultistepBase`. However, this base class may only contain elements that are relevant for both types of time integration schemes in the sense of true "is-a" relationships. An alternative to avoid `TimeIntBDFBase` and `TimeIntAdamsBase` could be to make use of composition rather than inheritance (e.g. by defining data structures that own the time integration constants, can return the scaling factor, can update the time integration constants, etc. (so exactly the functionality that we have in this PR in two modules (ConvDiff, IncNS) currently). I did not think about all the implications of such a design, though.

include/exadg/time_integration/time_int_multistep_base.h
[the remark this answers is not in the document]
nfehn: With the explanations you gave in https://github.com/exadg/exadg/pull/261#issuecomment-1468530690, I would be happy if we could insert `former` and use `initialize_former_multistep_dof_vectors()`.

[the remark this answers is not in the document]
nfehn: Regarding the inheritance discussion, I think it could make sense to have both functions in the base class for multistep methods, supposing that we find intuitive names.

jh66637:
Regarding the second bullet point: This is what is done in the current ```BDFBase``` class already. ```extra``` is not necessarily used, but it lives in the base class and is updated every time. So in my opinion, the crucial question is where to store the coefficients. We could move them to intermediate base classes as you suggest, if I understand correctly. But this means that in the intermediate class ```TimeIntAdamsBase``` we have ```bdf```, ```am``` and ```ab```. This is because it is still necessary to keep track of ```bdf``` constants to compute mesh velocities. (In the BDF case we would still have ```extra``` and ```bdf``` in any case). 

I think introducing something like ```MultistepCoeffManager``` which lives e.g. in ```IncNS::BDF``` would just shift these problems into another class.

Instead, we might want to think about moving the computation of the mesh velocity (and anything related) into an additional class instead (everything except setting the velocity to the corresponding operator is the same for all cases I can think of). This class should live in ```MultistepBase```. Then using intermediate base classes would be a solution (except of the remaining issue with ```extra```). But this means, we somehow have to manage ```bdf``` twice.

nfehn:
> But this means that in the intermediate class `TimeIntAdamsBase` we have `bdf`, `am` and `ab`.

If `bdf` is needed for ALE in case of Adams-type time integration, I see a reason to introduce these coefficients in `TimeIntAdamsBase`. My point here is to make sure that inheritance as a design concept is a reasonable tool for the functionality that we want to realize, not so much whether `TimeIntAdamsBase` is an optimal name (I would consider e.g. `TimeIntAdamsWithBDFForALEMeshVelocityBase` inferior, because of the length of this name, even though one might argue it described the numerical method more precisely). Having `am` and `ab` in `TimeIntAdamsBase` is in my opinion still better than introducing `am` and `ab` into `TimeIntBDF`. I do not see a reasonable argument for the latter.

> This is what is done in the current `BDFBase` class already. `extra` is not necessarily used.

I don't agree here. All derived BDF classes make use of `extra` (potentially depending on the parameters). Hence, it makes sense to have this in the base class and use inheritance. I see that one could argue whether the name `TimeIntBDFBase` is then still optimal. But this is just a matter of having reasonable short names, consider `TimeIntBDFWithExtrapolationForExplicitTermsBase` as an alternative. Coefficients `am` and `ab` will not be used for BDF-type time integration schemes, at least not for the schemes that we have currently in ExaDG. Hence, such schemes must not derive from base classes that own these coefficients according to my understanding of how to use inheritance.

> Then using intermediate base classes would be a solution (except of the remaining issue with extra). But this means, we somehow have to manage bdf twice.

I consider the problem of treating `bdf` twice in two different derived classes of the same type of BDF time integration (ConvDiff and IncNS, which is the case currently) more detrimental than once in `TimeIntBDFBase` and once in `TimeIntAdamsBase`. As you describe, the problem with ALE and somewhat redundant code could potentially be improved in the future. However, I remember that we struggled a lot in the past with realizing the ALE content outside the time integrator (when we started with ALE for incompressible flows). I am definitely open to new attempts in this direction, but I believe rather strongly that this should not be done within the present PR. I think the present PR should find the best realizable solution under the constraint that we want to support ALE (which is not explicitly part of the names for the time integration classes, but considered an additional "feature" of these time integration classes).

nfehn:
I think we need a design that you describe as "alternative" in the description of this PR, but with two modifications:
-  `bdf, extra, am, ab` coefficients shifted to the respective base classes (only those coefficients needed for a particular time stepping scheme). If the scheme supports ALE with need for BDF coefficients, we can add those coefficients as well.
-  a common function `initialize_former_multistep_dof_vectors()` in the multistep base class (in order to avoid code duplication and make use of the current interface of the current `BDFBase` class, that seems to work for Adams-type methods as well.

include/exadg/time_integration/adams_moulton_time_integration.h
```diff
@@ -0,0 +1,93 @@
...
+                                      bool const         start_with_low_order_method);
+
+  double
+  get_gamma0() const;
+
+  double
+  get_alpha(unsigned int const i) const;
+
+  /*
+   *  This function updates the time integrator constants of the AB scheme
```
nfehn, line 46: AB->AM (check also all other occurrences in this class)

include/exadg/time_integration/time_int_adams_bashforth_base.cpp
```diff
@@ -0,0 +1,63 @@
...
+  if(this->adaptive_time_stepping == false) // constant time steps
+  {
+    ab.update(this->time_step_number);
+    bdf.update(this->time_step_number);
+  }
+  else // adaptive time stepping
+  {
+    ab.update(this->time_step_number, this->time_steps);
+    bdf.update(this->time_step_number, this->time_steps);
+  }
```
nfehn, line 60: I think this if-else-logic should be implemented/shifted to the implementation of the time integration constants. It looks like we have the same code structure in every time integration scheme. This would be a natural way to avoid code duplication between BDF/AB/AM time integration schemes.

include/exadg/time_integration/time_int_adams_bashforth_moulton_base.cpp
```diff
@@ -0,0 +1,69 @@
...
+    ab.update(this->time_step_number - 1);
+    am.update(this->time_step_number);
+    bdf.update(this->time_step_number);
+  }
+  else
+  {
+    ab.update(this->time_step_number - 1, this->time_steps);
+    am.update(this->time_step_number, this->time_steps);
+    bdf.update(this->time_step_number);
+  }
```
nfehn, line 66: as above https://github.com/exadg/exadg/pull/261/files#r1146450962

include/exadg/time_integration/time_int_adams_moulton_base.cpp
```diff
@@ -0,0 +1,63 @@
...
+  if(this->adaptive_time_stepping == false) // constant time steps
+  {
+    am.update(this->time_step_number);
+    bdf.update(this->time_step_number);
+  }
+  else // adaptive time stepping
+  {
+    am.update(this->time_step_number, this->time_steps);
+    bdf.update(this->time_step_number, this->time_steps);
+  }
```
nfehn, line 60: as above https://github.com/exadg/exadg/pull/261/files#r1146450962

include/exadg/time_integration/time_int_bdf_base.cpp
```diff
@@ -18,360 +18,52 @@
...
+    bdf.update(this->time_step_number);
+    extra.update(this->time_step_number);
   }
   else // adaptive time stepping
   {
-    bdf.update(time_step_number, time_steps);
-    extra.update(time_step_number, time_steps);
+    bdf.update(this->time_step_number, this->time_steps);
+    extra.update(this->time_step_number, this->time_steps);
   }
```
nfehn, line 66: same as other comments https://github.com/exadg/exadg/pull/261/files#r1146450962

include/exadg/time_integration/time_int_bdf_base.cpp
```diff
@@ -18,360 +18,52 @@
...
-    extra(order_, start_with_low_order_),
-    start_with_low_order(start_with_low_order_),
-    adaptive_time_stepping(adaptive_time_stepping_),
-    time_steps(order_, -1.0)
-{
-}
-
-template<typename Number>
-void
-TimeIntBDFBase<Number>::setup(bool const do_restart)
```
nfehn, line 53: it would be good if all this code would not show up as code changes. Essentially, it looks like a renaming `TimeIntBDFBase` -> `TimeIntMultistepBase` took place. Then, some functionality has been extracted from `TimeIntMultistepBase` into a new class `TimeIntBDFBase`. It would be better if some names are non-sense for now and are changed in an upcoming PR for the sake of showing in this PR on the code that has been changed. I mean that `TimeIntBDFBase` means -- for the time being -- a base class for multistep time integration, and you introduce in this PR another class `TimeIntBDF : public TimeIntBDFBase`.

jh66637: Can we instead have one PR with two seperate commits?

include/exadg/time_integration/time_int_bdf_base.cpp
jh66637: [a9b700b](https://github.com/exadg/exadg/pull/261/commits/a9b700bed99547adf81d45ffa8e08a6b0d4b48bd)

nfehn:
As a general comment, it looks like the different classes for time integrator constants all have kind of the same structure. This would call for introducing inheritance also for the constants, `TimeIntegratorConstantsBase`. 

Overall, many thanks to @jh66637 for implementing Adams-Bashfort/Moulton time integrators! I think this has not been said before, sorry for that and that you had to wait quite some time.

I would be happy if you could arrange the code such that only the BDF-specific functions, which have been extracted from the base class, show up as code changes.

nfehn:
Let's stick to the arguments for now. Isn't the reason for the changed lines of code actually the new file `time_int_multistep_base.h` introduced in the first commit? I think you could even introduce all the final class names in the present PR and just use "old" file names, to be updated in the future (I do not really care about filenames). I might be wrong, though.

nfehn:
> If you want to, you can change it as a follow-up.

I consider this a bit unfair. Let me explain this in more detail. You have a point in the sense that one could argue it should have been me introducing the base class when I had implemented BDF and EXT coefficients some years ago. Probably, I did not receive any code review at that time, and probably I did not think that there is interest for other time integration routines with the same structure (maybe I did not even realize that Adams schemes can be implemented with this logic). So I can offer to introduce a base class now and that you rebase your code once this is done. I don't see why we should now introduce new classes so that more classes need to be touched again when introducing the base class. Another aspect is the following one: Until recently, I could not imagine how much time it takes to do code review for many other developers, working on vastly different subjects. I can only ask you for patience and that we are all willing to invest the time it needs. So if you do not want to introduce the inheritance here, your PR will probably get merged nevertheless, because it is an improvement for ExaDG and this is our rule for merging PRs or not (Is it a step forward for ExaDG?). However, this argument should not be used against code reviewers in my opinion. Yes, I would do  these changes by myself afterwards (if I find the time) because I believe such a refactoring improves again ExaDG. The downside is that this takes me some time. You are currently closer to the time integration content than me. This implies that if I do it by myself, this time will not be available for reviewing new code coming from other developers. I am willing to contribute "passively" as a reviewer or "actively" as a coder. My point is to find the optimum/compromise within the constraints I have explained above.

include/exadg/time_integration/extrapolation_scheme.cpp
```diff
@@ -165,35 +165,20 @@ ExtrapolationConstants::set_adaptive_time_step(unsigned int const          curre
...
-  {
-    set_adaptive_time_step(order, time_steps);
-  }
+  unsigned int update_order =
+    (current_order <= order && start_with_low_order == true) ? current_order : order;
+
+  if(adaptive)
+    set_adaptive_time_step(update_order, time_steps);
+  else
+    set_constant_time_step(update_order);
```
nfehn, line 181: I am against such refactoring changes in a PR such as the present one. You could argue that you are a 100% sure that the code does the same as before, and I might even agree on this. However, we had cases in the past where the developers and code reviewers were proven wrong and where the code was broken afterwards for some subtle reasons. I do not want to block the present PR for this particular reason. However, (and also in light of e.g. other PRs such as PR #381) I think we need to agree for the future that refactoring of the code should be subject of separate PRs and not appear within a big PR with 1,600 new lines of code. I have seen these 7 lines of code at least 4 times in your PR. Doesn't this call for a base class that implements exactly this logic? Ideally, and as written elsewhere, we would introduce this base class in a PR prior to the present PR and adapt BDF and EXT coefficients in a first step. Afterwards, the present PR would build upon this new design.

nfehn:
another suggestion to better manage the present PR is to introduce the new classes for the time integration constants based on the base class introduced in PR #395 (without any usage of these classes for now).

include/exadg/time_integration/extrapolation_scheme.cpp
jh66637: You asked me to refactor the code in several comments above.

jh66637:
> I consider this a bit unfair. Let me explain this in more detail. You have a point in the sense that one could argue it should have been me introducing the base class when I had implemented BDF and EXT coefficients some years ago.

My point is not that you did something wrong. My point is that the present PR is already lengthy, it works, has several new features, and adding yet another base class can also be done afterward. I can not stop you from merging https://github.com/exadg/exadg/pull/395 before this one, but I think I spent enough work on the present PR (and also refactored some parts of code upon request).

nfehn:
The decision whether we merge #395 first will be based on whether we prefer in terms of code development (i) to first remove code duplication in two instances and then introduce new features, or (ii) introduce duplicated code in two more instances before removing the duplicated code finally in four instances.

As a maintainer, it is hard for me to accept that you ask me to do the second step in procedure (ii) by myself (see comment https://github.com/exadg/exadg/pull/261#issuecomment-1481593158), but that you also don't appreciate if I contribute actively in the sense of procedure (i) ("I cannot stop you from ...), which I consider personally more goal-oriented overall for ExaDG.

PR #262, opened 2022-10-13
Introduce TimeControl

jh66637:
As discussed in https://github.com/exadg/exadg/pull/245

### DIFF
--- a/applications/incompressible_navier_stokes/beltrami/application.h
+++ b/applications/incompressible_navier_stokes/beltrami/application.h
@@ -247,27 +247,27 @@ private:
     PostProcessorData<dim> pp_data;
 
     // write output for visualization of results
-    pp_data.output_data.write_output     = this->output_parameters.write;
+    pp_data.output_data.time_control_data.is_active        = this->output_parameters.write;
+    pp_data.output_data.time_control_data.start_time       = start_time;
+    pp_data.output_data.time_control_data.trigger_interval = (end_time - start_time) / 10.0;
     pp_data.output_data.directory        = this->output_parameters.directory + "vtu/";
     pp_data.output_data.filename         = this->output_parameters.filename;
-    pp_data.output_data.start_time       = start_time;
-    pp_data.output_data.interval_time    = (end_time - start_time) / 10;
     pp_data.output_data.write_divergence = false;
     pp_data.output_data.degree           = this->param.degree_u;
 
     // calculation of velocity error
-    pp_data.error_data_u.analytical_solution_available = true;
+    pp_data.error_data_u.time_control_data.is_active        = true;
+    pp_data.error_data_u.time_control_data.start_time       = start_time;
+    pp_data.error_data_u.time_control_data.trigger_interval = (end_time - start_time) / 10.0;
     pp_data.error_data_u.analytical_solution.reset(new AnalyticalSolutionVelocity<dim>(viscosity));
-    pp_data.error_data_u.error_calc_start_time    = start_time;
-    pp_data.error_data_u.error_calc_interval_time = (end_time - start_time) / 10;
-    pp_data.error_data_u.name                     = "velocity";
+    pp_data.error_data_u.name = "velocity";
 
     // ... pressure error
-    pp_data.error_data_p.analytical_solution_available = true;
+    pp_data.error_data_p.time_control_data.is_active        = true;
+    pp_data.error_data_p.time_control_data.start_time       = start_time;
+    pp_data.error_data_p.time_control_data.trigger_interval = (end_time - start_time) / 10.0;
     pp_data.error_data_p.analytical_solution.reset(new AnalyticalSolutionPressure<dim>(viscosity));
-    pp_data.error_data_p.error_calc_start_time    = start_time;
-    pp_data.error_data_p.error_calc_interval_time = (end_time - start_time) / 10;
-    pp_data.error_data_p.name                     = "pressure";
+    pp_data.error_data_p.name = "pressure";
 
     std::shared_ptr<PostProcessorBase<dim, Number>> pp;
     pp.reset(new PostProcessor<dim, Number>(pp_data, this->mpi_comm));
--- a/applications/incompressible_navier_stokes/orr_sommerfeld/include/perturbation_energy.h
+++ b/applications/incompressible_navier_stokes/orr_sommerfeld/include/perturbation_energy.h
@@ -27,8 +27,8 @@
 
 // ExaDG
 #include <exadg/matrix_free/integrators.h>
+#include <exadg/postprocessor/time_control.h>
 #include <exadg/utilities/create_directories.h>
-#include <exadg/utilities/numbers.h>
 #include <exadg/utilities/print_functions.h>
 
 namespace ExaDG
@@ -38,24 +38,19 @@ namespace IncNS
 struct PerturbationEnergyData
 {
   PerturbationEnergyData()
-    : calculate(false),
-      calculate_every_time_steps(std::numeric_limits<unsigned int>::max()),
-      directory("output/"),
-      filename("orr_sommerfeld"),
-      omega_i(0.0),
-      h(1.0),
-      U_max(1.0)
+    : directory("output/"), filename("orr_sommerfeld"), omega_i(0.0), h(1.0), U_max(1.0)
   {
   }
 
   void
   print(dealii::ConditionalOStream & pcout)
   {
-    if(calculate == true)
+    if(time_control_data.is_active)
     {
       pcout << "  Calculate perturbation energy:" << std::endl;
-      print_parameter(pcout, "Calculate perturbation energy", calculate);
-      print_parameter(pcout, "Calculate every time steps", calculate_every_time_steps);
+      // only implemented for unsteady case
+      time_control_data.print(pcout, true /*unsteady*/);
+
       print_parameter(pcout, "Directory", directory);
       print_parameter(pcout, "Filename", filename);
       print_parameter(pcout, "Amplification omega_i", omega_i);
@@ -64,13 +59,13 @@ struct PerturbationEnergyData
     }
   }
 
-  bool         calculate;
-  unsigned int calculate_every_time_steps;
-  std::string  directory;
-  std::string  filename;
-  double       omega_i;
-  double       h;
-  double       U_max;
+  TimeControlData time_control_data;
+
+  std::string directory;
+  std::string filename;
+  double      omega_i;
+  double      h;
+  double      U_max;
 };
 
 /*
@@ -109,81 +104,71 @@ public:
     quad_index  = quad_index_in;
     data        = data_in;
 
-    if(data.calculate)
+    time_control.setup(data.time_control_data);
+
+    if(data.time_control_data.is_active)
       create_directories(data.directory, mpi_comm);
   }
 
   void
-  evaluate(VectorType const & velocity, double const & time, int const & time_step_number)
+  evaluate(VectorType const & velocity, double const time, bool const unsteady)
   {
-    if(data.calculate)
-    {
-      if(Utilities::is_unsteady_timestep(time_step_number))
-      {
-        do_evaluate(velocity, time, time_step_number);
-      }
-      else
-      {
-        AssertThrow(false,
-                    dealii::ExcMessage(
-                      "Calculation of perturbation energy for "
-                      "Orr-Sommerfeld problem only makes sense for unsteady problems."));
-      }
-    }
+    AssertThrow(unsteady,
+                dealii::ExcMessage(
+                  "Calculation of perturbation energy for "
+                  "Orr-Sommerfeld problem only makes sense for unsteady problems."));
+    do_evaluate(velocity, time);
   }
 
 private:
   void
-  do_evaluate(VectorType const & velocity, double const time, unsigned int const time_step_number)
+  do_evaluate(VectorType const & velocity, double const time)
   {
-    if((time_step_number - 1) % data.calculate_every_time_steps == 0)
+    Number perturbation_energy = 0.0;
+
+    integrate(*matrix_free, velocity, perturbation_energy);
+
+    if(!initial_perturbation_energy_has_been_calculated)
     {
-      Number perturbation_energy = 0.0;
+      initial_perturbation_energy = perturbation_energy;
 
-      integrate(*matrix_free, velocity, perturbation_energy);
+      initial_perturbation_energy_has_been_calculated = true;
+    }
 
-      if(!initial_perturbation_energy_has_been_calculated)
-      {
-        initial_perturbation_energy = perturbation_energy;
-
-        initial_perturbation_energy_has_been_calculated = true;
-      }
-
-      // write output file
-      if(dealii::Utilities::MPI::this_mpi_process(mpi_comm) == 0)
-      {
-        // clang-format off
+    // write output file
+    if(dealii::Utilities::MPI::this_mpi_process(mpi_comm) == 0)
+    {
+      // clang-format off
         unsigned int l = matrix_free->get_dof_handler(dof_index)
                            .get_triangulation().n_global_levels() - 1;
-        // clang-format on
+      // clang-format on
 
-        std::string filename =
-          data.directory + data.filename + "_l" + dealii::Utilities::int_to_string(l);
+      std::string filename =
+        data.directory + data.filename + "_l" + dealii::Utilities::int_to_string(l);
 
-        std::ofstream f;
-        if(clear_files == true)
-        {
-          f.open(filename.c_str(), std::ios::trunc);
-          f << "Perturbation energy: E = (1,(u-u_base)^2)_Omega" << std::endl
-            << "Error:               e = |exp(2*omega_i*t) - E(t)/E(0)|" << std::endl;
+      std::ofstream f;
+      if(clear_files == true)
+      {
+        f.open(filename.c_str(), std::ios::trunc);
+        f << "Perturbation energy: E = (1,(u-u_base)^2)_Omega" << std::endl
+          << "Error:               e = |exp(2*omega_i*t) - E(t)/E(0)|" << std::endl;
 
-          f << std::endl << "  Time                energy              error" << std::endl;
+        f << std::endl << "  Time                energy              error" << std::endl;
 
-          clear_files = false;
-        }
-        else
-        {
-          f.open(filename.c_str(), std::ios::app);
-        }
-
-        Number const rel   = perturbation_energy / initial_perturbation_energy;
-        Number const error = std::abs(std::exp(2 * data.omega_i * time) - rel);
-
-        unsigned int const precision = 12;
-        f << std::scientific << std::setprecision(precision) << std::setw(precision + 8) << time
-          << std::setw(precision + 8) << perturbation_energy << std::setw(precision + 8) << error
-          << std::endl;
+        clear_files = false;
       }
+      else
+      {
+        f.open(filename.c_str(), std::ios::app);
+      }
+
+      Number const rel   = perturbation_energy / initial_perturbation_energy;
+      Number const error = std::abs(std::exp(2 * data.omega_i * time) - rel);
+
+      unsigned int const precision = 12;
+      f << std::scientific << std::setprecision(precision) << std::setw(precision + 8) << time
+        << std::setw(precision + 8) << perturbation_energy << std::setw(precision + 8) << error
+        << std::endl;
     }
   }
 
@@ -241,6 +226,8 @@ private:
       }
     }
   }
+
+  TimeControl time_control;
 
   MPI_Comm const mpi_comm;
 
--- a/applications/incompressible_navier_stokes/periodic_hill/include/postprocessor.h
+++ b/applications/incompressible_navier_stokes/periodic_hill/include/postprocessor.h
@@ -86,10 +86,10 @@ public:
   }
 
   void
-  do_postprocessing(VectorType const & velocity,
-                    VectorType const & pressure,
-                    double const       time,
-                    int const          time_step_number)
+  do_postprocessing(VectorType const &     velocity,
+                    VectorType const &     pressure,
+                    double const           time,
+                    types::time_step const time_step_number)
   {
     Base::do_postprocessing(velocity, pressure, time, time_step_number);
 
@@ -104,7 +104,17 @@ public:
     }
 
     // line plot statistics
-    line_plot_calculator_statistics->evaluate(velocity, pressure, time, time_step_number);
+    if(line_plot_calculator_statistics->time_control_statistics.time_control.needs_evaluation(
+         time, time_step_number))
+    {
+      line_plot_calculator_statistics->evaluate(velocity, pressure);
+    }
+
+    if(line_plot_calculator_statistics->time_control_statistics.write_preliminary_results(
+         time, time_step_number))
+    {
+      line_plot_calculator_statistics->write_output();
+    }
   }
 
 private:
--- a/include/exadg/compressible_navier_stokes/postprocessor/output_generator.cpp
+++ b/include/exadg/compressible_navier_stokes/postprocessor/output_generator.cpp
@@ -26,7 +26,6 @@
 #include <exadg/compressible_navier_stokes/postprocessor/output_generator.h>
 #include <exadg/postprocessor/write_output.h>
 #include <exadg/utilities/create_directories.h>
-#include <exadg/utilities/numbers.h>
 
 namespace ExaDG
 {
@@ -99,8 +98,7 @@ write_output(OutputData const &                              output_data,
 }
 
 template<int dim, typename Number>
-OutputGenerator<dim, Number>::OutputGenerator(MPI_Comm const & comm)
-  : mpi_comm(comm), output_counter(0), reset_counter(true)
+OutputGenerator<dim, Number>::OutputGenerator(MPI_Comm const & comm) : mpi_comm(comm)
 {
 }
 
@@ -114,10 +112,9 @@ OutputGenerator<dim, Number>::setup(dealii::DoFHandler<dim> const & dof_handler_
   mapping     = &mapping_in;
   output_data = output_data_in;
 
-  // reset output counter
-  output_counter = output_data.start_counter;
+  time_control.setup(output_data_in.time_control_data);
 
-  if(output_data.write_output == true)
+  if(output_data_in.time_control_data.is_active)
   {
     create_directories(output_data.directory, mpi_comm);
 
@@ -140,10 +137,15 @@ OutputGenerator<dim, Number>::setup(dealii::DoFHandler<dim> const & dof_handler_
                          output_data.degree,
                          output_data.directory,
                          output_data.filename,
-                         output_counter,
+                         time_control.get_counter(),
                          mpi_comm);
     }
 
+    // write grid
+    if(output_data.write_grid)
+    {
+      write_grid(dof_handler->get_triangulation(), output_data.directory, output_data.filename);
+    }
 
     // processor_id
     if(output_data.write_processor_id)
@@ -162,68 +164,20 @@ void
 OutputGenerator<dim, Number>::evaluate(
   VectorType const &                              solution_conserved,
   std::vector<SolutionField<dim, Number>> const & additional_fields,
-  double const &                                  time,
-  int const &                                     time_step_number)
+  double const                                    time,
+  bool const                                      unsteady)
 {
-  dealii::ConditionalOStream pcout(std::cout,
-                                   dealii::Utilities::MPI::this_mpi_process(mpi_comm) == 0);
+  print_write_output_time(time, time_control.get_counter(), unsteady, mpi_comm);
 
-  if(output_data.write_output == true)
-  {
-    if(Utilities::is_unsteady_timestep(time_step_number))
-    {
-      // small number which is much smaller than the time step size
-      double const EPSILON = 1.0e-10;
-
-      // In the first time step, the current time might be larger than start_time. In that
-      // case, we first have to reset the counter in order to avoid that output is written every
-      // time step.
-      if(reset_counter)
-      {
-        if(time > output_data.start_time)
-        {
-          output_counter +=
-            int((time - output_data.start_time + EPSILON) / output_data.interval_time);
-        }
-        reset_counter = false;
-      }
-
-
-      if(time > (output_data.start_time + output_counter * output_data.interval_time - EPSILON))
-      {
-        pcout << std::endl
-              << "OUTPUT << Write data at time t = " << std::scientific << std::setprecision(4)
-              << time << std::endl;
-
-        write_output<dim, Number, VectorType>(output_data,
-                                              *dof_handler,
-                                              *mapping,
-                                              solution_conserved,
-                                              additional_fields,
-                                              output_counter,
-                                              mpi_comm);
-
-        ++output_counter;
-      }
-    }
-    else
-    {
-      pcout << std::endl
-            << "OUTPUT << Write " << (output_counter == 0 ? "initial" : "solution") << " data"
-            << std::endl;
-
-      write_output<dim, Number, VectorType>(output_data,
-                                            *dof_handler,
-                                            *mapping,
-                                            solution_conserved,
-                                            additional_fields,
-                                            output_counter,
-                                            mpi_comm);
-
-      ++output_counter;
-    }
-  }
+  write_output<dim, Number, VectorType>(output_data,
+                                        *dof_handler,
+                                        *mapping,
+                                        solution_conserved,
+                                        additional_fields,
+                                        time_control.get_counter(),
+                                        mpi_comm);
 }
+
 
 template class OutputGenerator<2, float>;
 template class OutputGenerator<2, double>;
--- a/include/exadg/compressible_navier_stokes/postprocessor/postprocessor.cpp
+++ b/include/exadg/compressible_navier_stokes/postprocessor/postprocessor.cpp
@@ -21,6 +21,7 @@
 
 #include <exadg/compressible_navier_stokes/postprocessor/postprocessor.h>
 #include <exadg/compressible_navier_stokes/spatial_discretization/operator.h>
+#include <exadg/utilities/numbers.h>
 
 namespace ExaDG
 {
@@ -89,9 +90,9 @@ PostProcessor<dim, Number>::setup(Operator<dim, Number> const & pde_operator)
 
 template<int dim, typename Number>
 void
-PostProcessor<dim, Number>::do_postprocessing(VectorType const & solution,
-                                              double const       time,
-                                              int const          time_step_number)
+PostProcessor<dim, Number>::do_postprocessing(VectorType const &     solution,
+                                              double const           time,
+                                              types::time_step const time_step_number)
 {
   /*
    * calculate derived quantities such as velocity, pressure, etc.
@@ -101,36 +102,60 @@ PostProcessor<dim, Number>::do_postprocessing(VectorType const & solution,
   /*
    *  write output
    */
-  output_generator.evaluate(solution, additional_fields, time, time_step_number);
+  if(output_generator.time_control.needs_evaluation(time, time_step_number))
+  {
+    output_generator.evaluate(solution,
+                              additional_fields,
+                              time,
+                              Utilities::is_unsteady_timestep(time_step_number));
+  }
 
   /*
    *  write pointwise output
    */
-  pointwise_output_generator.evaluate(solution, time, time_step_number);
+  if(pointwise_output_generator.time_control.needs_evaluation(time, time_step_number))
+  {
+    pointwise_output_generator.evaluate(solution,
+                                        time,
+                                        Utilities::is_unsteady_timestep(time_step_number));
+  }
   /*
    *  calculate error
    */
-  error_calculator.evaluate(solution, time, time_step_number);
+  if(error_calculator.time_control.needs_evaluation(time, time_step_number))
+    error_calculator.evaluate(solution, time, Utilities::is_unsteady_timestep(time_step_number));
 
   /*
    *  calculation of lift and drag coefficients
    */
-  lift_and_drag_calculator.evaluate(velocity, pressure, time);
+  if(lift_and_drag_calculator.time_control.needs_evaluation(time, time_step_number))
+    lift_and_drag_calculator.evaluate(velocity, pressure, time);
 
   /*
    *  calculation of pressure difference
    */
-  pressure_difference_calculator.evaluate(pressure, time);
+  if(pressure_difference_calculator.time_control.needs_evaluation(time, time_step_number))
+    pressure_difference_calculator.evaluate(pressure, time);
 
   /*
    *  calculation of kinetic energy
    */
-  kinetic_energy_calculator.evaluate(velocity, time, time_step_number);
+  if(kinetic_energy_calculator.time_control.needs_evaluation(time, time_step_number))
+  {
+    kinetic_energy_calculator.evaluate(velocity,
+                                       time,
+                                       Utilities::is_unsteady_timestep(time_step_number));
+  }
 
   /*
    *  calculation of kinetic energy spectrum
    */
-  kinetic_energy_spectrum_calculator.evaluate(velocity, time, time_step_number);
+  if(kinetic_energy_spectrum_calculator.time_control.needs_evaluation(time, time_step_number))
+  {
+    kinetic_energy_spectrum_calculator.evaluate(velocity,
+                                                time,
+                                                Utilities::is_unsteady_timestep(time_step_number));
+  }
 }
 
 template<int dim, typename Number>
@@ -206,31 +231,37 @@ template<int dim, typename Number>
 void
 PostProcessor<dim, Number>::calculate_additional_vectors(VectorType const & solution)
 {
-  if((pp_data.output_data.write_output == true && pp_data.output_data.write_pressure == true) ||
+  if((pp_data.output_data.time_control_data.is_active == true &&
+      pp_data.output_data.write_pressure == true) ||
      pp_data.calculate_pressure == true)
   {
     navier_stokes_operator->compute_pressure(pressure, solution);
   }
 
-  if((pp_data.output_data.write_output == true && pp_data.output_data.write_velocity == true) ||
-     (pp_data.output_data.write_output == true && pp_data.output_data.write_vorticity == true) ||
-     (pp_data.output_data.write_output == true && pp_data.output_data.write_divergence == true) ||
+  if((pp_data.output_data.time_control_data.is_active == true &&
+      pp_data.output_data.write_velocity == true) ||
+     (pp_data.output_data.time_control_data.is_active == true &&
+      pp_data.output_data.write_vorticity == true) ||
+     (pp_data.output_data.time_control_data.is_active == true &&
+      pp_data.output_data.write_divergence == true) ||
      pp_data.calculate_velocity == true)
   {
     navier_stokes_operator->compute_velocity(velocity, solution);
   }
 
-  if(pp_data.output_data.write_output == true && pp_data.output_data.write_temperature == true)
+  if(pp_data.output_data.time_control_data.is_active &&
+     pp_data.output_data.write_temperature == true)
   {
     navier_stokes_operator->compute_temperature(temperature, solution);
   }
 
-  if(pp_data.output_data.write_output == true && pp_data.output_data.write_vorticity == true)
+  if(pp_data.output_data.time_control_data.is_active && pp_data.output_data.write_vorticity == true)
   {
     navier_stokes_operator->compute_vorticity(vorticity, velocity);
   }
 
-  if(pp_data.output_data.write_output == true && pp_data.output_data.write_divergence == true)
+  if(pp_data.output_data.time_control_data.is_active == true &&
+     pp_data.output_data.write_divergence == true)
   {
     navier_stokes_operator->compute_divergence(divergence, velocity);
   }
--- a/include/exadg/incompressible_navier_stokes/postprocessor/line_plot_calculation_statistics.cpp
+++ b/include/exadg/incompressible_navier_stokes/postprocessor/line_plot_calculation_statistics.cpp
@@ -50,27 +50,30 @@ LinePlotCalculatorStatistics<dim, Number>::LinePlotCalculatorStatistics(
 
 template<int dim, typename Number>
 void
-LinePlotCalculatorStatistics<dim, Number>::setup(
-  LinePlotDataStatistics<dim> const & line_plot_data_in)
+LinePlotCalculatorStatistics<dim, Number>::setup(LinePlotDataStatistics<dim> const & data_in)
 {
   // initialize data
-  data = line_plot_data_in;
+  data = data_in;
 
-  if(data.statistics_data.calculate == true)
+  AssertThrow(Utilities::is_valid_timestep(
+                data_in.time_control_data_statistics.write_preliminary_results_every_nth_time_step),
+              dealii::ExcMessage("write_preliminary_results_every_nth_time_step has to be set."));
+  time_control_statistics.setup(data_in.time_control_data_statistics);
+
+  if(data_in.time_control_data_statistics.time_control_data.is_active)
   {
-    AssertThrow(data.line_data.lines.size() > 0, dealii::ExcMessage("Empty data"));
+    AssertThrow(data.lines.size() > 0, dealii::ExcMessage("Empty data"));
 
     // allocate data structures
-    velocity_global.resize(data.line_data.lines.size());
-    pressure_global.resize(data.line_data.lines.size());
-    global_points.resize(data.line_data.lines.size());
-    cells_global_velocity.resize(data.line_data.lines.size());
-    cells_global_pressure.resize(data.line_data.lines.size());
+    velocity_global.resize(data.lines.size());
+    pressure_global.resize(data.lines.size());
+    global_points.resize(data.lines.size());
+    cells_global_velocity.resize(data.lines.size());
+    cells_global_pressure.resize(data.lines.size());
 
     unsigned int line_iterator = 0;
-    for(typename std::vector<std::shared_ptr<Line<dim>>>::iterator line =
-          data.line_data.lines.begin();
-        line != data.line_data.lines.end();
+    for(typename std::vector<std::shared_ptr<Line<dim>>>::iterator line = data.lines.begin();
+        line != data.lines.end();
         ++line, ++line_iterator)
     {
       // Resize global variables for number of points on line
@@ -89,43 +92,23 @@ LinePlotCalculatorStatistics<dim, Number>::setup(
       cells_global_pressure[line_iterator].resize((*line)->n_points);
     }
 
-    create_directories(data.line_data.directory, mpi_comm);
+    create_directories(data.directory, mpi_comm);
   }
 }
 
 template<int dim, typename Number>
 void
-LinePlotCalculatorStatistics<dim, Number>::evaluate(VectorType const &   velocity,
-                                                    VectorType const &   pressure,
-                                                    double const &       time,
-                                                    unsigned int const & time_step_number)
+LinePlotCalculatorStatistics<dim, Number>::evaluate(VectorType const & velocity,
+                                                    VectorType const & pressure)
 {
-  if(data.statistics_data.calculate)
-  {
-    // EPSILON: small number which is much smaller than the time step size
-    double const EPSILON = 1.0e-10;
+  do_evaluate(velocity, pressure);
+}
 
-    if((time > data.statistics_data.sample_start_time - EPSILON) &&
-       (time < data.statistics_data.sample_end_time + EPSILON) &&
-       (time_step_number % data.statistics_data.sample_every_timesteps == 0))
-    {
-      // evaluate statistics
-      do_evaluate(velocity, pressure);
-
-      // write intermediate output
-      if(time_step_number % data.statistics_data.write_output_every_timesteps == 0)
-      {
-        do_write_output();
-      }
-    }
-
-    // write final output
-    if((time > data.statistics_data.sample_end_time - EPSILON) && write_final_output)
-    {
-      do_write_output();
-      write_final_output = false;
-    }
-  }
+template<int dim, typename Number>
+void
+LinePlotCalculatorStatistics<dim, Number>::write_output() const
+{
+  do_write_output();
 }
 
 template<int dim, typename Number>
@@ -135,9 +118,8 @@ LinePlotCalculatorStatistics<dim, Number>::initialize_cell_data(VectorType const
 {
   // Save data related to all adjacent cells for a given point along the line.
   unsigned int line_iterator = 0;
-  for(typename std::vector<std::shared_ptr<Line<dim>>>::iterator line =
-        data.line_data.lines.begin();
-      line != data.line_data.lines.end();
+  for(typename std::vector<std::shared_ptr<Line<dim>>>::iterator line = data.lines.begin();
+      line != data.lines.end();
       ++line, ++line_iterator)
   {
     // make sure that line type is correct
@@ -295,9 +277,8 @@ LinePlotCalculatorStatistics<dim, Number>::do_evaluate(VectorType const & veloci
 
   // Iterator for lines
   unsigned int line_iterator = 0;
-  for(typename std::vector<std::shared_ptr<Line<dim>>>::iterator line =
-        data.line_data.lines.begin();
-      line != data.line_data.lines.end();
+  for(typename std::vector<std::shared_ptr<Line<dim>>>::iterator line = data.lines.begin();
+      line != data.lines.end();
       ++line, ++line_iterator)
   {
     bool evaluate_velocity = false;
@@ -510,19 +491,17 @@ template<int dim, typename Number>
 void
 LinePlotCalculatorStatistics<dim, Number>::do_write_output() const
 {
-  if(dealii::Utilities::MPI::this_mpi_process(mpi_comm) == 0 &&
-     data.statistics_data.calculate == true)
+  if(dealii::Utilities::MPI::this_mpi_process(mpi_comm) == 0)
   {
-    unsigned int const precision = data.line_data.precision;
+    unsigned int const precision = data.precision;
 
     // Iterator for lines
     unsigned int line_iterator = 0;
-    for(typename std::vector<std::shared_ptr<Line<dim>>>::const_iterator line =
-          data.line_data.lines.begin();
-        line != data.line_data.lines.end();
+    for(typename std::vector<std::shared_ptr<Line<dim>>>::const_iterator line = data.lines.begin();
+        line != data.lines.end();
         ++line, ++line_iterator)
     {
-      std::string filename_prefix = data.line_data.directory + (*line)->name;
+      std::string filename_prefix = data.directory + (*line)->name;
 
       for(typename std::vector<std::shared_ptr<Quantity>>::const_iterator quantity =
             (*line)->quantities.begin();
--- a/include/exadg/incompressible_navier_stokes/postprocessor/line_plot_calculation_statistics_homogeneous.cpp
+++ b/include/exadg/incompressible_navier_stokes/postprocessor/line_plot_calculation_statistics_homogeneous.cpp
@@ -54,26 +54,31 @@ LinePlotCalculatorStatisticsHomogeneous<dim, Number>::setup(
 {
   data = data_in;
 
-  if(data.statistics_data.calculate)
+  AssertThrow(Utilities::is_valid_timestep(
+                data_in.time_control_data_statistics.write_preliminary_results_every_nth_time_step),
+              dealii::ExcMessage("write_preliminary_results_every_nth_time_step has to be set."));
+  time_control_statistics.setup(data_in.time_control_data_statistics);
+
+  if(data_in.time_control_data_statistics.time_control_data.is_active)
   {
     AssertThrow(dim == 3, dealii::ExcMessage("Not implemented."));
 
-    AssertThrow(data.line_data.lines.size() > 0, dealii::ExcMessage("Empty data"));
+    AssertThrow(data.lines.size() > 0, dealii::ExcMessage("Empty data"));
 
-    global_points.resize(data.line_data.lines.size());
-    cells_and_ref_points_velocity.resize(data.line_data.lines.size());
-    cells_and_ref_points_pressure.resize(data.line_data.lines.size());
-    cells_and_ref_points_ref_pressure.resize(data.line_data.lines.size());
+    global_points.resize(data.lines.size());
+    cells_and_ref_points_velocity.resize(data.lines.size());
+    cells_and_ref_points_pressure.resize(data.lines.size());
+    cells_and_ref_points_ref_pressure.resize(data.lines.size());
 
-    velocity_global.resize(data.line_data.lines.size());
-    wall_shear_global.resize(data.line_data.lines.size());
-    reynolds_global.resize(data.line_data.lines.size());
-    pressure_global.resize(data.line_data.lines.size());
-    reference_pressure_global.resize(data.line_data.lines.size());
+    velocity_global.resize(data.lines.size());
+    wall_shear_global.resize(data.lines.size());
+    reynolds_global.resize(data.lines.size());
+    pressure_global.resize(data.lines.size());
+    reference_pressure_global.resize(data.lines.size());
 
     // make sure that line type is correct
     std::shared_ptr<LineHomogeneousAveraging<dim>> line_hom =
-      std::dynamic_pointer_cast<LineHomogeneousAveraging<dim>>(data.line_data.lines[0]);
+      std::dynamic_pointer_cast<LineHomogeneousAveraging<dim>>(data.lines[0]);
     AssertThrow(line_hom.get() != 0,
                 dealii::ExcMessage("Invalid line type, expected LineHomogeneousAveraging<dim>"));
     averaging_direction = line_hom->averaging_direction;
@@ -82,9 +87,8 @@ LinePlotCalculatorStatisticsHomogeneous<dim, Number>::setup(
                 dealii::ExcMessage("Take the average either in x, y or z-direction"));
 
     unsigned int line_iterator = 0;
-    for(typename std::vector<std::shared_ptr<Line<dim>>>::iterator line =
-          data.line_data.lines.begin();
-        line != data.line_data.lines.end();
+    for(typename std::vector<std::shared_ptr<Line<dim>>>::iterator line = data.lines.begin();
+        line != data.lines.end();
         ++line, ++line_iterator)
     {
       // make sure that line type is correct
@@ -126,9 +130,8 @@ LinePlotCalculatorStatisticsHomogeneous<dim, Number>::setup(
       if(cell->is_locally_owned())
       {
         line_iterator = 0;
-        for(typename std::vector<std::shared_ptr<Line<dim>>>::iterator line =
-              data.line_data.lines.begin();
-            line != data.line_data.lines.end();
+        for(typename std::vector<std::shared_ptr<Line<dim>>>::iterator line = data.lines.begin();
+            line != data.lines.end();
             ++line, ++line_iterator)
         {
           AssertThrow((*line)->quantities.size() > 0,
@@ -202,9 +205,8 @@ LinePlotCalculatorStatisticsHomogeneous<dim, Number>::setup(
       if(cell->is_locally_owned())
       {
         line_iterator = 0;
-        for(typename std::vector<std::shared_ptr<Line<dim>>>::iterator line =
-              data.line_data.lines.begin();
-            line != data.line_data.lines.end();
+        for(typename std::vector<std::shared_ptr<Line<dim>>>::iterator line = data.lines.begin();
+            line != data.lines.end();
             ++line, ++line_iterator)
         {
           for(typename std::vector<std::shared_ptr<Quantity>>::iterator quantity =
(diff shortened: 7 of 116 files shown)


include/exadg/postprocessor/kinetic_energy_spectrum.cpp
```diff
@@ -403,100 +408,54 @@ KineticEnergySpectrumCalculator<dim, Number>::setup(
...
-  unsigned int const time_step_number)
+void
+KineticEnergySpectrumCalculator<dim, Number>::evaluate(VectorType const & velocity,
+                                                       double const       time,
+                                                       int const          time_step_number)
 {
-  bool evaluate = false;
+  AssertThrow(time_control.get_last_timestep_number_to_evaluate() == time_step_number,
+              dealii::ExcMessage(
+                "evaluate should only be called once needs_evaluation() returned true."));
```
nfehn, line 426: can we get rid of this assert? My hope would be that we no longer have to pass `time_step_number` to `evaluate()` functions. We can achieve this, apart from the assert, by adding a boolean parameter `evaluate(..., bool const steady_problem)` only for those postprocessing modules that are implemented also for steady problems. The function is then called with `evaluate(..., time_step_number < 0)` (or whatever the convention for steady problems is). If this is not implemented (as in this case here), the function `evaluate()` just contains the vector and the time.

include/exadg/postprocessor/time_control.h
```diff
@@ -0,0 +1,59 @@
...
+
+  unsigned int
+  get_last_timestep_number_to_evaluate() const;
+
+private:
+  // small number which is much smaller than the time step size
+  double const EPSILON;
+  mutable bool reset_counter = true;
+
+  bool                 calculate;
```
nfehn, line 51: I wonder whether a slightly different name like `activate` would be better here, just because we are decoupling the time control from the "calculate" part in this PR and the name "calculate" might therefore no longer be clear.

include/exadg/postprocessor/time_control.cpp
```diff
@@ -0,0 +1,126 @@
...
+}
+
+unsigned int
+TimeControl::get_counter() const
+{
+  return counter;
+}
+
+unsigned int
+TimeControl::get_last_timestep_number_to_evaluate() const
```
nfehn, line 127: I would like to get rid of this function.

nfehn (commented):
Looks very good overall. Many thanks for the initiative!

include/exadg/postprocessor/kinetic_energy_spectrum.cpp
jh66637: We can do this if we assume nobody is using this in external projects. Not changing the interface results in output at every timestep. If this is ok for you I will remove it.

nfehn: That's acceptable in my opinion. For external projects nothing is "falsified", just an if-statement has to be introduced. I think there are good reasons why we make the changes of this PR, i.e. it turned out that we had a bad design before.

include/exadg/postprocessor/time_control.cpp
```diff
@@ -0,0 +1,130 @@
...
+    if(reset_counter)
+    {
+      counter += int((time - start_time + EPSILON) / calculate_every_time_interval);
+      reset_counter = false;
+    }
+
+    if((time > (start_time + counter * calculate_every_time_interval - EPSILON)))
+    {
+      evaluate = true;
+      ++counter;
```
jh66637, line 103: Question: should we do this manually? There is no guarantee that evaluate is called after it is checked if evaluation is needed.

include/exadg/postprocessor/time_control.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+void
+TimeControl::reinit(bool const         calculate_,
+                    double const       start_time_,
+                    double const       calculate_every_time_interval_,
+                    unsigned int const calculate_every_time_steps_)
+{
+  calculate = calculate_;
+
+  if(calculate_)
```
jh66637, line 52: Question: I never used ExaDG for unsteady problems. Is this valid also in the unsteady case?

include/exadg/postprocessor/time_control.cpp
jh66637: [d0bc416](https://github.com/exadg/exadg/pull/262/commits/d0bc416c46ebf80a9ec7966064f64303d03a538b)

include/exadg/postprocessor/kinetic_energy_spectrum.cpp
jh66637: [d0bc416](https://github.com/exadg/exadg/pull/262/commits/d0bc416c46ebf80a9ec7966064f64303d03a538b)

include/exadg/postprocessor/time_control.h
```diff
@@ -0,0 +1,59 @@
...
+  unsigned int
+  get_counter() const;
+
+  unsigned int
+  get_last_timestep_number_to_evaluate() const;
+
+private:
+  // small number which is much smaller than the time step size
+  double const EPSILON;
+  mutable bool reset_counter = true;
```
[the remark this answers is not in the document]
jh66637: [04abea3](https://github.com/exadg/exadg/pull/262/commits/04abea3a0673d1aa9a1b1bf4ae66388e4686860d)

include/exadg/postprocessor/time_control.cpp
nfehn: Do you mean the boolean parameter or the check whether the parameters are consistent.

include/exadg/postprocessor/time_control.h
jh66637: [89b7335](https://github.com/exadg/exadg/pull/262/commits/89b73350e0d87538534d8a900f21b72ff6b5abee)
should completely decouple from calculate

include/exadg/postprocessor/time_control.cpp
jh66637: Sorry something went wrong...

include/exadg/postprocessor/time_control.cpp
```diff
@@ -0,0 +1,127 @@
...
+{
+  is_active = is_active_;
+
+  if(is_active_)
+  {
+    if(trigger_every_time_steps_ == std::numeric_limits<unsigned int>::max() ||
+       trigger_every_time_steps_ == 0)
+    {
+      AssertThrow(trigger_every_time_interval_ > 0.0 &&
+                    trigger_every_time_interval_ < std::numeric_limits<double>::max(),
```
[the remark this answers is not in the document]
jh66637: I seem not to be able to mark more lines. What I am referring to with the questions are the two asserts in the reinit function

include/exadg/postprocessor/time_control.cpp
nfehn: Do you mean incrementing the counter? What is the solution that you have in mind?

include/exadg/postprocessor/time_control.cpp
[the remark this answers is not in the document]
nfehn: I think these checks mean that the user sets exactly on option (time steps or time interval) in case that the postprocessing routine is active. So we also need it for the unsteady case. However, it does not really make sense in the steady case. We just want to activate a postprocessing routine without specifying the number of time steps or a time interval.

include/exadg/postprocessor/time_control.cpp
```diff
@@ -0,0 +1,127 @@
...
+}
+
+// If reinit() is not called needs_evaluation() will always evaluate to false
+bool
+TimeControl::needs_evaluation(double const time, unsigned int const time_step_number) const
+{
+  if(is_active == false)
+    return false;
+
+  if(time_step_number < 0) // steady problem (time_step_number = -1)
```
nfehn, line 81: I would change the comment here: "Steady problems are defined by a negative time step number"

include/exadg/postprocessor/time_control.cpp
nfehn: Principally, the postprocessing routine could take care of counter. However, it is then more difficult the fix the problem with the restart, and all the postprocessing routines need a counter. So I would accept the currrent solution, maybe with a comment to make users aware of this subtlety.

jh66637: A different option would be a member function ```increment_counter()```
```
time_control.increment_counter();
```
 or ```operator++```.
```
++time_control
```

include/exadg/postprocessor/time_control.cpp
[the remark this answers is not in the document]
jh66637: So the question is: Should the user care about setting the time_interval or time_step option such that the asserts are fulfilled also in the steady case. Is this currently done?

include/exadg/postprocessor/time_control.cpp
jh66637: In my opinion we should be explicit about this anyway, see:
https://github.com/exadg/exadg/pull/263
If you agree, we should merge this PR before this one.

include/exadg/postprocessor/time_control.cpp
```diff
@@ -0,0 +1,127 @@
...
+  {
+    if(time > start_time && (time_step_number - 1) % trigger_every_time_steps == 0)
+    {
+      evaluate = true;
+    }
+  }
+  else if(trigger_every_time_interval > 0.0)
+  {
+    // The current time might be larger than start_time. In that case, we first have to
+    // reset the counter in order to avoid that output is written every time step.
```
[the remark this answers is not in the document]
jh66637: [2b52f5f](https://github.com/exadg/exadg/pull/262/commits/2b52f5f19db3fa8d1322580689996a1e4c4f0955)

include/exadg/postprocessor/time_control.cpp
nfehn: Yes, but I think this variant can also lead to errors. Think about implementing a new postprocessing routine but forgetting to increment `TimeControl`` at  the end of the `evaluate()` function. Then, the postprocessing will be invoked after every time step and the parameters (time step interval and time interval) are just ignored. Since counter is not incremented, all output files have the same output number/counter in the filename (but output is written every timestep!). For the current variant, I would say it is less likely to forget calling `evaluate()` after the if-statement. What do you think? I would personally prefer the variant that does require one line of code for every postprocessing routine, given that both variants can be misused.

include/exadg/postprocessor/time_control.cpp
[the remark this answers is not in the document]
nfehn: The asserts should not be checked for steady problems in my opinion. It is probably difficult to pass the information steady/unsteady to the `reinit()` function. It would be simpler to include the checks in the `needs_evaluation()` function.

include/exadg/postprocessor/time_control.cpp
jh66637: OK. I think I will keep it as it is (the increment in TimeControl) :)

include/exadg/postprocessor/time_control.cpp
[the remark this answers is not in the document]
jh66637: [f6b2eae](https://github.com/exadg/exadg/pull/262/commits/f6b2eae4cde8451c53c0d307d844518d4d702abc)

applications/incompressible_navier_stokes/orr_sommerfeld/include/perturbation_energy.h
```diff
@@ -134,7 +133,9 @@ class PerturbationEnergyCalculator
 
 private:
   void
-  do_evaluate(VectorType const & velocity, double const time, unsigned int const time_step_number)
+  do_evaluate(VectorType const &     velocity,
+              double const           time,
+              types::time_step const time_step_number)
   {
     if((time_step_number - 1) % data.calculate_every_time_steps == 0)
```
nfehn, line 140: I think we would also have to adjust this code here.

include/exadg/compressible_navier_stokes/postprocessor/output_generator.cpp
```diff
@@ -162,69 +166,43 @@ void
...
+          << "OUTPUT << Write " << (time_control.get_counter() == 0 ? "initial" : "solution")
+          << " data" << std::endl;
+
+    write_output<dim, Number, VectorType>(output_data,
+                                          *dof_handler,
+                                          *mapping,
+                                          solution_conserved,
+                                          additional_fields,
+                                          time_control.get_counter(),
+                                          mpi_comm);
```
nfehn, line 201: The function call seems to be the same as above for the unsteady case. So we should pull this out of the if-else-branch and call the function only once.

include/exadg/incompressible_navier_stokes/postprocessor/output_generator.cpp
```diff
@@ -185,75 +189,49 @@ OutputGenerator<dim, Number>::setup(NavierStokesOperator const &    navier_stoke
...
+
+    write_output<dim>(output_data,
+                      *dof_handler_velocity,
+                      *dof_handler_pressure,
+                      *mapping,
+                      velocity,
+                      pressure,
+                      additional_fields,
+                      time_control.get_counter(),
+                      mpi_comm);
```
nfehn, line 234: Here, we should do the same as in the other comment above, i.e. write the function calls only once for steady/unsteady problems.

include/exadg/postprocessor/error_calculation.cpp
```diff
@@ -119,57 +114,44 @@ ErrorCalculator<dim, Number>::setup(dealii::DoFHandler<dim> const &   dof_handle
   mapping     = &mapping_in;
   error_data  = error_data_in;
 
+  time_control.reinit(
+    error_data_in.analytical_solution_available, // TODO: do we realy always activate error calc if
```
nfehn, line 118: Yes, I think this is the meaning of this variable. Maybe it is better if you rename this variable into `calculate` as for the other postprocessing modules.

include/exadg/postprocessor/error_calculation.cpp
```diff
@@ -119,57 +114,44 @@ ErrorCalculator<dim, Number>::setup(dealii::DoFHandler<dim> const &   dof_handle
...
-                                       double const &     time,
-                                       int const &        time_step_number)
+                                       double const       time,
+                                       bool const         unsteady)
 {
   dealii::ConditionalOStream pcout(std::cout,
                                    dealii::Utilities::MPI::this_mpi_process(mpi_comm) == 0);
 
   if(error_data.analytical_solution_available)
   {
```
nfehn, line 139: I think we the other comment above, we should be able to remove this.

include/exadg/postprocessor/error_calculation.cpp
```diff
@@ -119,57 +114,44 @@ ErrorCalculator<dim, Number>::setup(dealii::DoFHandler<dim> const &   dof_handle
...
+      do_evaluate(solution, time);
     }
     else
     {
       pcout << std::endl
             << "Calculate error for " << error_data.name << " for "
-            << (error_counter == 0 ? "initial" : "solution") << " data:" << std::endl;
+            << (time_control.get_counter() == 0 ? "initial" : "solution") << " data:" << std::endl;
 
       do_evaluate(solution, time);
```
nfehn, line 154: We should write this function call only once for steady/unsteady problems.

include/exadg/postprocessor/output_generator_scalar.cpp
```diff
@@ -119,53 +123,29 @@ OutputGenerator<dim, Number>::setup(dealii::DoFHandler<dim> const & dof_handler_
...
+  else
+  {
+    pcout << std::endl
+          << "OUTPUT << Write " << (time_control.get_counter() == 0 ? "initial" : "solution")
+          << " data" << std::endl;
 
-      ++output_counter;
-    }
+    write_output<dim>(
+      output_data, *dof_handler, *mapping, solution, time_control.get_counter(), mpi_comm);
```
nfehn, line 148: We should write this function call only once for steady/unsteady problems.

include/exadg/structure/postprocessor/output_generator.cpp
```diff
@@ -122,49 +126,29 @@ OutputGenerator<dim, Number>::setup(dealii::DoFHandler<dim> const & dof_handler_
...
+  else
+  {
+    pcout << std::endl
+          << "OUTPUT << Write " << (time_control.get_counter() == 0 ? "initial" : "solution")
+          << " data" << std::endl;
 
-      ++output_counter;
-    }
+    write_output<dim>(
+      output_data, *dof_handler, *mapping, solution, time_control.get_counter(), mpi_comm);
```
nfehn, line 151: call this function only once

include/exadg/time_integration/solver_info_data.h
```diff
@@ -54,10 +55,11 @@ struct SolverInfoData
     print_parameter(pcout, "Interval time steps", interval_time_steps);
   }
 
+  // TODO: here we might also be able to use timecontrol
```
nfehn, line 58: Principally yes, but note the comment below regarding the variable `old_time_step_number`.

include/exadg/utilities/numbers.h
```diff
@@ -24,18 +24,29 @@
...
-is_unsteady_timestep(unsigned int const timestep)
+is_unsteady_timestep(types::time_step const timestep)
 {
   return (timestep != numbers::steady_timestep);
 }
+inline bool
+is_valid_timestep(types::time_step const timestep)
+{
+  return (timestep != numbers::invalid_timestep);
+}
```
nfehn, line 49: Do we need this function? It looks like it is not used?

include/exadg/utilities/numbers.h
```diff
@@ -24,18 +24,29 @@
...
-namespace numbers
+namespace types
 {
-unsigned int const steady_timestep = -1;
+using time_step = unsigned int;
 }
 
+namespace numbers
+{
+types::time_step const invalid_timestep = std::numeric_limits<unsigned int>::max();
```
nfehn, line 34: needed?

include/exadg/utilities/numbers.h
```diff
@@ -24,18 +24,29 @@
...
+types::time_step const steady_timestep  = std::numeric_limits<unsigned int>::max() - 1;
```
nfehn, line 35: can we here use `dealii::numbers::invalid_unsigned_int`?

include/exadg/postprocessor/error_calculation.cpp
nfehn: In the application files, it might still be useful to call this function `analytical_solution_available`, and doing the assignment `error_data.calculate = analytical_solution_available;`

include/exadg/postprocessor/error_calculation.cpp
```diff
@@ -119,57 +114,44 @@ ErrorCalculator<dim, Number>::setup(dealii::DoFHandler<dim> const &   dof_handle
   mapping     = &mapping_in;
   error_data  = error_data_in;
 
+  time_control.reinit(
+    error_data_in.analytical_solution_available, // TODO: do we realy always activate error calc if
+                                                 // analytical sol is avail?
+    error_data_in.error_calc_start_time,
+    std::numeric_limits<double>::max(), // TODO: currently placeholder
```
nfehn, line 121: should be fine like this

include/exadg/postprocessor/kinetic_energy_calculation.cpp
```diff
@@ -47,6 +46,13 @@ KineticEnergyCalculator<dim, Number>::setup(dealii::MatrixFree<dim, Number> cons
   quad_index  = quad_index_in;
   data        = kinetic_energy_data_in;
 
+  time_control.reinit(kinetic_energy_data_in.calculate,
+                      0.0,                                // TODO: currently placeholder
+                      std::numeric_limits<double>::max(), // TODO: currently placeholder
+                      -1.0,                               // TODO:currently placeholder
```
nfehn, line 52: I think we should equip `kinetic_energy_data` with `start_time`, `time_interval`, etc.

The default setting for `kinetic_energy_data` should be to do it every time step, so that existing code is not changed.

include/exadg/postprocessor/kinetic_energy_spectrum.cpp
```diff
@@ -350,6 +349,12 @@ KineticEnergySpectrumCalculator<dim, Number>::setup(
 
   clear_files = data.clear_file;
 
+  time_control.reinit(data_in.calculate,
+                      data_in.start_time,
+                      std::numeric_limits<double>::max(), // TODO: currently placeholder
```
nfehn, line 354: should be fine like this

include/exadg/postprocessor/output_generator_scalar.cpp
```diff
@@ -74,8 +72,14 @@ OutputGenerator<dim, Number>::setup(dealii::DoFHandler<dim> const & dof_handler_
   mapping     = &mapping_in;
   output_data = output_data_in;
 
-  // reset output counter
-  output_counter = output_data.start_counter;
+  time_control.reinit(output_data_in.write_output,
+                      output_data_in.start_time,
+                      std::numeric_limits<double>::max(), // TODO: currently placeholder
```
nfehn, line 77: should be fine like this

include/exadg/postprocessor/output_generator_scalar.cpp
```diff
@@ -74,8 +72,14 @@ OutputGenerator<dim, Number>::setup(dealii::DoFHandler<dim> const & dof_handler_
...
+                      output_data_in.interval_time,
+                      dealii::numbers::invalid_unsigned_int); // TODO: currently placeholder
```
nfehn, line 79: I think here we should also work with `numeric_limits::max()`. However, I am not sure because we have to make sure that the asserts in `TimeControl::needs_evaluation()` do not cause problems.

include/exadg/postprocessor/time_control.cpp
```diff
@@ -0,0 +1,142 @@
...
+    trigger_interval(-1.0),
+    trigger_every_time_steps(dealii::numbers::invalid_unsigned_int)
+{
+}
+
+TimeControl::TimeControl() : EPSILON(1.0e-10), reset_counter(true), counter(0)
+{
+}
+
+// TODO: should not be used anymore!
```
nfehn, line 46: agree

include/exadg/time_integration/restart_data.h
```diff
@@ -60,11 +61,12 @@ struct RestartData
     }
   }
 
+  // TODO: use time_control
```
nfehn, line 64: the functionality is slightly different here. We also have the wall time in order to trigger restart. Hence, it isfine if we have a separate implementation here.

include/exadg/time_integration/solver_info_data.h
nfehn: We also have `wall_time` here, so it might be ok to not change this implementation.

include/exadg/postprocessor/time_control.cpp
```diff
@@ -0,0 +1,142 @@
...
+    {
+      evaluate = true;
+      ++counter;
+    }
+  }
+  else
+  {
+    AssertThrow(false,
+                dealii::ExcMessage("Invalid parameters specified. Use either "
+                                   "trigger_interval > 0.0 or trigger_every_time_steps > 0."));
```
nfehn, line 123: we have to think about this error message.

include/exadg/postprocessor/pointwise_output_generator_base.cpp
```diff
@@ -127,6 +126,12 @@ PointwiseOutputGeneratorBase<dim, Number>::setup_base(
 #ifdef DEAL_II_WITH_HDF5
   pointwise_output_data = pointwise_output_data_in;
 
+  time_control.reinit(pointwise_output_data_in.write_output,
+                      pointwise_output_data_in.start_time,
+                      pointwise_output_data_in.end_time,
+                      pointwise_output_data_in.interval_time,
+                      0);
```
nfehn, line 133: why 0?

include/exadg/incompressible_navier_stokes/postprocessor/output_generator.cpp
```diff
@@ -132,8 +131,13 @@ OutputGenerator<dim, Number>::setup(NavierStokesOperator const &    navier_stoke
   mapping                = &mapping_in;
   output_data            = output_data_in;
 
+  time_control.reinit(output_data_in.write_output,
+                      output_data_in.start_time,
+                      std::numeric_limits<double>::max(),
+                      output_data_in.interval_time,
+                      dealii::numbers::invalid_unsigned_int);
```
nfehn, line 138: we should only use ExaDG types here. Internally, `ExaDG::numbers::time_step` might potentially use `dealii::numbers::invalid_unsigned_int`.

nfehn:
The big open question for this PR is which default values to choose for `trigger_interval` and `trigger_time_steps`. Before, we had negative values to deactivate a certain option. However, this does not work if we use `unsigned int` as a type. We could potentially switch to `std::numeric_limits::max()` to deactivate a certain option (since this value will never be reached during time integration).

include/exadg/postprocessor/time_control.cpp
```diff
@@ -0,0 +1,142 @@
...
+#include <exadg/postprocessor/time_control.h>
+
+namespace ExaDG
+{
+TimeControlData::TimeControlData()
+  : is_active(false),
+    start_time(std::numeric_limits<double>::max()),
+    end_time(std::numeric_limits<double>::max()),
+    trigger_interval(-1.0),
+    trigger_every_time_steps(dealii::numbers::invalid_unsigned_int)
```
nfehn, line 38: we need to rethink these default values

jh66637:
I have been working on this this today but didn't push. I think some of the comments are obsolete. Sorry for not updating before your review

include/exadg/postprocessor/time_control.cpp
jh66637: should be OK

include/exadg/incompressible_navier_stokes/postprocessor/output_generator.cpp
jh66637: I did that already.

jh66637:
I still have some stuff to do to finish here, let me ping you once i think this is ready

include/exadg/postprocessor/pointwise_output_generator_base.cpp
jh66637: this was just to ensure everything compiles before refracturing

include/exadg/postprocessor/time_control.cpp
jh66637: should be done

include/exadg/postprocessor/output_generator_scalar.cpp
jh66637: should be done

include/exadg/utilities/numbers.h
jh66637: I think we have to make sure that steady_timestep and invalid_unsigned_int is not the same to avoid confusion. Therefore, steady_timestep can not be std::nl::max(). I just added the invalid_timestep to ensure it is clear that the values is occupied but we can also get rid of it.

include/exadg/utilities/numbers.h
jh66637: Same as next comment. I just did this to be consistent but I didnt use it.

applications/incompressible_navier_stokes/orr_sommerfeld/include/perturbation_energy.h
jh66637: done

jh66637:
@nfehn: I still have to care about the tests and check if something unexpected (obviously the output changes) triggers in the tests, but other than that, I am happy with the current state.
I'm not sure with which clang-format version the code is checked. I seem to have a different version locally, and I don't want to fix everything by hand for this PR :smile:

peterrum:
Use https://github.com/hyperdeal/hyperdeal/blob/master/contrib/utilities/download_clang_format.

nfehn:
We should not mix indentation changes with code changes in this PR, and we should introduce this as a separate PR and not make references to hyperdeal here.

jh66637:
I think there's a missunderstanding. Indentation is the same as requires by ExaDG. It's just that hyper deal provides a script do get the correct binary Also used by the indentation test of ExaDG. If you are referring to the commit message: I will squash everything together anyways.

include/exadg/incompressible_navier_stokes/postprocessor/postprocessor.h
```diff
@@ -73,17 +75,47 @@ class PostProcessor : public PostProcessorBase<dim, Number>
...
   MPI_Comm const mpi_comm;
 
 private:
+  void
+  initialize_additional_vectors();
+
+  // TODO: FOR A MODULAR DESIGN THERE SHOULD PROBALY BE A CLASS CALLED MEAN_VECTOR_CALCULATION IN
+  // WHICH RELEVANT SAMPLES CAN JUST BE SUBMITTED.
+  void
+  compute_mean_velocity(VectorType & mean_velocity, VectorType const & velocity);
```
jh66637, line 93: I think there should be a calculator for this. But I don't want to do it in this PR since it is already quite big.

include/exadg/incompressible_navier_stokes/postprocessor/postprocessor.cpp
```diff
@@ -101,53 +104,267 @@ PostProcessor<dim, Number>::setup(Operator const & pde_operator)
...
+                                              VectorType const &     pressure,
+                                              double const           time,
+                                              types::time_step const time_step_number)
 {
+  // TODO: NOTE THAT CALLING THE FUNCTION HERE RESULTS IN SLOWER CODE. THIS IS REPORTED AN RESOLVED
+  // IN PR #245. SO WE SHOULD ONLY MERGE THIS PR IF ALSO #245 IS READY
+  /*
+   * calculate derived quantities such as vorticity, divergence, etc.
+   */
+  calculate_additional_vectors(velocity, time, time_step_number);
```
jh66637, line 117: I pulled out the calculation of the additional vectors such that the interface of evaluate() can be as in all other cases. This is not a good idea without https://github.com/exadg/exadg/pull/245 since now additional vectors are computed every time-step. I think we should keep it this way and merge this PR and https://github.com/exadg/exadg/pull/245 at the same time, once https://github.com/exadg/exadg/pull/245 is ready.

include/exadg/incompressible_navier_stokes/postprocessor/postprocessor.cpp
```diff
@@ -101,53 +104,267 @@ PostProcessor<dim, Number>::setup(Operator const & pde_operator)
...
+  }
+
+  if(pp_data.output_data.mean_velocity.is_active == true)
+  {
+    // TODO: THIS IS A QUESTION IN GENERAL: SHOULD THIS ASSERT BE CHECKED INSIDE OR OUTSIDE? I TEND
+    // TO DO IT INSIDE THE FUNCTIONS SINCE THE INTERFACE WILL LOOK THE SAME AND THE USER DOES NOT
+    // HAVE TO CARE DURING IMPLEMENTATION
+    AssertThrow(Utilities::is_unsteady_timestep(time_step_number),
+                dealii::ExcMessage(
+                  "Calculating mean velocity does not make sense for unsteady problems."));
```
jh66637, line 350: I didn't do this in a consequent way, since it was already inconsistent before. I would be in favor to move those asserts to the relevant ```evaluate()``` functions, which then all need a ```bool const unsteady```. This way the interface is more general, and we don't have to think about writing an assent if a certain calculator is reused somewhere elese.

jh66637:
One of the two relates PRs seems to break compute_mean_velocity. I suppose this is not tested. I think best is to compute the turbulent channel for compressible and incompressible with all output on. With this we should be safe, nothing breaks. Can you confirm this @nfehn

jh66637:
I fixed the mean velocity output with [0d443d3](https://github.com/exadg/exadg/pull/262/commits/0d443d37e2f5fba50265e38f3823c534cc50afab)

applications/incompressible_navier_stokes/beltrami/application.h
```diff
@@ -247,27 +247,29 @@ class Application : public ApplicationBase<dim, Number>
...
-    pp_data.error_data_u.error_calc_interval_time = (end_time - start_time) / 10;
-    pp_data.error_data_u.name                     = "velocity";
+    pp_data.error_data_u.name = "velocity";
 
     // ... pressure error
-    pp_data.error_data_p.analytical_solution_available = true;
+    pp_data.error_data_p.time_control_data.is_active        = true;
+    pp_data.error_data_p.time_control_data.start_time       = start_time;
+    pp_data.error_data_p.time_control_data.trigger_interval = (end_time - start_time) / 10.0;
+    pp_data.error_data_p.analytical_solution_available      = true;
```
nfehn, line 270: `analytical_solution_available` and `time_control_data.is_active` is a redundancy in my opinion and should be avoided.

include/exadg/incompressible_navier_stokes/postprocessor/line_plot_calculation_statistics.cpp
```diff
@@ -50,27 +50,33 @@ LinePlotCalculatorStatistics<dim, Number>::LinePlotCalculatorStatistics(
...
-LinePlotCalculatorStatistics<dim, Number>::setup(
-  LinePlotDataStatistics<dim> const & line_plot_data_in)
+LinePlotCalculatorStatistics<dim, Number>::setup(LinePlotData<dim> const & line_plot_data_in)
 {
   // initialize data
   data = line_plot_data_in;
 
-  if(data.statistics_data.calculate == true)
+  AssertThrow(
+    line_plot_data_in.time_control_data.trigger_every_sub_time_step !=
```
nfehn, line 59: In my opinion, `trigger_every_sub_time_step` should not be integrated into `TimeControl`. It does not have a meaning for most use cases of `TimeControl`. Also the name should be clearer, e.g. `write_to_file_every_time_steps`.

nfehn: An alternative is to introduce `TimeControlStatistics` (containing an additional variable) building upon `TimeControl` via inheritance or composition.

include/exadg/incompressible_navier_stokes/postprocessor/line_plot_data.h
```diff
@@ -170,48 +193,6 @@ struct LineHomogeneousAveraging : Line<dim>
...
 };
 
-struct StatisticsData
-{
-  StatisticsData()
-    : calculate(false),
-      sample_start_time(0.0),
-      sample_end_time(1.0),
-      sample_every_timesteps(1),
-      write_output_every_timesteps(100)
```
nfehn, line 180: Because of this variable, I think we should introduce `TimeControlStatistics`. `TimeControl` should only contain the minimal set of parameters.

include/exadg/incompressible_navier_stokes/postprocessor/line_plot_calculation_statistics.cpp
jh66637: This is fine with me

nfehn: composition is probably easier to realize because of `LinePlotData` and `LinePlotDataStatistics`.

jh66637: I think I got rid of ```LinePlotDataStatistics```

include/exadg/incompressible_navier_stokes/postprocessor/postprocessor.cpp
nfehn: That a good idea. But it should read "steady problems" in the message, right?

include/exadg/incompressible_navier_stokes/postprocessor/postprocessor.h
nfehn: Good idea. Definitely not in this PR, full agreement from my side.

include/exadg/postprocessor/error_calculation.cpp
```diff
@@ -119,59 +114,40 @@ ErrorCalculator<dim, Number>::setup(dealii::DoFHandler<dim> const &   dof_handle
...
 ErrorCalculator<dim, Number>::evaluate(VectorType const & solution,
-                                       double const &     time,
-                                       int const &        time_step_number)
+                                       double const       time,
+                                       bool const         unsteady)
 {
+  AssertThrow(
+    error_data.analytical_solution_available,
+    dealii::ExcMessage(
+      "Function can only be called if analytical solution is given. If you don't have an analytical solution you probably accidently set time_control.is_active or didn't check if the function should be called."));
```
nfehn, line 132: From a user perspective, I would like to avoid this redundancy. I think a user having to write an application.h file is irritated about these two parameters and their meaning.

include/exadg/incompressible_navier_stokes/postprocessor/postprocessor.cpp
jh66637: Yes :)

include/exadg/postprocessor/error_calculation.cpp
jh66637: I will check if the analytical solution field was set instead

include/exadg/postprocessor/lift_and_drag_calculation.cpp
```diff
@@ -101,6 +122,13 @@ void
   Force = dealii::Utilities::MPI::sum(Force, mpi_comm);
 }
 
+void
+LiftAndDragData::print(dealii::ConditionalOStream & pcout, bool const unsteady) const
+{
+  free_print(pcout, unsteady, *this);
```
nfehn, line 128: why is this realized like this and not like it is done for other modules?

include/exadg/postprocessor/pointwise_output_generator_base.cpp
```diff
@@ -127,7 +87,13 @@ PointwiseOutputGeneratorBase<dim, Number>::setup_base(
 #ifdef DEAL_II_WITH_HDF5
   pointwise_output_data = pointwise_output_data_in;
 
-  if(pointwise_output_data.write_output && pointwise_output_data.evaluation_points.size() > 0)
+  AssertThrow(pointwise_output_data_in.time_control_data.counter_start == 0,
```
nfehn, line 90: what is `counter_start`? This looks also like a solution for a special problem inserted into a rather generic data structure.

include/exadg/postprocessor/pointwise_output_generator_base.cpp
```diff
@@ -136,9 +102,10 @@ PointwiseOutputGeneratorBase<dim, Number>::setup_base(
     componentwise_result.reinit(pointwise_output_data.evaluation_points.size());
 
     // number of samples to write into file
-    n_out_samples = 1 + static_cast<unsigned int>(std::ceil(
-                          (pointwise_output_data.end_time - pointwise_output_data.start_time) /
-                          pointwise_output_data.interval_time));
+    n_out_samples = 1 + static_cast<unsigned int>(
+                          std::ceil((pointwise_output_data.time_control_data.end_time -
+                                     pointwise_output_data.time_control_data.start_time) /
+                                    pointwise_output_data.time_control_data.trigger_interval));
```
nfehn, line 108: This looks like `trigger_every_time_steps` of `TimeControl` cannot be used for this module? I think this calls for a special solution for this particular postprocessing module.

include/exadg/postprocessor/lift_and_drag_calculation.cpp
jh66637: I will change it to be consistent. I am just used to have free prints instead of members and didn't think about the ExaDG conventions during implementation.

include/exadg/postprocessor/pointwise_output_generator_base.cpp
jh66637: It's the counter's initial value.  I think it is generic to be able to specify the initial value of the counter, even though one might not need it often.

include/exadg/postprocessor/pressure_difference_calculation.cpp
```diff
@@ -25,10 +25,37 @@
...
 #include <exadg/postprocessor/pressure_difference_calculation.h>
 #include <exadg/utilities/create_directories.h>
+#include <exadg/utilities/print_functions.h>
 #include <exadg/vector_tools/point_value.h>
 
 namespace ExaDG
 {
+template<int dim>
+void
+free_print(dealii::ConditionalOStream &        pcout,
```
nfehn, line 35: I would prefer to realize this as a member function.

include/exadg/postprocessor/time_control.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+namespace ExaDG
+{
+TimeControlData::EvalType
+get_unsteady_evaluation_type(TimeControlData const & data)
+{
+  bool const interval = (data.trigger_interval > 0.0);
+  bool const timestep = (data.trigger_every_time_steps != dealii::numbers::invalid_unsigned_int);
+
+  // only one of bothe can be true
```
nfehn, line 40: typo -> both

include/exadg/postprocessor/time_control.h
```diff
@@ -0,0 +1,99 @@
...
+  TimeControlData();
+
+  bool         is_active;
+  double       start_time;
+  double       end_time;
+  double       trigger_interval;
+  unsigned int trigger_every_time_steps;
+  unsigned int trigger_every_sub_time_step;
+  bool         force_final_evaluation;
+  unsigned int counter_start;
```
nfehn, line 50: I believe these parameters three should not be part of `TimeControlData`

include/exadg/postprocessor/pressure_difference_calculation.cpp
jh66637: [e482ac8](https://github.com/exadg/exadg/pull/262/commits/e482ac8101da49a4f0c27e76d4da9a75f683a836)

include/exadg/postprocessor/lift_and_drag_calculation.cpp
jh66637: [e482ac8](https://github.com/exadg/exadg/pull/262/commits/e482ac8101da49a4f0c27e76d4da9a75f683a836)

include/exadg/postprocessor/time_control.h
nfehn: I am currently not sure about `force_final_evaluation`. The problem with `counter_start` is that the counter is reset automatically in case of a restart for example. So I see a redundancy here. As written elsewhere, `sub_time_step` is related to statistical averaging and should be realized somewhere else in my opinion.

include/exadg/postprocessor/error_calculation.cpp
jh66637: [52282d1](https://github.com/exadg/exadg/pull/262/commits/52282d18bb5e741a06b37576a457006882c8a534)

include/exadg/postprocessor/time_control.h
```diff
@@ -0,0 +1,99 @@
...
+#include <deal.II/base/conditional_ostream.h>
+
+// ExaDG
+#include <exadg/utilities/numbers.h>
+
+namespace ExaDG
+{
+struct TimeControlData
+{
+  TimeControlData();
```
nfehn, line 41: where are the default values? trigger_interval and trigger_every_time_steps should be set to invalid values by default in my opinion.

include/exadg/postprocessor/time_control.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+
+  return evaluate;
+}
+
+unsigned int
+TimeControl::get_counter() const
+{
+  // the counter is incremented during needs_evaluation(), thus counter is ahead by 1
+  AssertThrow(counter - 1 != dealii::numbers::invalid_unsigned_int,
```
nfehn, line 196: is problematic if counter is of type unsigned. Did you think about this?

include/exadg/utilities/numbers.h
nfehn: Regarding this discussion: what is in your opinion the default value for `trigger_every_time_steps`, which should be an "invalid" value in my opinion as written in another comment?
For completeness: I realized this before by using `int` instead of `unsigned int` and using `-1` as an invalid default value.

include/exadg/postprocessor/time_control.h
jh66637: They are set in the constructor in the cpp file.

include/exadg/postprocessor/time_control.cpp
jh66637: Why is it problematic? Overflows are caught by the assert.

include/exadg/postprocessor/time_control.h
```diff
@@ -0,0 +1,99 @@
...
+
+  bool         is_active;
+  double       start_time;
+  double       end_time;
+  double       trigger_interval;
+  unsigned int trigger_every_time_steps;
```
nfehn, line 47: I would argue that you have to use `types::time_step` here, both variables are combined with each other via mathematical operators (addition, modulo) in `needs_evaluation`. The current implementation allows to use different types, implying implicit type conversions. This could potentially become very ugly in terms of steady_timestep, invalid_timestep, etc. and the values choosing for these constants.

include/exadg/utilities/numbers.h
jh66637: trigger_every_time_steps is a number and not a time step. It is set invalid in the constructor. I don't see a problem in this regard.

include/exadg/postprocessor/time_control.h
jh66637: I would argue that trigger_every_time_step is not a time_step but a number. And we are safe to do everything if we keep time_step also an unsigned int, which makes sense since a negative time_step can not exist. This is also true for the steady case and this is why I implemented the numbers::invalid_time_step in the first place.

include/exadg/postprocessor/time_control.cpp
nfehn: what exactly is behind dealii::numbers::invalid_unsigned_int?

jh66637: static_cast<unsigned  int>(-1)

include/exadg/postprocessor/pointwise_output_generator_base.cpp
nfehn: See arguments given here: https://github.com/exadg/exadg/pull/262#discussion_r1000353552

So it is a redundancy. A user does not know whether he was to set `counter_start` correctly in case of a restart, even though this is done automatically inside the code. For me, clarity from a user perspective must be a primary goal. Also, you do not take `counter_start` into account when resetting the counter in `needs_evaluation()`. This is at least an inconsistency/bug, even if we agreed to introduce `counter_start`.

include/exadg/postprocessor/time_control.h
nfehn: You argue that every_time_steps is a number (which would allow a negative value), but at the same time you argue with time_step, which cannot be negative. So I think this point is unclear so far?

jh66637: Its a positive number. Evaluating every -10th time-step does not make any sense.

include/exadg/postprocessor/time_control.cpp
nfehn: I see. So this is a "complicated" way to write `counter > 0`?

nfehn: Again, I believe this implementation would have to take care of `counter_start` (but maybe I misunderstand the code?), a parameter which I am against introducing actually (also for such reasons).

include/exadg/postprocessor/time_control.h
nfehn: I do not want to dispute this. But for me this implies it has to be of type `time_step` also because you do calculations between both variables. Put differently, why do you introduce `types::time_step` here if you do not want to make use of the type in the sense that we might change the concrete type in the future, where everything still has to work in such a case.

include/exadg/postprocessor/time_control.cpp
jh66637: In this particular case, yes. I can change the assert.
I think this has nothing to do with counter_start. The counter is also invalid if the user forgot to call ```needs_evaluation()``` if the counter is 0 in the beginning. But this is OK, since we agreed that we can expect the user to call the function before any evaluation.

include/exadg/postprocessor/time_control.h
jh66637: I can't think about any possible reason that this changes. Nevertheless:
[5e6d7a9](https://github.com/exadg/exadg/pull/262/commits/5e6d7a969014aa50aaba54a91cffe275f7145346)

include/exadg/utilities/numbers.h
jh66637: Now that also trigger_every_timestep is of type types::time_step this function is used

include/exadg/incompressible_navier_stokes/postprocessor/postprocessor.cpp
jh66637: [1c3277b](https://github.com/exadg/exadg/pull/262/commits/1c3277b32be3cc0569f9ebeb89a79db1744ef87d)

include/exadg/postprocessor/time_control.h
jh66637: I think expressing tolerance in terms of ```end_time - start_time``` will not help since the critical value here is ```time_step_size```. We could cache ```last_time``` and determine the ```time_step_size``` with this. 

If this is not an argument for enforce_final_evaluation, which purpose has the variable?

nfehn: maybe also `time_step_size`: `if(time + 0.01*time_step_size > end_time) -> needs_evaluation = true;`

nfehn: But is this realistic? Say a time step size cannot be solved faster than t_wall,dt=1e-5 (hardware limits), then such a simulation with 1e12 time steps still runs t_wall=1e7 seconds, or 115 days! I am willing to find a better solution ...

include/exadg/incompressible_navier_stokes/postprocessor/line_plot_calculation_statistics.cpp
jh66637: Yes, but I can undo it if trigger_every_sub_time_step should live somewhere else

nfehn: I think we need

```c++
TimeControlStatistics
{

bool write_results()
{
  if(time_control.get_counter() % write_preliminary_results_every_nth_evaluation or reached_end_time())
  { return true; }
}

private:
TimeControl time_control;
double write_preliminary_results_every_nth_evaluation; 

};

Comment (nfehn) [reply include/exadg/incompressible_navier_stokes/postprocessor/line_plot_calculation_statistics.cpp:59]: This version only requires to define another Data class, while the class `TimeControl` would be reused for statistics problems:

```c++
TimeControlStatisticsData
{
private:
TimeControlData time_control_data;
unsigned int write_preliminary_results_every_nth_evaluation; 

};

// not sure where to realize this exactly?
bool write_results()
{
  if(time_control.get_counter() % statistics_data.write_preliminary_results_every_nth_evaluation
     or time_control.reached_end_time())
  { do_write_results; }
}
```

jh66637: Sth like this? [b6a68c1](https://github.com/exadg/exadg/pull/262/commits/b6a68c149459d736ac598093701f85b89118f7a8)

nfehn: Yes, looks good overall. I am not sure if `write_preliminary_results()` is understandable from a very general `TimeControl` perspective, but we can still change this. This new solution has the advantage that it does not duplicate code.

include/exadg/postprocessor/time_control.h
jh66637: Let's keep it a fixed tolerance and address it some other time

include/exadg/postprocessor/pointwise_output_generator_base.cpp
jh66637: The module uses TimeControl. We just have to know before writing how much memory we need to allocate. Therefore, I use the data that is in TimeControl. You are right, this is only implemented for time intervals. This is because we do not know how much memory to allocate for the time step case and adaptive timestepping.

If it is OK I will add an assert and keep the rest

nfehn: Yes, that good. I think an assert will avoid errors.

jh66637: [68b7a04](https://github.com/exadg/exadg/pull/262/commits/68b7a047785a84d53e2c2a7ef610e44ad09c7993)

include/exadg/postprocessor/time_control.cpp
```diff
@@ -0,0 +1,162 @@
...
+  {
+    if((time_step_number - 1) % time_control_data.trigger_every_time_steps == 0)
+    {
+      evaluate = true;
+      ++counter;
+    }
+  }
+  else if(get_unsteady_evaluation_type(time_control_data) == TimeControlData::EvalType::Interval)
+  {
+    // In case of a restart, time might be significantly larger than start_time in the first time
```
nfehn, line 126: this increases the counter in case of a restart, starting e.g. with `counter = 13` (due to a previous simulation from which we want to restart), which is then incremented by one again and again during the simulation.
Hence, can we remove counter_start?

jh66637: Absolutely, I thought the user has to do it manually

jh66637:
I cleaned up and squashed the changes after code review into a new commit. If you approve all the changes, I will squash both commits before merging.

applications/incompressible_navier_stokes/periodic_hill/include/postprocessor.h
```diff
@@ -104,7 +104,13 @@ class MyPostProcessor : public PostProcessor<dim, Number>
     }
 
     // line plot statistics
-    line_plot_calculator_statistics->evaluate(velocity, pressure, time, time_step_number);
+    if(line_plot_calculator_statistics->time_control_statistics.needs_evaluation(time,
+                                                                                 time_step_number))
+      line_plot_calculator_statistics->evaluate(
```
nfehn, line 109: I such cases I would introduce brackets since its a bit hard to read in my opinion. Just as a minor comment for the future. No need to realize this here before merging.

include/exadg/compressible_navier_stokes/postprocessor/postprocessor.cpp
```diff
@@ -101,36 +102,52 @@ PostProcessor<dim, Number>::do_postprocessing(VectorType const & solution,
...
+                                       Utilities::is_unsteady_timestep(time_step_number));
 
   /*
    *  calculation of kinetic energy spectrum
    */
-  kinetic_energy_spectrum_calculator.evaluate(velocity, time, time_step_number);
+  if(kinetic_energy_spectrum_calculator.time_control.needs_evaluation(time, time_step_number))
+    kinetic_energy_spectrum_calculator.evaluate(velocity,
+                                                time,
+                                                Utilities::is_unsteady_timestep(time_step_number));
```
nfehn, line 150: From these lines, I get the impression that the check whether it is a steady or unsteady time step should be done outside the postprocessor, and `TimeControl` should not take care of this steady/unsteady check, but should address unsteady, time-dependent problems only. Again, this should be done in a subsequent PR. I just write this here so that we do not forget about this.

include/exadg/incompressible_navier_stokes/postprocessor/line_plot_data.h
```diff
@@ -242,55 +226,42 @@ struct LinePlotData
...
   {
-    line_data.print(pcout);
-
-    if(statistics_data.calculate)
+    if(time_control_data_statistics.time_control_data.is_active)
     {
-      statistics_data.print(pcout);
+      this->print_base();
+      // only makes sense in unsteady case
+      time_control_data_statistics.time_control_data.print(pcout, true);
```
nfehn, line 259: This implies that `write_preliminary_results_...` will not be printed. Ideally, we would also introduce a print function for `time_control_data_statistics`.

include/exadg/incompressible_navier_stokes/postprocessor/postprocessor.cpp
nfehn: I think the computations are only invoked if the output module is activated. So I think this is acceptable as an intermediate solution.

include/exadg/postprocessor/error_calculation.h
```diff
@@ -52,24 +49,19 @@ struct ErrorCalculationData
   void
   print(dealii::ConditionalOStream & pcout, bool unsteady)
   {
-    print_parameter(pcout, "Calculate error", analytical_solution_available);
-    if(analytical_solution_available == true && unsteady == true)
+    print_parameter(pcout, "Error calculation", unsteady == true && analytical_solution);
+    if(unsteady == true && time_control_data.is_active)
```
nfehn, line 53: I am not sure why this is done only for unsteady problems. Another reminder for the future ...

include/exadg/incompressible_navier_stokes/postprocessor/line_plot_data.h
jh66637: The print function is there. I just didn't call it correctly. 
[1be0d8e](https://github.com/exadg/exadg/pull/262/commits/1be0d8ee3d133733e1e6a3c80f10cb3142708837)

include/exadg/postprocessor/error_calculation.h
jh66637: This was also not clear for me.

applications/incompressible_navier_stokes/periodic_hill/include/postprocessor.h
jh66637: I prefer not to care about this lateron:
[0989d33](https://github.com/exadg/exadg/pull/262/commits/0989d3386318e0be33f226c31c569eb4519196af)

include/exadg/postprocessor/time_control.cpp
```diff
@@ -0,0 +1,183 @@
...
+    // step of the restarted simulation. Hence, we have to reconstruct the counter at the end of the
+    // previous simulation, in order to obtain the correct behavior at the beginning of the
+    // restarted simulation.
+    if(reset_counter)
+    {
+      unsigned int counter_increment =
+        ((time - time_control_data.start_time + EPSILON) / time_control_data.trigger_interval);
+
+      if(counter_increment > 0)
+        simulation_was_restarted = true;
```
nfehn, line 140: The name `simulation_was_restarted` is a bit misleading. It can be completely fine that a restarted simulation has counter_increment=0. So we have to be careful with the naming. TimeControl can not provide information about restart or not.

include/exadg/postprocessor/time_control.h
```diff
@@ -0,0 +1,103 @@
...
+struct TimeControlData
+{
+  TimeControlData();
+
+  bool             is_active;
+  double           start_time;
+  double           end_time;
+  double           trigger_interval;
+  types::time_step trigger_every_time_steps;
+  bool             force_final_evaluation;
```
nfehn, line 48: can be removed

include/exadg/postprocessor/time_control.h
```diff
@@ -0,0 +1,103 @@
...
+
+  bool
+  is_restarted() const;
+
+private:
+  // small number which is much smaller than the time step size
+  double const         EPSILON;
+  mutable bool         reset_counter;
+  mutable unsigned int counter;
+  mutable bool         forced_final_evaluation;
```
nfehn, line 92: can be removed

include/exadg/postprocessor/time_control.cpp
jh66637: In that case: Any idea how to check if a simulation is restarted in PointwiseOutputGenerator?

include/exadg/postprocessor/time_control_statistics.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+bool
+TimeControlStatistics::write_preliminary_results(types::time_step const time_step_number) const
+{
+  if(Utilities::is_valid_timestep(
+       time_control_data_statistics.write_preliminary_results_every_nth_time_step))
+    if((time_step_number - 1) %
+           (time_control_data_statistics.write_preliminary_results_every_nth_time_step) ==
+         0 ||
+       time_control.reached_end_time())
```
nfehn, line 76: I think this is the wrong logic: consider the time interval t = [0, 50] for a simulation, with sampling performed during the interval t=[20,50] (a typical setting for turbulent flows). The present code also wants to write output for the time interval t=[0,20], which does not make sense. The logic should be like this

`if((time_control.needs_evaluation() and (time_control.get_counter() % write_preliminary_every_nth_evaluation == 0) or time_control.reached_end_time())`

Note the naming `nth_evaluation` instead of `nth_time_step`! `nth_evaluation` makes sure that we only write new output if the statistics evaluation has been invoked before in the same time step.

`time_control.needs_evaluation()` in the above if-statement makes sure that we only consider the interval t=[20,50].

The problem with this is that `needs_evaluation()` will again increment the counter inside `TimeControl`. This calls for incrementing the counter manually to avoid such side-effects.

include/exadg/postprocessor/time_control_statistics.cpp
```diff
@@ -0,0 +1,83 @@
...
+}
+
+bool
+TimeControlStatistics::needs_evaluation(double const           time,
+                                        types::time_step const time_step_number) const
+{
+  if(time_control.needs_evaluation(time, time_step_number))
+    return true;
+  if(write_preliminary_results(time_step_number))
+    return true;
```
nfehn, line 57: I think we should not mix the logic like this. It is unclear in my opinion what this function does. Writing of preliminary results should only be invoked if `needs_evaluation()` returned true, so the logic has to be kind of nested. This should also allow to avoid the "double increment problem" described here https://github.com/exadg/exadg/pull/262/files#r1001517833. If you nest the functionality, there is not need to call `needs_evaluation()` a second time, with the side effect of incrementing the counter again.

nfehn:
I looked again through all the changes. In summary, there are two problems remaining: 1) `simulation_was_restarted` cannot be provided by `TimeControl`, 2) the logic of how `TimeControlStatistics` works needs to be changed in my opinion.

include/exadg/postprocessor/pointwise_output_generator_base.cpp
```diff
@@ -68,52 +59,26 @@ template class PointwiseOutputDataBase<3>;
...
-      write_time(time);
-      do_evaluate(solution);
-      ++counter;
-    }
-  }
+  AssertThrow(unsteady, dealii::ExcMessage("Only implemented for the unsteady case."));
+
+  AssertThrow(time_control.is_restarted() == false,
+              dealii::ExcMessage(
+                "Only implemented in the case that the simulation is not restarted"));
```
nfehn, line 69: To get rid of `is_restarted()`, check that `time_control.get_counter() == 0` in the first time step of a simulation.

include/exadg/postprocessor/time_control.cpp
nfehn: see here https://github.com/exadg/exadg/pull/262/files#r1001530656

include/exadg/postprocessor/time_control_statistics.cpp
nfehn: See also the comment here https://github.com/exadg/exadg/pull/262#discussion_r1001521153

include/exadg/postprocessor/time_control_statistics.cpp
jh66637: needs_evaluation() should also return true if no result is written, but an evaluation is needed inside the evaluate() function.  I will rewrite some code, but the interface should stay the same in my opinion.

nfehn: The problem with the current realization is that the evaluation of the postprocessing routine (`do_evaluate()`) is invoked whenever `write_preliminary_results()` returns true. But we do not want to have this, which is clearly a side effect. If you need - in addition to the above function - separate if-statements inside `evaluate()`, like

```c++
evaluate()
{
if(time_control.needs_evaluation())
  do_evaluate();
if(write_preliminary_results())
  do_write_results();
}
```
there is no use in defining `TimeControlStatistics::needs_evaluation()` in my opinion. You can write the above two if-statements in `Postprocessor::do_postprocessing()` directly and the logic is much clearer.

include/exadg/incompressible_navier_stokes/postprocessor/line_plot_calculation_statistics_homogeneous.cpp
```diff
@@ -314,44 +316,20 @@ LinePlotCalculatorStatisticsHomogeneous<dim, Number>::setup(
...
-    }
-
-    // write final output
-    if((time > data.statistics_data.sample_end_time - EPSILON) && write_final_output)
-    {
-      do_write_output();
-      write_final_output = false;
-    }
-  }
+  if(write_statistics)
```
nfehn, line 331: Suggestion 1: Write 

```c++
if(time_control.get_counter() % write_preliminary_results_every_nth_evaluation == 0 or 
   time_control.reached_end_time())
  do_write_output();
```

`TimeControlStatistics` should not implement a function called `needs_evaluation()` in my opinion. With this suggestion, you might need to adjust the if-statement in the postprocessor
```c++
if(time_control.needs_evaluation() or time_control.reached_end_time())
  postprocessing_module->evaluate();
```

This implements the logic described here https://github.com/exadg/exadg/pull/262#discussion_r1001521153.

Suggestion 2: The alternative is to completely separate the functionality of `do_evaluate()` and `write_preliminary_results(). This would lead to

```c++
if(time_control.needs_evaluation())
  postprocessing_module->evaluate();

if(time_control_statistic.write_results())
  postprocessing_module->write_results();
```

jh66637:
I think I addressed all concerns. If you think this can be merged, I will squash everything. If there is still something to do, let me know.

include/exadg/postprocessor/time_control.h
```diff
@@ -0,0 +1,99 @@
...
+
+class TimeControl
+{
+public:
+  TimeControl();
+
+  void
+  setup(TimeControlData const & time_control_data);
+
+  /// If reinit() is not called needs_evaluation() will always evaluate to false
```
nfehn, line 71: this comment should be removed I guess

include/exadg/postprocessor/time_control.cpp
```diff
@@ -0,0 +1,173 @@
...
+{
+}
+
+void
+TimeControl::setup(TimeControlData const & time_control_data_in)
+{
+  time_control_data = time_control_data_in;
+}
+
+// If setup() is not called needs_evaluation() will always evaluate to false
```
nfehn, line 87: Don't know if we need this comment here. I think it should actually be quite clear.

include/exadg/postprocessor/time_control.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+namespace ExaDG
+{
+TimeControlData::EvalType
+get_unsteady_evaluation_type(TimeControlData const & data)
+{
+  bool const interval = (data.trigger_interval > 0.0);
+  bool const timestep = Utilities::is_valid_timestep(data.trigger_every_time_steps);
+
+  // only one of both can be true
```
nfehn, line 40: This comment irritates me, because of "can". A user can specify both variables in "application.h". I would better say "We do not allow users to specify both a time interval and a number of time steps."

include/exadg/postprocessor/time_control.cpp
```diff
@@ -0,0 +1,173 @@
...
+  if(!Utilities::is_unsteady_timestep(time_step_number))
+  {
+    ++counter;
+    return true;
+  }
+
+  // unsteady evaluation
+  AssertThrow(get_unsteady_evaluation_type(time_control_data) != TimeControlData::EvalType::Invalid,
+              dealii::ExcMessage(
+                "Unsteady evaluation not possible since TimeControlData::EvalType::Invalid"));
```
nfehn, line 104: This is hard to interpret for a user, since a user does not know what ``TimeControlData::EvalType::Invalid`` is internally. The question is do we need the assert here or would it be better to place the assert in `get_unsteady_evaluation_type()`? Following this strategy, no comment would be needed in `get_unsteady_evaluation_type()`, since the assert message already documents the logic. Further, we could get rid of `TimeControlData::EvalType::Invalid`, since we call below `get_unsteady_evaluation_type()`, automatically invoking the exception in case the parameters are invalid.

include/exadg/postprocessor/time_control_statistics.cpp
```diff
@@ -0,0 +1,85 @@
...
+      return true;
+    }
+
+    if((time_step_number - 1) %
+         (time_control_data_statistics.write_preliminary_results_every_nth_time_step) ==
+       0)
+    {
+      return true;
+    }
+  }
```
nfehn, line 80: I want to make a suggestion below how to write this code more elegantly and safer in my opinion. I do not like so much the style with many returns inside a function. Why? Changing the order of the last two if-statements will in general change the logic of the code (!), but this is very hard to see and to understand for someone that has not written the code. Hence, this bug might easily be introduced in the future during refactoring.

```c++
  if(Utilities::is_valid_timestep(
       time_control_data_statistics.write_preliminary_results_every_nth_time_step))
  {
    // if we are outside the interval bounds
    if((time <
       time_control_data_statistics.time_control_data.start_time - time_control.get_epsilon())
      or final_output_written)
    {
      return false;
     }
    // check whether results have to be written
    else if((time_step_number - 1) %
         (time_control_data_statistics.write_preliminary_results_every_nth_time_step) ==
       0 or time_control.reached_end_time())
    {
      // make sure that we do no longer write results once the final output has been written
      if(time_control.reached_end_time())
        final_output_written = true;

      return true;
    }
  }
```

include/exadg/postprocessor/time_control.cpp
jh66637: The assert can not be in ```get_unsteady_evaluation_type()``` since it is OK to have ```EvalType::Invalid``` if ```is_active=false``` or if it is a steady simulation. I will rename the parameters and change the assert such that it is more clear.

include/exadg/postprocessor/time_control_statistics.cpp
jh66637: Im OK with both versions

jh66637:
I addressed all requested changes and also rebased https://github.com/exadg/exadg/pull/245 such that it could be directly merged after this one.

include/exadg/postprocessor/time_control.cpp
nfehn: I understand the implementation like this: `get_unsteady_evaluation_type()` is never called in case of `Is_active = false` or in case of a steady simulation. Therefore, could you point me to the place in the code where this happens?

jh66637: I currently do not have access to a Computer. But you can look e.g. for the Assert in the setup() function of PointwiseOutput generator if I recall correctly.

nfehn: I looked at pointwise output generator, but do not see a conflict here if `get_unsteady_evaluation_type()` cared about the assert.

include/exadg/postprocessor/pointwise_output_generator_base.cpp
```diff
@@ -127,7 +95,18 @@ PointwiseOutputGeneratorBase<dim, Number>::setup_base(
 #ifdef DEAL_II_WITH_HDF5
   pointwise_output_data = pointwise_output_data_in;
 
-  if(pointwise_output_data.write_output && pointwise_output_data.evaluation_points.size() > 0)
+  AssertThrow(
+    (get_unsteady_evaluation_type(pointwise_output_data_in.time_control_data) ==
+     TimeControlData::UnsteadyEvalType::Interval) ||
+      (get_unsteady_evaluation_type(pointwise_output_data_in.time_control_data) ==
+       TimeControlData::UnsteadyEvalType::None),
```
jh66637, line 102: Here we also allow EvalType::None

include/exadg/postprocessor/time_control.cpp
jh66637: https://github.com/exadg/exadg/pull/262/files#r1008219228

PR #263, opened 2022-10-13
Be explicit about unsteady problems

jh66637:
@nfehn 
Can you check if I forgot something?
Also I didn't find where the timestep is set -1, can you give me a hint?

### DIFF
--- a/applications/incompressible_navier_stokes/orr_sommerfeld/include/perturbation_energy.h
+++ b/applications/incompressible_navier_stokes/orr_sommerfeld/include/perturbation_energy.h
@@ -28,6 +28,7 @@
 // ExaDG
 #include <exadg/matrix_free/integrators.h>
 #include <exadg/utilities/create_directories.h>
+#include <exadg/utilities/numbers.h>
 #include <exadg/utilities/print_functions.h>
 
 namespace ExaDG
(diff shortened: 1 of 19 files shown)


nfehn:
> Also I didn't find where the timestep is set -1, can you give me a hint?

Do you mean at the place where the function is called? I assume the parameter is justed not used in the `DriverSteadyProblem` classes when calling the postprocessor, see e.g. https://github.com/exadg/exadg/blob/b0f0d540fbdc8937fedaf9a5ed3d8930ffdc9464/include/exadg/convection_diffusion/time_integration/driver_steady_problems.cpp#L154, since the parameter `time_step_number` has a default argument. Or do I misunderstand something.

nfehn (approved):

nfehn:
I think we can merge as soon as the code compiles without errors

jh66637:
If you can confirm I didn't miss any place. I'm not confident here^^

nfehn:
I think it looks good. We will go through it again anyway when adjusting the postprocessing modules according to `TimeControl`. Of course, it makes sense to test your new implementation for a single example with postprocessing (e.g. output or error computation).

jh66637:
Is there a test or test case that covers an unsteady problem? If yes, which one, then I can test it

nfehn:
The tests in ExaDG that you can run with ctest should include error calculations after a certain time interval, which would serve as a test here.

PR #264, opened 2022-10-16
Enum utilities

jh66637:
+++ REQUIRES C++17 +++

I thought I share the way I included MagicEnum (https://github.com/Neargye/magic_enum) in other projects to circumvent the boilerplate code for ```enum_to_string()``` and ```string_to_enum()```.

I am not sure if some features are already C++17 only in ExaDG or if this should wait. If users are forced to use C++17 anyway, I would be happy if you think this is worth merging :smile:


peterrum:
I use better_enum in some projects. See https://github.com/peterrum/dealii-parameter-handler-enum as example.

jh66637:
I would prefer to use magic enum if ExaDG is already limited to C++17. This way, there is no need of using BETTER_ENUM macros all over the pace and to define enums in classes.

include/exadg/utilities/enum_utilities.h
```diff
@@ -19,39 +19,39 @@
...
-#include <exadg/solvers_and_preconditioners/preconditioners/enum_types.h>
+#include <deal.II/base/exceptions.h>
 
 namespace ExaDG
 {
-namespace Elementwise
+/**
+ * Collection of enums utilities.
+ */
+namespace Utilities
```
nfehn, line 34: I like that you have wrapped these functions. This should allow to easily change the code in case this enum functionality becomes part of `std` some time in the future.

include/exadg/utilities/print_functions.h
```diff
@@ -82,7 +85,10 @@ template<typename ParameterType>
 void
 print_value(dealii::ConditionalOStream const & pcout, ParameterType const value)
 {
-  pcout << value << std::endl;
+  if constexpr(std::is_enum_v<ParameterType> || magic_enum::is_scoped_enum_v<ParameterType>)
+    pcout << Utilities::enum_to_string(value) << std::endl;
```
nfehn, line 89: Is there a possibility to also hide `std::is_enum_v` and `magic_enum` here (with the goal that all the `magic_enum` occurrences are wrapped in ExaDG? What is the meaning of this line of code (it is hard to read for me right now)? Comments would help here.

jh66637: Sure, I can add e.g. a function ```Utilities::is_enum<Type>()``` and add a comment in its definition

nfehn: sound great

include/exadg/utilities/enum_utilities.h
```diff
@@ -19,39 +19,39 @@
...
-  return string_type;
+template<typename Enum>
+void
+string_to_enum(Enum & enum_type, std::string const & enum_name)
+{
+  auto casted_enum = magic_enum::enum_cast<Enum>(enum_name);
+  if(casted_enum.has_value())
+    enum_type = casted_enum.value();
+  else
+    AssertThrow(false, dealii::ExcMessage("Could not convert string to enum."));
```
jh66637, line 51: TODO: add input string to enum_name

jh66637:
I am currently not using magic enums in combination with input files. If the conversion fails 
https://github.com/exadg/exadg/pull/264/files#diff-727606788c84a76ec8ca28b2e3c28f63db2ff2384d9de6538834f7a7fefcbcf1R47-R51
is triggered. So currently there is no indication which conversion fails. this is however, easy to integrate. I added a TODO for myself. 

In my opinion, it would be perfect to have something like ```ExaDG::Patterns::Enum``` instead of ```dealii::Patterns::Selection``` (which does the work internally). I shortly looked at the code needed for this and couldn't find a fast solution. We could however also wrap the ```add_parameter``` calls. somehow

peterrum:
I have played with `betterenum` and `ParameterHandler` a while ago. This is what I came up for that: https://github.com/peterrum/dealii-parameter-handler-enum/blob/master/include/deal.II/base/patterns_enum.h. It works quite well. Maybe something similar can also be applied to `magicenum`.

jh66637:
If we agree on the design regarding parameter handling, I will apply it to various places in the code before merging.
You can see the current state of the interface in the ```poisson/gaussian``` app. Personally, I like it this way, but I'm happy for feedback.

@peterrum thanks for showing how to work the parameter handler with enums :)

peterrum:
> @peterrum thanks for showing how to work the parameter handler with enums :)

Glad, I could help!

I am in favor to make the move towards magicenum, since it makes life much easier - if we are willing to make some compromises, e.g., no spaces in parameter names! Are there any concerns regarding CPP17? @kronbichler @nfehn?

nfehn:
> I am in favor to make the move towards magicenum, since it makes life much easier - if we are willing to make some compromises, e.g., no spaces in parameter names! Are there any concerns regarding CPP17? @kronbichler @nfehn?

Not from my side, but I am probably not most qualified/experienced to answer this.

nfehn:
Parameter parsing should be an independent PR.

kronbichler:
I am happy to require C++17 if it simplifies our life. That has been around for long enough to be robust and provided by enough compilers. Regarding bundling magicenum, I guess you included the relevant copyright notice etc to make sure we do it right? And indeed, we need to discuss parameter names at some point.

peterrum:
> I am in favor to make the move towards magicenum, since it makes life much easier - if we are willing to make some compromises, e.g., no spaces in parameter names!

> Not from my side, but I am probably not most qualified/experienced to answer this.

> I am happy to require C++17 if it simplifies our life.

I guess this is a "go ahead" from our side. Let us know if you have time to work on this PR in the next days; if now feel free to suggest how others can help!

jh66637:
Thank you for the fast feedback. I think there is not much left to do. I will resolve the merge requests and incorporate some documentation. As suggested by @nfehn I will place a second PR in which the enum integration into the parameter handler can be discussed, but I think @kronbichler and @peterrum (from which the enum integration is motivated) like the direction this is going. I think I will make it until the beginning of next week.

jh66637:
> Regarding bundling magicenum, I guess you included the relevant copyright notice etc to make sure we do it right? 

I think I did it the right way: I added the MIT license file, the README and the header of the project. 

From my side, this is ready to merge. I would be glad to discuss the parameter handling with enums in https://github.com/exadg/exadg/pull/382 soon to get a clean integration of enums also in this regard.

bundled/README.md
```diff
@@ -0,0 +1,6 @@
+# This folder contains third party projects used in AeroAcousticDG
```
peterrum, line 1: ?

bundled/README.md
```diff
@@ -0,0 +1,6 @@
...
+**Included projects have different copy rights compared to ExaDG. Their licences allow each project to be included here.**
```
peterrum, line 2: Q: Why do we bundle? Would git submodule the better approach so that we can update the package when needed? Have you modified the header file somehow?

bundled/README.md
jh66637: I fixed this

bundled/README.md
jh66637: I didn't modify the header at all. Personally, I don't like git submodules too much, since the git commands get lengthy. Since the submodule will probably not need a lot of updates, this is probably not the case here?

Also I don't know how to deal with submodules etc, so I would need some help here.

peterrum: Let's keep it as it is!

peterrum (approved):
Looks very good from my side! Thanks a lot! I would suggest that we merge so that we can proceed with https://github.com/exadg/exadg/pull/382 and https://github.com/exadg/exadg/pull/367!

include/exadg/utilities/enum_patterns.h
```diff
@@ -0,0 +1,75 @@
...
+{
+namespace Patterns
+{
+namespace Tools
+{
+/**
+ * Converter class for structs of type T that fulfill is_convertible<T>.
+ */
+template<class T>
+struct Convert<T, typename std::enable_if<ExaDG::Utilities::is_enum<T>()>::type>
```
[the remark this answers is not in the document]
jh66637: This part of code has been shifted to https://github.com/exadg/exadg/pull/382

peterrum (approved):
If there a no objects, let's merge! This was a tremendous work by @jh66637! Thank you!

nfehn:
I can later take a final look at this and then merge.

bundled/magic_enum/README.md
```diff
@@ -0,0 +1,250 @@
+```text
```
nfehn, line 1: I am not sure whether copying the readme is the correct procedure? The readme text we have here will be out-dated soon, so why not just putting a link to the true readme of magic-enum? A different aspect that I am not sure about is whether we actually have the rights to copy the content of a readme file?

include/exadg/utilities/print_functions.h
nfehn: let us check once more why this discussion has been resolved

bundled/magic_enum/README.md
nfehn: when writing this comment, I have not seen that source code is copied as well. So maybe this is the standard procedure ... Then you can probably forget about this comment.

jh66637: Yes, the README is in sync since the source code is also copied. As far as I know, this is OK since the project is under the MIT Licence. As I mentioned in https://github.com/exadg/exadg/pull/264#discussion_r1133288168 I would need some help if we take a different route (with submodules).

include/exadg/utilities/print_functions.h
jh66637: It seems I introduced mentioned function but did not apply it here. Sorry for that. I fixed this.

jh66637:
@nfehn I rebased and squashed again. If we do not want to copy the code of the TPL, I need some help here.

PR #268, opened 2022-10-31
Simplex support for Poisson operator

necioglu:
Continuing the work in #131 and in #110.

A working copy of the Poisson/Sine application is provided.

### DIFF
--- a/applications/poisson/sine/application.h
+++ b/applications/poisson/sine/application.h
@@ -28,14 +28,6 @@ namespace ExaDG
 {
 namespace Poisson
 {
-enum class GridType
-{
-  Hypercube,
-  Simplex
-};
-
-GridType GRID_TYPE = GridType::Hypercube;
-
 double const FREQUENCY            = 3.0 * dealii::numbers::PI;
 bool const   USE_NEUMANN_BOUNDARY = true;
 
@@ -160,6 +152,7 @@ private:
     // SPATIAL DISCRETIZATION
     this->param.grid.triangulation_type = TriangulationType::Distributed;
     this->param.grid.mapping_degree     = 3;
+    this->param.grid.element_type       = ElementType::Hypercube;
     this->param.spatial_discretization  = SpatialDiscretization::DG;
     this->param.IP_factor               = 1.0e0;
 
@@ -192,14 +185,14 @@ private:
     unsigned int n_cells_1d =
       std::max((unsigned int)2, this->param.grid.n_subdivisions_1d_hypercube);
 
-    if(GRID_TYPE == GridType::Hypercube)
+    if(this->param.grid.element_type == ElementType::Hypercube)
     {
       dealii::GridGenerator::subdivided_hyper_cube(*this->grid->triangulation,
                                                    n_cells_1d,
                                                    left,
                                                    right);
     }
-    else if(GRID_TYPE == GridType::Simplex)
+    else if(this->param.grid.element_type == ElementType::Simplex)
     {
       dealii::GridGenerator::subdivided_hyper_cube_with_simplices(*this->grid->triangulation,
                                                                   n_cells_1d,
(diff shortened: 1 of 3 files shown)


applications/poisson/sine/application.h
```diff
@@ -158,8 +150,9 @@ class Application : public ApplicationBase<dim, n_components, Number>
     this->param.right_hand_side = true;
 
     // SPATIAL DISCRETIZATION
-    this->param.grid.triangulation_type = TriangulationType::Distributed;
-    this->param.grid.mapping_degree     = 3;
+    this->param.grid.triangulation_type = TriangulationType::Serial;
+    this->param.grid.mapping_degree     = 2;
+    this->param.grid.element_type       = ElementType::Simplex;
```
nfehn, line 155: The default setup, which ExaDG users can see, should be Hypercube (with Multigrid) for now since the use of simplex is currently restricted.

applications/poisson/sine/input.json
```diff
@@ -7,7 +7,7 @@
     "Resolution": {
         "RunType": "RefineHAndP",
         "DegreeMin": "1",
-        "DegreeMax": "15",
+        "DegreeMax": "2",
```
nfehn, line 10: The default setup should be Hypercube (with Multigrid), so that we do not have to change this parameter.

applications/poisson/sine/application.h
```diff
@@ -286,8 +279,11 @@ class Application : public ApplicationBase<dim, n_components, Number>
     pp_data.output_data.write_output       = this->output_parameters.write;
     pp_data.output_data.directory          = this->output_parameters.directory + "vtu/";
     pp_data.output_data.filename           = this->output_parameters.filename;
-    pp_data.output_data.write_higher_order = true;
-    pp_data.output_data.degree             = this->param.degree;
+    pp_data.output_data.write_higher_order = false;
+    if(pp_data.output_data.write_higher_order == true and dim == 3 and
+       this->param.grid.element_type == ElementType::Simplex)
+      AssertThrow(false, dealii::ExcMessage("Can't use higher order output with 3D Simplices."));
```
nfehn, line 285: I think this can be written more elegantly:

```c++
    if(this->param.grid.element_type == ElementType::Simplex and dim == 3)
      AssertThrow(pp_data.output_data.write_higher_order == false, dealii::ExcMessage("Can't use higher order output with 3D Simplices."));
```

include/exadg/grid/grid.h
```diff
@@ -72,8 +74,19 @@ class Grid
...
+    }
+    else if(data.element_type == ElementType::Simplex)
+    {
+      mapping =
+        std::make_shared<dealii::MappingFE<dim>>(dealii::FE_SimplexP<dim>(data.mapping_degree));
+    }
+    else
+    {
+      AssertThrow(false, dealii::ExcMessage("Invalid parameter element_type."));
+    }
```
nfehn, line 89: very good

include/exadg/grid/grid_data.h
```diff
@@ -27,14 +27,20 @@
 
 namespace ExaDG
 {
+enum class ElementType
+{
+  Hypercube,
+  Simplex
+};
 struct GridData
```
nfehn, line 35: please introduce one empty line here

applications/poisson/sine/input.json
nfehn: This is probably the reason why the tests failed. So we should undo these changes so that we can merge the rest of this PR.

applications/poisson/sine/application.h
necioglu: The following is also an option:

```suggestion
    if(dim == 3 and this->param.grid.element_type == ElementType::Simplex)
       pp_data.output_data.write_higher_order = false;
    else
       pp_data.output_data.write_higher_order = true;
```

I like this more, because it makes switching between element types easier.

applications/poisson/sine/application.h
necioglu: done!

include/exadg/grid/grid_data.h
necioglu: done!

applications/poisson/sine/application.h
nfehn: sound good as well, but we still need the assert in the code, at best at a central place so that we only have to write this once. The (minor) problem with your suggestion is that from the code one cannot see that the combination Simplex with `write_higher_order` does not work. One might guess this, but it is not entirely clear (and also not enforced).

nfehn (approved):

necioglu:
If you think everything is ready to merge, I will squash all commits into one.

nfehn:
yes, looks good from my side

PR #269, opened 2022-11-03
make Grid::create_triangulation more flexible

nfehn:
This PR is a preparation to enable the construction of coarse triangulations required for "global coarsening multigrid" inside the `Grid` class.

I am not sure whether the public function `create_triangulation()` needs the parameter `global_refinements`. I think we need to equip `GridData` with a boolean parameter `construct_coarse_triangulations` for global coarsening. This could all happen inside `create_triangulation`, so that the parameter `global_refinements` could be deduced from `GridData`? Then, probably only the private function `do_create_triangulation()` needs a parameter `global_refinements`.

Then, we need to extend the public function `create_triangulation()` that loops over all global and local refinements and creates the coarser triangulations for global coarsening. How would we implement this? First reduce global refinement level until 0, and then reduce the entries in `vector_local_refinements` by one until all vector entries are 0?

~~Apart from the vector of coarse triangulations, we might also need to transfer the vector of mappings associated to coarser triangulations to the `Grid` class. I am not sure about this. Looks like multigrid could take care of this once we have the vector of triangulations?~~ Not that easy because of moving mesh implementation (see e.g. function `get_dynamic_mapping()`).

@peterrum I would be happy if you could share your opinion.

### DIFF
--- a/include/exadg/grid/grid.h
+++ b/include/exadg/grid/grid.h
@@ -24,11 +24,13 @@
 
 // deal.II
 #include <deal.II/distributed/fully_distributed_tria.h>
+#include <deal.II/distributed/repartitioning_policy_tools.h>
 #include <deal.II/distributed/tria.h>
 #include <deal.II/fe/fe_simplex_p.h>
 #include <deal.II/fe/mapping_fe.h>
 #include <deal.II/fe/mapping_q.h>
 #include <deal.II/grid/grid_tools.h>
+#include <deal.II/multigrid/mg_transfer_global_coarsening.h>
 
 // ExaDG
 #include <exadg/grid/enum_types.h>
@@ -37,6 +39,50 @@
 
 namespace ExaDG
 {
+/**
+ * A class to use for the deal.II coarsening functionality, where we try to
+ * balance the mesh coarsening with a minimum granularity and the number of
+ * partitions on coarser levels.
+ */
+template<int dim, int spacedim = dim>
+class BalancedGranularityPartitionPolicy
+  : public dealii::RepartitioningPolicyTools::Base<dim, spacedim>
+{
+public:
+  BalancedGranularityPartitionPolicy(unsigned int const n_mpi_processes)
+    : n_mpi_processes_per_level{n_mpi_processes}
+  {
+  }
+
+  virtual dealii::LinearAlgebra::distributed::Vector<double>
+  partition(dealii::Triangulation<dim, spacedim> const & tria_coarse_in) const override
+  {
+    dealii::types::global_cell_index const n_cells = tria_coarse_in.n_global_active_cells();
+
+    // TODO: We hard-code a grain-size limit of 200 cells per processor
+    // (assuming linear finite elements and typical behavior of
+    // supercomputers). In case we have fewer cells on the fine level, we do
+    // not immediately go to 200 cells per rank, but limit the growth by a
+    // factor of 8, which limits makes sure that we do not create too many
+    // messages for individual MPI processes.
+    unsigned int const grain_size_limit =
+      std::min<unsigned int>(200, 8 * n_cells / n_mpi_processes_per_level.back() + 1);
+
+    dealii::RepartitioningPolicyTools::MinimalGranularityPolicy<dim, spacedim> partitioning_policy(
+      grain_size_limit);
+    dealii::LinearAlgebra::distributed::Vector<double> const partitions =
+      partitioning_policy.partition(tria_coarse_in);
+
+    // The vector 'partitions' contains the partition numbers. To get the
+    // number of partitions, we take the infinity norm.
+    n_mpi_processes_per_level.push_back(static_cast<unsigned int>(partitions.linfty_norm()) + 1);
+    return partitions;
+  }
+
+private:
+  mutable std::vector<unsigned int> n_mpi_processes_per_level;
+};
+
 template<int dim>
 class Grid
 {
@@ -103,9 +149,47 @@ public:
 
     if(data.create_coarse_triangulations)
     {
-      // TODO
-      // construct all the coarser triangulations
+      // in case of a serial or distributed triangulation, deal.II can automatically generate the
+      // coarse grid triangulations
+      if(data.triangulation_type == TriangulationType::Serial or
+         data.triangulation_type == TriangulationType::Distributed)
+      {
+        coarse_triangulations =
+          dealii::MGTransferGlobalCoarseningTools::create_geometric_coarsening_sequence(
+            *triangulation,
+            BalancedGranularityPartitionPolicy<dim>(
+              dealii::Utilities::MPI::n_mpi_processes(triangulation->get_communicator())));
+      }
+      else if(data.triangulation_type ==
+              TriangulationType::FullyDistributed) // we need to generate the coarse triangulations
+                                                   // explicitly
+      {
+        // TODO
+        // construct all the coarser triangulations
+      }
+      else
+      {
+        AssertThrow(false, dealii::ExcMessage("Invalid parameter triangulation_type."));
+      }
     }
+  }
+
+  std::shared_ptr<dealii::Triangulation<dim> const>
+  get_triangulation() const
+  {
+    return triangulation;
+  }
+
+  std::vector<std::shared_ptr<dealii::Triangulation<dim> const>> const &
+  get_coarse_triangulations() const
+  {
+    return coarse_triangulations;
+  }
+
+  std::shared_ptr<dealii::Mapping<dim> const>
+  get_mapping() const
+  {
+    return mapping;
   }
 
   /**
@@ -116,7 +200,7 @@ public:
   /**
    * a vector of coarse triangulations required for global coarsening multigrid
    */
-  std::vector<std::shared_ptr<dealii::Triangulation<dim>>> coarse_triangulations;
+  std::vector<std::shared_ptr<dealii::Triangulation<dim> const>> coarse_triangulations;
 
   /**
    * dealii::GridTools::PeriodicFacePair's.
--- a/include/exadg/grid/grid.h
+++ b/include/exadg/grid/grid.h
@@ -144,6 +144,11 @@ public:
     unsigned int const                                        global_refinements,
     std::vector<unsigned int> const & vector_local_refinements = std::vector<unsigned int>())
   {
+    do_create_triangulation(data,
+                            create_coarse_triangulation,
+                            global_refinements,
+                            vector_local_refinements);
+
     // coarse triangulations need to be created for global coarsening multigrid
     if(data.create_coarse_triangulations)
     {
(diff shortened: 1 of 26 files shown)


include/exadg/grid/grid.h
```diff
@@ -80,30 +126,69 @@ class Grid
...
+            BalancedGranularityPartitionPolicy<dim>(
+              dealii::Utilities::MPI::n_mpi_processes(triangulation->get_communicator())));
+      }
+      else if(data.triangulation_type ==
+              TriangulationType::FullyDistributed) // we need to generate the coarse triangulations
+                                                   // explicitly
+      {
+        // TODO
+        // construct all the coarser triangulations
+      }
```
peterrum, line 156: I don't think this is the correct logic. To me it looks like, you are creating the active mesh first and based on that you are creating the coarser one. The logic should be c`reate coarse mesh`-`copy it`-`refine mesh`-`copy it`-....

peterrum, line 156: I don't think this is the correct logic. To me it looks like, you are creating the active mesh first and based on that you are creating the coarser one. The logic should be c`reate coarse mesh`-`copy it`-`refine mesh`-`copy it`-....

nfehn: I should mention this also here on github: @necioglu will implement this functionality. So what I write here is kind of a „platzhalter“ for this PR.

nfehn:
Now, everything should work again (which already worked before). We should now have the infrastructure in ExaDG to set up the coarse triangulations in `ExaDG::Grid`. Please review.

peterrum:
> Does this procedure also hold for local refinements (and not just global refinements)?

It works but the reduction rate of DoFs between levels might not be optimal in this case depending how the mesh was created. For globally refined meshes, it does not make any difference.

Generally there are three options:
1) take the refinement history (as stated in https://github.com/exadg/exadg/pull/269#discussion_r1013107434)
2) take the final mesh and perform global coarsening until we reach the coarsest level (this is done by `create_geometric_coarsening_sequence`). However, `p:f:T` and generally `Triangulation` with simplices do not support coarsening, yet.
3) take the final mesh and implicitly extract the levels without coarsening and write an additional function `TriangulationDescription::Utilities::create_description_from_triangulation()`.

Getting started, the first option makes the most sense, since one would start with globally refined meshes. To make option two and three working would be great as well. For 2), one has to modify `Triangulation` and generally teach AMR `p:f:T`.

For option 3), one can use an option like this (track if cells are active or not with the help of ghosted vectors), however, I don't know if I have missed some corer cases (in particular I am not sure if we need to check one-regularity or this is guaranteed when globally coarsening a mesh that has one-regularity):

```cpp
#include <deal.II/distributed/tria.h>

#include <deal.II/grid/grid_generator.h>

#include <deal.II/numerics/data_out.h>


using namespace dealii;


template <int dim>
void
print_mesh(const Triangulation<dim> &tria)
{
  DataOut<dim> data_out;
  data_out.attach_triangulation(tria);
  data_out.build_patches(1);
  data_out.write_vtu_in_parallel("mesh_active.vtu", tria.get_communicator());
}

template <int dim>
void
print_mesh(const Triangulation<dim> &tria,
           const std::vector<LinearAlgebra::distributed::Vector<double>>
             &refinement_state)
{
  DataOut<dim> data_out;
  data_out.attach_triangulation(tria);

  const auto next_cell = [&](const auto &tria, const auto &cell_in) {
    auto cell = cell_in;

    while (true)
      {
        cell++;

        if (cell == tria.end())
          break;

        if (cell->is_locally_owned_on_level() &&
            refinement_state[cell->level()][cell->global_level_cell_index()])
          return cell;
      }

    return tria.end();
  };

  // output mesh
  const auto first_cell = [&](const auto &tria) {
    const auto cell = tria.begin();

    if (cell == tria.end())
      return cell;

    if (cell->is_locally_owned_on_level() &&
        refinement_state[cell->level()][cell->global_level_cell_index()])
      return cell;

    return next_cell(tria, cell);
  };

  data_out.set_cell_selection(first_cell, next_cell);

  data_out.build_patches(1);

  unsigned int level = 0;

  for (unsigned int l = 1; l < refinement_state.size(); ++l)
    for (const auto i : refinement_state[l])
      if (i == 1.0)
        level = l;

  level = Utilities::MPI::max(level, tria.get_communicator());

  data_out.write_vtu_in_parallel("mesh_level_" + std::to_string(level) + ".vtu",
                                 MPI_COMM_WORLD);
}

template <int dim>
void
test(const unsigned int n_local_refinements)
{
  const MPI_Comm comm = MPI_COMM_WORLD;

  // create locally refined mesh
  parallel::distributed::Triangulation<dim> tria(
    comm,
    Triangulation<dim>::none,
    parallel::distributed::Triangulation<dim>::construct_multigrid_hierarchy);

  GridGenerator::subdivided_hyper_cube(tria, 2, -1.0, 1.0);

  for (unsigned int i = 0; i < n_local_refinements; ++i)
    {
      for (auto cell : tria.active_cell_iterators())
        if (cell->is_locally_owned())
          {
            bool flag = true;
            for (int d = 0; d < dim; d++)
              if (cell->center()[d] > 0.0)
                flag = false;
            if (flag)
              cell->set_refine_flag();
          }
      tria.execute_coarsening_and_refinement();
    }

  // write it out normally
  print_mesh(tria);

  // data structure to track if a cell is "active" or not
  std::vector<LinearAlgebra::distributed::Vector<double>> refinement_state(
    tria.n_global_levels());
  for (unsigned int l = 0; l < tria.n_global_levels(); ++l)
    refinement_state[l].reinit(
      tria.global_level_cell_index_partitioner(l).lock());

  // ... copy the state of the actually acive cells
  for (const auto &cell : tria.active_cell_iterators())
    if (cell->is_locally_owned())
      refinement_state[cell->level()][cell->global_level_cell_index()] = 1.0;

  // print active mesh with alternative function
  print_mesh(tria, refinement_state);

  // perform global coarsening by modifying the values in the vectors
  for (unsigned int ll = 0; ll < refinement_state.size() - 1; ++ll)
    {
      auto refinement_state_temp = refinement_state;

      for (unsigned int l = 0; l < refinement_state.size() - ll; ++l)
        {
          refinement_state[l].update_ghost_values();
          refinement_state_temp[l] = 0.0;
        }

      for (unsigned int l = 0; l < refinement_state.size() - 1 - ll; ++l)
        for (const auto parent_cell : tria.cell_iterators_on_level(l))
          if (parent_cell->is_locally_owned_on_level() == true &&
              parent_cell->has_children())
            {
              // cell has been visited already -> nothing to do
              if (refinement_state_temp[parent_cell->level()]
                                       [parent_cell
                                          ->global_level_cell_index()] != 0.0)
                continue;

              // check if all chilren are active
              bool can_become_active = true;

              for (const auto cell : parent_cell->child_iterators())
                if (refinement_state[cell->level()]
                                    [cell->global_level_cell_index()] == 0)
                  can_become_active = false;

              if (can_become_active)
                {
                  // if yes: make children inactive and make this cell active
                  refinement_state_temp[parent_cell->level()]
                                       [parent_cell
                                          ->global_level_cell_index()] = 1.0;
                  for (const auto cell : parent_cell->child_iterators())
                    refinement_state_temp[cell->level()]
                                         [cell->global_level_cell_index()] =
                                           -1.0; // indicate that cell has been
                                                 // visitied and its state has
                                                 // changed
                }
              else
                {
                  // if no: copy old state
                  refinement_state_temp
                    [parent_cell->level()]
                    [parent_cell->global_level_cell_index()] =
                      refinement_state[parent_cell->level()]
                                      [parent_cell->global_level_cell_index()];
                  for (const auto cell : parent_cell->child_iterators())
                    refinement_state_temp
                      [cell->level()][cell->global_level_cell_index()] =
                        refinement_state[cell->level()]
                                        [cell->global_level_cell_index()];
                }
            }

      for (unsigned int l = 0; l < refinement_state.size() - ll; ++l)
        {
          for (auto &i : refinement_state_temp[l])
            if (i == -1.0)
              i = 0.0; // not needed any more

          // notify owning cells about changing state
          refinement_state_temp[l].compress(VectorOperation::add);

          // copy new state
          refinement_state[l].zero_out_ghost_values();
          refinement_state[l].copy_locally_owned_data_from(
            refinement_state_temp[l]);
        }

      // print level mesh
      print_mesh(tria, refinement_state);
    }
}

int
main(int argc, char *argv[])
{
  Utilities::MPI::MPI_InitFinalize mpi_initialization(argc, argv, 1);

  test<2>(7);
}
```

nfehn:
@peterrum Do you think we can merge this state so that @necioglu can continue on this basis?

include/exadg/structure/preconditioners/multigrid_preconditioner.h
```diff
@@ -59,14 +59,16 @@ class MultigridPreconditioner : public MultigridPreconditionerBase<dim, Number>
...
-             dealii::FiniteElement<dim> const &          fe,
-             std::shared_ptr<dealii::Mapping<dim> const> mapping,
-             ElasticityOperatorBase<dim, Number> const & pde_operator,
-             bool const                                  nonlinear_operator,
-             Map const &                                 dirichlet_bc,
-             PeriodicFacePairs const &                   periodic_face_pairs);
+  initialize(
+    MultigridData const &                                                  mg_data,
+    dealii::Triangulation<dim> const *                                     tria,
+    std::vector<std::shared_ptr<dealii::Triangulation<dim> const>> const & coarse_triangulations,
```
peterrum, line 65: Maybe we could pass in `Grid` directly here?

peterrum (approved):
I just have a single point. But we don't need to address this now, since it is independent from @necioglu's work.

include/exadg/structure/preconditioners/multigrid_preconditioner.h
nfehn: I actually followed the same goal and I tried this today in the morning, but gave it up in the end because there are places in the code where the mapping parameter passed to multigrid is `get_dynamic_mapping(mapping, moving_mesh)`. Hence, this problem is better addressed separately.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -56,30 +56,31 @@ MultigridPreconditionerBase<dim, Number>::MultigridPreconditionerBase(MPI_Comm c
...
+  std::shared_ptr<dealii::Mapping<dim> const>                            mapping,
+  bool const                                                             operator_is_singular,
+  Map const &                                                            dirichlet_bc,
+  PeriodicFacePairs const &                                              periodic_face_pairs)
 {
   this->data = data;
 
   this->triangulation = tria;
 
+  this->coarse_triangulations = coarse_triangulations;
```
nfehn, line 72: calling here `dealii::MGTransferGlobalCoarseningTools::create_geometric_coarsening_sequence()` instead of the assignment of the vector of shared pointers fixes the bug introduced in this PR (see issue #303).

nfehn: Is the copy of a std::vector of shared pointers problematic?

PR #271, opened 2022-11-07
fix compilation error in flow past sphere test case

nfehn:
this test case is not up-to-date regarding e.g. PR #154 or other recent changes such as those in the grid class.

@kronbichler Please check in particular the function `create_grid()` in application.h. If I understand your code correctly, it is not possible to apply a `refine_global()` at the end of the grid generation function?

### DIFF
--- a/applications/incompressible_navier_stokes/flow_past_cylinder/application.h
+++ b/applications/incompressible_navier_stokes/flow_past_cylinder/application.h
@@ -324,8 +324,6 @@ private:
   void
   create_grid() final
   {
-    this->refine_level = this->param.grid.n_refine_global;
-
     auto const lambda_create_coarse_triangulation = [&](dealii::Triangulation<dim, dim> & tria) {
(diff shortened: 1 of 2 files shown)


kronbichler (approved):
Thank you for fixing up my old code. This looks good. Regarding the question: I manually select certain parts of the mesh for the adaptive refinement, which is done after the global refinement in order to limit the fine mesh elements to the regions of interest. So I assume we can't do it afterwards; we could look into this again in the future if we identify this as a bottleneck.

nfehn:
My question is more from the perspective of code design and whether this test case would be an interesting example for global coarsening multigrid with a fully-distributed triangulation, where (for now) we want to construct the coarse level meshes manually. The question then is how this code fits into the procedure outlined in #269, i.e. can we implement this grid with a `vector_local_refinements` (one entry for each material ID) in combination with subsequent global refinements.

PR #272, opened 2022-11-11
Fix global_coarsening.cpp

peterrum:
@necioglu Could you give this a try? Thx.


necioglu:
> @necioglu Could you give this a try? Thx.

Now it works fine with simplices as well.

PR #274, opened 2022-11-17
follow-up to #245

### LINKED ISSUE #284: Bug in Postprocessor
I think a bug has been introduced in #245, see comment https://github.com/exadg/exadg/pull/245/files#r1031496463

nfehn:
I think the design of `SolutionField` needs to be further improved. 

On the one hand, certain names are not intuitive regarding their meaning or what the functions actually do.

On the other hand, solution field has been misused for certain purposes, which should be avoided. E.g. implementing `get_vector_reference()` and at the same time also the lambda function `recompute_solution_field()` is a mixture of concepts in my opinion (i.e. it would be better to follow one or the other approach for clarity). Also the free functions `evaluate_get()` are irritating in my opinion due to the order of arguments and which variables are returned by these functions. Finally, passing the source vector to `SolutionField` via a `reinit()` function and using the vector later somewhere is unclear and should be avoided. The source vector should in my opinion directly appear in the interface of the `evaluate()` or `evaluate_get()` functions.

I also think there might be subtle bugs in the present implementation that are hopefully removed (at least partly) by this PR.

~~We still need to find a better solution for the "mean velocity" problem/computation.~~ should be resolved now

@jh66637 FYI. I am happy about a review.

Latest update:

closes #284

### DIFF
--- a/include/exadg/compressible_navier_stokes/postprocessor/postprocessor.cpp
+++ b/include/exadg/compressible_navier_stokes/postprocessor/postprocessor.cpp
@@ -112,17 +112,16 @@ PostProcessor<dim, Number>::do_postprocessing(VectorType const &     solution,
     {
       velocity.evaluate(solution);
       additional_fields_vtu.push_back(&velocity);
-
-      if(pp_data.output_data.write_vorticity)
-      {
-        vorticity.evaluate(velocity.get());
-        additional_fields_vtu.push_back(&vorticity);
-      }
-      if(pp_data.output_data.write_divergence)
-      {
-        divergence.evaluate(velocity.get());
-        additional_fields_vtu.push_back(&divergence);
-      }
+    }
+    if(pp_data.output_data.write_vorticity)
+    {
+      vorticity.evaluate(velocity.evaluate_get(solution));
+      additional_fields_vtu.push_back(&vorticity);
+    }
+    if(pp_data.output_data.write_divergence)
+    {
+      divergence.evaluate(velocity.evaluate_get(solution));
+      additional_fields_vtu.push_back(&divergence);
     }
     if(pp_data.output_data.write_temperature)
     {
@@ -224,44 +223,44 @@ PostProcessor<dim, Number>::initialize_derived_fields()
     };
 
     velocity.reinit();
+  }
 
-    // vorticity
-    if(pp_data.output_data.write_vorticity)
-    {
-      AssertThrow(pp_data.output_data.write_velocity == true,
-                  dealii::ExcMessage("You need to activate write_velocity."));
+  // vorticity
+  if(pp_data.output_data.write_vorticity)
+  {
+    AssertThrow(pp_data.output_data.write_velocity == true,
+                dealii::ExcMessage("You need to activate write_velocity."));
 
-      vorticity.type              = SolutionFieldType::vector;
-      vorticity.name              = "vorticity";
-      vorticity.dof_handler       = &navier_stokes_operator->get_dof_handler_vector();
-      vorticity.initialize_vector = [&](VectorType & dst) {
-        navier_stokes_operator->initialize_dof_vector_dim_components(dst);
-      };
-      vorticity.recompute_solution_field = [&](VectorType & dst, VectorType const & src) {
-        navier_stokes_operator->compute_vorticity(dst, src);
-      };
+    vorticity.type              = SolutionFieldType::vector;
+    vorticity.name              = "vorticity";
+    vorticity.dof_handler       = &navier_stokes_operator->get_dof_handler_vector();
+    vorticity.initialize_vector = [&](VectorType & dst) {
+      navier_stokes_operator->initialize_dof_vector_dim_components(dst);
+    };
+    vorticity.recompute_solution_field = [&](VectorType & dst, VectorType const & src) {
+      navier_stokes_operator->compute_vorticity(dst, src);
+    };
 
-      vorticity.reinit();
-    }
+    vorticity.reinit();
+  }
 
-    // divergence
-    if(pp_data.output_data.write_divergence)
-    {
-      AssertThrow(pp_data.output_data.write_velocity == true,
-                  dealii::ExcMessage("You need to activate write_velocity."));
+  // divergence
+  if(pp_data.output_data.write_divergence)
+  {
+    AssertThrow(pp_data.output_data.write_velocity == true,
+                dealii::ExcMessage("You need to activate write_velocity."));
 
-      divergence.type              = SolutionFieldType::scalar;
-      divergence.name              = "velocity_divergence";
-      divergence.dof_handler       = &navier_stokes_operator->get_dof_handler_scalar();
-      divergence.initialize_vector = [&](VectorType & dst) {
-        navier_stokes_operator->initialize_dof_vector_scalar(dst);
-      };
-      divergence.recompute_solution_field = [&](VectorType & dst, VectorType const & src) {
-        navier_stokes_operator->compute_divergence(dst, src);
-      };
+    divergence.type              = SolutionFieldType::scalar;
+    divergence.name              = "velocity_divergence";
+    divergence.dof_handler       = &navier_stokes_operator->get_dof_handler_scalar();
+    divergence.initialize_vector = [&](VectorType & dst) {
+      navier_stokes_operator->initialize_dof_vector_scalar(dst);
+    };
+    divergence.recompute_solution_field = [&](VectorType & dst, VectorType const & src) {
+      navier_stokes_operator->compute_divergence(dst, src);
+    };
 
-      divergence.reinit();
-    }
+    divergence.reinit();
   }
 
   // temperature
--- a/include/exadg/compressible_navier_stokes/postprocessor/postprocessor.cpp
+++ b/include/exadg/compressible_navier_stokes/postprocessor/postprocessor.cpp
@@ -196,9 +196,8 @@ PostProcessor<dim, Number>::initialize_derived_fields()
     pressure.type              = SolutionFieldType::scalar;
     pressure.name              = "pressure";
     pressure.dof_handler       = &navier_stokes_operator->get_dof_handler_scalar();
-    pressure.initialize_vector = [&](std::shared_ptr<VectorType> & dst) {
-      dst = std::make_shared<VectorType>();
-      navier_stokes_operator->initialize_dof_vector_scalar(*dst);
+    pressure.initialize_vector = [&](VectorType & dst) {
+      navier_stokes_operator->initialize_dof_vector_scalar(dst);
     };
     pressure.recompute_solution_field = [&](VectorType & dst, VectorType const & src) {
       navier_stokes_operator->compute_pressure(dst, src);
@@ -217,9 +216,8 @@ PostProcessor<dim, Number>::initialize_derived_fields()
     velocity.type              = SolutionFieldType::vector;
     velocity.name              = "velocity";
     velocity.dof_handler       = &navier_stokes_operator->get_dof_handler_vector();
-    velocity.initialize_vector = [&](std::shared_ptr<VectorType> & dst) {
-      dst = std::make_shared<VectorType>();
-      navier_stokes_operator->initialize_dof_vector_dim_components(*dst);
+    velocity.initialize_vector = [&](VectorType & dst) {
+      navier_stokes_operator->initialize_dof_vector_dim_components(dst);
     };
     velocity.recompute_solution_field = [&](VectorType & dst, VectorType const & src) {
       navier_stokes_operator->compute_velocity(dst, src);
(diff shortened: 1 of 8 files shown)


include/exadg/compressible_navier_stokes/postprocessor/postprocessor.cpp
```diff
@@ -181,70 +214,87 @@ PostProcessor<dim, Number>::initialize_additional_vectors()
...
-
-  // vorticity
-  if(pp_data.output_data.write_vorticity)
-  {
-    vorticity.type        = SolutionFieldType::vector;
-    vorticity.name        = "vorticity";
-    vorticity.dof_handler = &navier_stokes_operator->get_dof_handler_vector();
-    navier_stokes_operator->initialize_dof_vector_dim_components(vorticity.get_vector_reference());
-    vorticity.recompute_solution_field = [&](VectorType & dst, VectorType const &) {
-      navier_stokes_operator->compute_vorticity(dst, velocity.get_vector());
```
nfehn, line 218: This should be avoided. If I see it correctly, the code enforces that `velocity` is set up if `write_vorticity`, but it does not enforce that the `velocity` vector is up-to-date.

include/exadg/incompressible_navier_stokes/postprocessor/postprocessor.cpp
```diff
@@ -195,116 +237,142 @@ PostProcessor<dim, Number>::do_postprocessing(VectorType const &     velocity,
...
-  // vorticity magnitude
-  if(pp_data.output_data.write_vorticity_magnitude)
-  {
-    vorticity_magnitude.type        = SolutionFieldType::scalar;
-    vorticity_magnitude.name        = "vorticity_magnitude";
-    vorticity_magnitude.dof_handler = &navier_stokes_operator->get_dof_handler_u_scalar();
-    navier_stokes_operator->initialize_vector_velocity_scalar(
-      vorticity_magnitude.get_vector_reference());
-    vorticity_magnitude.recompute_solution_field = [&](VectorType & dst, VectorType const &) {
-      navier_stokes_operator->compute_vorticity_magnitude(dst, evaluate_get(vorticity));
```
nfehn, line 253: This should be avoided. In my opinion it is hard to see that the vector passed to `evaluate()`/`evaluate_get()` is by-passed in case of `vorticity_magnitude`, given that the `reinit_additional_fields()` function says `vorticity_magnitude.reinit(velocity);` (velocity, not vorticity!).

include/exadg/incompressible_navier_stokes/postprocessor/postprocessor.cpp
```diff
@@ -195,116 +237,142 @@ PostProcessor<dim, Number>::do_postprocessing(VectorType const &     velocity,
...
-  // streamfunction
-  if(pp_data.output_data.write_streamfunction)
-  {
-    streamfunction.type        = SolutionFieldType::scalar;
-    streamfunction.name        = "streamfunction";
-    streamfunction.dof_handler = &navier_stokes_operator->get_dof_handler_u_scalar();
-    navier_stokes_operator->initialize_vector_velocity_scalar(
-      streamfunction.get_vector_reference());
-    streamfunction.recompute_solution_field = [&](VectorType & dst, VectorType const &) {
-      navier_stokes_operator->compute_streamfunction(dst, evaluate_get(vorticity));
```
nfehn, line 269: as above

include/exadg/incompressible_navier_stokes/postprocessor/postprocessor.cpp
```diff
@@ -195,116 +237,142 @@ PostProcessor<dim, Number>::do_postprocessing(VectorType const &     velocity,
...
 
-    this->additional_fields_vtu.push_back(&mean_velocity);
+    mean_velocity.reinit();
   }
 
   // cfl
   if(pp_data.output_data.write_cfl)
   {
-    cfl_vector.type                     = SolutionFieldType::cellwise;
-    cfl_vector.name                     = "cfl_relative";
```
nfehn, line 307: I wonder why there has not been an initialization of the vector in the previous/current code version?

jh66637: I was wondering about the same: https://github.com/exadg/exadg/pull/245#discussion_r996444168

include/exadg/incompressible_navier_stokes/postprocessor/postprocessor.cpp
```diff
@@ -195,116 +237,142 @@ PostProcessor<dim, Number>::do_postprocessing(VectorType const &     velocity,
...
      pp_data.output_data.write_vorticity_magnitude)
   {
-    vorticity.type        = SolutionFieldType::vector;
-    vorticity.name        = "vorticity";
-    vorticity.dof_handler = &navier_stokes_operator->get_dof_handler_u();
-    navier_stokes_operator->initialize_vector_velocity(vorticity.get_vector_reference());
+    vorticity.type              = SolutionFieldType::vector;
+    vorticity.name              = "vorticity";
+    vorticity.dof_handler       = &navier_stokes_operator->get_dof_handler_u();
+    vorticity.initialize_vector = [&](std::shared_ptr<VectorType> & dst) {
```
jh66637, line 249: Initializing the vector by passing a lambda is not very intuitive in my opinion, but now it is not possible to abuse the ```get_reference_vector``` getter anymore, so I am OK with this. 

What I don't get is why the shared_ptr is introduced here. I think this should not be necessary.

include/exadg/postprocessor/solution_field.h
```diff
@@ -109,28 +128,11 @@ class SolutionField : public dealii::Subscriptor
   dealii::DoFHandler<dim> const * dof_handler;
 
 private:
-  bool               is_available;
-  VectorType         vector;
-  VectorType const * solution;
+  mutable bool                        is_available;
+  mutable std::shared_ptr<VectorType> solution_vector;
+  VectorType const *                  src_vector;
```
jh66637, line 133: I think src_vector is not used in your current implementations anymore.

include/exadg/postprocessor/solution_field.h
```diff
@@ -41,45 +41,61 @@ class SolutionField : public dealii::Subscriptor
...
-    return vector;
+    AssertThrow(is_available,
+                dealii::ExcMessage("You are trying to access a Vector that is not available."));
+
+    return *solution_vector;
   }
 
   VectorType const &
-  get_vector() const
+  evaluate_get(VectorType const & src)
```
jh66637, line 94: I liked that evaluate_get() didn't need a src vector before. With this we had to be careful during initialization, but in the rest of the code we don't have to care about what's needed if we want the evaluated quantity.
Now (to get the evaluated quantity) it might be necessary to call e.g.
```
qunatity.evaluate_get(other_qunatity.evaluate_get(third_quantity))
```
instead of 
```
evaluate_get(quantity)
```

include/exadg/incompressible_navier_stokes/postprocessor/postprocessor.cpp
```diff
@@ -329,16 +397,16 @@ PostProcessor<dim, Number>::compute_mean_velocity(VectorType &       mean_veloci
...
-  vorticity.reinit(velocity);
-  divergence.reinit(velocity);
-  velocity_magnitude.reinit(velocity);
-  vorticity_magnitude.reinit(velocity);
-  streamfunction.reinit(velocity);
-  q_criterion.reinit(velocity);
-  cfl_vector.reinit(velocity);
-  mean_velocity.reinit(velocity);
+  vorticity.invalidate();
+  divergence.invalidate();
```
jh66637, line 403: I like that the ```reinit()``` function gets independent of velocity. I still think it is possible to use ```evaluate_get()``` without passing the vector that is needed for evaluation (see the comment above).

jh66637:
Thanks for the PR. I think this resolves some problematic parts of previous implementation :) 
I still have some comments which --- I think --- could improve the interface for the user.

include/exadg/postprocessor/solution_field.h
nfehn: agree

include/exadg/postprocessor/solution_field.h
nfehn: I think this is the main point where we disagree. I like the first option, as it is clear what the input or src-vector is. The second option allows some misuse (wanted or unwanted) that I would like to avoid.

jh66637: I'm not sure how such a misuse could look like. Do you have an example?

In my opinion, it is more safe to provide the correct src-vector within the lambda at only one point in the code. 
On the other hand: If the user has to specify the src-vector every time he wants to access the field, it is more likely that an unwanted error is introduced. This might also lead to code bloat if the user wants to access the field in an arbitrary function, since this function also needs the src-vector(s) as input.

nfehn: I am referring to versions of the lambda function where the src parameter remained empty and some other variable of the PostProcessor class has been used instead. In my opinion it is safer to not operate on member variables but on the src, dst vectors passed to a function. I also believe it is easier to read/understand the code. I probably have a strong procedural perspective and believe that this is still a recommendable way of thinking for many problems we are dealing with in the context of PDE solvers. Its fine what we do not have identical perspectives here.

nfehn:
I changed the implementation of the mean velocity computation in the postprocessor. I am relatively happy with that now, but please check @jh66637 . As a next step, I can remove again the shared::ptr, which I introduced yesterday in order to not store the mean_velocity redundantly in two classes.

include/exadg/incompressible_navier_stokes/postprocessor/postprocessor.cpp
nfehn: I removed it again here 655b8f49f28c45ea520cf681d308c7edc8801b99. It was an intermediate solution because of the mean_velocity implementation.

Regarding the lambda's, I want to do it consistently, so I see a strong analogy between initializing and recomputing the vector. Before, we needed a function `get_vector_reference()`, which I would like to avoid.

include/exadg/compressible_navier_stokes/postprocessor/postprocessor.cpp
```diff
@@ -94,15 +94,44 @@ PostProcessor<dim, Number>::do_postprocessing(VectorType const &     solution,
...
+      additional_fields_vtu.push_back(&pressure);
+    }
+    if(pp_data.output_data.write_velocity)
+    {
+      velocity.evaluate(solution);
+      additional_fields_vtu.push_back(&velocity);
+
+      if(pp_data.output_data.write_vorticity)
+      {
+        vorticity.evaluate(velocity.get());
```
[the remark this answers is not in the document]
nfehn: I don't understand this. I think we we have such a dependency, which is why I placed these two if-statements inside the if-statement for the velocity vector.

We can certainly debate which logic is the better one, i.e. should the user be forced to activate `write_velocity` in order to be able to do `write_vorticity`, or should the user just set `write_vorticity` and the code has to care about how to realize this. If you want, I can again write these if-statements outside the velocity, and then write `vorticity.evaluate(velocity.evaluate_get(solution));`, which will not perform unnecessary computations in case the velocity has already been evaluated before. Should I change this again?

include/exadg/incompressible_navier_stokes/postprocessor/postprocessor.cpp
nfehn: Before, `vorticity_magnitude` received `velocity` in the reinit function (`vorticity_magnitude.reinit(velocity)`), but the above lambda uses `vorticity` as source vector. This is why I proposed to remove the vector from the reinit-functions and explicitly pass the src-vector to all `evaluate()/evaluate_get()` functions. I do not see the disadvantage that the `evaluate()/evaluate_get()` functions have a parameter. This way, the "flow of information" and the dependency between derived fields becomes clear. I can certainly make a mistake by passing the wrong vector. I argue, however, that it is easier to find this mistake that in the current situation where the lambda accesses some vectors where I do not fully understand where they come from or when they are updated.

include/exadg/postprocessor/solution_field.h
jh66637: What I had in mind was to remove the src parameter from the lambda completely and capturing the required src-vector in the lambda definition as it was done, e.g. for the vorticity_magnitude. This way, there is no need for a member variable in the SolutionField class. 
I can relate with both ways, but still prefer the version without handing over the src-vector every time.

include/exadg/compressible_navier_stokes/postprocessor/postprocessor.cpp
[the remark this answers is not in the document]
jh66637: Yes, from my perspective this should be changed. This also means that the previous velocity.evaluate(solution) can be removed again. 
I think similar code also appears in the incompressible NSE post-processor.

include/exadg/incompressible_navier_stokes/postprocessor/postprocessor.cpp
jh66637: Looks good to me :)

include/exadg/postprocessor/solution_field.h
nfehn: I think capturing the src-vector is difficult since the src-vector is not a member variable of `PostProcessor` in most cases and, therefore, not available during setup/initialization.

include/exadg/compressible_navier_stokes/postprocessor/postprocessor.cpp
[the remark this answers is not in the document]
nfehn: I changed back to the previous logic here: https://github.com/exadg/exadg/pull/274/commits/be5a1044ec6717e6e986b4bed0e1074ee36243fe

include/exadg/postprocessor/solution_field.h
jh66637: True. I think it's OK for me the second way :)

nfehn:
I have tested the code by running an example that activates output of additional fields, where I identified another bug. 

I think I will then merge this PR once the checks have been successful.

PR #276, opened 2022-11-21
Instantiate dim+1 components

jh66637:
I need the mass operator for a system of equations with dim+1 components.

### DIFF
--- a/include/exadg/operators/mass_operator.cpp
+++ b/include/exadg/operators/mass_operator.cpp
@@ -90,11 +90,18 @@ template class MassOperator<2, 1, double>;
 template class MassOperator<3, 1, float>;
 template class MassOperator<3, 1, double>;
 
-// vector
+// dim components
 template class MassOperator<2, 2, float>;
 template class MassOperator<2, 2, double>;
 
 template class MassOperator<3, 3, float>;
 template class MassOperator<3, 3, double>;
+
+// dim + 1 components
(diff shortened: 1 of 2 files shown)


nfehn:
Changes look good. 

The more subtle changes in this PR are those instantiating `OperatorBase` with dim+1 components. I am not 100% sure if this has been taken into account when implementing `OperatorBase` or whether there might be places that depend on the number of components and where dim+1 might become problematic (hopefully not). This should definitely be checked before merging.

May I ask why you need this? For compressible Navier-Stokes (dim+2 components) with explicit time stepping we do not need it. It exists for incompressible Navier-Stokes with implicit time stepping and to realize a modular implementation with mass, convective, and diffusive operators. The inverse mass matrix for explicit time stepping is not precompiled, but can be used with any number of components.

jh66637:
Actually, I was able to reconstruct my code such that I don't need it anymore. This was intended to be used for the ExaDG acoustics module I am currently developing (which has pressure and velocity as components).
So you think this is valid for the future? If yes, I can loop into it when I find some time. If not, I will close the PR.

nfehn:
I was wondering whether we have to extend this whenever a new number of components is realized. So if you currently do not need it, we could close this PR (but of course this should not complicate your code too much).

jh66637:
@nfehn: I checked the code of OperatorBase and MassOperator and couldn't find any places where this might become problematic.

Issue #279, opened 2022-11-23
ExaDG does not seem to work with current dealii master


nfehn:
Reported by @necioglu

PR #280, opened 2022-11-23
remove templates from make_sparsity_pattern

necioglu:
Following PR of dealii  https://github.com/dealii/dealii/pull/14433 removes the templates from https://github.com/exadg/exadg/blob/f7c3c2fe427a4c625f22eb9a5da1be35821e9b3f/include/exadg/operators/operator_base.cpp#L789. 

Made Exadg compatible with the changes. Thanks to @bergbauer 

Closes https://github.com/exadg/exadg/issues/279

### DIFF
--- a/include/exadg/operators/operator_base.cpp
+++ b/include/exadg/operators/operator_base.cpp
@@ -786,8 +786,7 @@ OperatorBase<dim, Number, n_components>::internal_init_system_matrix(
   else if(is_dg && !is_mg)
     dealii::DoFTools::make_flux_sparsity_pattern(dof_handler, dsp);
   else if(/*!is_dg &&*/ is_mg)
-    dealii::MGTools::make_sparsity_pattern<dim, dim, dealii::DynamicSparsityPattern, Number>(
-      dof_handler, dsp, this->level, *this->constraint);
+    dealii::MGTools::make_sparsity_pattern(dof_handler, dsp, this->level, *this->constraint);
   else /* if (!is_dg && !is_mg)*/
     dealii::DoFTools::make_sparsity_pattern(dof_handler, dsp, *this->constraint);
 


peterrum:
@nfehn Do we care about keeping compatibility with release 9.4?

kronbichler:
> @nfehn Do we care about keeping compatibility with release 9.4?

Shouldn't we still be compatible? I don't see immediately why we would have needed to specify the template arguments and not rely on the automatic deduction for the 9.4 release.

nfehn:
I think we decided to only test against dealii:master, right?

nfehn (approved):

peterrum:
> I think we decided to only test against dealii:master, right?

No, we have not. We decided on dropping the testing of 9.3 a few weeks before the 9.4 release. I don't mind if we keep it like this. But there will be a time when we need to start versioning once we have more users.

nfehn:
But everything is green, so where would such an incompatibility with dealii-9.4 show up? 

Principally, we should for now decide for a solution that allows most efficient progress. dealii promises to not make incompatible changes, but in practice it happens every few weeks to months. So I don't know what's best.

kronbichler:
>  dealii promises to not make incompatible changes, but in practice it happens every few weeks to months.

While I agree that we should (a) make sure to enable quick progress, but also (b) at some point have the facilities to support multiple versions of deal.II, I do not understand the issue here. Has someone tested that the current code actually is *not compatible* with version 9.4?  In my experience, 80% of the issues we have with deal.II are things where a change we make is still compatible with the old version in deal.II (like I suppose is the case here, but I didn't test it). This helps us to make the code better overall.

nfehn:
I just wanted  to mention that this type of changes in ExaDG happen regularily. I do not criticize this. So I definitely appreciate overall progress in deal.II, but at the same time we have to find a good solution for ExaDG, and for that I think we need to take into account that such incompatible changes happen in deal.II.

Issue #277, opened 2022-11-22
dof_handler.distribute_mg_dofs()


nfehn:
What should we do with this line: https://github.com/exadg/exadg/blob/dc2c483d3ca9fe275ce5e6e268985755fcce02fc/include/exadg/poisson/spatial_discretization/operator.cpp#L134?

The if-statement checking for hypercube-elements is not the right logic in my opinion. I think we need this only for local-smoothing multigrid. Hence, we should change the if-statement accordingly, by deducing from the list of parameters whether local-smoothing multigrid is used or not.

As discussed already somewhere else (but I do not remember where), this function call would be better placed inside the multigrid preconditioner. Then, it would be intuitively clear for new developers (like @necioglu) where this is needed, i.e., we would achieve a more modular design.

@kronbichler @peterrum It would be great if you could share your opinion here so that we can solve this now.

nfehn:
see also #133

nfehn:
@necioglu see also the discussion here: https://github.com/exadg/exadg/pull/110#discussion_r765046008

kronbichler:
I would simply move the code to the multigrid class, in other words do:
```
  if (!dof_handler.has_level_dofs())
    const_cast<DoFHandler<dim> &>(dof_handler).distribute_mg_dofs();
```
and delete the code here. I cannot promise I will get around to implement and test this in the next two days (given that I started looking into it in January), but I would assume it should be worth a try.

nfehn:
I agree, @necioglu . However, the variant by @kronbichler might look more generic and we should place this code in the multigrid-preconditioner. @kronbichler Could you explain the reasoning behind your implementation and explain it in terms of local-smoothing versus global-coarsening multigrid? Also, what happens if this code would be called multiple times? For Navier-Stokes, several equations (viscous step and penalty step) might want to set up a multigrid preconditioner for the velocity dof-handler. In `SpatialOperator` we have "good control" that the function is only called once. So we have to make sure that shifting the code to multigrid cannot introduce unwanted side effects.

kronbichler:
That's why I suggest to guard the code inside the multigrid preconditioner by the check `if (!dof_handler.has_level_dofs())` - this ensures that the enumeration is only compute if it does not yet exist.

nfehn:
Ah, so you would additionally make sure that global-coarsening multigrid does not run into this branch?

kronbichler:
Yes, we should only do this if necessary by the multigrid algorithm, to clarify the intent.

[...]

Issue #133, opened 2022-01-12
Multigrid solvers: Remove unnecessary distribute_mg_dofs


kronbichler:
In the standard multigrid code path, we currently unconditionally call `distribute_mg_dofs`, https://github.com/exadg/exadg/blob/f2ff9d113b0898c8eb5d089b2ee5f0d3112c9297/include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp#L808
However, in case of c or p transfer we switch to another `DoFHandler`, so we only need it for those `DoFHandler` objects where there is a subsequent h-coarsening. As cph multigrid seems to be our favorite nowadays, this would avoid allocating memory for the DoF indices on the finest levels with many DoFs, translating into some slight improvement in memory consumption. I have not checked yet how easy it would be to figure this out, but it seems like a useful optimization.

kronbichler:
Discovered while working on #132.

nfehn:
One solution to this could be that in the function

https://github.com/exadg/exadg/blob/97f92fdc00e514b892e167cdf2f3bcbe4f0f6a0f/include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp#L134

we store for each p-level the information whether this p-level needs `distribute_mg_dofs()` or not. If I understand it correctly, a p-level needs `distribute_mg_dofs()` in case it is followed by h-coarsening. Note however, that it is not always the highest or lowest p-level, because c-transfer is also treated as a p-level. Consider as an example chp-coarsening. In this case, the second-finest p-level should require `distribute_mg_dofs()`  since this level is followed by h-coarsening. (In case of phc-coarsening, it would be the second lowest p-level taht needs `distribute_mg_dofs()`. For all other cases in which p- and c-coarsening are done next to each other, it should be either the finest or the coarsest p-level that needs `distribute_mg_dofs()` ).

nfehn:
See also comment https://github.com/exadg/exadg/issues/339#issuecomment-1433131216, highlighting that MG constraints need to be discussed in this context as well.

@kronbichler, could you rethink this issue based on the recent insights and changes? (#339, #342)

PR #283, opened 2022-11-24
remove distribute_mg_dofs from poisson spatial operator

necioglu:
Closes https://github.com/exadg/exadg/issues/277

distrubute_mg_dofs() function call was unnecessary, since it is already called in: https://github.com/exadg/exadg/blob/97f92fdc00e514b892e167cdf2f3bcbe4f0f6a0f/include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp#L485

It was also causing problems because it was being called regardless if one uses `local_smoothing`or `global_coarsening`. For `global_coarsening` it is not needed.

### DIFF
--- a/include/exadg/poisson/spatial_discretization/operator.cpp
+++ b/include/exadg/poisson/spatial_discretization/operator.cpp
@@ -130,10 +130,6 @@ Operator<dim, n_components, Number>::distribute_dofs()
 
   dof_handler.distribute_dofs(*fe);
 
-  // TODO: might need adjustments for simplices
-  if(this->grid->triangulation->all_reference_cells_are_hyper_cube())
-    dof_handler.distribute_mg_dofs();
-
   // affine constraints only relevant for continuous FE discretization
   if(param.spatial_discretization == SpatialDiscretization::CG)
   {


nfehn:
did you rebase on exadg/master?

kronbichler:
This looks good to me. This should also resolve #133, right?

kronbichler (approved):

nfehn:
#133 is something different in my opinion

nfehn:
@necioglu I am not sure if "Solving issue:" (in your first comment) will close issue #277 automatically when merging the present PR. But writing "Closes #277" should work.

kronbichler:
> https://github.com/exadg/exadg/issues/133 is something different in my opinion

Yes, I agree now on a second thought.

nfehn:
ok, thanks. Does someone understand the failing check with preCICE?

kronbichler:
It seems to be some technical/github problem:
```
503 Egress is over the account limit.
2022-11-24 13:14:22 ERROR 503: Egress is over the account limit..
```

nfehn:
can we somehow trigger it again, or should we merge the current state?

Issue #128, opened 2022-01-11
Provide quadrature point index in VariableCoefficients class


kronbichler:
Looking at https://github.com/exadg/exadg/pull/122#discussion_r782232750, I realized that https://github.com/exadg/exadg/blob/01939cbd3441629dacd624d5cef041cbb9904db6/include/exadg/operators/variable_coefficients.h#L40-L42 hard-codes the number of quadrature points to `degree + 1`, independent of what a possible use e.g. in https://github.com/exadg/exadg/blob/01939cbd3441629dacd624d5cef041cbb9904db6/include/exadg/incompressible_navier_stokes/spatial_discretization/operators/viscous_operator.cpp#L76 would request: The `quad_index` in that class could be signalling a different number of points, e.g. a 3/2 quadrature rule.

nfehn:
@kronbichler would you like to provide a PR for this problem? It would be required here https://github.com/exadg/exadg/pull/287#issuecomment-1328771261.

PR #287, opened 2022-11-25
generalize InterfaceCoupling towards both surface- and volume-coupled problems

nfehn:
`InterfaceCoupling` currently depends on `ContainerInterfaceData`, which itself depends on a set of boundary IDs. However, the implementation of `InterfaceCoupling` via `deallii::RemotePointEvaluation` is actually more generally and would allow to treat both surface- and volume-coupled problems.

Therefore, I introduce in this PR the class `CouplingDataBase`, a data structure storing the quadrature points and solution values for both surface- and volume-coupled problems.

Two classes derive from `CouplingDataBase`:
1. `CouplingDataSurface` (formerly `ContainerInterfaceData`)
2. `CouplingDataVolume` (a new class to be filled in the future, i.e. this PR or subsequent PR(s))

With these changes, `InterfaceCoupling` only depends on `CouplingDataBase` and can be used for both surface- and volume-coupled problems. For this reason, I also performed the renaming
- `InterfaceCoupling` -> `SolutionInterpolator`

@kronbichler @peterrum For surface coupled problems, we use the three parameters {face index, q-point index, vectorized array index} to obtain the global vector index accessing the array of q-points and solution values. For volume-coupled problems with the coupling data to be used by `MatrixFree` cell/face/boundary-face loops, I do not know whether we want to have the basic data type `Tensor<rank, dim, VectorizedArray<Number>>` instead of `Tensor<rank, dim, Number>` without unrolling the vectorized array (which we do for surface-coupled problems because of the interface of `dealii::Function<dim>`). If you agree on this, I would have to change `CouplingDataBase` a bit in the sense that the template `number_type` needs to have a more general meaning, i.e. not just `double` vs. `float` but also cover `VectorizedArray`.

@jh66637 FYI

### DIFF
--- a/include/exadg/fluid_structure_interaction/driver.cpp
+++ b/include/exadg/fluid_structure_interaction/driver.cpp
@@ -111,7 +111,7 @@ Driver<dim, Number>::setup_interface_coupling()
 
     if(application->fluid->get_parameters().mesh_movement_type == IncNS::MeshMovementType::Poisson)
     {
-      structure_to_ale = std::make_shared<InterfaceCoupling<1, dim, Number>>();
+      structure_to_ale = std::make_shared<SolutionInterpolator<1, dim, Number>>();
       structure_to_ale->setup(fluid->ale_poisson_operator->get_container_interface_data(),
                               structure->pde_operator->get_dof_handler(),
                               *application->structure->get_grid()->mapping,
@@ -121,7 +121,7 @@ Driver<dim, Number>::setup_interface_coupling()
     else if(application->fluid->get_parameters().mesh_movement_type ==
             IncNS::MeshMovementType::Elasticity)
     {
-      structure_to_ale = std::make_shared<InterfaceCoupling<1, dim, Number>>();
+      structure_to_ale = std::make_shared<SolutionInterpolator<1, dim, Number>>();
       structure_to_ale->setup(
         fluid->ale_elasticity_operator->get_container_interface_data_dirichlet(),
         structure->pde_operator->get_dof_handler(),
@@ -151,7 +151,7 @@ Driver<dim, Number>::setup_interface_coupling()
       application->structure->get_boundary_descriptor()->neumann_cached_bc);
     auto const marked_vertices_structure = get_marked_vertices_via_boundary_ids(tria, boundary_ids);
 
-    structure_to_fluid = std::make_shared<InterfaceCoupling<1, dim, Number>>();
+    structure_to_fluid = std::make_shared<SolutionInterpolator<1, dim, Number>>();
     structure_to_fluid->setup(fluid->pde_operator->get_container_interface_data(),
                               structure->pde_operator->get_dof_handler(),
                               *application->structure->get_grid()->mapping,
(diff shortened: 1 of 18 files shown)


nfehn:
For volume coupling: I plan to use plain arrays (as in `CouplingDataBase`) in order to interact with `dealii::RemotePointEvaluation`, and then to reshuffle the data into `ExaDG::VariableCoefficients` data structures which are based on `VectorizedArray<Number>` in order to perform matrix-free operator evaluation.

kronbichler:
This sounds like a good plan. As far as I understand, the current primary focus of the class is the regular case with frequent application, i.e., use inside a `MatrixFree::loop` to be used as a matrix-vector product/operator evaluation with many iterations on a given set of coefficients (which might include coefficients for convection-diffusion problems in the future, a case we don't cover through this mechanism at the moment), including building preconditioners. The question is whether we also think about the case with irregular quadrature points that do not form a tensor product, e.g., because we need to compute sub-cell quadrature formulas to ensure smooth a integrand over the domain of integration, or the case where we use the data just for a simple evaluation of a right hand side like the exchange of some surface data in a partitioned FSI computation. It could be that we wouldn't choose to store data in a separate class and directly consume whatever arrived to compute integrals, so my question might be asking for something this class will not need to do.

nfehn:
> like the exchange of some surface data in a partitioned FSI computation

This is currently done by `CouplingDataSurface`, and I would principally plan to continue like this. Of course we could also think about refactoring this functionality. The logic there is different because we operate only on a set of boundary IDs and we also need to interact with `dealii::Function`.

For `CouplingDataVolume`, the primary use case are matrix-free loops over cells/faces/boundary faces.

nfehn:
> It could be that we wouldn't choose to store data in a separate class and directly consume whatever arrived to compute integrals

I did not completely understand which scenario/use case you are referring to here? Do you mean not reshuffling data from `std::vector` to `dealii::Table` with `VectorizedArray` inside, as required by `MatrixFree` loops?

To further explain the use case I have in mind: I am thinking of some partitioned coupling between single-field solvers via a surface or the volume. Exchange of data is done rarely as compared to iterating on the individual fields in order to compute their single-field solutions, e.g. via preconditioned Krylov solvers.

kronbichler:
> I did not completely understand which scenario/use case you are referring to here? Do you mean not reshuffling data from std::vector to dealii::Table with VectorizedArray inside, as required by MatrixFree loops?

Yes exactly, I was thinking of a case where we would not re-shuffle data but request the data from a pair (`cell`, `quadrature_point`) that describes the quadrature formula, and then directly compute a product with test functions or whatever we need. This is not the main goal you describe here, so my initial request seems to fall outside the scope of the proposed class (and class hierarchy). All I wanted to say is that we do not necessarily need to limit ourselves to the `FEEvaluation` (`CellIntegrator`, `FaceIntegrator`) with vectorization across cells. This more flexible setup might cost some performance if the operator evaluation is done many times, but OK or even faster if done only once. Or we might have no choice when we have no tensor product, but in the latter case we might still pick a suitable vectorized scheme, to be compared with the work by @bergbauer.

nfehn:
It could be that a single-field problem that uses data from another field evaluates this field only once per time step, e.g. because of explicit time stepping. In this case one can probably argue that the field with explicit time stepping is much cheaper anyways and that the costs are spent somewhere else (in the implicit solver). Of course, one could also think of purely explicit time stepping for all fields, but then one would probably implement the PDE solvers or the coupling between them differently.

nfehn:
@kronbichler I implemented something here https://github.com/exadg/exadg/pull/287/commits/b71d0954ec3d04c4b6def926d1ab907bcf9c1029

However, this looks like very basic `MatrixFree` functionality and I wonder whether this is already available in dealii in some form? In particular, I mean the functions to translate certain indices (cell/face, q-point, vectorized-array) into a global index for a `std::vector`.

Currently, the number of quadrature points per cell or face is uninitialized. How would I access this information via `MatrixFree`, see issue #128?

PR #288, opened 2022-11-25
[WIP] choose mesh smoothing

necioglu:
Changed the  mesh smoothing to [none](https://github.com/necioglu/dealii/blob/bdae22407dc0425211e9f4772fa2255dae1e315b/include/deal.II/grid/tria.h#L1155), the distributed triangulation setting to [default_setting](https://github.com/necioglu/dealii/blob/bdae22407dc0425211e9f4772fa2255dae1e315b/include/deal.II/distributed/tria.h#L302) and the triangulation description setting to [default_setting](https://github.com/necioglu/dealii/blob/bdae22407dc0425211e9f4772fa2255dae1e315b/include/deal.II/grid/tria_description.h#L304) for global coarsening, since the options limit_level_differences_at_vertices and construct_multigrid_hierarchy are not needed for global coarsening. For local smoothing they are set.

Also added an Assert for using use_global_coarsening but not activating create_coarse_triangulations.

### DIFF
--- a/applications/poisson/sine/application.h
+++ b/applications/poisson/sine/application.h
@@ -174,6 +174,11 @@ private:
     this->param.multigrid_data.coarse_problem.preconditioner =
       MultigridCoarseGridPreconditioner::AMG;
     this->param.multigrid_data.coarse_problem.solver_data.rel_tol = 1.e-3;
+
+    AssertThrow(this->param.grid.create_coarse_triangulations ==
+                  this->param.multigrid_data.use_global_coarsening,
+                dealii::ExcMessage(
+                  "Can't use global_coarsening without activating create_coarse_triangulations"))
   }
 
   void
@@ -280,10 +285,7 @@ private:
     pp_data.output_data.directory                   = this->output_parameters.directory + "vtu/";
     pp_data.output_data.filename                    = this->output_parameters.filename;
     pp_data.output_data.write_higher_order          = true;
-    if(this->param.grid.element_type == ElementType::Simplex and dim == 3)
-      AssertThrow(pp_data.output_data.write_higher_order == false,
-                  dealii::ExcMessage("Can't use higher order output with 3D Simplices."));
-    pp_data.output_data.degree = this->param.degree;
+    pp_data.output_data.degree                      = this->param.degree;
 
     pp_data.error_data.time_control_data.is_active = true;
     pp_data.error_data.analytical_solution.reset(new Solution<dim>());
@@ -291,6 +293,10 @@ private:
 
     std::shared_ptr<PostProcessorBase<dim, Number>> pp;
     pp.reset(new PostProcessor<dim, Number>(pp_data, this->mpi_comm));
+
+    if(this->param.grid.element_type == ElementType::Simplex and dim == 3)
+      AssertThrow(pp_data.output_data.write_higher_order == false,
+                  dealii::ExcMessage("Can't use higher order output with 3D Simplices."));
 
     return pp;
   }
--- a/applications/poisson/sine/application.h
+++ b/applications/poisson/sine/application.h
@@ -174,11 +174,6 @@ private:
     this->param.multigrid_data.coarse_problem.preconditioner =
       MultigridCoarseGridPreconditioner::AMG;
     this->param.multigrid_data.coarse_problem.solver_data.rel_tol = 1.e-3;
-
-    AssertThrow(this->param.grid.create_coarse_triangulations ==
-                  this->param.multigrid_data.use_global_coarsening,
-                dealii::ExcMessage(
-                  "Can't use global_coarsening without activating create_coarse_triangulations"))
   }
 
   void
--- a/include/exadg/grid/grid.cpp
+++ b/include/exadg/grid/grid.cpp
@@ -72,19 +72,10 @@ BalancedGranularityPartitionPolicy<dim, spacedim>::partition(
 template<int dim>
 Grid<dim>::Grid(const GridData & data, MPI_Comm const & mpi_comm)
 {
-  typename dealii::parallel::distributed::Triangulation<dim>::Settings distributed_settings;
-
   if(data.create_coarse_triangulations)
-  {
-    mesh_smoothing       = dealii::Triangulation<dim>::none;
-    distributed_settings = dealii::parallel::distributed::Triangulation<dim>::default_setting;
-  }
+    mesh_smoothing = dealii::Triangulation<dim>::none;
   else
-  {
     mesh_smoothing = dealii::Triangulation<dim>::limit_level_difference_at_vertices;
-    distributed_settings =
-      dealii::parallel::distributed::Triangulation<dim>::construct_multigrid_hierarchy;
-  }
 
   // triangulation
   if(data.triangulation_type == TriangulationType::Serial)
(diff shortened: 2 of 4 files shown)


necioglu:
I have noticed an error in this approach. The default case is now a triangulation with limit_level_differences_at_vertices, since we decide with the variable use_global_coarsening. This approach is correct, if a multigrid preconditioner is used. But if the user is not using a multigrid preconditioner, limit_level_differences_at_vertices is set, although it is not needed. However the grid class doesn't know, if a preconditioner is used. 

If we plan on making the global_coarsening case the default, this problem would also be solved. If not, I think adding 3 user parameters in gridData and checking the correct combinations with asserts in application.h would be best solution. What do you think @nfehn?

applications/poisson/sine/application.h
```diff
@@ -174,6 +174,11 @@ class Application : public ApplicationBase<dim, n_components, Number>
     this->param.multigrid_data.coarse_problem.preconditioner =
       MultigridCoarseGridPreconditioner::AMG;
     this->param.multigrid_data.coarse_problem.solver_data.rel_tol = 1.e-3;
+
+    AssertThrow(this->param.grid.create_coarse_triangulations ==
+                  this->param.multigrid_data.use_global_coarsening,
+                dealii::ExcMessage(
+                  "Can't use global_coarsening without activating create_coarse_triangulations"))
```
nfehn, line 181: This Assert is not placed optimally here in application. A user does not want to add this Assert to other test cases.

include/exadg/grid/grid.cpp
```diff
@@ -72,18 +72,32 @@ BalancedGranularityPartitionPolicy<dim, spacedim>::partition(
 template<int dim>
 Grid<dim>::Grid(const GridData & data, MPI_Comm const & mpi_comm)
 {
+  typename dealii::parallel::distributed::Triangulation<dim>::Settings distributed_settings;
+
+  if(data.create_coarse_triangulations)
```
nfehn, line 77: if you think `create_coarse_triangulations` is a dependent parameter, it should not be part of `GridData`, but instead be passed to the constructor of `Grid` as an argument. When calling the constructor, this parameter has to be correctly deduced from the other `Parameters`. There are examples in ExaDG where we added public functions to `Parameters` in order to obtain certain information already contained in other parameters.

nfehn:
@necioglu I very much appreciate your thoughts. As you describe, our goal is a solution where we do not have implicit assumptions in the code, but where the deduction of certain parameters and dealii-settings is very clear from the parameter lists we provide in ExaDG. We probably still have to think about the optimal solution.

include/exadg/grid/grid.cpp
```diff
@@ -72,18 +72,32 @@ BalancedGranularityPartitionPolicy<dim, spacedim>::partition(
 template<int dim>
 Grid<dim>::Grid(const GridData & data, MPI_Comm const & mpi_comm)
 {
+  typename dealii::parallel::distributed::Triangulation<dim>::Settings distributed_settings;
```
nfehn, line 75: if this is only used for parallel::distributed::Triangulation, it should only appear in the corresponding if-branch.

include/exadg/grid/grid.cpp
```diff
@@ -233,15 +247,22 @@ Grid<dim>::do_create_triangulation(
...
-        triangulation->get_communicator(),
-        group_size,
-        dealii::Triangulation<dim>::none,
-        dealii::TriangulationDescription::construct_multigrid_hierarchy);
+      create_description_from_triangulation_in_groups<dim, dim>(serial_grid_generator,
+                                                                serial_grid_partitioner,
+                                                                triangulation->get_communicator(),
+                                                                group_size,
+                                                                mesh_smoothing,
+                                                                triangulation_description_setting);
```
nfehn, line 265: this will change the code in my opinion. Is this intended?

include/exadg/grid/grid.cpp
necioglu: moved it to the corresponding if-branch.

include/exadg/grid/grid.cpp
necioglu: Yes, the reason behind this is the same as setting the distributed triangulation setting to the default setting. If construct multigrid hierarchy is activated, limit level differences at vertices is set automatically: https://github.com/dealii/dealii/blob/69313620a085f92bc4f8e619424eeeba47927840/source/grid/tria_description.cc#L822 

In both cases the code changes.

include/exadg/grid/grid.h
```diff
@@ -81,6 +81,11 @@ class Grid
   std::shared_ptr<dealii::Mapping<dim> const>
   get_mapping() const;
 
+  /**
+   * dealii::Triangulation::MeshSmoothing
+   */
+  typename dealii::Triangulation<dim>::MeshSmoothing mesh_smoothing;
```
peterrum, line 87: Do we need to store this info?

peterrum (approved):

include/exadg/grid/grid.h
nfehn: It is not only used in the constructor, but also in do_create_triangulation()

necioglu: I have only created the parameter to avoid code duplication, since `mesh_smoothing` is needed in the constructor of the grid class and in `do_create_triangulation`. I can change it and paste a copy of the if-else in `do_create_triangulation`.

nfehn: copying the code is not an option. The only option I see is a small private function get_mesh_smoothing() returning mesh_smoothing.

nfehn: Then, the member variable can be removed as suggested by @peterrum

include/exadg/grid/grid.cpp
```diff
@@ -233,15 +246,22 @@ Grid<dim>::do_create_triangulation(
...
-        serial_grid_partitioner,
-        triangulation->get_communicator(),
-        group_size,
-        dealii::Triangulation<dim>::none,
-        dealii::TriangulationDescription::construct_multigrid_hierarchy);
+      create_description_from_triangulation_in_groups<dim, dim>(serial_grid_generator,
+                                                                serial_grid_partitioner,
+                                                                triangulation->get_communicator(),
+                                                                group_size,
+                                                                mesh_smoothing,
```
peterrum, line 263:
```suggestion
                                                                triangulation->get_mesh_smoothing(),
```

necioglu: Thanks! This works, but we rely on `triangulation_description_setting` changing the mesh smoothing from none to limit_level_difference_at_vertices in case of local smoothing.

peterrum: Why is it `none`? I don't understand this.

necioglu: Because if I do `triangulation->get_mesh_smoothing()`, I get the mesh smoothing of the triangulation built in:
https://github.com/exadg/exadg/blob/22aebf621ea018865c6838eda526dd15b6a3e8bb/include/exadg/grid/grid.cpp#L90-L91
which is `none`. The following code sniplet

```
    std::cout << "Mesh Smoothing before create_description_from_triangulation_in_groups"
              << std::endl;
    switch(triangulation->get_mesh_smoothing())
    {
      case dealii::Triangulation<dim>::none:
        std::cout << "  none" << std::endl;
        break;
      case dealii::Triangulation<dim>::limit_level_difference_at_vertices:
        std::cout << "  limit_level_difference_at_vertices" << std::endl;
        break;
    }

    // TODO SIMPLEX: this will not work in case of simplex meshes
    auto const description = dealii::TriangulationDescription::Utilities::
      create_description_from_triangulation_in_groups<dim, dim>(
        serial_grid_generator,
        serial_grid_partitioner,
        triangulation->get_communicator(),
        group_size,
        triangulation->get_mesh_smoothing(),
        dealii::TriangulationDescription::Settings::construct_multigrid_hierarchy);

    triangulation->create_triangulation(description);

    std::cout << "Mesh Smoothing after create_description_from_triangulation_in_groups"
              << std::endl;
    switch(triangulation->get_mesh_smoothing())
    {
      case dealii::Triangulation<dim>::none:
        std::cout << "  none" << std::endl;
        break;
      case dealii::Triangulation<dim>::limit_level_difference_at_vertices:
        std::cout << "  limit_level_difference_at_vertices" << std::endl;
        break;
    }
```
gives this output:
```
Mesh Smoothing before create_description_from_triangulation_in_groups
  none
Mesh Smoothing after create_description_from_triangulation_in_groups
  limit_level_difference_at_vertices
```

nfehn (approved):

PR #295, opened 2022-12-12
Create coarse triangulations for FullyDistributed

necioglu:
Create the coarse triangulations needed for global coarsening multigrid manually in grid class.

### DIFF
--- a/applications/poisson/sine/application.h
+++ b/applications/poisson/sine/application.h
@@ -185,24 +185,23 @@ private:
     unsigned int n_cells_1d =
       std::max((unsigned int)2, this->param.grid.n_subdivisions_1d_hypercube);
 
-    if(this->param.grid.element_type == ElementType::Hypercube)
+    auto const lambda_create_coarse_triangulation = [&](dealii::Triangulation<dim, dim> & tria)
     {
-      dealii::GridGenerator::subdivided_hyper_cube(*this->grid->triangulation,
-                                                   n_cells_1d,
-                                                   left,
-                                                   right);
-    }
-    else if(this->param.grid.element_type == ElementType::Simplex)
-    {
-      dealii::GridGenerator::subdivided_hyper_cube_with_simplices(*this->grid->triangulation,
-                                                                  n_cells_1d,
-                                                                  left,
-                                                                  right);
-    }
-    else
-    {
-      AssertThrow(false, ExcNotImplemented());
-    }
+      if(this->param.grid.element_type == ElementType::Hypercube)
+      {
+        dealii::GridGenerator::subdivided_hyper_cube(tria, n_cells_1d, left, right);
+      }
+      else if(this->param.grid.element_type == ElementType::Simplex)
+      {
+        dealii::GridGenerator::subdivided_hyper_cube_with_simplices(tria, n_cells_1d, left, right);
+      }
+      else
+      {
+        AssertThrow(false, ExcNotImplemented());
+      }
+    };
+
+    this->grid->create_triangulation(this->param.grid, lambda_create_coarse_triangulation);
 
     if(mesh_type == MeshType::Cartesian)
     {
@@ -250,8 +249,6 @@ private:
         }
       }
     }
-
-    this->grid->triangulation->refine_global(this->param.grid.n_refine_global);
   }
 
 
--- a/applications/poisson/sine/application.h
+++ b/applications/poisson/sine/application.h
@@ -185,8 +185,7 @@ private:
     unsigned int n_cells_1d =
       std::max((unsigned int)2, this->param.grid.n_subdivisions_1d_hypercube);
 
-    auto const lambda_create_coarse_triangulation = [&](dealii::Triangulation<dim, dim> & tria)
-    {
+    auto const lambda_create_coarse_triangulation = [&](dealii::Triangulation<dim, dim> & tria) {
       if(this->param.grid.element_type == ElementType::Hypercube)
       {
         dealii::GridGenerator::subdivided_hyper_cube(tria, n_cells_1d, left, right);
@@ -201,7 +200,9 @@ private:
       }
     };
 
-    this->grid->create_triangulation(this->param.grid, lambda_create_coarse_triangulation);
+    this->grid->create_triangulation(this->param.grid,
+                                     lambda_create_coarse_triangulation,
+                                     this->param.grid.n_refine_global);
 
     if(mesh_type == MeshType::Cartesian)
     {
--- a/include/exadg/grid/grid.cpp
+++ b/include/exadg/grid/grid.cpp
@@ -190,16 +190,19 @@ Grid<dim>::create_triangulation(
       }
 
       // undo local refinements
-      while(*std::max_element(refine_local.begin(), refine_local.end()) != 0)
+      if(refine_local.size() > 0)
       {
-        for(size_t material_id = 0; material_id < refine_local.size(); material_id++)
+        while(*std::max_element(refine_local.begin(), refine_local.end()) != 0)
         {
-          if(refine_local[material_id] > 0)
+          for(size_t material_id = 0; material_id < refine_local.size(); material_id++)
           {
-            refine_local[material_id]--;
+            if(refine_local[material_id] > 0)
+            {
+              refine_local[material_id]--;
+            }
           }
+          lambda_create_coarse_levels(0, refine_local);
         }
-        lambda_create_coarse_levels(0, refine_local);
       }
     }
     else
--- a/include/exadg/grid/grid.cpp
+++ b/include/exadg/grid/grid.cpp
@@ -169,9 +169,10 @@ Grid<dim>::create_triangulation(
         triangulation->n_global_levels());
 
       // we start with the finest level
-      unsigned int level = triangulation->n_global_levels() - 1;
+      unsigned int              level        = triangulation->n_global_levels() - 1;
+      std::vector<unsigned int> refine_local = vector_local_refinements;
 
-      // lambda function for creating and placing the coarse triangulations
+      // lambda function for creating the coarse triangulations
       auto const lambda_create_level_triangulation = [&](unsigned int              refine_global,
                                                          std::vector<unsigned int> refine_local) {
         auto level_tria = std::make_shared<dealii::parallel::fullydistributed::Triangulation<dim>>(
@@ -180,18 +181,16 @@ Grid<dim>::create_triangulation(
         do_create_triangulation(
           level_tria, data, create_coarse_triangulation, refine_global, refine_local);
 
-        coarse_triangulations[level] = level_tria;
-
-        level--;
+        return level_tria;
       };
-
-      // start with the finest level
-      std::vector<unsigned int> refine_local = vector_local_refinements;
 
       // undo global refinements
       for(int refine_global = global_refinements; refine_global >= 0; --refine_global)
       {
-        lambda_create_level_triangulation(refine_global, refine_local);
+        coarse_triangulations[level] =
+          lambda_create_level_triangulation(refine_global, refine_local);
+
+        level--;
       }
 
       // undo local refinements
@@ -206,7 +205,9 @@ Grid<dim>::create_triangulation(
               refine_local[material_id]--;
             }
           }
-          lambda_create_level_triangulation(0, refine_local);
+          coarse_triangulations[level] = lambda_create_level_triangulation(0, refine_local);
+
+          level--;
         }
       }
     }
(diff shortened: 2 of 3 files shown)


include/exadg/grid/grid.cpp
```diff
@@ -160,10 +160,47 @@ Grid<dim>::create_triangulation(
...
+      while(refine_local != std::vector<unsigned int>(refine_local.size(), 0))
+      {
+        for(size_t material_id = 0; material_id < refine_local.size(); material_id++)
+        {
+          if(refine_local[material_id] > 0)
+          {
+            refine_local[material_id]--;
+          }
+        }
+        lambda_create_coarse_levels(0, refine_local);
```
peterrum, line 202: I don't understand the logic here. Could you explain this?

necioglu: We use the `do_create_triangulation` repeatedly to create the coarse triangulations needed. First we "undo" the global refinements, i.e., we reduce the global refinement by one each time until there is no global refinement and save the triangulation to the `coarse_triangulations` vector. Afterwards we "undo" the `local_refinements` by checking if the `refine_local` vector is all zeros, i.e., there are no local refinements left. If there are local refinements, we reduce each entity by one,build the triangulation, save it and continue reducing until there is none.

include/exadg/grid/grid.cpp
```diff
@@ -160,10 +160,47 @@ Grid<dim>::create_triangulation(
...
 
-      AssertThrow(false, ExcNotImplemented());
+      // start with the finest level
+      std::vector<unsigned int> refine_local = vector_local_refinements;
+
+      // undo global refinements
+      for(int it = global_refinements; it >= 0; --it)
+      {
+        lambda_create_coarse_levels(it, refine_local);
+      }
```
peterrum, line 190:
```suggestion
      std::vector<unsigned int> refine_local = vector_local_refinements;
      for(auto & i : refine_local)
        i += global_refinements;
```

include/exadg/grid/grid.cpp
```diff
@@ -160,10 +160,47 @@ Grid<dim>::create_triangulation(
...
+
+      // undo local refinements
+      while(refine_local != std::vector<unsigned int>(refine_local.size(), 0))
```
peterrum, line 193: Please use `std::max(...)!=0`.

include/exadg/grid/grid.cpp
peterrum: Makes sense. I have suggested a way to simplify the code a bit.

include/exadg/grid/grid.cpp
necioglu: I don't see how this would do the same thing as the original code.

include/exadg/grid/grid.cpp
necioglu: done!

include/exadg/grid/grid.cpp
necioglu: Yes that makes sense if we always have a local refinement, but what happens if the `refine_local` vector is empty? Then the `global_refinements` are also disregarded, and no `coarse_triangulations` are created.

peterrum: Can that be empty? If yes let's keep it like this!

necioglu: Yes, it is the default case.

https://github.com/exadg/exadg/blob/7edf07c5a30582accdd72ce2af833caa51382107/include/exadg/grid/grid.h#L68-L73

bergbauer (approved):
LGTM!

include/exadg/grid/grid.cpp
```diff
@@ -160,10 +160,50 @@ Grid<dim>::create_triangulation(
...
-      // TODO
-      // construct all the coarser triangulations
+      // resize the empty coarse triangulations vector
+      coarse_triangulations = std::vector<std::shared_ptr<dealii::Triangulation<dim> const>>(
+        triangulation->n_global_levels());
+
+      unsigned int coarse_level = triangulation->n_global_levels() - 1;
+
+      // lambda function for creating and placing the coarse triangulations
+      auto const lambda_create_coarse_levels = [&](unsigned int              global_refinement,
```
nfehn, line 170: the name should be `refine_global` for consistency with the second argument `refine_local`.

include/exadg/grid/grid.cpp
```diff
@@ -160,10 +160,50 @@ Grid<dim>::create_triangulation(
     {
       // we need to generate the coarse triangulations explicitly
 
-      // TODO
-      // construct all the coarser triangulations
+      // resize the empty coarse triangulations vector
+      coarse_triangulations = std::vector<std::shared_ptr<dealii::Triangulation<dim> const>>(
+        triangulation->n_global_levels());
+
+      unsigned int coarse_level = triangulation->n_global_levels() - 1;
```
nfehn, line 167: The name `coarse_level` is misleading, because the lambda `create_coarse_triangulation` refers to the coarsest level. Hence, there is a mismatch in notation. Just use `level` instead, and add a comment that you start, by convention, with the highest level.

include/exadg/grid/grid.cpp
nfehn, line 170: The name `lambda_create_coarse_levels` does also not fit very well, since the lambda function creates the triangulation just for one level, not for all. The current name reads like the function creates all coarse levels at once. I suggest `lambda_create_level_triangulation` for consistency with the other names.

include/exadg/grid/grid.cpp
```diff
@@ -160,10 +160,50 @@ Grid<dim>::create_triangulation(
...
+
+      // lambda function for creating and placing the coarse triangulations
+      auto const lambda_create_coarse_levels = [&](unsigned int              global_refinement,
+                                                   std::vector<unsigned int> refine_local) {
+        auto new_tria = std::make_shared<dealii::parallel::fullydistributed::Triangulation<dim>>(
```
nfehn, line 172: I would then use `level_tria` here, because `new` is very general and can mean a lot.

include/exadg/grid/grid.cpp
```diff
@@ -144,6 +142,8 @@ Grid<dim>::create_triangulation(
     // coarse grid triangulations
     if(data.triangulation_type == TriangulationType::Serial)
     {
+      AssertThrow(triangulation->all_reference_cells_are_hyper_cube(), dealii::ExcNotImplemented());
```
nfehn, line 145: In my opinion, this needs some more explanations. As a new user/developer of ExaDG, I can deduce that Serial requires hypercubes, but I would like to know the reason, e.g. that this is a current limitation of deal.II. Why does simplex not work here? I think this should be mentioned here. Imagine deal.II becomes more powerful. How should a developer of ExaDG know in 1-2 years why we have this assert in the code and how should he decide whether it can be removed or not. A developer might remove it and then realize it works, but he will then ask himself whether this is just luck and whether it will work in all cases ...

necioglu: That is why I used `dealii::ExcNotImplemented()` and not `ExaDG::ExcNotImplemented()`. I thought this made clear, that currently this is not allowed, because of dealii and not because of ExaDG. However, I agree that adding more explanation would be helpful.

include/exadg/grid/grid.cpp
necioglu: I agree. I wanted to keep our naming similar to dealii. See the equivalent function in dealii using the variable name `new_tria`: https://github.com/dealii/dealii/blob/07b5a85db9d97aac577b56873c6c34ff4f3bf73a/include/deal.II/multigrid/mg_transfer_global_coarsening.templates.h#L2439-L2451

include/exadg/grid/grid.cpp
necioglu: changed in fe49537a6dd436ad6b49130feb488cd584e5d308.

necioglu: changed in fe49537a6dd436ad6b49130feb488cd584e5d308.

include/exadg/grid/grid.cpp
necioglu: changed in fe49537a6dd436ad6b49130feb488cd584e5d308.

include/exadg/grid/grid.cpp
necioglu: changed in fe49537a6dd436ad6b49130feb488cd584e5d308.

include/exadg/grid/grid.cpp
```diff
@@ -160,10 +164,51 @@ Grid<dim>::create_triangulation(
...
+
+        level--;
+      };
 
-      AssertThrow(false, ExcNotImplemented());
+      // start with the finest level
+      std::vector<unsigned int> refine_local = vector_local_refinements;
+
+      // undo global refinements
+      for(int it = global_refinements; it >= 0; --it)
```
nfehn, line 192: a last suggestion for renaming would be to use `refine_global` here instead of `it`.

nfehn:
I made one more comment. Apart from that, it looks very good from my side!

include/exadg/grid/grid.cpp
necioglu: done!

include/exadg/grid/grid.cpp
```diff
@@ -160,10 +164,51 @@ Grid<dim>::create_triangulation(
...
+                                                         std::vector<unsigned int> refine_local) {
+        auto level_tria = std::make_shared<dealii::parallel::fullydistributed::Triangulation<dim>>(
+          triangulation->get_communicator());
+
+        do_create_triangulation(
+          level_tria, data, create_coarse_triangulation, refine_global, refine_local);
+
+        coarse_triangulations[level] = level_tria;
+
+        level--;
```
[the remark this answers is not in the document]
nfehn: wouldn't it be better if the lambda returned the coarse triangulation and filling the vector is done outside the lambda?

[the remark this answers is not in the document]
necioglu: done!

include/exadg/grid/grid.cpp
```diff
@@ -160,10 +164,52 @@ Grid<dim>::create_triangulation(
...
 
-      // TODO
-      // construct all the coarser triangulations
+      // resize the empty coarse triangulations vector
+      coarse_triangulations = std::vector<std::shared_ptr<dealii::Triangulation<dim> const>>(
+        triangulation->n_global_levels());
+
+      // we start with the finest level
+      unsigned int              level        = triangulation->n_global_levels() - 1;
+      std::vector<unsigned int> refine_local = vector_local_refinements;
```
nfehn, line 173: Why do you define these variables here? Wouldn't it make more sense to shift these lines to line 187 below.

nfehn:
Some of the checks were not successful. Overall, I think this PR can be merged with only some minor changes, so it would be good to make progress here and complete this PR.

include/exadg/grid/grid.cpp
necioglu: done!

necioglu: I also moved lines 167-169.

necioglu:
> Some of the checks were not successful. Overall, I think this PR can be merged with only some minor changes, so it would be good to make progress here and complete this PR.

I wasn't sure on what to do the about the failing checks. They didn't seemed like they were caused because of this PR, since the master is/was failing also. I rebased and pushed now.

nfehn:
looks good from my side

applications/poisson/sine/application.h
```diff
@@ -185,24 +185,24 @@ class Application : public ApplicationBase<dim, n_components, Number>
...
+      }
+      else
+      {
+        AssertThrow(false, ExcNotImplemented());
+      }
+    };
+
+    this->grid->create_triangulation(this->param.grid,
+                                     lambda_create_coarse_triangulation,
+                                     this->param.grid.n_refine_global);
```
nfehn, line 205: According to my current understanding, this change here is problematic in terms of the bug discussed in issue #313 because the part with Neumann boundary conditions below is done after this function call, which is not allowed, because the `coarse_grid_triangulations` for global-coarsening multigrid then have the wrong boundary conditions. So I conclude the bug has not been introduced by changes in #269, but PR #269 did not update this particular test case (poisson/sine) so that it can be used for global coarsening multigrid.

necioglu: I am looking into this!

nfehn: I moved the Neumann part into the lambda and it works. Im already working on it.

necioglu: Okay. Thanks!

nfehn: The problem is that - by the current design - we can not solve the problem with curvilinear mesh in poisson/sine correctly, see issue #319.

PR #296, opened 2022-12-12
Interior penalty parameter for simplex

necioglu:
Add the interior penalty parameter for simplices.

K. Shahbazi, An explicit expression for the penalty parameter of the interior penalty method, Journal of Computational Physics 205, 401 – 407, 2005.

Equation 7.

### DIFF
--- a/include/exadg/compressible_navier_stokes/spatial_discretization/kernels_and_operators.h
+++ b/include/exadg/compressible_navier_stokes/spatial_discretization/kernels_and_operators.h
@@ -960,27 +960,6 @@ public:
   typedef dealii::Tensor<2, dim, dealii::VectorizedArray<Number>> tensor;
   typedef dealii::Point<dim, dealii::VectorizedArray<Number>>     point;
 
-  ElementType
-  get_element_type(FaceIntegratorScalar & integrator) const
-  {
-    if(integrator.get_matrix_free()
-         .get_dof_handler()
-         .get_triangulation()
-         .all_reference_cells_are_simplex())
-    {
-      return ElementType::Simplex;
-    }
-    else if(integrator.get_matrix_free()
-              .get_dof_handler()
-              .get_triangulation()
-              .all_reference_cells_are_hyper_cube())
-    {
-      return ElementType::Hypercube;
-    }
-    else
-      AssertThrow(false, dealii::ExcMessage("Only hypercube or simplex elements are supported."));
-  }
-
   ViscousOperator() : matrix_free(nullptr), degree(1)
   {
   }
@@ -1036,7 +1015,11 @@ public:
     scalar tau =
       std::max(fe_eval_m.read_cell_data(array_penalty_parameter),
                fe_eval_p.read_cell_data(array_penalty_parameter)) *
-      IP::get_penalty_factor<dim, Number>(degree, get_element_type(fe_eval_m), data.IP_factor) * nu;
+      IP::get_penalty_factor<dim, Number>(
+        degree,
+        get_element_type(fe_eval_m.get_matrix_free().get_dof_handler().get_triangulation()),
+        data.IP_factor) *
+      nu;
 
     return tau;
   }
@@ -1047,7 +1030,11 @@ public:
   {
     scalar tau =
       fe_eval.read_cell_data(array_penalty_parameter) *
-      IP::get_penalty_factor<dim, Number>(degree, get_element_type(fe_eval), data.IP_factor) * nu;
+      IP::get_penalty_factor<dim, Number>(
+        degree,
+        get_element_type(fe_eval.get_matrix_free().get_dof_handler().get_triangulation()),
+        data.IP_factor) *
+      nu;
 
     return tau;
   }
--- a/include/exadg/compressible_navier_stokes/spatial_discretization/kernels_and_operators.h
+++ b/include/exadg/compressible_navier_stokes/spatial_discretization/kernels_and_operators.h
@@ -960,13 +960,21 @@ public:
   typedef dealii::Tensor<2, dim, dealii::VectorizedArray<Number>> tensor;
   typedef dealii::Point<dim, dealii::VectorizedArray<Number>>     point;
 
-  bool
-  using_simplex(FaceIntegratorScalar & integrator) const
+  ElementType
+  get_element_type(FaceIntegratorScalar & integrator) const
   {
-    return integrator.get_matrix_free()
-      .get_dof_handler()
-      .get_triangulation()
-      .all_reference_cells_are_simplex();
+    if(integrator.get_matrix_free()
+         .get_dof_handler()
+         .get_triangulation()
+         .all_reference_cells_are_simplex())
+      return ElementType::Simplex;
+    else if(integrator.get_matrix_free()
+              .get_dof_handler()
+              .get_triangulation()
+              .all_reference_cells_are_hyper_cube())
+      return ElementType::Hypercube;
+    else
+      AssertThrow(false, dealii::ExcMessage("Only hypercube or simplex elements are supported."));
   }
 
   ViscousOperator() : matrix_free(nullptr), degree(1)
@@ -1024,7 +1032,7 @@ public:
     scalar tau =
       std::max(fe_eval_m.read_cell_data(array_penalty_parameter),
                fe_eval_p.read_cell_data(array_penalty_parameter)) *
-      IP::get_penalty_factor<dim, Number>(degree, using_simplex(fe_eval_m), data.IP_factor) * nu;
+      IP::get_penalty_factor<dim, Number>(degree, get_element_type(fe_eval_m), data.IP_factor) * nu;
 
     return tau;
   }
@@ -1035,7 +1043,7 @@ public:
   {
     scalar tau =
       fe_eval.read_cell_data(array_penalty_parameter) *
-      IP::get_penalty_factor<dim, Number>(degree, using_simplex(fe_eval), data.IP_factor) * nu;
+      IP::get_penalty_factor<dim, Number>(degree, get_element_type(fe_eval), data.IP_factor) * nu;
 
     return tau;
   }
--- a/include/exadg/compressible_navier_stokes/spatial_discretization/kernels_and_operators.h
+++ b/include/exadg/compressible_navier_stokes/spatial_discretization/kernels_and_operators.h
@@ -1017,7 +1017,8 @@ public:
                fe_eval_p.read_cell_data(array_penalty_parameter)) *
       IP::get_penalty_factor<dim, Number>(
         degree,
-        get_element_type(fe_eval_m.get_matrix_free().get_dof_handler().get_triangulation()),
+        get_element_type(
+          fe_eval_m.get_matrix_free().get_dof_handler(data.dof_index).get_triangulation()),
         data.IP_factor) *
       nu;
 
@@ -1028,13 +1029,13 @@ public:
     scalar
     get_penalty_parameter(FaceIntegratorScalar & fe_eval) const
   {
-    scalar tau =
-      fe_eval.read_cell_data(array_penalty_parameter) *
-      IP::get_penalty_factor<dim, Number>(
-        degree,
-        get_element_type(fe_eval.get_matrix_free().get_dof_handler().get_triangulation()),
-        data.IP_factor) *
-      nu;
+    scalar tau = fe_eval.read_cell_data(array_penalty_parameter) *
+                 IP::get_penalty_factor<dim, Number>(
+                   degree,
+                   get_element_type(
+                     fe_eval.get_matrix_free().get_dof_handler(data.dof_index).get_triangulation()),
+                   data.IP_factor) *
+                 nu;
 
     return tau;
   }
--- a/include/exadg/compressible_navier_stokes/spatial_discretization/kernels_and_operators.h
+++ b/include/exadg/compressible_navier_stokes/spatial_discretization/kernels_and_operators.h
@@ -1012,9 +1012,10 @@ public:
     scalar
     get_penalty_parameter(FaceIntegratorScalar & fe_eval_m, FaceIntegratorScalar & fe_eval_p) const
   {
-    scalar tau = std::max(fe_eval_m.read_cell_data(array_penalty_parameter),
-                          fe_eval_p.read_cell_data(array_penalty_parameter)) *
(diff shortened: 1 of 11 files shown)


include/exadg/operators/interior_penalty_parameter.h
```diff
@@ -94,11 +94,16 @@ calculate_penalty_parameter(
...
+template<int dim, typename Number>
 Number
-get_penalty_factor(unsigned int const degree, Number const factor = 1.0)
+get_penalty_factor(unsigned int const                      degree,
+                   dealii::MatrixFree<dim, Number> const & matrix_free,
+                   Number const                            factor = 1.0)
 {
-  return factor * (degree + 1.0) * (degree + 1.0);
+  if(matrix_free.get_dof_handler().get_triangulation().all_reference_cells_are_simplex())
+    return factor * (degree + 1.0) * (degree + dim) / dim;
```
peterrum, line 104: Could you add a source here?

include/exadg/compressible_navier_stokes/spatial_discretization/kernels_and_operators.h
```diff
@@ -1012,9 +1012,10 @@ class ViscousOperator
     scalar
     get_penalty_parameter(FaceIntegratorScalar & fe_eval_m, FaceIntegratorScalar & fe_eval_p) const
   {
-    scalar tau = std::max(fe_eval_m.read_cell_data(array_penalty_parameter),
-                          fe_eval_p.read_cell_data(array_penalty_parameter)) *
-                 IP::get_penalty_factor<Number>(degree, data.IP_factor) * nu;
+    scalar tau =
+      std::max(fe_eval_m.read_cell_data(array_penalty_parameter),
+               fe_eval_p.read_cell_data(array_penalty_parameter)) *
+      IP::get_penalty_factor<dim, Number>(degree, fe_eval_m.get_matrix_free(), data.IP_factor) * nu;
```
peterrum, line 1018: I must admit that I don't really like that `get_penalty_factor` gets a `MatrixFree` object but I don't have a better solution.

include/exadg/operators/interior_penalty_parameter.h
necioglu: Updated the PR description, or did you mean add a source to the code?

peterrum: Could you add the reference to the source code?

necioglu: done!

peterrum (approved):
Looks good from my side.

include/exadg/compressible_navier_stokes/spatial_discretization/kernels_and_operators.h
necioglu: I have also implemented the mentioned alternative solution. But I decided for this one, since the `calculate_penalty_parameter` function also gets a `MatrixFree` function: https://github.com/exadg/exadg/blob/7edf07c5a30582accdd72ce2af833caa51382107/include/exadg/operators/interior_penalty_parameter.h#L41-L44

But I am okay with both solutions.

bergbauer: I agree with Peter on this point. With the `calculate_penalty_parameter` function it is different because this function needs different kinds of information from the `MatrixFree` object. As long as the only information necessary for `get_penalty_factor` is if simplex or hypercube, we should determine this at the calling side and only pass this information.

necioglu: Okay, I will change it accordingly.

bergbauer (approved):

include/exadg/compressible_navier_stokes/spatial_discretization/kernels_and_operators.h
```diff
@@ -960,6 +960,15 @@ class ViscousOperator
...
   typedef dealii::Point<dim, dealii::VectorizedArray<Number>>     point;
 
+  bool
+  using_simplex(FaceIntegratorScalar & integrator) const
+  {
+    return integrator.get_matrix_free()
+      .get_dof_handler()
+      .get_triangulation()
+      .all_reference_cells_are_simplex();
+  }
```
nfehn, line 970: I see this code multiple times, so we should not realize this design.

include/exadg/compressible_navier_stokes/spatial_discretization/kernels_and_operators.h
```diff
@@ -1012,9 +1021,10 @@ class ViscousOperator
     scalar
     get_penalty_parameter(FaceIntegratorScalar & fe_eval_m, FaceIntegratorScalar & fe_eval_p) const
   {
-    scalar tau = std::max(fe_eval_m.read_cell_data(array_penalty_parameter),
-                          fe_eval_p.read_cell_data(array_penalty_parameter)) *
-                 IP::get_penalty_factor<Number>(degree, data.IP_factor) * nu;
+    scalar tau =
+      std::max(fe_eval_m.read_cell_data(array_penalty_parameter),
+               fe_eval_p.read_cell_data(array_penalty_parameter)) *
+      IP::get_penalty_factor<dim, Number>(degree, using_simplex(fe_eval_m), data.IP_factor) * nu;
```
nfehn, line 1027: I do not like the interface with `using_simplex`, since this does not follow the "open-closed" principle (open for extensions, closed for modifications). We have `ExaDG::ElementType`, which should be passed here.

include/exadg/operators/interior_penalty_parameter.h
```diff
@@ -89,16 +89,21 @@ calculate_penalty_parameter(
...
 Number
-get_penalty_factor(unsigned int const degree, Number const factor = 1.0)
+get_penalty_factor(unsigned int const degree, bool const use_simplex, Number const factor = 1.0)
 {
-  return factor * (degree + 1.0) * (degree + 1.0);
+  // use penalty factor for simplex elements according to Shahbazi (2005)
+  if(use_simplex)
+    return factor * (degree + 1.0) * (degree + dim) / dim;
+  else
+    return factor * (degree + 1.0) * (degree + 1.0);
```
nfehn, line 106: We then also have to introduce a reference here for quad/hex for completeness. You can find the original reference in my PhD thesis. Maybe it is good to also provide the Journal, Issue number, etc. as you did in the description of this PR.

include/exadg/compressible_navier_stokes/spatial_discretization/kernels_and_operators.h
necioglu: Yes, I also didn't like it while I was copy-pasting the function. But I allowed myself doing it, since the Kernels in operators `diffusive_operator`, `viscous_operator` etc. have multiple identical functions and as far as I see no base class. 

As a short term solution, I can suggest pushing the function to `interior_penalty_parameter.h` as a free standing function.

include/exadg/operators/interior_penalty_parameter.h
necioglu: addressed in afa84cdcbb31076a780301311fece2f03099716e.

include/exadg/incompressible_navier_stokes/spatial_discretization/operators/viscous_operator.h
```diff
@@ -151,16 +168,18 @@ class ViscousKernel
   void
   reinit_face(IntegratorFace & integrator_m, IntegratorFace & integrator_p) const
   {
-    tau = std::max(integrator_m.read_cell_data(array_penalty_parameter),
-                   integrator_p.read_cell_data(array_penalty_parameter)) *
-          IP::get_penalty_factor<Number>(degree, data.IP_factor);
+    tau =
+      std::max(integrator_m.read_cell_data(array_penalty_parameter),
+               integrator_p.read_cell_data(array_penalty_parameter)) *
+      IP::get_penalty_factor<dim, Number>(degree, get_element_type(integrator_m), data.IP_factor);
```
nfehn, line 174: looks good in this form!

include/exadg/incompressible_navier_stokes/spatial_discretization/operators/viscous_operator.h
```diff
@@ -54,6 +54,23 @@ class ViscousKernel
...
+         .all_reference_cells_are_simplex())
+      return ElementType::Simplex;
+    else if(integrator.get_matrix_free()
+              .get_dof_handler()
+              .get_triangulation()
+              .all_reference_cells_are_hyper_cube())
+      return ElementType::Hypercube;
+    else
+      AssertThrow(false, dealii::ExcMessage("Only hypercube or simplex elements are supported."));
+  }
```
nfehn, line 72: We now need to realize this in a way that code duplication is avoided. What is the reason why this function is duplicated? (As a minor comment, I would prefer bracket around the return-statements for better readability.)

include/exadg/operators/interior_penalty_parameter.h
```diff
@@ -89,16 +93,26 @@ calculate_penalty_parameter(
...
 Number
-get_penalty_factor(unsigned int const degree, Number const factor = 1.0)
+get_penalty_factor(unsigned int const degree, ElementType element_type, Number const factor = 1.0)
 {
-  return factor * (degree + 1.0) * (degree + 1.0);
+  if(element_type == ElementType::Simplex)
+    // use penalty factor for simplex elements according to K. Shahbazi, An explicit expression for
+    // the penalty parameter of the interior penalty method, Journal of Computational Physics 205,
+    // 401–407, 2005.
+    return factor * (degree + 1.0) * (degree + dim) / dim;
```
nfehn, line 110: Please introduce brackets here (and below)

include/exadg/operators/interior_penalty_parameter.h
```diff
@@ -89,16 +93,26 @@ calculate_penalty_parameter(
...
+  else
```
nfehn, line 111: should be `else if(element_type == Hypercube)`. An additional else branch with `AssertThrow(false,...)` is required below.

include/exadg/compressible_navier_stokes/spatial_discretization/kernels_and_operators.h
nfehn: Which functions do you mean exactly? If this is the case, we should also aim to improve this, but we should not introduce new duplicated code. A free function makes sense in my opinion.

necioglu: Most of the Kernel functions are duplicated in different Kernels. Here are some random examples from `DiffusiveKernel` and `ViscousKernel`: 

**I will put the function in `interior_penalty_parameter.h`, but the best would be to create a base class for the Kernels, which is I think outside of the purpose of this PR.**

https://github.com/exadg/exadg/blob/3ea21b101c8b3a421a56fc706a38705e484a2837/include/exadg/incompressible_navier_stokes/spatial_discretization/operators/viscous_operator.h#L83-L88

https://github.com/exadg/exadg/blob/3ea21b101c8b3a421a56fc706a38705e484a2837/include/exadg/convection_diffusion/spatial_discretization/operators/diffusive_operator.h#L80-L85

https://github.com/exadg/exadg/blob/3ea21b101c8b3a421a56fc706a38705e484a2837/include/exadg/incompressible_navier_stokes/spatial_discretization/operators/viscous_operator.h#L120-L132

https://github.com/exadg/exadg/blob/3ea21b101c8b3a421a56fc706a38705e484a2837/include/exadg/convection_diffusion/spatial_discretization/operators/diffusive_operator.h#L87-L99

nfehn: I agree that such a refactoring of Kernels is not related to the present PR and should, therefore, not be done here.

necioglu: Moved the function to `grid.h`, as discussed.

include/exadg/poisson/spatial_discretization/laplace_operator.h
```diff
@@ -60,14 +62,15 @@ class LaplaceKernel
   void
   reinit(dealii::MatrixFree<dim, Number> const & matrix_free,
          LaplaceKernelData const &               data_in,
-         unsigned int const                      dof_index)
+         unsigned int const                      dof_index_in)
   {
-    data = data_in;
+    data           = data_in;
+    data.dof_index = dof_index_in;
```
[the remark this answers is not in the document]
necioglu: https://github.com/exadg/exadg/blob/3ea21b101c8b3a421a56fc706a38705e484a2837/include/exadg/poisson/spatial_discretization/operator.cpp#L241

The alternative I had in mind was adding the following line bellow the above shown line:
`laplace_operator_data.kernel_data.dof_index=get_dof_index();`

But this raises the question why do we have the `dof_index` both in `laplace_operator_data` and in `laplace_operator_data.kernel_data`. I am not saying that the current solution in this PR solves this, but it hides it. 

I think I need a clear understanding of the difference between the Laplace Kernel and the Laplace Operator and which one should hold the dof_index. Maybe you think the proposed solution is sufficient enough? Let's talk about this today in our meeting!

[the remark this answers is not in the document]
necioglu: With 0766fec9c7e2572d0abcef0c031c67edae51f3ae we pass the `dof_index` to the Kernel functions, which need them. The `dof_index` is no more duplicated in Kerneldata. Please have a look @nfehn @bergbauer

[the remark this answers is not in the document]
nfehn: the design looks good now.

nfehn:
I wonder whether figuring out the element type in every `reinit_face()` function results in a visible performance penalty? @kronbichler @peterrum could you comment on this? Principally, this is an information that holds for all elements of a mesh throughout a whole simulation, so it could be determined once in a setup phase.

peterrum:
> I wonder whether figuring out the element type in every reinit_face() function results in a visible performance penalty? @kronbichler @peterrum could you comment on this? Principally, this is an information that holds for all elements of a mesh throughout a whole simulation, so it could be determined once in a setup phase.

@nfehn  I fully agree that calling that function is not the right way to go. In a follow-up PR, we should pre-determine the cell types. Once we go to mixed meshes, we will also need a logic to determine the type of cells on-the-fly. See also my earlier comment: https://github.com/exadg/exadg/pull/296#discussion_r1045890482

PR #298, opened 2022-12-13
Interface and Volume Coupling

jh66637:
@peterrum and I recently wrote a class that gets the data from (cell, quadrature) pairs as suggested by @kronbichler in https://github.com/exadg/exadg/pull/287.
This is a draft PR which aims to integrate this class into ExaDG. 

Currently, the class is only used for non-matching and overset discretizations in the Poisson case. I successfully tested it for volume coupling, though. Instead of removing the other overset implementation, (which uses Cached DBCs) I added another test case.

To check whether the approach works, I added a test case which does the same thing but uses the overset implementation once. Using GMRES in both cases, I got following Iteration counts.

Iterations GMRES (Matching vs ArtificialNonmatching)
None:  2522  vs 2521
PointJacobi: 1053 vs 1053
BlockJacobi: 688 vs 687
Multigrid: 11 vs 15

@peterrum @nfehn Any obligations before extending to incompressible Navier Stokes?

### DIFF
--- a/applications/poisson/overset_grids2/CMakeLists.txt
+++ b/applications/poisson/overset_grids2/CMakeLists.txt
@@ -1,0 +1,32 @@
+#########################################################################
+# 
+#                 #######               ######  #######
+#                 ##                    ##   ## ##
+#                 #####   ##  ## #####  ##   ## ## ####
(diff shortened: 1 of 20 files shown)


jh66637:
This is going to be a part of deal.II. So I will close present PR.

PR #301, opened 2022-12-23
Convergence problems Structure::DriverQuasiStatic

nfehn:
It helps to extrapolate the displacements in order to avoid invalid deformation states in the Newton solver, causing the solution to abort when initializing the Chebyshev smoother for the linearized problem.

The option `adjust_load_increment` does not seem to work in a robust manner. This topic should be addressed independently of this PR. The implementation has probably not sufficiently been tested, e.g. `++re_try_counter;` could not have any effect in the current implement when written in the try-block after the call to the solver.

It is important to check the validity of the deformation state (det_F > 0) for all geometric multigrid levels, and then only update the linearized operator and preconditioners/smoothers in case the deformation is valid.

The implementation of the Newton solver and linear solvers should be generalized such that one can independently call the update of the linearized operator/preconditioner and call the solver for the linearized problem. However, I would suggest to do this in a follow-up PR in order to concentrate on the `Structure` module in the present PR.

I also adjusted the application structure/bar. The Newton tolerances with 1.e-10 appear to be too tight for some problems, which is why I increased them to 1.e-8.

### DIFF
--- a/include/exadg/solvers_and_preconditioners/newton/newton_solver.h
+++ b/include/exadg/solvers_and_preconditioners/newton/newton_solver.h
@@ -80,12 +80,11 @@ public:
       linear_operator.set_solution_linearization(solution);
 
       // determine whether to update the operator/preconditioner of the linearized problem
-      bool const threshold_exceeded = (newton_iterations % update.threshold_newton_iter == 0) ||
-                                      (linear_iterations_last > update.threshold_linear_iter);
+      bool const update_now =
+        update.do_update and (newton_iterations % update.update_every_newton_iter == 0);
 
       // solve linear problem
-      linear_iterations_last =
-        linear_solver.solve(increment, residual, update.do_update && threshold_exceeded);
+      linear_iterations_last = linear_solver.solve(increment, residual, update_now);
 
       // damped Newton scheme
       double             omega         = 1.0; // damping factor (begin with 1)
@@ -129,6 +128,19 @@ public:
                 dealii::ExcMessage(
                   "Newton solver failed to solve nonlinear problem to given tolerance. "
                   "Maximum number of iterations exceeded!"));
+
+    // update linear operator and preconditioner once converged
+    // TODO: we currently re-solve the linear problem, which is actually un-necessary work
+    // given that we only want to update the operator/preconditioner
+    if(update.do_update and update.update_once_converged)
+    {
+      increment = 0.0;
+      nonlinear_operator.evaluate_residual(residual, solution);
+      residual *= -1.0;
+      // set linearization point
+      linear_operator.set_solution_linearization(solution);
+      linear_solver.solve(increment, residual, true /*update*/);
+    }
 
     return std::tuple<unsigned int, unsigned int>(newton_iterations, linear_iterations);
   }
--- a/include/exadg/solvers_and_preconditioners/newton/newton_solver.h
+++ b/include/exadg/solvers_and_preconditioners/newton/newton_solver.h
@@ -46,8 +46,7 @@ public:
     : solver_data(solver_data_in),
       nonlinear_operator(nonlinear_operator_in),
       linear_operator(linear_operator_in),
-      linear_solver(linear_solver_in),
-      linear_iterations_last(0)
+      linear_solver(linear_solver_in)
   {
   }
 
@@ -84,7 +83,7 @@ public:
         update.do_update and (newton_iterations % update.update_every_newton_iter == 0);
 
       // solve linear problem
-      linear_iterations_last = linear_solver.solve(increment, residual, update_now);
+      unsigned int const n_iter_linear = linear_solver.solve(increment, residual, update_now);
 
       // damped Newton scheme
       double             omega         = 1.0; // damping factor (begin with 1)
@@ -121,7 +120,7 @@ public:
 
       // increment iteration counter
       ++newton_iterations;
-      linear_iterations += linear_iterations_last;
+      linear_iterations += n_iter_linear;
     }
 
     AssertThrow(norm_r <= this->solver_data.abs_tol || norm_r / norm_r_0 <= solver_data.rel_tol,
@@ -150,8 +149,6 @@ private:
   NonlinearOperator & nonlinear_operator;
   LinearOperator &    linear_operator;
   LinearSolver &      linear_solver;
-
-  unsigned int linear_iterations_last;
 };
 
 } // namespace Newton
--- a/include/exadg/structure/spatial_discretization/operators/nonlinear_operator.cpp
+++ b/include/exadg/structure/spatial_discretization/operators/nonlinear_operator.cpp
@@ -58,7 +58,7 @@ template<int dim, typename Number>
 bool
 NonLinearOperator<dim, Number>::valid_deformation(VectorType const & displacement) const
 {
-  std::vector<Number> dst(1, 0.0);
+  Number dst = 0.0;
 
   // dst has to remain zero for a valid deformation state
   this->matrix_free->cell_loop(&This::cell_loop_valid_deformation,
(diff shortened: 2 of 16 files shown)


nfehn:
see also https://github.com/dealii/dealii/issues/14609

nfehn:
I decided to remove the option `adjust_load_increment` in its current implementation, since this seems to be untested / unvalidated code. So it might be better to start from something clean.

@kronbichler @peterrum I experienced that it helps a lot to simply not update the preconditioner (Chebyshev smoother). This is in line with the interpretation that invalid or close-to-invalid deformation states occur in some Newton iterations.

What about an update-heuristic that updates the preconditioner once the Newton solver converged (where we can be sure to have a valid deformation state)? Other than that, we would never update the preconditioner.

kronbichler (approved):
I think this makes sense. I wonder if we would ever run into a case where the initial state of a Newton iteration would be too far away from the converged state so that the number of iterations would increase. I could imagine that some metric that tracks the size of the diagonal entries might be able to determine these cases, i.e., we check whether the ratio between the largest entry on the diagonal and the smallest one becomes too big.

I made the experience in similar circumstances that one needs to adaptively select the strategy, sometimes do one and sometimes the other variant. Being able to catch bad cases and do something else (like reduce the load step) is definitely a step in the right direction.

nfehn:
> I wonder if we would ever run into a case where the initial state of a Newton iteration would be too far away from the converged state so that the number of iterations would increase.

I definitely see a large increase in iteration counts in case the preconditioner is not updated from one load step to the next or after a few load steps (this occurs, of course, only for large deformations). Lagging the preconditioner inside one Newton solve could be robust. I am not sure what you mean by initial state of a Newton iteration? The initial Newton iteration is problematic in my opinion because this state has already applied the new Dirichlet value with a larger load factor. So I currently believe one would really have to update the preconditioner once the Newton solver reported convergence. My hope would be that the preconditioner is then up-to-date enough so that the next load step can be solved in a robust manner. Unfortunately, we can not do the update in the next Newton solver call in the very beginning (as explained above), which would be easier to realize from an implementation point of view. So we would have to equip the Newton solver with a new update rule in order to realize the procedure described here.

I don't know how your idea with the diagonal entries would work or would be implemented. So if we want to follow this, I think it would be best to have a discussion in person.

nfehn:
I realized in https://github.com/exadg/exadg/pull/301/commits/8fc298916c5fe176bd30b53db3fe31906a554e06 the udpate rule described above (update linear operator/preconditioner once Newton reported convergence).

nfehn:
@kronbichler, @peterrum I achieve considerable improvements in solver robustness if I check the validity of the deformation state before updating the spatial operators, smoothers, etc. for the multigrid preconditioner inside the linearized solver of the Newton solver. It is important to do this check separately for each multigrid level and not just on the finest level (see text below)! See the newest commit https://github.com/exadg/exadg/pull/301/commits/4223d4aa277562b51ed8d9a2d64537bdd2c8f81e.

**Very interestingly:** Originally, I only checked the validity of the deformation state for the actual PDE operator (i.e. on the finest level). Then, I did not detect an invalid deformation state, but the solver did also not converge! An unexpected behavior. Afterwards, I checked the deformation state for every single multigrid level. Then, I identified invalid deformation states and the solver converged if I did not update the linearized operator for these multigrid levels in such a case (with invalid deformation). From this observation, I deduce that the transfer of the displacement vector (point of linearization of Newton solver) to coarser multigrid levels (p- or h-multigrid) leads to invalid deformation states. We should think about whether this is plausible, or whether the function doing the multigrid transfer for the displacement vector from fine to coarse levels does something wrong!? This is the function doing the transfer of the displacement vector to coarser levels (essentially involving `transfer->interpolate()`): https://github.com/exadg/exadg/blob/1351668d0146368bcd04596cb9d5b9ae566746a1/include/exadg/structure/preconditioners/multigrid_preconditioner.cpp#L155-L173

Do we need some "structure-preserving" multigrid transfer for nonlinear structural mechanics? Or is the problem unavoidable and related to finite element spaces/Lagrange polynomials?

nfehn:
I think this is ready for review now (@kronbichler @peterrum). The implementation of the Newton solver / linear solvers in ExaDG should be adapted as mentioned at the top. However, I would suggest this should be done in a follow-up PR.

peterrum:
> Do we need some "structure-preserving" multigrid transfer for nonlinear structural mechanics? Or is the problem unavoidable and related to finite element spaces/Lagrange polynomials?

The explanation sounds reasonable. There has been some structure-preserving interpolation schemes proposed by Satheesh et.al.:  https://arxiv.org/pdf/2211.16507.pdf. I read it last year; it looks interesting and one could probably do some simplifications for our case that we know that meshes are nested.

> I think this is ready for review now (@kronbichler @peterrum). The implementation of the Newton solver / linear solvers in ExaDG should be adapted as mentioned at the top. However, I would suggest this should be done in a follow-up PR.

I will take a look later in the evenging.

include/exadg/solvers_and_preconditioners/newton/newton_solver.h
```diff
@@ -80,12 +80,11 @@ class Solver
       linear_operator.set_solution_linearization(solution);
 
       // determine whether to update the operator/preconditioner of the linearized problem
-      bool const threshold_exceeded = (newton_iterations % update.threshold_newton_iter == 0) ||
-                                      (linear_iterations_last > update.threshold_linear_iter);
+      bool const update_now =
+        update.do_update and (newton_iterations % update.update_every_newton_iter == 0);
```
peterrum, line 84: Why did you change the logic here?

include/exadg/solvers_and_preconditioners/newton/newton_solver.h
```diff
@@ -80,12 +80,11 @@ class Solver
...
 
       // solve linear problem
-      linear_iterations_last =
-        linear_solver.solve(increment, residual, update.do_update && threshold_exceeded);
+      linear_iterations_last = linear_solver.solve(increment, residual, update_now);
```
peterrum, line 87: Now you don't need the temporal `linear_iterations_last` anymore.

include/exadg/structure/spatial_discretization/operators/nonlinear_operator.cpp
```diff
@@ -54,12 +54,39 @@ NonLinearOperator<dim, Number>::evaluate_nonlinear(VectorType & dst, VectorType
                           true);
 }
 
+template<int dim, typename Number>
+bool
+NonLinearOperator<dim, Number>::valid_deformation(VectorType const & displacement) const
+{
+  std::vector<Number> dst(1, 0.0);
```
peterrum, line 61:
```suggestion
  Number dst = 0.0;
```

include/exadg/structure/spatial_discretization/operators/nonlinear_operator.cpp
```diff
@@ -267,6 +294,45 @@ NonLinearOperator<dim, Number>::do_cell_integral(IntegratorCell & integrator) co
...
+      // material displacement gradient
+      tensor const Grad_d = integrator.get_gradient(q);
+
+      // material deformation gradient
+      tensor const F     = get_F<dim, Number>(Grad_d);
+      scalar const det_F = determinant(F);
+      for(unsigned int v = 0; v < det_F.size(); ++v)
+      {
+        // if deformation is invalid, add a positive value to dst
+        if(det_F[v] < 0.0)
```
peterrum, line 329:
```suggestion
        if(det_F[v] <= 0.0)
```

peterrum (approved):

include/exadg/solvers_and_preconditioners/newton/newton_solver.h
nfehn: I think this feature (`threshold_linear_iter`) was never used, so I removed it.

include/exadg/solvers_and_preconditioners/newton/newton_solver.h
nfehn: yes, thanks, at least not as a class variable.

include/exadg/structure/spatial_discretization/operators/nonlinear_operator.cpp
nfehn: agree

include/exadg/structure/spatial_discretization/operators/nonlinear_operator.cpp
nfehn: agree

PR #302, opened 2023-01-04
Fix checkout version

bergbauer:


peterrum (approved):

peterrum:
Let me merge. The change is correct; we are on this in deal.II (see: https://github.com/dealii/dealii/issues/14637).

PR #309, opened 2023-01-23
adapt config_trilinos.sh since we need trilinos 12.14 or newer

nfehn:
I already adjusted the wiki page on [Installation](https://github.com/exadg/exadg/wiki/Installation) (using Trilinos 13.0.0 since dealii forces us to use 12.14 or newer). 

@bergbauer provided new settings for the configure script for Trilinos, forming the present PR.


peterrum (approved):

bergbauer (approved):
To write it down somewhere. The `Trilinos_ENABLE_SEACASExodus:BOOL=ON` option allows us to read Exodus meshes.

nfehn:
@bergbauer I will add this information to the wiki.

PR #311, opened 2023-01-25
flow past sphere: add missing parameter

### LINKED ISSUE #303: Multigrid Preconditioner with global coarsening and h-MG fails
As an example, see the failing test case of the flow past a sphere.

As far as I can relate, this is because the ```Grid``` class fails to build the coarse_triangulations (in function ```create_triangulation()```) . Since the creation is just a function call to a deal.II function, I suppose the function is either called with wrong arguments, a different function should be called, or something is broken inside deal.II. I am new to the whole topic of using Multigrid in deal.II, so maybe someone who is more experienced can have a look. 

The current work-around for me is to set ```MultigridType::cpMG``` instead of ```MultigridType::cphMG```.

jh66637:
Test fails if run out of the box, due to ```*hMG``` without ```param.grid.create_coarse_triangulations```

Resolves #303 

Thanks to @necioglu @nfehn @peterrum

### DIFF
--- a/applications/incompressible_navier_stokes/flow_past_sphere/application.h
+++ b/applications/incompressible_navier_stokes/flow_past_sphere/application.h
@@ -138,6 +138,8 @@ public:
     this->param.grid.triangulation_type = TriangulationType::Distributed;
     this->param.grid.mapping_degree     = this->param.degree_u;
     this->param.degree_p                = DegreePressure::MixedOrder;
+    // we need to create coarse triangulations since we use global coarsening multigrid
+    this->param.grid.create_coarse_triangulations = true;
 
     // convective term
     if(this->param.formulation_convective_term == FormulationConvectiveTerm::DivergenceFormulation)


applications/incompressible_navier_stokes/flow_past_sphere/application.h
```diff
@@ -135,9 +135,10 @@ class Application : public ApplicationBase<dim, Number>
     this->param.rel_tol_steady = 1.e-8;
 
     // SPATIAL DISCRETIZATION
-    this->param.grid.triangulation_type = TriangulationType::Distributed;
-    this->param.grid.mapping_degree     = this->param.degree_u;
-    this->param.degree_p                = DegreePressure::MixedOrder;
+    this->param.grid.triangulation_type           = TriangulationType::Distributed;
+    this->param.grid.mapping_degree               = this->param.degree_u;
+    this->param.degree_p                          = DegreePressure::MixedOrder;
+    this->param.grid.create_coarse_triangulations = true;
```
nfehn, line 141: Please add a comment like "we need to create coarse triangulations since we use global coarsening multigrid" in order to help other users in understanding the reason behind this setting

PR #312, opened 2023-01-25
Generalize pointwise output generator for any VectorType and update to deal.II

jh66637:
I extended the class to be able to use it with ```BlockVectors```. For scalar blocks, the output data type changes in ```deal.II```, so I added a function overload for ```write_quantity()```


nfehn:
If `VectorType` is introduced as a template, I think the template `Number` should be removed.

nfehn (approved):

PR #315, opened 2023-01-25
Make FDA benchmark work out of the box

jh66637:
I currently investigated the precursor setup in ExaDG. For this I loked at the FDA benchmark and the backward facing step. Both applications did not work out of the box. 

* For the FDA benchmark, the wrong triangulation is used to set periodicity.
* For both benchmarks, evaluating e.g. the line plot data fails since the shared pointer for the calculator is only created if it is needed but is accessed in any case.

### DIFF
--- a/applications/incompressible_navier_stokes/fda_benchmark/application.h
+++ b/applications/incompressible_navier_stokes/fda_benchmark/application.h
@@ -461,7 +461,7 @@ private:
     }
 
     dealii::GridTools::collect_periodic_faces(
-      *this->grid->triangulation, 0 + 10, 1 + 10, 2, this->grid_pre->periodic_faces);
+      *this->grid_pre->triangulation, 0 + 10, 1 + 10, 2, this->grid_pre->periodic_faces);
     this->grid_pre->triangulation->add_periodicity(this->grid_pre->periodic_faces);
 
     // perform global refinements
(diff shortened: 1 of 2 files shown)


applications/incompressible_navier_stokes/fda_benchmark/application.h
```diff
@@ -461,7 +461,7 @@ class Application : public ApplicationBasePrecursor<dim, Number>
     }
 
     dealii::GridTools::collect_periodic_faces(
-      *this->grid->triangulation, 0 + 10, 1 + 10, 2, this->grid_pre->periodic_faces);
+      *this->grid_pre->triangulation, 0 + 10, 1 + 10, 2, this->grid_pre->periodic_faces);
```
nfehn, line 464: very good!

applications/incompressible_navier_stokes/fda_benchmark/include/postprocessor.h
```diff
@@ -141,16 +141,19 @@ class PostProcessorFDA : public PostProcessor<dim, Number>
     }
 
     // evaluation of results along lines
-    if(line_plot_calculator_statistics->time_control_statistics.time_control.needs_evaluation(
-         time, time_step_number))
+    if(line_plot_calculator_statistics)
```
nfehn, line 144: what does this if-statement mean? `line_plot_calculator_statistics` is not a boolean variable, right?

nfehn (approved):

applications/incompressible_navier_stokes/fda_benchmark/include/postprocessor.h
jh66637: Yes, its the same as:
```
if(line_plot_calculator_statistics!=nullptr)
```

Issue #317, opened 2023-01-26
Global coarsening multigrid doesn't work for IncNS/Vortex


necioglu:
I tried global coarsening multigrid with the vortex application and ran into an error during setup. I have added the following to the application in master:
```
this->param.grid.create_coarse_triangulations=true;
this->param.multigrid_data_viscous.use_global_coarsening= true;
this->param.multigrid_data_pressure_poisson.use_global_coarsening=true;
```
I have also changed the `create_grid` function to incorporate the grid class functions, so that the `coarse_triangulations` can be created. I ran into an Assert during `initialize_chebyshev_smoother_coarse_grid`. See Stacktrace bellow:
```
--------------------------------------------------------
An error occurred in line <637> of file </scratch/necioglu/workspace/dealii/include/deal.II/lac/solver_cg.h> in function
    void dealii::internal::SolverCG::IterationWorker<VectorType, MatrixType, PreconditionerType, <template-parameter-1-4> >::do_iteration(unsigned int) [with VectorType = dealii::LinearAlgebra::distributed::Vector<float>; MatrixType = ExaDG::MultigridOperatorBase<2, float>; PreconditionerType = ExaDG::JacobiPreconditioner<ExaDG::MultigridOperatorBase<2, float> >; <template-parameter-1-4> = int]
The violated condition was: 
    std::abs(p_dot_A_dot_p) != 0.
Additional information: 
    A piece of code is attempting a division by zero. This is likely going
    to lead to results that make no sense.

Stacktrace:
-----------
#0  /scratch/necioglu/workspace/exadg/build-debug/libexadg.so: void dealii::SolverCG<dealii::LinearAlgebra::distributed::Vector<float, dealii::MemorySpace::Host> >::solve<ExaDG::MultigridOperatorBase<2, float>, ExaDG::JacobiPreconditioner<ExaDG::MultigridOperatorBase<2, float> > >(ExaDG::MultigridOperatorBase<2, float> const&, dealii::LinearAlgebra::distributed::Vector<float, dealii::MemorySpace::Host>&, dealii::LinearAlgebra::distributed::Vector<float, dealii::MemorySpace::Host> const&, ExaDG::JacobiPreconditioner<ExaDG::MultigridOperatorBase<2, float> > const&)
[...]

PR #320, opened 2023-01-27
Redesign Grid class by using free functions

### LINKED ISSUE #313: Global coarsening multigrid does not give grid-independent iteration counts
Iteration counts strongly increase for increasingly fine meshes.

The problem might be related to the multigrid transfer (in particular h-transfer) or how we construct the coarse triangulations for global coarsening MG.

### LINKED ISSUE #319: Enforce all applications to generate the grid via Grid::create_triangulation()?
It is currently possible to generate the grid in an application.h file without using `Grid::create_triangulation()`, since `Grid::triangulation` is a public member. This may cause troubles such as global coarsening multigrid not working because the coarse grid triangulations are not created (correctly).

The solution is to make the member `triangulation` in `Grid` private and thereby force every application to use `Grid::create_triangulation()`. However, one restriction of this design is that `Application::create_grid()` must end with the call  `Grid::create_triangulation()`. Currently, we have applications for which this design is hard to realize, because they apply a mesh deformation based on the fine mesh. However, the lambda passed to `Grid::create_triangulation()` only generates the coarse mesh and refinement is done inside `Grid::create_triangulation()`.

The most simple solution to avoid all these problems would be that we do not have to create the coarse grid triangulations manually in `Grid`, but somewhat generically.

nfehn:
... and letting the `lambda_create_triangulation()` function do the refinements inside the `Application` class.

A series of PRs and issues (#313, #317, #319) has revealed that the `Grid` class is currently not well-designed.

For flexibility and generality, the application class has to take care of the global/local refinements (also because our variant with a vector of local refinements for different material IDs is just one particular way of doing local refinements). The solution is to define free utility functions (in order to avoid code duplications) that can be called by applications.

We should check and make sure that the error message if using global-coarsening multigrid but not having initialized the vector `coarse_triangulations` in  the application, clearly describes to a user the source of error and what needs to be done to avoid this.

In the future, `GridData` needs to be refactored as well. For example, the newly introduced function `GridUtilities::create_triangulation()` "by-passes" the parameter / member variable `grid_data.n_refine_global` (describing the finest mesh) and instead uses the parameter `global_refinements` (describing the refinement of any intermediate mesh level for global coarsening multigrid). This is some form of non-uniqueness of parameters in the code that is hard to understand and should better be avoided. Also, the parameter `bool GridData::create_coarse_triangulations` needs to be removed, since it is a dependent parameter that can be deduced from other parameters such as `MultigridData` (see #303).

@necioglu @richardschu FYI

closes #313 
closes #319


richardschu:
Hi Niklas,
checked out your branch today, tested it with my iliac bifurcation test-application and exadg-applications/applications/lung. The examples you adapted in this PR look fine, where the Poisson/sine just maps vertices? For the lung, the Grid class stores the mapping, such that I could not use global coarsening multigrid. What is planned in this regard?
Best, R

nfehn:
I have not thought about this in detail so far. Prinicipally, my hope is that the current design is flexible enough. Whether the functionality of the existing `ExaDG::GridUtility` functions should be extended, or whether to write separate free functions for other applications (which you mention above) is unclear to me. Where exactly do you think is the critical aspect in getting the lung problem running? Is the `Grid` class not flexible enough? In any case, I think we should get such an application running in a first step, and then discuss the design.

richardschu:
As far as I can tell one might want to initialize the mapping with a displacement vector via
MappingDoFVector<dim, Number>::initialize_mapping_q_cache(undeformed_mapping,
                                                              displacement_vector,
                                                              dof_handler);
per level. So the Grid class might need a std::vector<std::shared_ptr<dealii::Mapping<dim>>> mappings; ?
Only reason I am thinking about this is the (somewhat unrelated?) error I get otherwise.

nfehn:
Would make sense in my opinion and be consistent with the design of a vector of coarse triangulations in the Grid class

nfehn:
but don‘t we have the mappings for coarse triangulations inside the multigrid preconditioner?

include/exadg/grid/grid_utilities.h
```diff
@@ -0,0 +1,252 @@
+/*
+ * grid_utilities.h
+ *
+ *  Created on: Jan 26, 2023
+ *      Author: fehn
+ */
```
peterrum, line 6:
```suggestion
/*  ______________________________________________________________________
 *
 *  ExaDG - High-Order Discontinuous Galerkin for the Exa-Scale
 *
 *  Copyright (C) 2023 by the ExaDG authors
 *
 *  This program is free software: you can redistribute it and/or modify
 *  it under the terms of the GNU General Public License as published by
 *  the Free Software Foundation, either version 3 of the License, or
 *  (at your option) any later version.
 *
 *  This program is distributed in the hope that it will be useful,
 *  but WITHOUT ANY WARRANTY; without even the implied warranty of
 *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 *  GNU General Public License for more details.
 *
 *  You should have received a copy of the GNU General Public License
 *  along with this program.  If not, see <https://www.gnu.org/licenses/>.
 *  ______________________________________________________________________
 */
```

include/exadg/grid/balanced_granularity_partition_policy.h
```diff
@@ -0,0 +1,61 @@
+/*
+ * balanced_granularity_partition_policy.h
+ *
+ *  Created on: Jan 26, 2023
+ *      Author: fehn
+ */
```
peterrum, line 6:
```suggestion
/*  ______________________________________________________________________
 *
 *  ExaDG - High-Order Discontinuous Galerkin for the Exa-Scale
 *
 *  Copyright (C) 2023 by the ExaDG authors
 *
 *  This program is free software: you can redistribute it and/or modify
 *  it under the terms of the GNU General Public License as published by
 *  the Free Software Foundation, either version 3 of the License, or
 *  (at your option) any later version.
 *
 *  This program is distributed in the hope that it will be useful,
 *  but WITHOUT ANY WARRANTY; without even the implied warranty of
 *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 *  GNU General Public License for more details.
 *
 *  You should have received a copy of the GNU General Public License
 *  along with this program.  If not, see <https://www.gnu.org/licenses/>.
 *  ______________________________________________________________________
 */
```

peterrum (approved):

include/exadg/grid/grid_utilities.h
peterrum: It seems that many of the files still have the old headers. Let's keep it like this and merge so that we can proceed with https://github.com/exadg/exadg/pull/323.

PR #323, opened 2023-01-28
Read external grid with an application to Possion/sine 

necioglu:
This PR introduces external mesh reading. Two example grids are provided for the Poisson/sine application. The following is done:

1. `Input_parameters.h` is added including the parameters required for reading an external grid. The function for reading the parameters from the input file is called in `application_base.h`.  I am not super happy with the naming of the variables, for example the name of the subsection being `Input`. I am open to suggestions. 
2. Two example grids are provided, one of them consist of simplex elements and the other one is a hypercube mesh. The hypercube mesh is the same grid as the application creates. One could run the tests with it and tests would pass. The simplex grid is an unstructured coarse grid, which has 70 elements. Both grids are created with Cubit and are in exodus file format. I can add the cubit journal files,used for creating the grids with a script, if necessary.
3. Let's merge this after #320, so that I can adapt the design accordingly. That is why I will mark it as draft for now.


richardschu:
Hi, I am currently working on something similar: reading in a coarse grid (tetrahedra or hexahedra, .msh file format) and then reconstruct a higher-order mapping via a deformation field given in an external file (per grid level? let's see...). Will fit those routines to your structures then -- I will let you know once I succeed matching the support points and actually mapping at least the finest grid! Do you plan something similar?

applications/poisson/sine/application.h
```diff
@@ -216,9 +217,26 @@ class Application : public ApplicationBase<dim, n_components, Number>
...
+      dealii::GridIn<dim> grid_in;
+
+      grid_in.attach_triangulation(tria);
+
+      grid_in.read_exodusii(this->input_parameters.input_grid_file);
+    };
+
+    if(this->input_parameters.read_external_grid)
+    {
+      this->grid->create_triangulation(this->param.grid,
```
peterrum, line 230: Wouldn't it make sense to introduce a version of `create_triangulation()` that accepts a file name?

peterrum:
> Hi, I am currently working on something similar: reading in a coarse grid (tetrahedra or hexahedra, .msh file format) and then reconstruct a higher-order mapping via a deformation field given in an external file (per grid level? let's see...). Will fit those routines to your structures then -- I will let you know once I succeed matching the support points and actually mapping at least the finest grid! Do you plan something similar?

I think reading the mesh and setting up the mapping

https://github.com/exadg/exadg/blob/55c597ff31bcd57ed0fe2dc2224815fad6d8cdc8/include/exadg/grid/grid.h#L119

are independent. Latter can be done in a second step once the grid is read.

richardschu:
> > Hi, I am currently working on something similar: reading in a coarse grid (tetrahedra or hexahedra, .msh file format) and then reconstruct a higher-order mapping via a deformation field given in an external file (per grid level? let's see...). Will fit those routines to your structures then -- I will let you know once I succeed matching the support points and actually mapping at least the finest grid! Do you plan something similar?
> 
> I think reading the mesh and setting up the mapping
> 
> https://github.com/exadg/exadg/blob/55c597ff31bcd57ed0fe2dc2224815fad6d8cdc8/include/exadg/grid/grid.h#L119
> 
> are independent. Latter can be done in a second step once the grid is read.

Sure, but would make sense not to have two implementations of the same thing, if we are both working on it ; )

necioglu:
> Hi, I am currently working on something similar: reading in a coarse grid (tetrahedra or hexahedra, .msh file format) and then reconstruct a higher-order mapping via a deformation field given in an external file (per grid level? let's see...). Will fit those routines to your structures then -- I will let you know once I succeed matching the support points and actually mapping at least the finest grid! Do you plan something similar?

I am trying to use simplex elements and with complicated grids, it makes sense to read an external grid file, rather than creating one using the `GridGenerator` functions provided by dealii. I guess you also have a similar reasoning for reading an external grid file.

For my use case, the mapping stays untouched. That is also why I agree wit the point made by @peterrum in https://github.com/exadg/exadg/pull/323#issuecomment-1407473802. If you think that your future PR will cover the goals of this PR in a similar way as well, I think it makes sense to merge this first and then merge a separate PR with the changes to the mapping. However if you are reading an external grid in a completely different way, we should find the better alternative between the solutions.

richardschu:
Exactly - I am reading hexa meshes from .msh and then I will add a displacement to them.
I thought you would also tie a higher-order mapping providing a displacement vector.
The grid read is almost the same: grid_in.read_msh(...), so, depends on file type and name only.
Then I will wate for your PR to be merged and adapt my stuff afterwards ; )

applications/poisson/sine/application.h
necioglu: Yes, I could design an overloaded `create_triangulation()` function for reading an external grid, and get rid of the `lambda_read_external_coarse_grid` function in the application. I think it would be a better design. I will do this after the changes to the grid class are made.

necioglu:
> I thought you would also tie a higher-order mapping providing a displacement vector. 

No, I am nıt changig anything on the mapping.

> The grid read is almost the same: grid_in.read_msh(...), so, depends on file type and name only. Then I will wate for your PR to be merged and adapt my stuff afterwards ; )

Let me know, if you have any suggestions to current design, so that your mapping changes can be easily add on top :)

include/exadg/utilities/input_parameters.h
```diff
@@ -0,0 +1,53 @@
...
+
+#ifndef INCLUDE_EXADG_UTILITIES_INPUT_PARAMETERS_H_
+#define INCLUDE_EXADG_UTILITIES_INPUT_PARAMETERS_H_
+
+// deal.II
+#include <deal.II/base/parameter_handler.h>
+
+namespace ExaDG
+{
+struct InputParameters
```
nfehn, line 30: The struct ``GridData`` has already been prepared for reading grid files with a string for the filename. So I think there is no need to intoduce another struct right now.

include/exadg/utilities/input_parameters.h
```diff
@@ -0,0 +1,53 @@
...
+    prm.enter_subsection(subsection_name);
+      prm.add_parameter("GridFile", input_grid_file,    "External input grid file.");
+      prm.add_parameter("ReadGrid", read_external_grid, "Decides whether a grid file is read.");
+    prm.leave_subsection();
+    // clang-format on
+  };
+
+  std::string input_grid_file;
+
+  bool read_external_grid;
```
nfehn, line 49: I am not sure if this boolean is really needed. The application, more precisely the function `create_grid()`, will call a certain function that reads a grid from a file, so the parameter `read_external_grid` is somewhat redundant. I consider it a special case that an application can both read in a file or construct the grid internally. In such a special case, the boolean parameter `read_external_grid` should only live inside this particular `Application`, but not be a general parameter in ExaDG in my opinion.

applications/poisson/sine/application.h
peterrum: That would be great. Reading an external mesh is such a common operation! The lambda created here would just be created in the overloaded function!

nfehn: I agree. We do not want to rewrite these lines of code for every application reading a grid from a file. This functionality should instead become some generic functionality of ExaDG. In accordance with #320, we would realize this as a free function in `GridUtilities`.

What we have to think about in this context is how to deal with different file formats. We should avoid introducing again and again new functions with new names for different file endings. Instead, the function taking a filename as argument should internally deduce the type of grid, with an assert if the file-type is not implemented or incorrect.

applications/poisson/sine/input.json
```diff
@@ -20,5 +20,9 @@
         "OutputDirectory": "output/poisson/sine/",
         "OutputName": "test",
         "WriteOutput": "false"
+    },
+    "Input": {
```
nfehn, line 24: Input as a name is unspecific. The filename is input.json, so everything in this file is "input" from a user persective. Why not using "Grid" in accordance with the class `ExaDG::Grid` and the parameters in `GridData`?

nfehn (changes requested):

applications/poisson/sine/application.h
nfehn: Why is there a need for a lambda in an overloaded function?

include/exadg/utilities/input_parameters.h
necioglu: I forgot about the the struct in `GridData`.  I will change the design accordingly.

include/exadg/utilities/input_parameters.h
necioglu: Okay, I see your point. During development, having the variable in the input file made sense for me, because I was switching between grids to check them. But yes, I agree it is special case!

applications/poisson/sine/input.json
necioglu: Yes, I also didn't like the naming. The name Grid sounds better.

applications/poisson/sine/application.h
necioglu: I also didn't understand why a lambda would be needed.

include/exadg/utilities/input_parameters.h
necioglu: addressed in ce5731610269cac685066d2a69b823c059586cd3.

include/exadg/utilities/input_parameters.h
necioglu: addressed in ce5731610269cac685066d2a69b823c059586cd3.

applications/poisson/sine/input.json
necioglu: addressed in ce5731610269cac685066d2a69b823c059586cd3.

applications/poisson/sine/application.h
necioglu: > What we have to think about in this context is how to deal with different file formats. We should avoid introducing again and again new functions with new names for different file endings. Instead, the function taking a filename as argument should internally deduce the type of grid, with an assert if the file-type is not implemented or incorrect.

To address this, we should use the [`grid_in.read(string,format)`](https://www.dealii.org/current/doxygen/deal.II/classGridIn.html#ad30d9d6f2ff9920c4285e795c4495826) function. But still, this function requires the file format as input. To solve this to options come two my mind: 

1. Adding a parameter to `input.json` and to `gridData` specifying the file format. 
2. Searching in the string for the file name extension.

nfehn: This reminds me of the different parameter file formats (json, prm, etc.). Some time ago, I introduced a function in deal.II which works for all the different input file formats, see https://github.com/dealii/dealii/pull/9741. I think we should strive for a similar solution in terms of reading different grid files (with different endings). This line shows how to extract the filename ending

https://github.com/dealii/dealii/blob/d77e5ebb0af4487fa3c503f97fd9d126f2eac18f/source/base/parameter_handler.cc#L573

So I think we should follow variant 2 and avoid an additional input parameter (variant 1), because this would again be redundant information.

necioglu: > What we have to think about in this context is how to deal with different file formats. We should avoid introducing again and again new functions with new names for different file endings. Instead, the function taking a filename as argument should internally deduce the type of grid, with an assert if the file-type is not implemented or incorrect.

This is done in 5b4d30d37476e26dfeab45c15911e31f36280fb3. Waiting for #320 to move the function to `GridUtilities`.

peterrum:
@necioglu Could you rebase?

applications/poisson/sine/application.h
```diff
@@ -216,9 +217,44 @@ class Application : public ApplicationBase<dim, n_components, Number>
...
+                    "You are trying to read a grid file, but the the string is empty."));
+
+      dealii::GridIn<dim> grid_in;
+
+      grid_in.attach_triangulation(tria);
+
+      std::string extension =
+        this->param.grid.grid_file.substr(this->param.grid.grid_file.find_last_of('.') + 1);
+
+      AssertThrow(!extension.empty(), dealii::ExcMessage("File extension was empty!"));
```
nfehn, line 232: "was" -> "is" would be better in my opinion. Again, the message is a bit unspecific. Since this functionality will become some utility function in ExaDG, the description should be more detailed.

include/exadg/grid/grid_data.h
```diff
@@ -67,6 +71,16 @@ struct GridData
...
 
+  void
+  add_parameters(dealii::ParameterHandler & prm, std::string const & subsection_name = "Grid")
+  {
+    // clang-format off
+    prm.enter_subsection(subsection_name);
+      prm.add_parameter("GridFile", grid_file, "External input grid file.");
+    prm.leave_subsection();
+    // clang-format on
+  }
```
nfehn, line 82: This design is not in line with what we currently do in ExaDG. Note that `OutputParameters` (depending on `dealii::ParameterHandler`) and `OutputDataBase` are two different classes. So the philosophy behind this design is that `dealii::ParameterHandler` should not be visible in "inner" data structures, e.g. `GridData`. When following the choice in this PR, we give up the flexbility to use something different than `dealii::ParameterHandler` (there are already other variants in deal.II itself). I currently see it this way: `Application` depends on `ParameterHandler` and `Grid`/`GridData`, but  `Grid`/`GridData` should not depend on `Application` and `ParameterHandler`.

I am not sure which design is the best one, but we definitely need to discuss this.

applications/poisson/sine/application.h
```diff
@@ -216,9 +217,44 @@ class Application : public ApplicationBase<dim, n_components, Number>
...
+      AssertThrow(!this->param.grid.grid_file.empty(),
+                  dealii::ExcMessage(
+                    "You are trying to read a grid file, but the the string is empty."));
+
+      dealii::GridIn<dim> grid_in;
+
+      grid_in.attach_triangulation(tria);
+
+      std::string extension =
+        this->param.grid.grid_file.substr(this->param.grid.grid_file.find_last_of('.') + 1);
```
nfehn, line 230: I think we should think about the name `grid_file`. The above line shows us that we get expressions like `grid.grid...`. So I suggest either `filename` or `file` to avoid this.

include/exadg/grid/grid_data.h
```diff
@@ -84,8 +98,8 @@ struct GridData
 
   unsigned int mapping_degree;
 
-  // TODO: path to a grid file
-  // std::string grid_file;
+  // path to a grid file
```
nfehn, line 101: This comment needs to be extended. E.g. 

"the filename needs to include a proper filename ending/extension so that we can internally deduce the correct type of file format."

applications/poisson/sine/application.h
necioglu: Yes, but `grid` could potentially have more external files in the future, for example an external mapping, see: https://github.com/exadg/exadg/pull/323#issuecomment-1407440353. That is why I decided for `grid_file` to distinguish this from a future `mapping_file`. Both `file` and `filename` were to unspecific for me.

What do you think about `triangulation_file`?

nfehn: According to my experience in software development, I would introduce a more complicated name only when needed. E.g. currently we do not know whether these mapping files become general ExaDG functionality, or are application specific, ... .

include/exadg/grid/grid_data.h
necioglu: Yes, this was the initial reason why I added `InputParameters`. Because it felt natural after seeing `OutputParameters` and I forgot the existence of `gridfile` in `gridData` :) 

I don't see it as a complete dependency, i.e., one could still use `gridData` without using the `add_parameters` function. If one wants to read an external grid, yes we are dependent, which makes sense.

What I could do to solve this is introducing `GridParameters` with a free function `add_parameters` with a write access to `GridData`? Or a free function in `GridUtilities`?

nfehn: In any case, I think one has the dependency software-wise.

I think I am against a write access to `GridData`. Maybe we wait some time with this aspect for other opinions before doing too many changes. I do not say that the current implementation has to be changed, but want to point to weaknesses from a design perspective.

applications/poisson/sine/application.h
necioglu: Moved the function to `GridUtilities` in d932484523cc4b0a6ccbe41963e79a51e5d7419.

applications/poisson/sine/application.h
```diff
@@ -216,9 +217,44 @@ class Application : public ApplicationBase<dim, n_components, Number>
       }
     };
 
-    this->grid->create_triangulation(this->param.grid,
-                                     lambda_create_coarse_triangulation,
-                                     this->param.grid.n_refine_global);
+    auto const lambda_read_external_coarse_grid = [&](dealii::Triangulation<dim, dim> & tria) {
+      AssertThrow(!this->param.grid.grid_file.empty(),
+                  dealii::ExcMessage(
+                    "You are trying to read a grid file, but the the string is empty."));
```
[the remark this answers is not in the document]
necioglu: changed in d932484523cc4b0a6ccbe41963e79a51e5d74191.

applications/poisson/sine/application.h
necioglu: changed in d932484523cc4b0a6ccbe41963e79a51e5d74191.

applications/poisson/sine/application.h
necioglu: changed in d932484523cc4b0a6ccbe41963e79a51e5d74191.

include/exadg/grid/grid_data.h
necioglu: changed in d932484523cc4b0a6ccbe41963e79a51e5d74191.

applications/poisson/sine/application.h
```diff
@@ -216,9 +217,44 @@ class Application : public ApplicationBase<dim, n_components, Number>
       }
     };
 
-    this->grid->create_triangulation(this->param.grid,
-                                     lambda_create_coarse_triangulation,
-                                     this->param.grid.n_refine_global);
+    auto const lambda_read_external_coarse_grid = [&](dealii::Triangulation<dim, dim> & tria) {
```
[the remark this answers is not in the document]
necioglu: changed in d932484523cc4b0a6ccbe41963e79a51e5d74191.

necioglu:
With eb1b600e7766219deb3b631a3d6fc1361ad5d65b, the `read_external_triangulation` has a `GridData` argument instead of a string argument. This forces the user to use the `file_name` in `GridData`, i.e., doesn't allow to use a simple string possibly defined in the application. 

What do you think about this? @nfehn @peterrum

include/exadg/grid/grid_data.h
necioglu: Only this is not addressed in the latest commits. I couldn't find a solution, where we:

-  don't allow write access to `GridData`
-  don't introduce new dependencies  
-  and keep `grid_file` string in `GridData`.

peterrum:
> With [eb1b600](https://github.com/exadg/exadg/commit/eb1b600e7766219deb3b631a3d6fc1361ad5d65b), the `read_external_triangulation` has a `GridData` argument instead of a string argument. This forces the user to use the `file_name` in `GridData`, i.e., doesn't allow to use a simple string possibly defined in the application.
> 
> What do you think about this? @nfehn @peterrum

I am fine with that.

include/exadg/grid/grid_data.h
nfehn: `OutputParameters` and `OutputDataBase` is one such realization, as mentioned above.

I am not sure what you mean by write access exactly. All these parameter structs have public members, implying write access. I am against something like `grid_data.set_filename(grid_parameters)`, since this introduces not only additional or unnecessary member functions, but also additional dependencies. However, an assignment `grid_data.filename = name` (just as we set all the parameters in application.h) is fine, isn't it? This has to happen in `Application::set_parameters()`, because from then on, `grid_data` needs to be considered `const`.

necioglu: > However, an assignment `grid_data.filename = name` (just as we set all the parameters in application.h) is fine, isn't it?

This I would have described as "write access", since we are writing on a parameter, and I thought you were against it. I will change the design to the suggested version!

nfehn: Wait a second. You do not have to change it to that version. I mainly want to have a discussion about this. I understand that you want to finish the present PR soon, so we can also take the current version and change it in the future.

nfehn: Regarding "write access": we make the member variables of `Parameters` public so that we do not always have to call setter/getter functions e.g. in `Application::set_parameters()`. At the same time, we do not want to misuse this "freedom" in the sense that we pass the structs to functions that can manipulate the parameters. Imagine I pass `gird_data` to two different functions during the setup phase, both having write access to `grid_data`. How do you know that the data in `grid_data` is reasonable after setup?

necioglu: > Wait a second. You do not have to change it to that version. I mainly want to have a discussion about this. I understand that you want to finish the present PR soon, so we can also take the current version and change it in the future.

Maybe it seemed like I wanted to finish this PR soon and that is why I accept your suggestion, but I am trying to contribute a good PR, not something to be changed in the future. That is why I am listening to suggestions and making changes accordingly.  

I liked the idea of having a separate file `grid_parameters.h` with `add_parameters` in it. I know I don't have to change it to that version.

peterrum: I am not completely whether I understood the discussion here. The discussion is whether to have a method called `add_parameters()` or to have a free function for each parameter struct similar to what we have for the enums. Personally, I normally use the first approach in my projects but the free-function approach is also fine if that is more inline with rest of ExaDG :+1:

peterrum (commented):
The PR looks overall good. Once the discussion regarding parameter is settled, let's merge. Out of curiosity, @necioglu is your goal to work with exodus files or why did you add the incomplete specialization?

include/exadg/grid/grid_data.h
necioglu: I guess we had a misunderstanding with @nfehn, and a short meeting solved it :+1:  We thought adding `grid_parameters.h` with `add_parameters()` in it would be a good approach here, also considering your comment @peterrum.

I will implement this solution!

necioglu: I have made the changes in 3f0998180eafacf5a49545f48f19b5f3e82f6f04.

necioglu:
> The PR looks overall good. Once the discussion regarding parameter is settled, let's merge. Out of curiosity, @necioglu is your goal to work with exodus files or why did you add the incomplete specialization?

@peterrum Thanks! Yes, I have been working with exodus files, but that was just because I used to work with them before and I was familiar with Cubit, exodus etc. 

If you mean this part:
```
  if(extension == "e" || extension == "exo")
    format = dealii::GridIn<dim>::Format::exodusii;
```
by incomplete specialization, the reason behind this is, that `parse_format()` doesn't work with the file ending `.e`. 

If you ask why we don't use `exodusIIData`, the reason is I didn't want to introduce it here since I didn't need the struct. I just left is a TODO, so that we can use it if needed in the future and we are aware of this.

peterrum (commented):
Looks good to me! I would suggest to go ahead with this PR :+1:

include/exadg/grid/grid_utilities.h
```diff
@@ -245,6 +246,53 @@ create_fine_and_coarse_triangulations(
...
+template<int dim>
+inline void
+read_external_triangulation(dealii::Triangulation<dim, dim> & tria, GridData const & data)
+{
+  AssertThrow(!data.file_name.empty(),
+              dealii::ExcMessage(
+                "You are trying to read a grid file, but the string, which is supposed to contain"
+                " the file, is empty. Most likely, you forgot to specify the file location in the"
+                " input file. The input file of the Poisson/sine application can be an example"
+                " showing how to set it correctly."));
```
nfehn, line 262: I would remove the sentence on Poisson/sine here.

include/exadg/grid/grid_utilities.h
```diff
@@ -245,6 +246,53 @@ create_fine_and_coarse_triangulations(
...
+ * the triangulation "tria".
+ */
+template<int dim>
+inline void
+read_external_triangulation(dealii::Triangulation<dim, dim> & tria, GridData const & data)
+{
+  AssertThrow(!data.file_name.empty(),
+              dealii::ExcMessage(
+                "You are trying to read a grid file, but the string, which is supposed to contain"
+                " the file, is empty. Most likely, you forgot to specify the file location in the"
```
nfehn, line 260: for clarity, this should read "contain the filename" in my opinion.

include/exadg/grid/grid_utilities.h
```diff
@@ -245,6 +246,53 @@ create_fine_and_coarse_triangulations(
   }
 }
 
+/**
+ * This functions reads an external triangulation defined in the file "file_in" and stores it in
```
nfehn, line 250: There is no argument file_in, so the comment needs to be adapted.

include/exadg/grid/grid_utilities.h
```diff
@@ -245,6 +246,53 @@ create_fine_and_coarse_triangulations(
...
+                "You are trying to read a grid file, but the string, which is supposed to contain"
+                " the file, is empty. Most likely, you forgot to specify the file location in the"
+                " input file. The input file of the Poisson/sine application can be an example"
+                " showing how to set it correctly."));
+
+  dealii::GridIn<dim> grid_in;
+
+  grid_in.attach_triangulation(tria);
+
+  // find the file extension from the given file_in string
```
nfehn, line 268: this comment is also not up to date.

include/exadg/grid/grid_utilities.h
```diff
@@ -245,6 +246,53 @@ create_fine_and_coarse_triangulations(
...
+
+  // TODO: check if the exodusIIData is needed
+  // typename dealii::GridIn<dim>::ExodusIIData exodusIIData;
+
+  grid_in.read(data.file_name, format);
+
+  AssertThrow(get_element_type(tria) == data.element_type,
+              dealii::ExcMessage("You are trying to read a grid file, but the element type of the"
+                                 " external grid file and the element type specified in the"
+                                 " application don't match. Most likely, you forgot to change the"
```
nfehn, line 291: The comment "specified in the application" is not precise enough. There is no need to set this parameter in the application file, since this parameter has a default value. Since we cannot know how GridData is used and will be used in the future, we should not refer to "application" here.

include/exadg/utilities/grid_parameters.h
```diff
@@ -0,0 +1,51 @@
...
+// deal.II
+#include <deal.II/base/parameter_handler.h>
+
+namespace ExaDG
+{
+struct GridParameters
+{
+  GridParameters() : file_name()
+  {
+  }
```
nfehn, line 34: not necessary in my opinion

include/exadg/utilities/grid_parameters.h
```diff
@@ -0,0 +1,51 @@
...
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *  ______________________________________________________________________
+ */
+
+#ifndef INCLUDE_EXADG_UTILITIES_GRID_PARAMETERS_H_
+#define INCLUDE_EXADG_UTILITIES_GRID_PARAMETERS_H_
```
nfehn, line 23: I think this file should be placed in the directory exadg/grid. In a similar direction, output_parameters.h should also not be in exadg/utilities.

include/exadg/utilities/grid_parameters.h
necioglu: I didn't know that std::string was automatically initialized to an empty one. I removed this.

necioglu:
@nfehn I have addressed everything in the last commits.

I didn't change `output_parameters.h`'s location.

include/exadg/grid/grid_utilities.h
```diff
@@ -245,6 +246,52 @@ create_fine_and_coarse_triangulations(
   }
 }
 
+/**
+ * This function reads an external triangulation. The path to the triangulation file should
+ * be defined in the input file and then stored in GridData. This function extracts it from
+ * GridData and stores it in the triangulation "tria".
```
nfehn, line 252: I am still not happy with this comment (and the ones below). Note that this function is not aware of any "input files". I can use this function in ExaDG without any input file! Hence, we are not allowed to refer to an input file here in my opinion.

include/exadg/grid/grid_utilities.h
```diff
@@ -245,6 +246,52 @@ create_fine_and_coarse_triangulations(
...
+ */
+template<int dim>
+inline void
+read_external_triangulation(dealii::Triangulation<dim, dim> & tria, GridData const & data)
+{
+  AssertThrow(!data.file_name.empty(),
+              dealii::ExcMessage(
+                "You are trying to read a grid file, but the string, which is supposed to contain"
+                " the file name, is empty. Most likely, you forgot to specify the file location in"
+                " the input file."));
```
nfehn, line 262: as above

include/exadg/grid/grid_utilities.h
```diff
@@ -245,6 +246,52 @@ create_fine_and_coarse_triangulations(
...
+
+  grid_in.attach_triangulation(tria);
+
+  // find the file extension from the file name
+  std::string extension = data.file_name.substr(data.file_name.find_last_of('.') + 1);
+
+  AssertThrow(!extension.empty(),
+              dealii::ExcMessage("You are trying to read a grid file, but the file extension is"
+                                 " empty. Check the input file to make sure that the file"
+                                 " extension is correctly defined after a full stop."));
```
nfehn, line 274: Another reference to input file. I might be that this function is not properly designed in terms of the single-responsibility principle. Please rethink how to design this function: which parameters are passed to this function? `GridData` contains much more information than what is needed here.

include/exadg/grid/grid_utilities.h
```diff
@@ -245,6 +246,52 @@ create_fine_and_coarse_triangulations(
...
+  // TODO: check if the exodusIIData is needed
+  // typename dealii::GridIn<dim>::ExodusIIData exodusIIData;
+
+  grid_in.read(data.file_name, format);
+
+  AssertThrow(get_element_type(tria) == data.element_type,
+              dealii::ExcMessage("You are trying to read a grid file, but the element type of the"
+                                 " external grid file and the element type specified in GridData"
+                                 " don't match. Most likely, you forgot to change the element_type"
+                                 " parameter of GridData to the desired element type."));
```
nfehn, line 292: I also do not see why this assert has to be placed in `read_external_triangulation()`. As the function name says, a grid file should be read (think about single-responsibility principle: who is responsible for reading the grid? who is responsible for checking consistency of `GridData` content?). It is in my opinion not clear that it is the responsibility of this function to check compatibility/consistency with other data structures of ExaDG (such as `GridData`, where it is questionable if this struct should be passed at all to this function). I understand the goal to write this code only once in the code, but the current solution appears to be sub-optimal in my opinion.

include/exadg/grid/grid_utilities.h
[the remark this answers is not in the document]
nfehn: done

include/exadg/grid/grid_parameters.h
```diff
@@ -0,0 +1,51 @@
...
+// deal.II
+#include <deal.II/base/parameter_handler.h>
+
+namespace ExaDG
+{
+struct GridParameters
+{
+  GridParameters()
+  {
+  }
```
nfehn, line 34: this could also be removed

include/exadg/poisson/user_interface/application_base.h
```diff
@@ -59,6 +60,7 @@ class ApplicationBase
   add_parameters(dealii::ParameterHandler & prm)
   {
     output_parameters.add_parameters(prm);
+    grid_parameters.add_parameters(prm);
```
nfehn, line 63: The solver `poisson/solver.h` contains the option to create a default input file, see the function `create_input_file(std::string const & input_file)`. If I remember correctly, the parameters are added to the input file in the order they are added in the code. Hence, if we want `grid_parameters` as an "input" to occur before "output" variables, we should probably change the order here. Note that the order is also different below when defining grid and output parameters as class members.

include/exadg/grid/grid_utilities.h
nfehn: If you are undecided regarding this point, I would suggest to keep it as is for now and potentially change it later.

include/exadg/grid/grid_utilities.h
necioglu: Yes, I agree that this function is independent from the input file. However, I think that we should explain somewhere, how this function was meant to be used. And I thought this would be a good location for this. Otherwise, the user might not know the possibility of defining the path in the input_file at all.

nfehn: Yes, I see your intention. But this is a clear indication that the design is not good enough right now. Imagine someone forgot to add `grid_parameters.add_parameters()` in `Structure::ApplicationBase`. One will then run into this assert, but adding the line with `file_name` to the input file will not help, because it is kind of a bug in the software (not being implemented so far) and the parameter will never be read from the input file. The error message of the above assert just causes more confusion. A function and its documentation in `grid/grid_utilities.h` has to work for all modules/applications in ExaDG, not just Poisson where this have already been implemented. But also for Poisson, the implementation might be changed in the future. How can you guarantee that the developer doing these changes will also adapt `grid_utilities.h`?

include/exadg/grid/grid_utilities.h
necioglu: I decided to pass `GridData` to this function because:

1. The user is forced to use `GridData`, which is in my opinion good. I could also go with a `std::string` argument, but then everyone could define the path in application directly. (Not that they can't do it now, but it is harder and therefore less likely.)
2. I can check the mismatch of element types bellow with an Assert.

Do you think that a `std::string` argument would be better?

include/exadg/grid/grid_utilities.h
necioglu: I see, what you mean. I think my comment above https://github.com/exadg/exadg/pull/323#discussion_r1105533234 explains why I decided to go for this solution. 


> I understand the goal to write this code only once in the code, but the current solution appears to be sub-optimal in my opinion.

Yes, this was my goal. And I can't think of another solution without duplicating code. Maybe something like `GridData.check()`, but I think this would complicate this PR.

include/exadg/grid/grid_parameters.h
necioglu: Just to be sure, you mean the initializer and not the class itself, right?

include/exadg/poisson/user_interface/application_base.h
necioglu: Yes you are right. I will look into this! I will also change the order in `input.json`, if needed.

include/exadg/grid/grid_utilities.h
necioglu: I understand your point better now! I think extending this comment with the whole procedure of reading a grid_file from the input_file would be too much here. I will remove input_file from this comment and adapt the wording accordingly.

include/exadg/grid/grid_utilities.h
necioglu: They are the same, that is why I said : 

> (Not that they can't do it now, but it is harder and therefore less likely.)

For now, I only need the `file_name` and `element_type`. The `element_type` is only needed for the Assert. I have checked the variables of `GridData` again, and it seems unlikely that this function would end up needing more variables. 

I will change to a `std::string` argument and omit the Assert for now. But I think the Assert is needed somewhere.

include/exadg/grid/grid_parameters.h
nfehn: the constructor

include/exadg/grid/grid_utilities.h
```diff
@@ -245,6 +246,49 @@ create_fine_and_coarse_triangulations(
   }
 }
 
+/**
+ * This function reads an external triangulation. The triangulation is extracted from
```
nfehn, line 250: I do not understand "triangulation is extracted from GridData"?

nfehn (approved):

include/exadg/grid/grid_utilities.h
nfehn: Let's keep this in mind for the future.

PR #336, opened 2023-02-15
Postprocess shear rate

richardschu:
Added the shear rate to the postprocessor of incompressible_navier_stokes in complete analogy to the routines available for the velocity divergence and set flag in the incompressible_navier_stokes/poiseuille example.


include/exadg/incompressible_navier_stokes/spatial_discretization/calculators/shear_rate_calculator.cpp
```diff
@@ -0,0 +1,93 @@
...
+    integrator_vector.reinit(cell);
+    integrator_vector.gather_evaluate(src, dealii::EvaluationFlags::gradients);
+
+    integrator_scalar.reinit(cell);
+
+    for(unsigned int q = 0; q < integrator_scalar.n_q_points; q++)
+    {
+      dealii::SymmetricTensor<2, dim, dealii::VectorizedArray<Number>> sym_grad_u =
+        integrator_vector.get_symmetric_gradient(q);
+      scalar shear_rate = std::sqrt(0.5 * scalar_product(sym_grad_u, sym_grad_u));
```
nfehn, line 77: Is there a common definition of how to define the shear rate? I remember the implementation of the Smagorinsky turbulence model where we have the factor of 2 instead of 0.5, see https://github.com/exadg/exadg/blob/ef064e0a5651ad46cee8205cfeccb203960443e5/include/exadg/incompressible_navier_stokes/spatial_discretization/turbulence_model.cpp#L282.

In case there are conventions involved, it could make sense to mention them and potentially give a reference to literature.

include/exadg/incompressible_navier_stokes/spatial_discretization/calculators/shear_rate_calculator.cpp
```diff
@@ -0,0 +1,93 @@
...
+  for(unsigned int cell = cell_range.first; cell < cell_range.second; ++cell)
+  {
+    integrator_vector.reinit(cell);
+    integrator_vector.gather_evaluate(src, dealii::EvaluationFlags::gradients);
+
+    integrator_scalar.reinit(cell);
+
+    for(unsigned int q = 0; q < integrator_scalar.n_q_points; q++)
+    {
+      dealii::SymmetricTensor<2, dim, dealii::VectorizedArray<Number>> sym_grad_u =
```
nfehn, line 75: for consistency, we could `typedef` the symmetric tensor in the class definition and reuse it here (as we do it for `scalar`).

nfehn:
Thanks @richardschu! Looks very good overall! I only have tiny comments on the implementation.

include/exadg/incompressible_navier_stokes/spatial_discretization/calculators/shear_rate_calculator.cpp
richardschu: Thanks for pointing that out, in fact, definitions with 0.5 and 2 are being used ... which is a mess.
I have been using 0.5 in the past (changing the constitutive parameters accordingly), BUT using 2 instead is (1) the one consistent with 1D Poiseuille flow, (2) more common (which is debatable?), (3) coincides with the code snippet you mention, and we have a good source to cite:
https://doi.org/10.1007/978-3-7643-7806-6
So, I will swap out the 0.5 for 2.0.

include/exadg/incompressible_navier_stokes/spatial_discretization/calculators/shear_rate_calculator.cpp
richardschu: I use now
typedef (...) symmetrictensor;
Normally, types would follow "SomeDefinedType" and variables would be "my_new_variable".
But for the scalar, vector, tensor with dealii::VectorizedArray, you used "scalar", "vector", "tensor" as type names.
What is the convention here?

nfehn: yes, I would also use a naming convention in line with `scalar`, `tensor` (small letters only).

include/exadg/compressible_navier_stokes/spatial_discretization/calculators.h
```diff
@@ -176,6 +176,81 @@ class DivergenceCalculator
   unsigned int quad_index;
 };
 
+template<int dim, typename Number>
+class ShearRateCalculator
```
nfehn, line 180: What is exactly the difference to the incompressible case? It looks like we have somewhat copied the code in the past for other calculators as well (divergence, vorticity). Since this has been a bad style, we should now work on removing such code redundancies. So the goal should be to define certain calculators in a general module (e.g. `exadg/operators` and living in the namespace `ExaDG`) and use them where possible, i.e. the same implementation for incompressible and compressible flow.

For the present PR I would suggest that we try to figure out whether we can use the same implementation for both compressible and incompressible (or what we have to change so that this is possible). In a follow-up PR we can then apply this to other calculators relevant for both compressible and incompressible flows.

richardschu: There is no difference whatsoever, but since I found the `calculators` defined separately, I added my code accordingly. I will have a look and keep you posted on further possibilities.

richardschu: I moved the 'calculators' (for now this contains only the shear rate) to 'include/operators/navier_stokes_calculators', copying the one from the incompressible Navier-Stokes directory. Changed the remaining files accordingly and tested w. 'poiseuille' examples in ('in'-)'compressible_navier_stokes'.
As you suggest: in a follow-up PR, we can pull over the rest of the calculators, i.e., vorticity, divergence, velocity magnitude and q-criterion.

include/exadg/compressible_navier_stokes/postprocessor/output_generator.h
```diff
@@ -74,6 +76,9 @@ struct OutputData : public OutputDataBase
   // write divergence of velocity field
   bool write_divergence;
 
+  // write divergence of velocity field
```
nfehn, line 79: copy-paste error from line 76 above

include/exadg/compressible_navier_stokes/postprocessor/postprocessor.cpp
```diff
@@ -263,6 +268,25 @@ PostProcessor<dim, Number>::initialize_derived_fields()
     divergence.reinit();
   }
 
+  // shear rate
+  if(pp_data.output_data.write_shear_rate)
+  {
+    AssertThrow(pp_data.output_data.write_velocity == true,
+                dealii::ExcMessage("You need to activate write_velocity."));
```
nfehn, line 275: I am currently not sure whether we need this Assert, because in the  very lengthy if-statement above you already added `|| pp_data.output_data.write_shear_rate` (which should be the correct logic, i.e. we need to re-compute the velocity, but we do not have to write vtu-output for velocity in order to be able to write output for the shear rate). Would you like to check this? Of course, the same would then apply to previous implementations for `vorticity`, `divergence`, and potentially other quantities ...

include/exadg/operators/navier_stokes_calculators.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+    integrator_scalar.reinit(cell);
+
+    for(unsigned int q = 0; q < integrator_scalar.n_q_points; q++)
+    {
+      symmetrictensor sym_grad_u = integrator_vector.get_symmetric_gradient(q);
+
+      // Shear rate definition according to Galdi et al., 2008
+      // ("Hemodynamical Flows: Modeling, Analysis and Simulation").
+      // trace(sym_grad_u^2) = sym_grad_u : sym_grad_u
```
nfehn, line 77: I do not fully understand this equation (or where to see the factor of 2 from this equation).

include/exadg/compressible_navier_stokes/postprocessor/postprocessor.cpp
richardschu: Yep those assertions can be removed safely, doing that for the shear rate, div(u) and vorticity. See screenshot, .pvtu file does not contain u, but the dependent quantities!
Screenshot from 2023-02-17 11-52-37

include/exadg/operators/navier_stokes_calculators.cpp
richardschu: One usually finds a definition (also in that cited source): shear_rate := sqrt(2 * trace(gradS(u)^2)), but since it is symmetric, we can equivalently use the contraction instead, which is cheaper to compute. I can also put the full formula?

include/exadg/operators/navier_stokes_calculators.cpp
```diff
@@ -74,7 +74,7 @@ ShearRateCalculator<dim, Number>::cell_loop(dealii::MatrixFree<dim, Number> cons
 
       // Shear rate definition according to Galdi et al., 2008
       // ("Hemodynamical Flows: Modeling, Analysis and Simulation").
-      // trace(sym_grad_u^2) = sym_grad_u : sym_grad_u
+      // sqrt(2*trace(sym_grad_u^2)) = sqrt(2*sym_grad_u : sym_grad_u)
```
richardschu, line 77: we can also leave this out completely if that is preferred?

include/exadg/compressible_navier_stokes/postprocessor/postprocessor.cpp
nfehn: thtanks for checking this

nfehn (approved):

PR #338, opened 2023-02-16
Fix volume/surface calculation for simplices in IP

### LINKED ISSUE #334: Wrong calculation of surface/volume for simplex elements in IP
The quadrature rules in `interior_penalty_parameter.h` should be adapted for simplices.

@nfehn @bergbauer This only solves the problem I showed in my interim presentation partially. (The problem of iteration counts rising for simplex elements with a MG Preconditioner for Poisson/Sine problem.) 

But this is a bug anyways, and I will fix it.

https://github.com/exadg/exadg/blob/7472268b8a32577812642a1fbc1ff8b210bc63b4/include/exadg/operators/interior_penalty_parameter.h#L57-L61

necioglu:
Closes #334.

The calculation of the surface and the volume for simplices was wrong, because of the wrong quadrature rule. This PR fixes this.

### DIFF
--- a/include/exadg/operators/interior_penalty_parameter.h
+++ b/include/exadg/operators/interior_penalty_parameter.h
@@ -62,7 +62,8 @@ calculate_penalty_parameter(
   dealii::FEValues<dim> fe_values(mapping, fe, quadrature, dealii::update_JxW_values);
 
   auto const face_quadrature =
-    reference_cells[0].template get_gauss_type_quadrature<dim - 1>(degree + 1);
+    reference_cells[0].face_reference_cell(0).template get_gauss_type_quadrature<dim - 1>(degree +
+                                                                                          1);
   dealii::FEFaceValues<dim> fe_face_values(mapping, fe, face_quadrature, dealii::update_JxW_values);
 
   for(unsigned int i = 0; i < n_cells; ++i)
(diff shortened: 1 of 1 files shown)


peterrum (approved):

necioglu:
Just noticed that I can't request a quadrature with `<dim-1>` with `get_gauss_type_quadrature`, because of: https://github.com/dealii/dealii/blob/52ddd1508c82e0ffbd5b38021f536ddd19e80715/source/grid/reference_cell.cc#L188

I was testing in release, not in debug and missed the Assert.

include/exadg/operators/interior_penalty_parameter.h
```diff
@@ -54,10 +54,15 @@ calculate_penalty_parameter(
...
+  auto const reference_cells =
+    matrix_free.get_dof_handler(dof_index).get_triangulation().get_reference_cells();
+  AssertThrow(reference_cells.size() == 1, dealii::ExcMessage("No mixed meshes allowed."));
+
+  auto const quadrature = reference_cells[0].template get_gauss_type_quadrature<dim>(degree + 1);
   dealii::FEValues<dim> fe_values(mapping, fe, quadrature, dealii::update_JxW_values);
 
-  dealii::QGauss<dim - 1>   face_quadrature(degree + 1);
+  auto const face_quadrature =
+    reference_cells[0].template get_gauss_type_quadrature<dim - 1>(degree + 1);
```
peterrum, line 65:
```suggestion
  auto const face_quadrature =
    reference_cells[0]face_reference_cell(0).template get_gauss_type_quadrature<dim - 1>(degree + 1);
```

necioglu: Thanks! Changing to this.

peterrum (approved):

PR #348, opened 2023-02-18
[WIP] Refactor navier stokes calculators

richardschu:
Follow-up PR; the `navier_stokes_calculators` now include not only the `ShearRateCalculator` but also the `DivergenceCalculator`, `VelocityMagnitudeCalculator` and `VorticityCalculator`. Remaining calculators are specialised for (in)compressible flow (solution conversion, streamfunction), but in Q-Criterion, we could adopt the incompressible case, or template a bool `is_compressible` as hinted at in `q_criterion_calculator.cpp`.

Also I want to add the wall shear stress, separate PR for that?

### DIFF
--- a/.github/workflows/indentation.yml
+++ b/.github/workflows/indentation.yml
@@ -12,12 +12,12 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-    - uses: DoozyX/clang-format-lint-action@v0.11
+    - uses: DoozyX/clang-format-lint-action@v0.14
       with:
         source: '.'
         exclude: ''
         extensions: 'h.in,h,cc,cpp'
-        clangFormatVersion: 6
+        clangFormatVersion: 11.1.0
         inplace: true
     - run: |
         git diff
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 *cbp
 .directory
 build/
+programs/
--- a/applications/fluid_structure_interaction/bending_wall/application.h
+++ b/applications/fluid_structure_interaction/bending_wall/application.h
@@ -291,14 +291,16 @@ private:
       SchurComplementPreconditioner::PressureConvectionDiffusion;
   }
 
-  void create_triangulation(dealii::Triangulation<2> & tria)
+  void
+  create_triangulation(dealii::Triangulation<2> & tria)
   {
     (void)tria;
 
     AssertThrow(false, dealii::ExcMessage("not implemented."));
   }
 
-  void create_triangulation(dealii::Triangulation<3> & tria)
+  void
+  create_triangulation(dealii::Triangulation<3> & tria)
   {
     std::vector<dealii::Triangulation<3>> tria_vec;
     tria_vec.resize(17);
--- a/applications/fluid_structure_interaction/cylinder_with_flag/application.h
+++ b/applications/fluid_structure_interaction/cylinder_with_flag/application.h
@@ -315,7 +315,8 @@ private:
       SchurComplementPreconditioner::PressureConvectionDiffusion;
   }
 
-  void create_triangulation(dealii::Triangulation<2> & tria)
+  void
+  create_triangulation(dealii::Triangulation<2> & tria)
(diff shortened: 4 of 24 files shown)


nfehn:
Yes, probably a separate PR for the wall shear stress to separate refactoring from new functionality.

include/exadg/operators/navier_stokes_calculators.h
```diff
@@ -72,6 +109,125 @@ class ShearRateCalculator
...
+            Range const &                           cell_range) const;
+
+  dealii::MatrixFree<dim, Number> const * matrix_free;
+
+  unsigned int dof_index_u;
+  unsigned int dof_index_u_scalar;
+  unsigned int quad_index;
+};
+
+template<int dim, typename Number, bool compressible_flow>
```
nfehn, line 189: I think `compressible_flow` should be realized as an additional parameter to the initialize function (with the goal to avoid template parameters as much as possible).

include/exadg/operators/navier_stokes_calculators.h
```diff
@@ -72,6 +109,125 @@ class ShearRateCalculator
...
+            VectorType const &                            src,
+            std::pair<unsigned int, unsigned int> const & cell_range) const;
+
+  dealii::MatrixFree<dim, Number> const * matrix_free;
+  unsigned int                            dof_index;
+  unsigned int                            quad_index;
+};
+
+template<int dim, typename Number>
+class VelocityMagnitudeCalculator
```
nfehn, line 149: We should rename this class to avoid misunderstandings, because it is also used to calculate the vorticity magnitude (if I remember correctly). So I suggest just `MagnitudeCalculator`.

nfehn:
Looks good from my side. Two questions remain before merging:

Is it correct that the only change in the code is adding the compressible case to the q-criterion calculation?

Did you test the new implementation for both incompressible and compressible examples?

include/exadg/operators/navier_stokes_calculators.h
richardschu: Did that, commit is incoming!

include/exadg/operators/navier_stokes_calculators.h
richardschu: Ok, I introduced it as a member variable and passed it as a bool const into the cell_loop, hoping that the compiler realizes that it is constant and gets rid of the if() at the quadrature point level. I rather did that than duplicating the cell_loop, which would be a safer option? What do you prefer?

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
```diff
@@ -1197,7 +1198,7 @@ void
 SpatialOperatorBase<dim, Number>::compute_vorticity_magnitude(VectorType &       dst,
                                                               VectorType const & src) const
 {
-  velocity_magnitude_calculator.compute(dst, src);
```
richardschu, line 1200: here is that point you mentioned @nfehn .

include/exadg/operators/navier_stokes_calculators.cpp
```diff
@@ -84,10 +144,232 @@ ShearRateCalculator<dim, Number>::cell_loop(dealii::MatrixFree<dim, Number> cons
...
+        {
+          Om[i][j] = 0.5 * (grad_u[i][j] - grad_u[j][i]);
+          S[i][j]  = 0.5 * (grad_u[i][j] + grad_u[j][i]);
+        }
+      }
+
+      // Q criterion for compressible flow is based on the deviatoric
+      // part of S (Kolar, V.: Compressibility Effect in Vortex Identification,
+      // AIAA, 2009. https://doi.org/10.2514/1.40131)
+      if(compressible_flow)
```
richardschu, line 329: here is the if at the quadrature point level:
1) leave it that way
2) duplicate cell loop inside `cell_loop` and have an if at the front
3) ?
What should we go for?

PR #349, opened 2023-02-22
[WIP] wall_shear_stress_calculator for incompressible Navier--Stokes

richardschu:
Added a `wall_shear_stress_calculator` for the `incompressible_navier_stokes` solver. Interpolation via dealii::FEFaceValues on boundaries with specified IDs, on the entire boundary or skip (see Poiseuille flow example). Output via dealii::DataOutFaces (in 3D only, since .vtk output is not available for 2D). This is in the `incompressible_navier_stokes` module, since I assumed constant viscosity and density.
For an iliac bifurcation, this might look something like (nodal values on the inlet are zero):
Screenshot from 2023-02-22 17-26-03


include/exadg/incompressible_navier_stokes/spatial_discretization/calculators/wall_shear_stress_calculator.cpp
```diff
@@ -0,0 +1,232 @@
...
+         write_on_boundary_IDs.find(cell->face(face)->boundary_id()) != write_on_boundary_IDs.end())
+      {
+        fe_face_values.reinit(cell, face);
+        fe_face_values[vector].get_function_gradients(src, velocity_gradients);
+
+        for(auto const i : fe_face_values.quadrature_point_indices())
+        {
+          unsigned int matching_idx = face_to_cell_index[face][i];
+
+#ifdef DEBUG
```
richardschu, line 181: do we even want to have this check here? ... could be useful if someone accidentally changes face orientation on the boundary at some point.

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
```diff
@@ -867,6 +872,13 @@ SpatialOperatorBase<dim, Number>::get_viscosity() const
   return param.viscosity;
 }
 
+template<int dim, typename Number>
+double
+SpatialOperatorBase<dim, Number>::get_density() const
```
richardschu, line 877: I did not need this in the end, should I remove it again?

include/exadg/incompressible_navier_stokes/spatial_discretization/calculators/wall_shear_stress_calculator.h
```diff
@@ -0,0 +1,86 @@
...
+    std::vector<dealii::types::boundary_id> const write_wall_shear_stress_on_IDs) const;
+
+private:
+  dealii::MatrixFree<dim, Number> const * matrix_free;
+
+  unsigned int                                              dof_index;
+  unsigned int                                              quad_index;
+  std::vector<std::vector<dealii::types::global_dof_index>> face_to_cell_index;
+  double                                                    dynamic_viscosity;
+  double const                                              rel_tol = 1.0e-5;
```
richardschu, line 79: Relative tolerance used for point matching combined with cell->diameter()

applications/incompressible_navier_stokes/poiseuille/application.h
```diff
@@ -476,17 +476,18 @@ class Application : public ApplicationBase<dim, Number>
...
-    pp_data.output_data.degree                    = this->param.degree_u;
-    pp_data.output_data.write_higher_order        = true;
+    pp_data.output_data.directory                      = this->output_parameters.directory + "vtu/";
+    pp_data.output_data.filename                       = this->output_parameters.filename;
+    pp_data.output_data.write_vorticity                = true;
+    pp_data.output_data.write_divergence               = true;
+    pp_data.output_data.write_shear_rate               = true;
+    pp_data.output_data.write_velocity_magnitude       = true;
+    pp_data.output_data.write_vorticity_magnitude      = true;
+    pp_data.output_data.write_wall_shear_stress_on_IDs = {dealii::numbers::invalid_boundary_id};
```
nfehn, line 486: We should implement it in a way such that this default initialization is not necessary.

include/exadg/incompressible_navier_stokes/postprocessor/output_generator.h
```diff
@@ -77,6 +78,9 @@ struct OutputData : public OutputDataBase
   // write vorticity magnitude
   bool write_vorticity_magnitude;
 
+  // write wall shear stress on IDs
+  std::vector<dealii::types::boundary_id> write_wall_shear_stress_on_IDs;
```
nfehn, line 82: I think the name should include "boundary", e.g. `_boundary_IDs` or short `_bids`.

include/exadg/incompressible_navier_stokes/spatial_discretization/calculators/wall_shear_stress_calculator.cpp
```diff
@@ -0,0 +1,232 @@
...
+}
+
+template<int dim, typename Number>
+void
+WallShearStressCalculator<dim, Number>::initialize(
+  dealii::MatrixFree<dim, Number> const & matrix_free_in,
+  unsigned int const                      dof_index_in,
+  unsigned int const                      quad_index_in,
+  double const                            kinematic_viscosity,
+  double const                            density)
```
nfehn, line 41: why not writing the interface with `dynamic_viscosity`?

include/exadg/incompressible_navier_stokes/postprocessor/postprocessor.cpp
```diff
@@ -160,6 +161,14 @@ PostProcessor<dim, Number>::do_postprocessing(VectorType const &     velocity,
       velocity_magnitude.evaluate(velocity);
       additional_fields_vtu.push_back(&velocity_magnitude);
     }
+    if(pp_data.output_data.write_wall_shear_stress_on_IDs.size() > 0)
+    {
+      wall_shear_stress.evaluate(velocity);
+      if(dim == 3)
+        surface_fields_vtu.push_back(&wall_shear_stress);
+      else
+        additional_fields_vtu.push_back(&wall_shear_stress);
```
nfehn, line 170: Right now, I do not understand this. Does it mean that a volume field is written in 2D for the wall shear stress? This would not be intuitive and require some explanation in my opinion.

include/exadg/incompressible_navier_stokes/spatial_discretization/calculators/wall_shear_stress_calculator.h
nfehn: I suggest to add this as a comment to the code

include/exadg/incompressible_navier_stokes/spatial_discretization/calculators/wall_shear_stress_calculator.h
```diff
@@ -0,0 +1,86 @@
...
+             unsigned int const                      quad_index_in,
+             double const                            kinematic_viscosity,
+             double const                            density);
+
+  void
+  compute_wall_shear_stress(
+    VectorType &                                  dst,
+    VectorType const &                            src,
+    std::shared_ptr<dealii::Mapping<dim> const>   mapping,
+    std::vector<dealii::types::boundary_id> const write_wall_shear_stress_on_IDs) const;
```
nfehn, line 70: I would also here include "boundary" in the name (see other comment).

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
nfehn: yes, should be removed again.

include/exadg/incompressible_navier_stokes/postprocessor/output_generator.cpp
```diff
@@ -200,6 +257,13 @@ OutputGenerator<dim, Number>::evaluate(
                     additional_fields,
                     time_control.get_counter(),
                     mpi_comm);
+
+  if(dim == 3 && surface_fields.size() > 0)
```
nfehn, line 261: I currently think this if statement should appear inside the function `write_surface_output()` from a design perspective.

include/exadg/incompressible_navier_stokes/postprocessor/output_generator.cpp
```diff
@@ -200,6 +257,13 @@ OutputGenerator<dim, Number>::evaluate(
...
+  {
+    // surface vtk output for dim == 3 only
```
nfehn, line 263: No need to mention the "vtk" info here in my opinion.

include/exadg/incompressible_navier_stokes/postprocessor/output_generator.cpp
```diff
@@ -114,6 +115,61 @@ write_output(
...
+void
+write_surface_output(
+  OutputData const &                                                    output_data,
+  dealii::Mapping<dim> const &                                          mapping,
+  std::vector<dealii::SmartPointer<SolutionField<dim, Number>>> const & surface_fields,
+  unsigned int const                                                    output_counter,
+  MPI_Comm const &                                                      mpi_comm)
+{
+  AssertThrow(dim == 3, dealii::ExcMessage("Surface vtk output for dim == 3 only."));
+  AssertThrow(surface_fields.size() > 0, dealii::ExcMessage("Provided surface fields empty."));
```
nfehn, line 128: can be removed when "if-statement" is implemented here.

include/exadg/incompressible_navier_stokes/postprocessor/output_generator.cpp
```diff
@@ -186,6 +242,7 @@ OutputGenerator<dim, Number>::evaluate(
   VectorType const &                                                    velocity,
   VectorType const &                                                    pressure,
   std::vector<dealii::SmartPointer<SolutionField<dim, Number>>> const & additional_fields,
+  std::vector<dealii::SmartPointer<SolutionField<dim, Number>>> const & surface_fields,
```
nfehn, line 245: It seems we currently have a mixture of designs. In some sense, this line breaks the modularity of the Navier-Stokes postprocessing design (we need to change interfaces to realize more/other functionality). According to the current function interface above, I would have expected that the new functionality is included in the existing `write_output()` function. Instead, a separate `write_surface_output()` function is introduced, which aims at a modular design somewhat deeper in the code. However, this raises the question why surface output is then not realized completely modular as a separate postprocessing module (e.g. `SurfaceOutputGenerator`, or a seperate function inside the existing `OutputGenerator`). In my opinion, both realizations are possible and should be discussed.

Some arguments below:

Another option could be to equip `SolutionField` with additional information (volume vs. surface) such that the if-branches can be made at the point where output is written. I am not sure if this makes the design better, because in the end, we also have to discuss the design in relation to the deal.II library (which has separate classes `DataOut` and `DataOutFaces`). This argument relates also to this discussion https://github.com/exadg/exadg/pull/349/files#r1126093917

What should be mentioned at this point and be part of the overall discussion is that I discussed with @c-p-schmidt some time ago that it would actually be better that `write_output()` would only receive `additional_fields` as argument and not `velocity` and `pressure` separately, because this would allow us to (maybe) use exactly the same `write_output()` function for all PDE modules in ExaDG (Poisson, ConvDiff, NavierStokes, etc.). This argument supports to keep `write_output()` and `write_surface_output()` separately (and separate arguments `volume_fields` and `surface_fields`), as is done currently.

include/exadg/incompressible_navier_stokes/spatial_discretization/calculators/wall_shear_stress_calculator.cpp
```diff
@@ -0,0 +1,232 @@
...
+                        dealii::ExcMessage("Face to cell match recovery not succesful."));
+          }
+#endif
+
+          dealii::Tensor<1, dim> const S_times_n =
+            dynamic_viscosity * (velocity_gradients[i] + transpose(velocity_gradients[i])) *
+            fe_face_values.normal_vector(i);
+          dealii::Tensor<1, dim> const wall_shear_stress =
+            S_times_n - scalar_product((S_times_n), fe_face_values.normal_vector(i)) *
+                          fe_face_values.normal_vector(i);
```
nfehn, line 210: I would introduce
```
normal = fe_face_values.normal_vector(i);
```
since it is used three times here.

include/exadg/incompressible_navier_stokes/spatial_discretization/calculators/wall_shear_stress_calculator.cpp
nfehn: What do you mean by "someone accidentally changes face orientation on the boundary"? A bug being introduced in deal.II?

include/exadg/incompressible_navier_stokes/spatial_discretization/calculators/wall_shear_stress_calculator.cpp
```diff
@@ -0,0 +1,232 @@
...
+  const dealii::FEValuesExtractors::Vector     vector(0);
+
+  // fill vector with entries on boundary via cell-loop or prepared container.
+  for(auto const & cell : matrix_free->get_dof_handler(dof_index).active_cell_iterators())
+  {
+    if(cell->is_locally_owned() == false)
+      continue;
+
+    if(cell->at_boundary() == false)
+      continue;
```
nfehn, line 157: Personally, I do not like the `if(.... == false) continue;` style (negations are always harder to read for humans than positive statements). I think this "style" originates from indentation aspects with 80 characters per line in deal.II and the fact that deal.II indentation introduces two tabs whenever using brackets. Note that all this does not apply for ExaDG.

include/exadg/incompressible_navier_stokes/spatial_discretization/calculators/wall_shear_stress_calculator.cpp
```diff
@@ -0,0 +1,232 @@
...
+  {
+    if(write_wall_shear_stress_on_IDs[0] == dealii::numbers::invalid_boundary_id)
+      write_on_all_boundary_IDs = true;
+  }
+  std::map<dealii::types::boundary_id, bool> write_on_boundary_IDs;
+  for(auto const id : write_wall_shear_stress_on_IDs)
+    write_on_boundary_IDs.insert({id, true});
+
+  unsigned int const dofs_per_component =
+    matrix_free->get_dof_handler(dof_index).get_fe(0).dofs_per_cell / dim;
```
nfehn, line 123: Can't deal.II provide the information about scalar dofs per cell directly? Where is this variable actually used?

include/exadg/incompressible_navier_stokes/spatial_discretization/calculators/wall_shear_stress_calculator.cpp
```diff
@@ -0,0 +1,232 @@
...
+      "Provide at least write_wall_shear_stress_on_IDs = {dealii::numbers::invalid_boundary_id}.")) bool
+    write_on_all_boundary_IDs = false;
+  if(write_wall_shear_stress_on_IDs.size() == 1)
+  {
+    if(write_wall_shear_stress_on_IDs[0] == dealii::numbers::invalid_boundary_id)
+      write_on_all_boundary_IDs = true;
+  }
+  std::map<dealii::types::boundary_id, bool> write_on_boundary_IDs;
+  for(auto const id : write_wall_shear_stress_on_IDs)
+    write_on_boundary_IDs.insert({id, true});
```
nfehn, line 120: Looks like `std::set` would be more appropriate?

include/exadg/incompressible_navier_stokes/spatial_discretization/calculators/wall_shear_stress_calculator.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+  unsigned int const dofs_per_component =
+    matrix_free->get_dof_handler(dof_index).get_fe(0).dofs_per_cell / dim;
+  unsigned int const    fe_degree = matrix_free->get_dof_handler(dof_index).get_fe(0).degree;
+  dealii::FESystem<dim> fe_system(dealii::FE_DGQ<dim>(fe_degree), dim);
```
nfehn, line 125: We need to keep in mind (and assert) that this will not work for simplex.

include/exadg/incompressible_navier_stokes/spatial_discretization/calculators/wall_shear_stress_calculator.cpp
```diff
@@ -0,0 +1,232 @@
...
+  AssertThrow(
+    write_wall_shear_stress_on_IDs.size() > 0,
+    dealii::ExcMessage(
+      "Provide at least write_wall_shear_stress_on_IDs = {dealii::numbers::invalid_boundary_id}.")) bool
+    write_on_all_boundary_IDs = false;
+  if(write_wall_shear_stress_on_IDs.size() == 1)
+  {
+    if(write_wall_shear_stress_on_IDs[0] == dealii::numbers::invalid_boundary_id)
+      write_on_all_boundary_IDs = true;
+  }
```
nfehn, line 117: I am against an implicit logic as realized here. I think this clarifies my original comment here https://github.com/exadg/exadg/pull/349/files#r1126058363. If we want to write on all boundaries, I think we have to list all of them in the application file.

From an application perspective, I would argue that no-slip wall boundaries (also referring to the picture you show in this PR) typically have the same boundary ID (the deal.II default value 0)?

include/exadg/incompressible_navier_stokes/spatial_discretization/calculators/wall_shear_stress_calculator.cpp
nfehn: I do not see the reason to introduce `write_on_boundary_IDs`

include/exadg/incompressible_navier_stokes/spatial_discretization/calculators/wall_shear_stress_calculator.cpp
```diff
@@ -0,0 +1,232 @@
...
+}
+
+template<int dim, typename Number>
+void
+WallShearStressCalculator<dim, Number>::compute_wall_shear_stress(
+  VectorType &                                  dst,
+  VectorType const &                            src,
+  std::shared_ptr<dealii::Mapping<dim> const>   mapping,
+  std::vector<dealii::types::boundary_id> const write_wall_shear_stress_on_IDs) const
```
nfehn, line 103: `std::set`?

include/exadg/incompressible_navier_stokes/spatial_discretization/calculators/wall_shear_stress_calculator.cpp
nfehn: std::cout output is problematic in ExaDG sections guarded by `debug`, because this could cause failing tests when running in debug mode. However, our policy in ExaDG is that tests should pass in both debug and release mode.

For code review right now, I think it would be easier for me if all debug sections were removed and were realized experimentally on a developer branch for now.

include/exadg/incompressible_navier_stokes/postprocessor/output_generator.cpp
c-p-schmidt: I'm not sure whether it works for all modules but it sounded very appealing to me. If this `additional_fields` data object consists of objects the final output routine is able to write to disc, then the `write_output()` method should mainly need to iterate over the provided data from `additional_fields` and choose the correct output method to write to disc.

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
richardschu: I did :)

include/exadg/incompressible_navier_stokes/spatial_discretization/calculators/wall_shear_stress_calculator.cpp
richardschu: If someone is manually changing the face orientation on the outer boundary, this influences the ordering of the points/DoFs -> so we would have non-matching points. @peterum commented on this, that this *should* never be the case, which is why I do the matching during `initialize`.

Regarding the debug sections and std::cout there:
do the checks compare the log files via diff?
If the face orientation is not changed on the boundary (by mistake?!), even debug mode does not output anything - it ONLY CHECKS, if the match was successful, but should still not output anything. So I think if the tests then fail here ... this actually tells us a lot just as we want, right?

Of course one could remove that entirely, but this is kind of a safety measure. We could also always match the points for each face, but that is considerably more effort. And that wall shear stress will lateron be needed for adapting boundary conditions, so it has to be computed at each time step!

So I would actually rather keep it.

include/exadg/incompressible_navier_stokes/spatial_discretization/calculators/wall_shear_stress_calculator.h
richardschu: I added a comment.

applications/incompressible_navier_stokes/poiseuille/application.h
richardschu: now, the default is: write on all  boundaries, if `write_wall_shear_stress == true`, and the user can specify a std::set of `boundary_id`s. I would like this more than just allowing a single one, since we might use the boundary ID more actual varying boundary conditions *on* the wall (thrombus growth, heated surface or whatever - some multiphysics application). In my mind this is just a bit more flexible.

include/exadg/incompressible_navier_stokes/postprocessor/output_generator.h
richardschu: I did, good point.

include/exadg/incompressible_navier_stokes/spatial_discretization/calculators/wall_shear_stress_calculator.cpp
richardschu: I changed it, this was for no special reason. I will actually rewrite this once we have a fixed setup with the variable viscosity!

include/exadg/incompressible_navier_stokes/postprocessor/postprocessor.cpp
richardschu: Yes, unfortunately. deal.II does not allow output of 2D `SurfaceOutput`. Hence I write a volumetric data. And yes, I agree that this is ugly =(

include/exadg/incompressible_navier_stokes/spatial_discretization/calculators/wall_shear_stress_calculator.cpp
richardschu: Ok, I am fine with that style!

richardschu: Same style question: should I even write `if( ... == true)`?

include/exadg/incompressible_navier_stokes/spatial_discretization/calculators/wall_shear_stress_calculator.cpp
richardschu: did not fine one in [classFiniteElement](https://www.dealii.org/current/doxygen/deal.II/classFiniteElement.html), and did not get lucky with get_sub_fe(component_mask).
This form here works for all ReferenceCell types, so that should be fine as far as I am concerned?

include/exadg/incompressible_navier_stokes/spatial_discretization/calculators/wall_shear_stress_calculator.cpp
richardschu: I changed to std::set, thanks, that spared me the std::map introduced here and from the user side its identical!

include/exadg/incompressible_navier_stokes/spatial_discretization/calculators/wall_shear_stress_calculator.cpp
richardschu: I did that for now. But one could also setup FE_DGSimplex and FE_DGQ and use them depending on `is_hyper_cube`. Problem is: I do not want to introduce functionality I have not tested, I do not want to introduce a temporary test now, and simplices are not yet ready for iNS.
Once we have that, we can refactor without a big hastle and test in one go!

include/exadg/incompressible_navier_stokes/spatial_discretization/calculators/wall_shear_stress_calculator.cpp
richardschu: I went for default all based on `write_wall_shear_stress`, or selected ones when IDs are specified, is that ok for you as well?

include/exadg/incompressible_navier_stokes/postprocessor/output_generator.cpp
richardschu: Not passing in `velocity` and `pressure` (and hence not needing the two  dof_handlers) and reusing the same `write_output` sounds quite elegant. This calls for some `Calculators` in `navier_stokes_calculators` then, right? I think adding the fields and then having surface and volume output separate in `utilities` or `postprocessor` would be good.

nfehn: calculators should not be necessary, because velocity/pressure are the primary variables for IncNS. One just needs to add velocity/pressure to the vector of solution fields.

include/exadg/incompressible_navier_stokes/spatial_discretization/calculators/wall_shear_stress_calculator.cpp
nfehn: I would skip `==true`

richardschu: passt, then I will adopt that style from now on!

include/exadg/incompressible_navier_stokes/spatial_discretization/calculators/wall_shear_stress_calculator.cpp
richardschu: Ok, I did that now.
So what we do in `initialize` is generate a face index to cell index map and later on, we can use this map to get indices from the face to the DoFs on the cell. 

This assumes that face orientations do not change, and are in fact similar to the first cell encountered on each subdomain. Since I do not know enough regarding changes in orientation or when they might occur (maybe even by accident?), I put the initial node matching into the cell loop to be re-done for each cell all the time in debug mode and Assert if it does not match.

In release mode, I skip this match on each face integration point to the corresponding cell integration point.

include/exadg/incompressible_navier_stokes/spatial_discretization/calculators/wall_shear_stress_calculator.cpp
```diff
@@ -0,0 +1,233 @@
...
+                dynamic_viscosity * (velocity_gradients[i] + transpose(velocity_gradients[i])) *
+                normal;
+              dealii::Tensor<1, dim> const wall_shear_stress =
+                S_times_n - scalar_product((S_times_n), normal) * normal;
+              for(unsigned int d = 0; d < dim; ++d)
+              {
+                if(src.get_partitioner()->in_local_range(
+                     dof_indices[matching_idx + d * dofs_per_component]))
+                  dst(dof_indices[matching_idx + d * dofs_per_component]) = wall_shear_stress[d];
+              }
```
kronbichler, line 215: This will only work for discontinuous elements such as `FESystem(FE_DGQ)` or `FESystem(FE_SimplexDGP)`, which is something you cannot control here. The reason is that deal.II orders unknowns entity-by-entity, i.e., first all vertices (in their order within a cell), then lines, then faces (aka 'quads'), then volumes (aka 'hexes', this naming also applies to tetrahedra). For discontinuous elements, this ends up being the expected ordering, because all DoFs are associated with the volume, but not for continuous elements where they (necessarily) live on vertices, lines, etc. - so overall you make assumptions on how unknowns are ordered.

Without full knowledge of the code (and all comments, sorry for that), I think that this is a fundamental issue on how this code is structured. As this is post-processing of a quantity that only lives on the boundary, the first question would be whether `DataOutFaces` could have been used (generating a different file). If output in the same file would be desired, the approach I would have taken is to solve a least-squares system that tries to make the polynomial solution representation of a discontinuous solution space (irrespective the underlying `FiniteElement`) **directly in the output module**. This means that I would intuitively not create a global vector first and then write it somewhere, but use the `DataPostprocessor` facilities of `deal.II` for that, see https://www.dealii.org/developer/doxygen/deal.II/classDataPostprocessor.html
Even though there might be some extensions (or performance optimizations) necessary in that module, the overall feeling is that we should not duplicate such functionality in ExaDG unless there are good reasons.

include/exadg/incompressible_navier_stokes/spatial_discretization/calculators/wall_shear_stress_calculator.cpp
necioglu: Maybe instead of the Assert in the beginning, we could put the Assert in the location where the code will fail, so that it is easier to extend the code in the future. Just an opinion, it is also fine if it stays at the beginning. I just wanted to jump into the discussion after seeing the word simplex.

include/exadg/incompressible_navier_stokes/spatial_discretization/calculators/wall_shear_stress_calculator.cpp
```diff
@@ -0,0 +1,233 @@
...
+                                           dealii::update_quadrature_points |
+                                             dealii::update_gradients |
+                                             dealii::update_normal_vectors);
+  dealii::FEValues<dim>     fe_values(*mapping,
+                                  fe_system,
+                                  quadrature,
+                                  dealii::update_quadrature_points);
+#else
+  dealii::FEFaceValues<dim> fe_face_values(
+    *mapping, fe_system, face_quadrature, dealii::update_gradients | dealii::update_normal_vectors);
```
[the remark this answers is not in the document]
richardschu: This was not leading to an error, since I only used 'FE_DGQ', I did not test with other element types, which is of course very restrictive - I will address this in the other comment section.

include/exadg/incompressible_navier_stokes/spatial_discretization/calculators/wall_shear_stress_calculator.cpp
richardschu: I think it might be good to work with `BoundaryMassOperator` proposed as part of #555.
will look at this at some point again :)

PR #305, opened 2023-01-20
introduce simplex meshes for ExaDG::Structure

nfehn:
in analogy to what we did for ExaDG::Poisson ...

@necioglu @peterrum FYI

closes #307

PR #351, opened 2023-02-23
Simplex support for incompressible Navier-Stokes

necioglu:
This PR introduces simplex elements to incompressible Navier-Stokes. As an application example, flow past cylinder (FPC) is chosen. The following is done:

- The spatial operator is adapted to work with simplices. The multigrid momentum and the multigrid projection preconditioners are also adjusted. ~~However, not all preconditiers work with simplices, for now. (see Asserts in `parameters.cpp`) The reason behind is that the inverse mass operator is not implemented for simplices in dealii. I will adress this in ExaDG in a seperate PR.~~ 
- Flow past cylinder is used as an example application working with simplices.
- For now, only BDFCoupled is supported, others are asserted.
- In the current master, the square test case of the FPC application doesn't work because of a mismatch in mesh smoothing.  This is addressed in this PR in `grid.h` of the FPC application.

### DIFF
--- a/applications/incompressible_navier_stokes/flow_past_cylinder/application.h
+++ b/applications/incompressible_navier_stokes/flow_past_cylinder/application.h
@@ -195,6 +195,8 @@ private:
     this->param.grid.triangulation_type = TriangulationType::Distributed;
     this->param.grid.mapping_degree     = this->param.degree_u;
     this->param.degree_p                = DegreePressure::MixedOrder;
+    this->param.grid.element_type       = ElementType::Hypercube;
+    this->param.grid.multigrid          = MultigridVariant::LocalSmoothing;
 
     // convective term
     if(this->param.formulation_convective_term == FormulationConvectiveTerm::DivergenceFormulation)
--- a/applications/incompressible_navier_stokes/flow_past_cylinder/application.h
+++ b/applications/incompressible_navier_stokes/flow_past_cylinder/application.h
@@ -331,7 +331,10 @@ private:
           unsigned int const                                       global_refinements,
           std::vector<unsigned int> const &                        vector_local_refinements) {
         (void)periodic_face_pairs;
-        create_coarse_grid<dim>(tria, this->grid->periodic_face_pairs, cylinder_type_string);
+        create_coarse_grid<dim>(tria,
+                                this->grid->periodic_face_pairs,
+                                cylinder_type_string,
+                                this->param.grid.element_type);
 
         if(vector_local_refinements.size() > 0)
           refine_local(tria, vector_local_refinements);
--- a/applications/incompressible_navier_stokes/flow_past_cylinder/include/grid.h
+++ b/applications/incompressible_navier_stokes/flow_past_cylinder/include/grid.h
@@ -1019,12 +1019,12 @@ double const h_x_1 = D / nele_x_middle_middle;
 double const h_x_0 = (X_1 - X_0) / nele_x_left;
 
 void
-do_create_coarse_triangulation(dealii::Triangulation<2> &        tria,
-                               std::vector<unsigned int> const & repetitions,
-                               dealii::Point<2> const &          p1,
-                               dealii::Point<2> const &          p2,
-                               ElementType const &               element_type,
-                               bool const                        colorize = false)
+create_subdivided_hyper_rectangle(dealii::Triangulation<2> &        tria,
+                                  std::vector<unsigned int> const & repetitions,
+                                  dealii::Point<2> const &          p1,
+                                  dealii::Point<2> const &          p2,
+                                  ElementType const &               element_type,
+                                  bool const                        colorize = false)
 {
   if(element_type == ElementType::Hypercube)
   {
@@ -1053,7 +1053,7 @@ create_trapezoid(dealii::Triangulation<2> & tria,
 
   dealii::Point<2> x_1 = x_0 + dealii::Point<2>(length, height);
 
-  do_create_coarse_triangulation(tmp, ref, x_0, x_1, element_type, false);
+  create_subdivided_hyper_rectangle(tmp, ref, x_0, x_1, element_type, false);
 
   for(dealii::Triangulation<2>::vertex_iterator v = tmp.begin_vertex(); v != tmp.end_vertex(); ++v)
   {
@@ -1155,52 +1155,52 @@ do_create_coarse_triangulation(dealii::Triangulation<2> & triangulation,
   std::vector<unsigned int> ref_middle_left  = {nele_x_middle_left, nele_y_middle};
 
   // left part
-  do_create_coarse_triangulation(left_bottom,
-                                 ref_left_bottom,
-                                 dealii::Point<2>(X_0, Y_0),
-                                 dealii::Point<2>(X_1, Y_1),
-                                 element_type,
-                                 false);
+  create_subdivided_hyper_rectangle(left_bottom,
+                                    ref_left_bottom,
+                                    dealii::Point<2>(X_0, Y_0),
+                                    dealii::Point<2>(X_1, Y_1),
+                                    element_type,
+                                    false);
 
-  do_create_coarse_triangulation(left_middle,
-                                 ref_left_middle,
-                                 dealii::Point<2>(X_0, Y_1),
-                                 dealii::Point<2>(X_1, Y_2),
-                                 element_type,
-                                 false);
+  create_subdivided_hyper_rectangle(left_middle,
+                                    ref_left_middle,
+                                    dealii::Point<2>(X_0, Y_1),
+                                    dealii::Point<2>(X_1, Y_2),
+                                    element_type,
+                                    false);
 
-  do_create_coarse_triangulation(left_top,
-                                 ref_left_top,
-                                 dealii::Point<2>(X_0, Y_2),
-                                 dealii::Point<2>(X_1, H),
-                                 element_type,
-                                 false);
+  create_subdivided_hyper_rectangle(left_top,
+                                    ref_left_top,
+                                    dealii::Point<2>(X_0, Y_2),
+                                    dealii::Point<2>(X_1, H),
+                                    element_type,
+                                    false);
 
   // merge left triangulations
   dealii::GridGenerator::merge_triangulations(left_bottom, left_middle, tmp);
   dealii::GridGenerator::merge_triangulations(tmp, left_top, left);
 
   // right part
-  do_create_coarse_triangulation(right_bottom,
-                                 ref_right_bottom,
-                                 dealii::Point<2>(X_2, Y_0),
-                                 dealii::Point<2>(L, Y_1),
-                                 element_type,
-                                 false);
+  create_subdivided_hyper_rectangle(right_bottom,
+                                    ref_right_bottom,
+                                    dealii::Point<2>(X_2, Y_0),
+                                    dealii::Point<2>(L, Y_1),
+                                    element_type,
+                                    false);
 
-  do_create_coarse_triangulation(right_middle,
-                                 ref_right_middle,
-                                 dealii::Point<2>(X_2, Y_1),
-                                 dealii::Point<2>(L, Y_2),
-                                 element_type,
-                                 false);
+  create_subdivided_hyper_rectangle(right_middle,
+                                    ref_right_middle,
+                                    dealii::Point<2>(X_2, Y_1),
+                                    dealii::Point<2>(L, Y_2),
+                                    element_type,
+                                    false);
 
-  do_create_coarse_triangulation(right_top,
-                                 ref_right_top,
-                                 dealii::Point<2>(X_2, Y_2),
-                                 dealii::Point<2>(L, H),
-                                 element_type,
-                                 false);
+  create_subdivided_hyper_rectangle(right_top,
+                                    ref_right_top,
+                                    dealii::Point<2>(X_2, Y_2),
+                                    dealii::Point<2>(L, H),
+                                    element_type,
+                                    false);
 
   // merge right triangulations
   dealii::GridGenerator::merge_triangulations(right_bottom, right_middle, tmp);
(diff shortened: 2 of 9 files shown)


include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
```diff
@@ -147,10 +145,14 @@ SpatialOperatorBase<dim, Number>::fill_matrix_free_data(
...
-  matrix_free_data.insert_quadrature(dealii::QGauss<1>(param.get_degree_p(param.degree_u) + 1),
+  matrix_free_data.insert_quadrature(fe_u->reference_cell().template get_gauss_type_quadrature<dim>(
+                                       param.degree_u + 1),
+                                     field + quad_index_u);
+  matrix_free_data.insert_quadrature(fe_p->reference_cell().template get_gauss_type_quadrature<dim>(
+                                       param.get_degree_p(param.degree_u) + 1),
                                      field + quad_index_p);
-  matrix_free_data.insert_quadrature(dealii::QGauss<1>(param.degree_u + (param.degree_u + 2) / 2),
+  matrix_free_data.insert_quadrature(fe_u->reference_cell().template get_gauss_type_quadrature<dim>(
+                                       param.degree_u + (param.degree_u + 2) / 2),
```
necioglu, line 155: Increased the `dim` of the quadrature from 1 to dim, keeping this in mind: https://github.com/exadg/exadg/pull/305#discussion_r1082353641

include/exadg/incompressible_navier_stokes/user_interface/parameters.cpp
```diff
@@ -540,6 +540,26 @@ Parameters::check(dealii::ConditionalOStream const & pcout) const
...
+                                   " is not supported with simplex elements."));
+
+    AssertThrow(not(preconditioner_velocity_block == MomentumPreconditioner::InverseMassMatrix),
+                dealii::ExcMessage(enum_to_string(preconditioner_velocity_block) +
+                                   " is not supported for simplex elements."));
+
+    AssertThrow(not(preconditioner_projection == PreconditionerProjection::InverseMassMatrix),
+                dealii::ExcMessage(enum_to_string(preconditioner_projection) +
+                                   " is not supported for simplex elements."));
+  }
```
necioglu, line 562: I have asserted everything, which don't work with simplices, here.  I couldn't decide if I should go for a design similar to `poisson/application.h `, where we have a simplex configuration and a hypercube configuration.

applications/incompressible_navier_stokes/flow_past_cylinder/application.h
```diff
@@ -391,9 +396,9 @@ class Application : public ApplicationBase<dim, Number>
     pp_data.output_data.time_control_data.is_active        = this->output_parameters.write;
     pp_data.output_data.time_control_data.start_time       = start_time;
     pp_data.output_data.time_control_data.trigger_interval = (end_time - start_time) / 20.0;
-    pp_data.output_data.directory          = this->output_parameters.directory + "vtu/";
-    pp_data.output_data.filename           = name;
-    pp_data.output_data.write_divergence   = true;
+    pp_data.output_data.directory        = this->output_parameters.directory + "vtu/";
+    pp_data.output_data.filename         = name;
+    pp_data.output_data.write_divergence = this->param.grid.element_type == ElementType::Hypercube;
```
peterrum, line 401: Maybe create an issue as reminder?

applications/incompressible_navier_stokes/flow_past_cylinder/include/grid.h
```diff
@@ -1391,18 +1484,38 @@ void
...
 
   switch(cylinder_type)
   {
     case circular:
+    {
+      AssertThrow(
+        element_type == ElementType::Hypercube,
+        dealii::ExcMessage(
+          "You are trying to create a grid with simplex elements for the circular cylinder case, "
+          "which is currently not supported."));
```
peterrum, line 1509: Why is this not working?

peterrum (commented):
Looks good :+1:

applications/incompressible_navier_stokes/flow_past_cylinder/include/grid.h
necioglu: I have tried this using `convert_hypercube_to_simplex_mesh()`, since the cylinder case has `hyper_shells`. But I had problems with manifold ids not matching. I didn't invest more time in this and moved on to reading an external grid for the circular case, which works fine.

applications/incompressible_navier_stokes/flow_past_cylinder/application.h
necioglu: Will do. This is also related to the invermassoperator not working for simplices.

include/exadg/grid/grid_data.h
```diff
@@ -44,6 +44,9 @@ struct GridData
   void
   check() const
   {
+    if(element_type == ElementType::Simplex)
+      AssertThrow(not(triangulation_type == TriangulationType::Distributed),
+                  dealii::ExcMessage("Can't use simplex elements with Distributed triangulation."))
```
nfehn, line 49: I think this should be suggested as a separate PR.

nfehn:
Could you please also comment on dual splitting / pressure-correction in the description of this PR?

Much of this PR seems to be related to the flow past cylinder test case. I would like to change the order of extending ExaDG in the sense that we first address the inverse mass matrix problem, then extend certain incompressible Navier-Stokes classes towards simplices (the present PR), and finally add an application example. The present PR should also add asserts for the postprocessing routines that will not work in combination with simplices. Moreover, I think the specific functions for dual splitting / pressure-correction not working for simplices should contain asserts (apart from the assert with BDFCoupledSolution in `Parameters`).

include/exadg/incompressible_navier_stokes/user_interface/parameters.cpp
nfehn: I think the asserts are also needed where the actual implementations happen. How can a developer be safe to remove such an assert in `Parameters` if this is the only assert on a certain topic/functionality?

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
```diff
@@ -317,24 +340,9 @@ SpatialOperatorBase<dim, Number>::distribute_dofs()
...
-      dealii::Utilities::pow(param.degree_u, dim - 1) * (param.degree_u + 1) * dim;
-  }
-  else
-  {
-    AssertThrow(false, dealii::ExcMessage("FE not implemented."));
-  }
-
-  unsigned int const ndofs_per_cell_pressure =
-    dealii::Utilities::pow(param.get_degree_p(param.degree_u) + 1, dim);
+  unsigned int ndofs_per_cell_velocity = fe_u->n_dofs_per_cell();
```
necioglu, line 343: This might not work for HDIV, see: https://github.com/exadg/exadg/pull/349#discussion_r1129664955

kronbichler: This should still work. The other comment was about using something like `fe_u->base_element(0)` to return the element of a scalar component, as all components are within a single monolithic system in Hdiv. (There is a separation in `FE_RaviartThomasNodal` from the polynomial point of view, but the final polynomials cover the whole space and the Piola transform couples them together.)

applications/incompressible_navier_stokes/flow_past_cylinder/include/grid.h
```diff
@@ -1018,20 +1018,42 @@ double const h_x_2 = (L - X_2) / nele_x_right;
 double const h_x_1 = D / nele_x_middle_middle;
 double const h_x_0 = (X_1 - X_0) / nele_x_left;
 
+void
+do_create_coarse_triangulation(dealii::Triangulation<2> &        tria,
```
nfehn, line 1022: The code is nicely structured when providing a separate function for creating a hyper-rectangle. 

Could you give the function a more specific name? e.g. `ExaDG::create_subdivided_hyper_rectangle()`

This function could be implemented at a more central place in ExaDG, since this is potentially relevant for many other applications as well, see e.g. the function `ExaDG::create_periodic_box()` in the `grid` folder.

include/exadg/incompressible_navier_stokes/preconditioners/multigrid_preconditioner_momentum.cpp
```diff
@@ -142,8 +142,23 @@ MultigridPreconditioner<dim, Number>::fill_matrix_free_data(
...
+  {
+    matrix_free_data.insert_quadrature(dealii::QGaussSimplex<dim>(this->level_info[level].degree() +
+                                                                  1),
+                                       "std_quadrature");
+  }
+  else
+  {
+    AssertThrow(false,
+                dealii::ExcMessage("Only pure hypercube or pure simplex meshes are implemented for "
+                                   "MomentumPreconditioner::Multigrid."));
```
nfehn, line 160: The notion `MomentumPreconditione::Multigrid` is not known by this class (it is introduced in an outer software layer like `IncNS::Parameters`). Since this class can be used independently of the parameter `MomentumPreconditioner`, I think this name should not appear here in the error message. I would instead write the class name, `IncNS::MultigridPreconditioner`.

include/exadg/incompressible_navier_stokes/preconditioners/multigrid_preconditioner_projection.cpp
```diff
@@ -113,8 +113,23 @@ MultigridPreconditionerProjection<dim, Number>::fill_matrix_free_data(
...
+  {
+    matrix_free_data.insert_quadrature(dealii::QGaussSimplex<dim>(this->level_info[level].degree() +
+                                                                  1),
+                                       "std_quadrature");
+  }
+  else
+  {
+    AssertThrow(false,
+                dealii::ExcMessage("Only pure hypercube or pure simplex meshes are implemented for "
+                                   "ProjectionPreconditioner::Multigrid."));
```
nfehn, line 131: as above

applications/incompressible_navier_stokes/flow_past_cylinder/application.h
nfehn: is this already resolved or issued,  @peterrum @necioglu ?

nfehn:
The changes look overall good in my opinion.

For ongoing work, it would help if you could provide information where dual splitting and pressure-correction are not working for simplex (in case you looked into this).

This PR shows simplex elements for the FPC test case. The "problem" with this test case in the current state of development (realizing a new method (simplex elements) for the IncNS module) is that we cannot really measure errors for this application. Did you run convergence tests using simplex elements for any of the IncNS application having an analytical solution (e.g. vortex)?

applications/incompressible_navier_stokes/flow_past_cylinder/application.h
necioglu: Inversemassoperator is now available for simplices, hence I changed the statement back to `...= true;`.

applications/incompressible_navier_stokes/flow_past_cylinder/include/grid.h
necioglu: > This function could be implemented at a more central place in ExaDG, since this is potentially relevant for many other applications as well, see e.g. the function `ExaDG::create_periodic_box()` in the `grid` folder.

I agree. I will rename it to the suggested name and leave it here for now. Because although the renaming is an improvement, I don't think it is the best name, since it is not clear from the name (at least for me) that this function is capable of creating both simplex and hyper-rectangle meshes.

include/exadg/incompressible_navier_stokes/preconditioners/multigrid_preconditioner_momentum.cpp
necioglu: Makes sense. Just for context, this assert was my inspiration:
https://github.com/exadg/exadg/blob/0f6563a5139715382f117cfa731f03e0d820353e/include/exadg/structure/preconditioners/multigrid_preconditioner.cpp#L141-L144

I am not familiar with that part of the code, thus I don't know if the wording of the assert is correct there. Maybe it needs to be reworded, as well?

necioglu:
> The changes look overall good in my opinion.

Thanks for the review. I have addressed your suggestions. 

> For ongoing work, it would help if you could provide information where dual splitting and pressure-correction are not working for simplex (in case you looked into this).

I have moved the asserts to the locations where dual-splitting and pressure-correction are not working for simplex. However, this should not imply that these are the only locations. These are the first occurrences. Therefore, a "global" assert in `parameters.cpp` may be the better option. Switching back, i.e., removing the last commit, would be easy for me.

necioglu:
> Did you run convergence tests using simplex elements for any of the IncNS application having an analytical solution (e.g. vortex)?

I only worked with applications with a reference solution: FPC (circular) with an external mesh and 3D TGV.

include/exadg/incompressible_navier_stokes/preconditioners/multigrid_preconditioner_momentum.cpp
nfehn: `Structure::MultigridPreconditioner` in the assert that you mention above fits to my comment above. In the assert, we mention the class that implements the functionality (`Structure::MultigridPreconditioner` or `IncNS::MultigridPreconditioner`), not the parameter in `MyModule::Parameters`.

nfehn:
> These are the first occurrences. Therefore, a "global" assert in parameters.cpp may be the better option.

A global assert in the parameters.cpp file still makes sense in my opinion (see also my comment above https://github.com/exadg/exadg/pull/351#issuecomment-1456170345).

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_dual_splitting.cpp
```diff
@@ -801,7 +801,10 @@ OperatorDualSplitting<dim, Number>::local_interpolate_velocity_dirichlet_bc_boun
   VectorType const &,
   Range const & face_range) const
 {
-  unsigned int const dof_index  = this->get_dof_index_velocity();
+  unsigned int const dof_index = this->get_dof_index_velocity();
+  AssertThrow(
+    matrix_free.get_dof_handler(dof_index).get_triangulation().all_reference_cells_are_hyper_cube(),
+    dealii::ExcMessage("This function is not supported for simplices."));
```
nfehn, line 807: Thanks for adding asserts here.

It would be slightly more precise to say that the function is only implemented for pure hypercube elements currently. There might be more variants in the future and, then, the current error message might be unclear.

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_pressure_correction.cpp
```diff
@@ -482,7 +482,10 @@ OperatorPressureCorrection<dim, Number>::local_interpolate_pressure_dirichlet_bc
   VectorType const &,
   Range const & face_range) const
 {
-  unsigned int const dof_index  = this->get_dof_index_pressure();
+  unsigned int const dof_index = this->get_dof_index_pressure();
+  AssertThrow(
+    matrix_free.get_dof_handler(dof_index).get_triangulation().all_reference_cells_are_hyper_cube(),
+    dealii::ExcMessage("This function is not supported for simplices."));
```
nfehn, line 488: as here https://github.com/exadg/exadg/pull/351/files#r1187174043

include/exadg/incompressible_navier_stokes/preconditioners/multigrid_preconditioner_momentum.cpp
```diff
@@ -142,8 +142,23 @@ MultigridPreconditioner<dim, Number>::fill_matrix_free_data(
...
+    matrix_free_data.insert_quadrature(dealii::QGaussSimplex<dim>(this->level_info[level].degree() +
+                                                                  1),
+                                       "std_quadrature");
+  }
+  else
+  {
+    AssertThrow(false,
+                dealii::ExcMessage("Only pure hypercube or pure simplex meshes are implemented for "
+                                   "IncNS::Multigrid."));
+  }
```
nfehn, line 161: In `IncNS::SpatialOperatorBase`, you have used `get_gauss_type_quadrature<dim>()`, which seems to handle both hypercube and simplex elements. Could you comment why you did not use this here?

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
```diff
@@ -147,10 +145,14 @@ SpatialOperatorBase<dim, Number>::fill_matrix_free_data(
   matrix_free_data.insert_constraint(&constraint_u_scalar, field + dof_index_u_scalar);
 
   // quadrature
-  matrix_free_data.insert_quadrature(dealii::QGauss<1>(param.degree_u + 1), field + quad_index_u);
-  matrix_free_data.insert_quadrature(dealii::QGauss<1>(param.get_degree_p(param.degree_u) + 1),
+  matrix_free_data.insert_quadrature(fe_u->reference_cell().template get_gauss_type_quadrature<dim>(
```
nfehn, line 148: I think it would be good to have one common style in ExaDG, see for comparison the `Poisson` module

https://github.com/exadg/exadg/blob/9aba12677bb4025a2608eeff5b5f0c396932527c/include/exadg/poisson/spatial_discretization/operator.cpp#L235-L241

and the `Structure` module

https://github.com/exadg/exadg/blob/9aba12677bb4025a2608eeff5b5f0c396932527c/include/exadg/structure/spatial_discretization/operator.cpp#L317-L323

include/exadg/incompressible_navier_stokes/preconditioners/multigrid_preconditioner_momentum.cpp
necioglu: There was no reason. I changed the other occurrence of `matrix_free_data.insert_quadrature(...)` to this `if-else` logic as well, to keep the design consistent.

nfehn: What is `IncNS::Multigrid`?

The `get_gauss_type_quadrature<dim>()` is actually more elegant in my opinion, so I will open an issue so that we have a reminder to change this in the future.

include/exadg/incompressible_navier_stokes/preconditioners/multigrid_preconditioner_projection.cpp
```diff
@@ -113,8 +113,23 @@ MultigridPreconditionerProjection<dim, Number>::fill_matrix_free_data(
...
+  {
+    matrix_free_data.insert_quadrature(dealii::QGaussSimplex<dim>(this->level_info[level].degree() +
+                                                                  1),
+                                       "std_quadrature");
+  }
+  else
+  {
+    AssertThrow(false,
+                dealii::ExcMessage("Only pure hypercube or pure simplex meshes are implemented for "
+                                   "IncNS::Multigrid."));
```
nfehn, line 131: as above https://github.com/exadg/exadg/pull/351/files#r1188565050

include/exadg/incompressible_navier_stokes/preconditioners/multigrid_preconditioner_momentum.cpp
nfehn: see issue #430

include/exadg/incompressible_navier_stokes/preconditioners/multigrid_preconditioner_projection.cpp
nfehn: can you change the name to the precise name of the class?

nfehn:
Comment https://github.com/exadg/exadg/pull/351#issuecomment-1537938139 seems to be open

nfehn:
looks ready-to-merge from my side. Let's wait for a final statement by  @bergbauer.

nfehn (approved):

PR #352, opened 2023-02-24
Omit creating the finest level again for Fullydistributed global coarsening multigrid

### LINKED ISSUE #329: Why re-creating the finest triangulation in case of global-coarsening MG?
See https://github.com/exadg/exadg/blob/4cd65d5c8973e4f3da199f3906155571b4c66191/include/exadg/grid/grid_utilities.h#L187-L192

I think on the finest mesh we do not have to re-create the triangulation via `lambda_create_triangulation`, but could instead let the first element of the vector point to the fine triangulation object created before.

necioglu:
Closes #329. 

Similar design to: https://github.com/dealii/dealii/blob/8447f2a1b8d4b12d7481f6dbee23db2b79d03502/include/deal.II/multigrid/mg_transfer_global_coarsening.templates.h#L2393-L2396

### DIFF
--- a/include/exadg/grid/grid_utilities.h
+++ b/include/exadg/grid/grid_utilities.h
@@ -257,8 +257,16 @@ create_coarse_triangulations(
     unsigned int              level        = fine_triangulation.n_global_levels() - 1;
     std::vector<unsigned int> refine_local = vector_local_refinements;
 
+    // make the last entry of the coarse_triangulations point to the fine_triangulation
+    coarse_triangulations[level].reset(&fine_triangulation, [](auto *) {
+      // empty deleter, since fine_triangulation is an external field
+      // and its destructor is called somewhere else
+    });
+
+    level--;
+
     // undo global refinements
-    for(int refine_global = data.n_refine_global; refine_global >= 0; --refine_global)
+    for(int refine_global = data.n_refine_global - 1; refine_global >= 0; --refine_global)
     {
       coarse_triangulations[level] = lambda_create_level_triangulation(refine_global, refine_local);
 


peterrum (approved):

include/exadg/grid/grid_utilities.h
```diff
@@ -263,8 +263,16 @@ create_coarse_triangulations(
...
+    coarse_triangulations[level].reset(&fine_triangulation, [](auto *) {
+      // empty deleter, since fine_triangulation is an external field
+      // and its destructor is called somewhere else
+    });
+
+    level--;
+
     // undo global refinements
-    for(int refine_global = data.n_refine_global; refine_global >= 0; --refine_global)
+    for(int refine_global = data.n_refine_global - 1; refine_global >= 0; --refine_global)
```
nfehn, line 275: `n_refine_global` might be zero, and the data type of  `n_refine_global` should be `unsigned int`, isn't it? Then, I am actually not sure whether this works correctly.

nfehn: Did you test for `n_refine_global = 0`? Could you provide an explanation of how the int-casts will be done here?

necioglu: I double checked just now and did a quick research on `int=unsigned int-unsigned int` operations.

The code works fine for `n_refine_global=0`: The `coarse_triangulations` vector has only one element, which is the fine triangulation.

The reason behind this is for  `data.n_refine_global=0` the loop variable `refine_global`, which is an `int`, becomes `-1` and we don't enter the loop at all. So, regarding functionality, there are no problems for the case `data.n_refine_global=0`. 

If I change the loop variable to `unsigned int refine_global`, then yes, the behavior becomes undefined and `refine_global` becomes a random positive number for negative results.

nfehn: ok. I still wonder whether it would make sense to write the cast to int explicitly (for clarity).

necioglu: Yes, or I could change `refine_global` to `unsigned int` and make changes accordingly. 

I didn't notice that it was `int` until you mentioned it. I agree that it is better to keep the loop variable as `unsigned int`.

kronbichler: I agree, keeping the variable `unsigned` is clearer, so I suggest you use the solution proposed last @necioglu . The case `unsigned int(0) - 1` is well-defined due to the implementation of signed integer numbers with two's complement https://en.wikipedia.org/wiki/Two%27s_complement, where `-1=4,294,967,295u`.

nfehn: I would have preferred the previous version (I mean the logic of the for-loop, ignoring data types) with an explicit cast to `int`. From the  current code version it is more difficult to see that we start the loop with the second finest level, not the finest level. The implementation is not longer "intuitive" in my opinion.

necioglu: In my opinion, we should use the cast to `int`, if it is needed to ensure functionality. In this case, since the operation is well-defined, we don't need it. But of course, I can understand that it might reduce confusion. 

I am fine with all the proposed solutions. But I don't know how to weigh in the different arguments. Maybe adding a comment would make the loop clearer?

necioglu: @nfehn @kronbichler How should I proceed here?

kronbichler (approved):
I think it is up to personal preference of which version to prefer. I would say the author of the code, i.e., @necioglu, should have the final word in such a case.

And just to make things complicated, let me express a third option what I would have done (without say that this should be done, I see my role here to express different options and let the author decide on what seems best). Since there are no data dependencies as far as I can see, I would use a loop starting from the lowest level (making sure `level` and `refine_global` match to what we intend to do), which avoids the issue with signed-ness of integers from a decreasing loop. I also think it requires attention to understand the role of the two variables `refine_global` and `level` and their respective role, so there is some complication here also from before. Likely not easy to fulfill the requirements of simplicity along all directions.

include/exadg/grid/grid_utilities.h
nfehn: Yes, this was my intention. It's just about being more explicit. As @necioglu has written, it worked correctly without such a cast. 

Overall, I agree that we need to come to an end in final time. I can live with different potential solutions.

nfehn:
In this particular case, I would suggest to go back to the original version of the author, because the solution we currently have originated from a mis-understanding of my review comment, as far as I see it.

I am also open to @kronbichler's suggestion. If we go for this variant, I want to express my wish to do it consistently with local refinements, for the sake of understanding/readability of the code. It might be that I did not correctly understand the comment, though.

nfehn (approved):

PR #353, opened 2023-02-26
Generalize variable coefficients to arbitrary types

### LINKED ISSUE #128: Provide quadrature point index in VariableCoefficients class

### LINKED ISSUE #289: Generalize VariableCoefficients to tensor-valued quantities
The data structure `VariableCoefficients` currently only covers scalar quantities. The functionality of this data structure should be extended to also cover tensor-valued quantities. Applications could be multiphysics-coupling (volume-coupling) between different fields operating on different triangulations (requiring a transfer of solution fields between the triangulations), or more complex structural mechanics problems with more complex material laws requiring e.g. history variables.

btemuer:
Especially in problems with complex (multi-)physics, scalar as well as tensor-valued (generally rank 2) variable coefficients play a significant role. Currently, the `VariableCoefficients` class provides the basic implementation for scalar-valued coefficients as tables with cell and quadrature indices.

This PR:
- generalizes the `VariableCoefficients` class (and the related `VariableCoefficientsCell` class) to hold coefficients of arbitrary type like tensors of different ranks, hence closes #289,
- fixes the calculation of the number of quadrature points by getting the `quad_index` alongside the `MatrixFree` object and extracting the quadrature formula from it, hence closes #128.

Also:
- removes the `n_q_points_1d` variable piped through in the structure module, as it is not needed for the number of quadrature points,
- passes in the `quad_index` into the viscous kernel for the purpose of getting the number of quadrature points from the `MatrixFree` object inside the `VariableCoefficients` class, 
- does not touch or modify the `cell_based_face_loop` type of variable coefficients which was and still is marked `TODO`.

This PR constitutes one of the first steps towards widely enabling variable coefficients in ExaDG. I will try to bring in more issues and PRs soon on this topic.

### DIFF
--- a/include/exadg/incompressible_navier_stokes/spatial_discretization/operators/viscous_operator.h
+++ b/include/exadg/incompressible_navier_stokes/spatial_discretization/operators/viscous_operator.h
@@ -63,7 +63,8 @@ public:
   void
   reinit(dealii::MatrixFree<dim, Number> const & matrix_free,
          ViscousKernelData const &               data,
-         unsigned int const                      dof_index)
+         unsigned int const                      dof_index,
+         unsigned int const                      quad_index)
   {
     this->data = data;
 
@@ -77,7 +78,7 @@ public:
     if(data.viscosity_is_variable)
     {
       // allocate vectors for variable coefficients and initialize with constant viscosity
-      viscosity_coefficients.initialize(matrix_free, degree, data.viscosity);
+      viscosity_coefficients.initialize(matrix_free, quad_index, data.viscosity);
     }
   }
 
@@ -495,7 +496,7 @@ private:
 
   mutable scalar tau;
 
-  VariableCoefficients<dim, Number> viscosity_coefficients;
+  VariableCoefficients<dealii::VectorizedArray<Number>> viscosity_coefficients;
 };
 
 } // namespace Operators
--- a/include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
+++ b/include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
@@ -407,7 +407,7 @@ SpatialOperatorBase<dim, Number>::initialize_operators(std::string const & dof_i
   viscous_kernel->reinit(*matrix_free,
                          viscous_kernel_data,
                          get_dof_index_velocity(),
-                         get_quad_index_velocity_linearized());
+                         get_quad_index_velocity_linear());
 
   dealii::AffineConstraints<Number> constraint_dummy;
   constraint_dummy.close();
--- a/include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
+++ b/include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
@@ -404,7 +404,10 @@ SpatialOperatorBase<dim, Number>::initialize_operators(std::string const & dof_i
   viscous_kernel_data.viscosity_is_variable        = param.use_turbulence_model;
   viscous_kernel_data.variable_normal_vector       = param.neumann_with_variable_normal_vector;
   viscous_kernel = std::make_shared<Operators::ViscousKernel<dim, Number>>();
-  viscous_kernel->reinit(*matrix_free, viscous_kernel_data, get_dof_index_velocity());
+  viscous_kernel->reinit(*matrix_free,
+                         viscous_kernel_data,
+                         get_dof_index_velocity(),
+                         get_quad_index_velocity_linearized());
 
   dealii::AffineConstraints<Number> constraint_dummy;
   constraint_dummy.close();
--- a/include/exadg/operators/variable_coefficients.h
+++ b/include/exadg/operators/variable_coefficients.h
@@ -23,117 +23,111 @@
 
 namespace ExaDG
 {
-template<int dim, typename Number>
+template<typename coefficient_type>
 class VariableCoefficientsCells
 {
-private:
-  typedef dealii::VectorizedArray<Number> scalar;
-
 public:
+  template<int dim, typename Number>
   void
   initialize(dealii::MatrixFree<dim, Number> const & matrix_free,
-             unsigned int const                      degree,
-             Number const &                          constant_coefficient)
+             unsigned int const                      quad_index,
+             coefficient_type const &                constant_coefficient)
   {
-    unsigned int const points_per_cell = dealii::Utilities::pow(degree + 1, dim);
+    reinit(matrix_free, quad_index);
 
-    coefficients_cell.reinit(matrix_free.n_cell_batches(), points_per_cell);
-    coefficients_cell.fill(dealii::make_vectorized_array<Number>(constant_coefficient));
+    fill(constant_coefficient);
   }
 
-  scalar
+  coefficient_type
   get_coefficient(unsigned int const cell, unsigned int const q) const
   {
     return coefficients_cell[cell][q];
   }
 
   void
-  set_coefficient(unsigned int const cell, unsigned int const q, scalar const & value)
+  set_coefficient(unsigned int const cell, unsigned int const q, coefficient_type const & value)
   {
     coefficients_cell[cell][q] = value;
   }
 
 private:
-  // variable coefficients
-  dealii::Table<2, scalar> coefficients_cell;
-};
-
-template<int dim, typename Number>
-class VariableCoefficients
-{
-private:
-  typedef dealii::VectorizedArray<Number> scalar;
-
-public:
+  template<int dim, typename Number>
   void
-  initialize(dealii::MatrixFree<dim, Number> const & matrix_free,
-             unsigned int const                      degree,
-             Number const &                          constant_coefficient)
+  reinit(dealii::MatrixFree<dim, Number> const & matrix_free, unsigned int const quad_index)
   {
-    unsigned int const points_per_cell = dealii::Utilities::pow(degree + 1, dim);
-    unsigned int const points_per_face = dealii::Utilities::pow(degree + 1, dim - 1);
-
-    // cells
-    coefficients_cell.reinit(matrix_free.n_cell_batches(), points_per_cell);
-
-    coefficients_cell.fill(dealii::make_vectorized_array<Number>(constant_coefficient));
-
-    // face-based loops
-    coefficients_face.reinit(matrix_free.n_inner_face_batches() +
-                               matrix_free.n_boundary_face_batches(),
-                             points_per_face);
-
-    coefficients_face.fill(dealii::make_vectorized_array<Number>(constant_coefficient));
-
-    coefficients_face_neighbor.reinit(matrix_free.n_inner_face_batches(), points_per_face);
-
-    coefficients_face_neighbor.fill(dealii::make_vectorized_array<Number>(constant_coefficient));
-
-    // TODO cell-based face loops
-    //    coefficients_face_cell_based.reinit(matrix_free.n_cell_batches()*2*dim,
-    //        points_per_face);
-    //
-    //    coefficients_face_cell_based.fill(dealii::make_vectorized_array<Number>(constant_coefficient));
+    coefficients_cell.reinit(matrix_free.n_cell_batches(), matrix_free.get_n_q_points(quad_index));
   }
 
-  scalar
+  void
+  fill(coefficient_type const & constant_coefficient)
+  {
+    coefficients_cell.fill(constant_coefficient);
+  }
+
+  // variable coefficients
+  dealii::Table<2, coefficient_type> coefficients_cell;
+};
+
+template<typename coefficient_type>
+class VariableCoefficients
+{
+public:
+  template<int dim, typename Number>
+  void
+  initialize(dealii::MatrixFree<dim, Number> const & matrix_free,
+             unsigned int const                      quad_index,
+             coefficient_type const &                constant_coefficient)
+  {
+    reinit(matrix_free, quad_index);
+
+    fill(constant_coefficient);
+  }
+
+  coefficient_type
   get_coefficient_cell(unsigned int const cell, unsigned int const q) const
   {
     return coefficients_cell[cell][q];
   }
 
   void
-  set_coefficient_cell(unsigned int const cell, unsigned int const q, scalar const & value)
+  set_coefficient_cell(unsigned int const       cell,
+                       unsigned int const       q,
+                       coefficient_type const & value)
   {
     coefficients_cell[cell][q] = value;
   }
 
-  scalar
+  coefficient_type
   get_coefficient_face(unsigned int const face, unsigned int const q) const
   {
     return coefficients_face[face][q];
   }
 
   void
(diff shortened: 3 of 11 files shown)


include/exadg/incompressible_navier_stokes/spatial_discretization/operators/viscous_operator.h
```diff
@@ -77,7 +78,12 @@ class ViscousKernel
     if(data.viscosity_is_variable)
     {
       // allocate vectors for variable coefficients and initialize with constant viscosity
-      viscosity_coefficients.initialize(matrix_free, degree, data.viscosity);
+      viscosity_coefficients.initialize(matrix_free.n_cell_batches(),
+                                        matrix_free.n_inner_face_batches(),
+                                        matrix_free.n_boundary_face_batches(),
+                                        matrix_free.get_n_q_points(quad_index),
+                                        matrix_free.get_n_q_points_face(quad_index),
```
peterrum, line 85: Personally, I am not sure if this change is the right way. I am not sure if the old interface is the right one either. If we want to support mixed meshes, cells will have different number of quadrature points so that providing here a fixed value won't work.

include/exadg/operators/variable_coefficients.h
```diff
@@ -23,144 +23,173 @@
 
 namespace ExaDG
 {
-template<int dim, typename Number>
+template<int dim, typename Number, int rank = 0>
 class VariableCoefficientsCells
 {
 private:
-  typedef dealii::VectorizedArray<Number> scalar;
+  using coefficient_t = dealii::Tensor<rank, dim, dealii::VectorizedArray<Number>>;
```
peterrum, line 30: I would make this whole expression the template argument of this class. I.e.,
```cpp
template<typename coefficient_t>
class VariableCoefficientsCells;
```

include/exadg/structure/material/library/st_venant_kirchhoff.h
```diff
@@ -64,7 +64,6 @@ class StVenantKirchhoff : public Material<dim, Number>
   typedef CellIntegrator<dim, dim, Number>                   IntegratorCell;
 
   StVenantKirchhoff(dealii::MatrixFree<dim, Number> const & matrix_free,
-                    unsigned int const                      n_q_points_1d,
```
peterrum, line 67: Nice cleanup :+1:

include/exadg/operators/variable_coefficients.h
btemuer: Yes, makes a whole lot sense, why not 😄 . I'll see to it.

include/exadg/incompressible_navier_stokes/spatial_discretization/operators/viscous_operator.h
btemuer: I see what you mean. The question is if we can keep using `dealii::Table` in that case as well. Maybe keeping the old interface is a good compromise right now to be able to adapt more easily later. What do you think?

include/exadg/structure/material/library/st_venant_kirchhoff.h
btemuer: Thanks for the review!

include/exadg/incompressible_navier_stokes/spatial_discretization/operators/viscous_operator.h
btemuer: I brought back the old interface, of course now passing in `quad_index` alongside the `MatrixFree` object.

include/exadg/operators/variable_coefficients.h
btemuer: This is done now, and `coefficient_t` is now a template parameter alongside `dim` and `Number`. Could have actually removed `dim` and `Number` as you suggested, however this would not work anymore as we reverted the interface to be initialized with a `MatrixFree<dim, Number>` object again.

include/exadg/operators/variable_coefficients.h
```diff
@@ -23,144 +23,160 @@
...
+    coefficients_cell.fill(constant_coefficient);
+  }
+
   // variable coefficients
-  dealii::Table<2, scalar> coefficients_cell;
+  dealii::Table<2, coefficient_t> coefficients_cell;
 };
 
-template<int dim, typename Number>
+template<int dim, typename Number, typename coefficient_t>
```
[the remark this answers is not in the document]
btemuer: Oh yes, you can. Not only `Number`, but `dim` as well. Should I?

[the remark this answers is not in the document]
peterrum: Yes. Even better!

peterrum (approved):

peterrum:
Do you have the indentation script? If not, you can get it via https://github.com/hyperdeal/hyperdeal/blob/master/contrib/utilities/download_clang_format.

btemuer:
> Do you have the indentation script? If not, you can get it via https://github.com/hyperdeal/hyperdeal/blob/master/contrib/utilities/download_clang_format.

Oops, sorry about that. It was the latest change that I did. I already have it.

peterrum (approved):

include/exadg/operators/variable_coefficients.h
```diff
@@ -23,144 +23,164 @@
 
 namespace ExaDG
 {
-template<int dim, typename Number>
+template<typename coefficient_t>
```
nfehn, line 26: I would like to be explicit about `t` (I guess you mean `coefficient_type`; but one could potentially also think of `coefficient_template`, so I think being explicit helps in general for code readability or when talking about the code).

nfehn:
@btemuer it's great that you tackle these issues. Many thanks! The PR looks overall very good from my side.

include/exadg/operators/variable_coefficients.h
kronbichler: In the C++ language, I do not think that the trailing word `_t` can be misunderstood, because it is part of the language specification, see here for one example: https://en.cppreference.com/w/cpp/language/types - it is just that @btemuer likely did not know all the project's habits. And to be explicit on our habits (that are different from the C++ standard), I agree that spelling out `t` does help, so :+1: from my end on the suggested change of `_type`.

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
```diff
@@ -404,7 +404,10 @@ SpatialOperatorBase<dim, Number>::initialize_operators(std::string const & dof_i
   viscous_kernel_data.viscosity_is_variable        = param.use_turbulence_model;
   viscous_kernel_data.variable_normal_vector       = param.neumann_with_variable_normal_vector;
   viscous_kernel = std::make_shared<Operators::ViscousKernel<dim, Number>>();
-  viscous_kernel->reinit(*matrix_free, viscous_kernel_data, get_dof_index_velocity());
+  viscous_kernel->reinit(*matrix_free,
+                         viscous_kernel_data,
+                         get_dof_index_velocity(),
+                         get_quad_index_velocity_linearized());
```
nfehn, line 410: I think this line is problematic. `ViscousKernel` is used for both the `ViscousOperator` (used e.g. to evaluate the nonlinear residual of the incompressible Navier-Stokes equations, using the standard quadrature rule with k+1 points for the viscous term) and the `MomentumOperator` (the (1,1)-block in the linearized system of equations). For the momentum operator (including the convective term), all operators (mass, convection, diffusion) have to use the same quadrature rule, so one can choose different quadrature rules (3/2-overintegration vs. standard quadrature rule with k+1 points). Hence, for the viscous kernel it is in general not clear which quadrature rule to use. 

- In the most general case, viscous kernel would have to support different quadrature rules and `VariableCoefficients` data for these different quadrature rules (if we want to support all the present functionality and parameter choices). This would increase storage requirements, but could be realized easily from an implementational point of view.
-  Another solution could be that the viscous operator always has to use the same quadrature rule as the convective operator in case a fully-implicit, monolithic solution approach is selected to solve the incompressible Navier-Stokes problem (increasing computational costs). This solution would only make sense if we no longer allow to select different quadrature rules for the convective term in the nonlinear residual and in the linearized operator (which is possible in the current code version).
-  Yet another solution could be to remove again the merging of kernels in `MomentumOperator`, which we introduced some time ago to improve performance compared to a sequential matrix-free evaluation with separate loops for mass, convective, viscous terms). Performance would deteriorate in this case. The benefit would be a strictly modular design with independent operators and a unique quadrature rule for the viscous term. Note, however, that such changes would require to rethink the multigrid preconditioner implementation.

I think the best / easiest solution is currently to use `get_quad_index_velocity_linear()` here and to assert that the standard quadrature is used for the linearized momentum operator if `viscous_kernel_data.viscosity_is_variable == true` with a comment that one would otherwise have to initialize `VariableCoefficients` for an additional quadrature rule using over-integration.

Overall, this points to a general design problem in ExaDG. Which class is responsible for quad-indices and the handling of variable coefficients? `Kernel` vs. `Operator`? @btemuer if you are interested, we could discuss in more detail what a good solution to this design problem could be. @richardschu this could also be relevant for your applications with variable viscosity. @kronbichler what do you think? You have accompanied the developments from the very beginning.

richardschu: Yep, finding this as the cause of a related error message when applying the `MomentumOperator` in the nonlinear case  took me more time than it should have ... in a PR I will open today, I put some assertions for the case of variable viscosity, since the default is using `QuadratureRuleLinearization::Overintegration32k`, leading to an index access out of range when evaluating the momentum operator (same as you propose; fits for an intermediate solution until we support one of the options you listed).

Regarding your posed concepts, I would go for option 1 and accept the increased storage efforts in case one wants to solve for the velocity in a nonlinear fashion. I am not sure if I would sacrifice better performance for the possibility of overintegration? The question is, how important overintegration really is? In this regard, I have unfortunately little to no experience, so what is your take on that? I have run my tests with `QuadratureRuleLinearization::Standard` so far and it seemed fine ... but again, I did not do intense testing. 

Comparing Options 2 and 3, how does the increased effort of using 3/2k instead of Standard compare to the merging or individual application of the kernels in the `MomentumOperator`?

kronbichler: To start with, let me note that **over-integration** is necessary in cases the simulation is under-resolved, because then we might aliasing errors that can make the simulation break. Thus, there is a need to support over-integration in general, but there might be applications where we can get around with fewer points (possible in the current setting). Hence, one cannot really compare the effort of 3k/2 vs standard and merging/not merging because the former is an accuracy requirement, whereas the latter is a software/performance issue without influence on the visible accuracy.

Regarding the design, my position is that option 3 of having separate global operators for different terms in the Navier-Stokes operator as the only way to implement operators in ExaDG would contradict the ambition to be an HPC code on CPU-type hardware. On GPUs, I could see the split making sense in certain implementations (in case one wants to save registers and would need to go out for global memory to get the data again anyway), but trends for GPUs also go towards caches. But who knows about the future, and performance modeling and measurement will be an essential ingredient on that path to choose whatever fits the hardware and available software best.
This does not mean that having the possibility to build operators from building blocks is wrong, it's quite handy for development, e.g. parts that are rarely used and thus not performance critical. But for important operators that do this in inner loops, I would be willing to spend considerable resources to find, implement and maintain better data locality variants.
Along this background, I think the first variant is the most appropriate one of the three options at this stage. It is not pretty and this is certainly requiring more improvement in the future. As a general direction, I suggest we prepare ourselves for being able to use standard quadrature formulas also for convective terms in "inner loops" (e.g. preconditioners), but over-integration and similar things in more outer loops. This fits very well in the first concept expressed by @nfehn above, as the memory consumption is likely not our primary concern. The overall goal is to perform the bulk of the work with approximate (but cheap) operator evaluation, say standard quadrature, and use the actually desired operator with best accuracy only rarely. This will require additional research and additional changes in the software layers once we go for it, but I believe that there are substantial gains to be made this way. This is one ingredient on our path to the future, we should consider also other options, including skew-symmetric formulations of convective terms or the like, to identify the best compromise in terms of accuracy & stability versus run-time. In particular with implicit time stepping, we have plenty of options.

nfehn: One should add to the discussion that we have observed in the past that using a standard quadrature rule for the linearized momentum operator reduces the costs for the operator evaluation, but increases iteration counts noticably. From these numerical experiments, it was not clear whether standard quadrature pays off in terms of overall efficiency. However, we left this option in the code so that one can later again easily investigate this topic.

nfehn: Option 1 seems to be preferred, so let's realize this one. Since there has been a bug in the code before the present PR, we should first fix this bug by adding an assert to the code (as described above), so that it is clear for the present PR which quadrature rule to plug in for the viscous kernel.

Afterwards, in a follow-up PR, we can realize the viscous kernel operating on multiple quadrature rules.

btemuer: Alright, I'm on it👍

nfehn: > Yep, finding this as the cause of a related error message when applying the MomentumOperator in the nonlinear case took me more time than it should have ... 

This should be reported in form of an issue as soon as such problems are encountered.

> in a PR I will open today, I put some assertions for the case of variable viscosity, since the default is using

Bugfixes should be proposed as a separate PR and not as a minor change within a long PR with several thousand lines of code being changed.

To make progress with the present PR, I suggest that @btemuer adds an assert where the MomentumOperator is set up, or @richardschu adds the assert in a separate PR, so that @btemuer can select the proper quadrature rule for the ViscousKernel to proceed with the present PR. 

All other changes such as multiple quadrature rules, nonlinear viscosity models etc. are to be discussed in the future.

btemuer: I've already added an assert in the parameter check thinking it would be equivalent. I also selected the proper quadrature rule. Can you check the latest commit?

nfehn (approved):

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
nfehn: looks good

btemuer:
Could be merged if there are no more points to discuss.

btemuer:
Can we merge @nfehn @peterrum ?

PR #355, opened 2023-02-27
Add git info of exadg

jh66637:
I am currently trying to trace back the commit where I seem to have introduced a bug in a related project. Having the info of the branch on which results were computed would be very helpful. This PR adds the info to the output.


cmake/macros/macro_git_info.cmake
```diff
@@ -0,0 +1,45 @@
...
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+#########################################################################
+
+FUNCTION(SERIALIZE_CMD_COMMAND CMD_COMMAND_STRING_WITH_ARGUMENTS CMD_OUTPUT)
```
peterrum, line 28: Just use: https://github.com/hyperdeal/hyperdeal/blob/8145d0a3e90aa479d0ea158991caf3886ec17dd1/CMakeLists.txt#L61

include/exadg/utilities/exadg_git_info.cpp
```diff
@@ -0,0 +1,41 @@
...
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *  ______________________________________________________________________
+ */
+
+#include <exadg/utilities/exadg_git_info.h>
+#include <exadg/utilities/exadg_git_revision.h>
+
+namespace ExaDG
```
peterrum, line 25: Could we keep the interface similar is it in deal.II: https://github.com/dealii/dealii/blob/master/include/deal.II/base/revision.h.in?

peterrum (commented):
Revision :1st_place_medal:

include/exadg/utilities/exadg_git_info.cpp
jh66637: @peterrum I can do that, but I would like to keep the additional ```.cpp``` file. Otherwise, after every commit, there might be a lot of headers that have to be rebuilt. I can unify this, such that also the header of the ```deal.II``` revision is included in the source file.
Any comments? :)

peterrum: Sure :+1:

include/exadg/utilities/print_general_infos.h
```diff
@@ -93,6 +94,17 @@ print_dealii_info(dealii::ConditionalOStream const & pcout)
...
 }
 
+// print ExaDG info
+inline void
+print_exadg_info(dealii::ConditionalOStream const & pcout)
+{
+  pcout << std::endl
+        << "ExaDG info:" << std::endl
+        << std::endl
+        << "  ExaDG git version " << GitRevision::get_exadg_git_shortrev() << " on branch "
```
nfehn, line 104: We should discuss why this is done differently than for deal.II a few lines above. It is not clear to me why we introduce `exadg_git_info.h`, `exadg_git_info.cpp` in this PR? One solution could be to change it on the deal.II side (with changes in the deal.II library itself and/or get-functions in ExaDG).

jh66637: I would like to include the file which contains the #defines (currently strings, but I will change it) in one source file only, to avoid recompilation of several files if the commit hash changes. Therefore, I introduced the get functions. It would be good to encapsulate the files for deal.ii as well, but it is not as important since the deal.ii library is probably less often recompiled if the focus is on developing code for ExaDG. Personally, I don't think this needs a change on the deal.ii side. 

I will update this PR as soon as possible such that it could be merged

jh66637:
@peterrum I updated such that ```deal.ii``` and ```ExaDG``` have the same interface. @nfehn To keep the revision header includes in a source file only I added ```print_general_infos.cpp``` in which I put all functions that are no templates. 

From my side, this would be ready to merge once all tests have passed. What do you think?

include/exadg/utilities/print_general_infos.cpp
```diff
@@ -0,0 +1,82 @@
...
+// print ExaDG info
+void
+print_exadg_info(dealii::ConditionalOStream const & pcout)
+{
+  pcout << std::endl
+        << "ExaDG info:" << std::endl
+        << std::endl
+        << "  ExaDG git version "
+        << EXADG_GIT_SHORTREV
+        //<< " on branch " << EXADG_GIT_BRANCH
```
nfehn, line 78: what about this comment here?

include/exadg/utilities/print_general_infos.h
```diff
@@ -24,42 +24,22 @@
 
 // deal.II
 #include <deal.II/base/conditional_ostream.h>
-#include <deal.II/base/revision.h>
 #include <deal.II/base/utilities.h>
 #include <deal.II/distributed/tria_base.h>
+#include <deal.II/fe/mapping_q.h>
```
nfehn, line 29: Why including mapping here? Is it really needed by this file or somewhere else?

include/exadg/utilities/print_general_infos.h
```diff
@@ -142,6 +116,7 @@ print_general_info(dealii::ConditionalOStream const & pcout,
   if(not(is_test))
   {
     print_dealii_info(pcout);
+    print_exadg_info(pcout);
```
nfehn, line 119: I would not print ExaDG info between deal.II and matrix-free info. Maybe we print ExaDG first, given that its the ExaDG software here and given that deal.II/matrix-free is additional third-party information.

include/exadg/utilities/print_general_infos.h
jh66637: It is needed in ```print_grid_info()``` and after splitting the files, there were errors that MappingQ is not known

include/exadg/utilities/print_general_infos.cpp
jh66637: I resolved this

include/exadg/utilities/print_general_infos.h
jh66637: Done :)

include/exadg/utilities/print_general_infos.h
nfehn: I see, we should in the future shift the other implementations also to the cpp file.

nfehn (approved):

nfehn:
Let's wait for @peterrum 's approval.

peterrum (approved):

PR #364, opened 2023-03-06
Bump up version of clang-format to 11.1

### LINKED ISSUE #173: Update to newer clang-format version
Having changed to our new Ubuntu OS at LNM with clang-format version 10, I receive clang-format errors, probably because the "indent check" uses an older version.

Here we should discuss whether we want to go to a newer clang-format version, and which version to choose. Also, it would be helpful if ExaDG could provide this tool somewhat automatically in the correct version in order to take this burden from ExaDG developers (I think deal.II has a script to install clang-format automatically).

peterrum:
closes #173

### DIFF
--- a/.github/workflows/indentation.yml
+++ b/.github/workflows/indentation.yml
@@ -12,12 +12,12 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-    - uses: DoozyX/clang-format-lint-action@v0.11
+    - uses: DoozyX/clang-format-lint-action@v0.14
       with:
         source: '.'
         exclude: ''
         extensions: 'h.in,h,cc,cpp'
-        clangFormatVersion: 6
+        clangFormatVersion: 11.1.0
         inplace: true
     - run: |
         git diff
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 *cbp
 .directory
 build/
+programs/
--- a/applications/fluid_structure_interaction/bending_wall/application.h
+++ b/applications/fluid_structure_interaction/bending_wall/application.h
@@ -291,14 +291,16 @@ private:
       SchurComplementPreconditioner::PressureConvectionDiffusion;
   }
 
-  void create_triangulation(dealii::Triangulation<2> & tria)
+  void
+  create_triangulation(dealii::Triangulation<2> & tria)
   {
     (void)tria;
 
     AssertThrow(false, dealii::ExcMessage("not implemented."));
   }
 
-  void create_triangulation(dealii::Triangulation<3> & tria)
+  void
+  create_triangulation(dealii::Triangulation<3> & tria)
   {
     std::vector<dealii::Triangulation<3>> tria_vec;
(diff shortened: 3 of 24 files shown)


.github/workflows/indentation.yml
```diff
@@ -12,12 +12,12 @@ jobs:
...
     steps:
     - uses: actions/checkout@v2
-    - uses: DoozyX/clang-format-lint-action@v0.11
+    - uses: DoozyX/clang-format-lint-action@v0.14
       with:
         source: '.'
         exclude: ''
         extensions: 'h.in,h,cc,cpp'
-        clangFormatVersion: 6
+        clangFormatVersion: 11.1.0
```
btemuer, line 20: I guess you also meant to write 11.1 in the issue title?

btemuer (commented):
I'd like to ask why we are not using `v11.0`, `v12.0` or any  other `.0` version, but `v11.1`. These are easily available on `apt` and probably on many package managers simply as `clang-format-11` etc. 

I guess it is so that we have the same version with deal.II, right? I had to use the scripts that deal.II provides to install `v11.1`, which I think we could also provide if we want to enforce `v11.1`.

peterrum:
> I guess it is so that we have the same version with deal.II, right?

Yes, it is. I think we have agreed on that a long time ago, if I remember correctly.

nfehn:
@peterrum can you adjust/extent the wiki (Coding conventions) regarding how to receive a proper clang-format version? Originally, I thought this PR is about introducing some scripts in ExaDG in analogy to deal.II/hyperdeal. What would be the workflow?

peterrum:
@nfehn I made the requested changes. I would suggest to merge soon so that we don't block other PRs too long.

peterrum:
What about merging so that other PRs are not blocked?

nfehn (approved):

nfehn:
The wiki page https://github.com/exadg/exadg/wiki/Installation#clang-format has not been updated. How can I use the new clang-format?

nfehn:
The wiki page coding conventions mentioned above (https://github.com/exadg/exadg/pull/364#issuecomment-1457736560) seems to be broken. Can someone else confirm this? @peterrum Did you change something or is it maybe still opened in your browser?

btemuer:
Yep, I can't access the coding conventions page in the wiki either. You can look at #371 for a solution. For now, you can run the script
1. `bash ./download_clang_format`  inside `contrib/utilities` to donwload and install,
2. modify `indent-all` as it says in #371,
3. go to the top-level exadg directory and run `bash contrib/utilities/indent-all`.

nfehn:
thanks @btemuer. I opened a PR #377

peterrum:
> The wiki page coding conventions mentioned above (https://github.com/exadg/exadg/pull/364#issuecomment-1457736560) seems to be broken. Can someone else confirm this? @peterrum Did you change something or is it maybe still opened in your browser?

Should be back now. For some reason the file has been renamed.

PR #366, opened 2023-03-07
[WIP] generalized Newtonian fluid for incompressible flow solvers

richardschu:
This PR enables variable viscosity for the incompressible flow solvers and demonstrates the expected convergence rates via a manufactured solution to be found in `applications/incompressible_navier_stokes/navier_stokes_manufactured/`.
Contributions are:
1) Implement variable viscosity due to generalizted Newtonian fluids based on the implementation of turbulence models, merging the implementations, which both precompute the viscosity coefficients per integration point (see #353 ). This motivated including a number of Assertions in `parameters.cpp`. 
2) fixes a bug in the face evaluation of the coefficient
3) demonstration in `navier_stokes_manufactured` (everything as expected)
4) automated convergence table computation by writing and reading the error files, which were already produced, but now giving them identifiers and a `run_id`. This gives flexibility in computing the individual runs to be put into the convergence table and also does not break the interfaces. The location of the convergence study files was hard-coded to `output/` already before.
Note that I used the new formatting already, will rebase once #364 is merged, such that the changes are easier compared!

PR #367, opened 2023-03-07
std::string to enum_type conversion for incompressible Navier Stokes

richardschu:
Split-off from #366 :
to allow the user defining an `enum_type` in the input.json, via, e.g., 

`prm.add_parameter("FormulationViscousTerm",
      		                     formulation_viscous_term_string,
      				     "Formulation of the viscous term.",
      				     dealii::Patterns::Selection("DivergenceFormulation|"
      							                           "LaplaceFormulation"));`

, we need then to parse this std::string to the actual predefined enum like so:

`string_to_enum(formulation_viscous_term, formulation_viscous_term_string);`

To avoid redifining these in all the individual application.h files individually, I did the work here for the incompressible Navier Stokes solver. Then, the user can easily use them as described above.

### DIFF
--- a/include/exadg/incompressible_navier_stokes/user_interface/enum_types.cpp
+++ b/include/exadg/incompressible_navier_stokes/user_interface/enum_types.cpp
@@ -59,6 +59,19 @@ enum_to_string(ProblemType const enum_type)
   return string_type;
 }
 
+void
+string_to_enum(ProblemType & enum_type, std::string const string_type)
+{
+  if(string_type == "Undefined")
+    enum_type = ProblemType::Undefined;
+  else if(string_type == "Steady")
+    enum_type = ProblemType::Steady;
+  else if(string_type == "Unsteady")
+    enum_type = ProblemType::Unsteady;
+  else
+    AssertThrow(false, dealii::ExcMessage("Unknown enum type. Not implemented."));
+}
+
 std::string
 enum_to_string(EquationType const enum_type)
 {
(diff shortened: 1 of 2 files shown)


nfehn:
In my opinion, we should aim at a solution proposed by @jh66637 in #264. The present PR shows how exhausting the present design is, 700 new lines of code without any meaning basically ... lines of code that need to be maintained. So I believe we should shift the discussion to #264.

nfehn:
Regarding parameter parsing

> `prm.add_parameter("FormulationViscousTerm", formulation_viscous_term_string, "Formulation of the viscous term.", dealii::Patterns::Selection("DivergenceFormulation|" "LaplaceFormulation"));`

we should think critically about using `dealii::Patterns::Selection`, because listing the different variants/names is essentially duplicated code. This will imply quite some maintenance work in practice to keep everything in sync. Due to this "code duplication", users might be irritated by the error message that they receive (the string in the input .json file might be identical to the enum declaration in the code, but the program terminates nevertheless!?).

peterrum:
@nfehn I would suggest to proceed independently here from the decision regarding `magicenum`.

jh66637:
I will try to propose a version tomorrow (but latest beginning next week) such that it can be discussed in detail :)

richardschu:
As far as I understand, this becomes superfluous with #382 , right? Thank you @jh66637 !

jh66637:
Yes, every enum<->string conversion should work out of the box with the utility function. 

The only thing that is not covered are the two newly introduced enums classes. I just mention it here such that the two classes are not forgotton :)

richardschu:
Beautiful stuff! I shall look into this once #382 is merged, to make the required adaptations.

richardschu:
So let's keep this one open to remind me :)

nfehn:
I am closing this PR now due to #382.

PR #368, opened 2023-03-07
automated convergence studies

richardschu:
split-off from #366 
automated convergence study analysis taking files generated by `ErrorCalculator`, now writing the files in a programmatic way: in the first call (when originally the files were only flushed), we now give a new run_id if a file with the same `filename_base` exists. 

This allows splitting up the studies in various parts, or just displaying all the error in one go after a single potentially small run for easier extraction of the data. Since the `ErrorCalculators` are in `utilities`, one just needs to adapt the individual `solver.h` files instead of postprocessing individual log files. After all, the `solver.h` were designed to run the studies already.

If no files or no directory exists, we print a message; if only a single file exists, there is a specific message as well, and if there are multiple files found, a convergence study is populated and printed.


richardschu:
I have to check the tests, the DIFF fails, I will do this tomorrow ; )

Issue #373, opened 2023-03-09
Enable Dirichlet BCs with ComponentMask for all multigrid implementations and all PDE modules


nfehn:
`ComponentMask` is currently only possible for the `Structure` module in combination with local-smoothing multigrid. This limitation should be relaxed (right now, first and foremost to use global-coarsening multigrid for the structure module).

I am currently starting to work on this. I think we have to forbid reimplementing such functions in derived multigrid classes. My suggestion is to equip `MultigridPreconditionerBase::initialize()` with a map for the `ComponentMask`, enlarging once more the interface, in the hope that this makes our complex multigrid implementation more flexible/general and less error prone.

Issue #324, opened 2023-02-02
Global coarsening MG does not work for structure/bar app


nfehn:
The problem occurs for both `Distributed::Triangulation` and `FullyDistributed::Triangulation` and both `Hypercube` and `Simplex` (of course, the latter being only available for fully-distributed tria's).

A main difference to the Poisson module (where this seems to work well) is that we use `ComponentMask` for boundary conditions in case of the structure module. Maybe this is not treated correctly in the multigrid preconditioner.

nfehn:
I changed the BCs (prescribing Dirichlet values in all directions; no need for ComponentMask), and it works. See also PR #325, providing more flexibility in changing BCs easily.

Hence, BCs with `ComponentMask` do not seem to be treated correctly by the multigrid preconditioner.

PR #376, opened 2023-03-09
initialize multigrid preconditioner with map for ComponentMask

nfehn:
addresses issue #373 partly. Global coarsening multigrid does still not make use of ComponentMask.


nfehn:
some tests were failing, which I fixed by commit https://github.com/exadg/exadg/pull/376/commits/3d4e420d2181b6b0e9032023208471f20b10e256. I have no idea, though, why this is necessary. The function `ConstraintUtil::add_constraints<dim>()` called somewhat below immediately returns in the DG case, so where are the constrained dofs actually used? For me this looks like they are not really used, so I consider this an unexpected behavior.  I would be happy if @kronbichler @peterrum had an idea where this comes from?

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -485,14 +493,41 @@ MultigridPreconditionerBase<dim, Number>::do_initialize_dof_handler_and_constrai
...
+        boundary_functions[it.first] = &zero_function;
+      }
+
+      auto const & mapping_dummy =
+        dof_handler->get_fe().reference_cell().template get_default_linear_mapping<dim>();
+
+      dealii::VectorTools::interpolate_boundary_values(mapping_dummy,
+                                                       *dof_handler,
+                                                       boundary_functions,
+                                                       *affine_constraints_own);
```
nfehn, line 527: We have a map of boundary ids and `ComponentMasks`, so this function can not be used any more. Can we solve this by just looping over the boundary IDs and calling `dealii::VectorTools::interpolate_boundary_values()` (for al single boundary id, function, and component mask) inside the loop? Any concerns?

See also https://github.com/dealii/dealii/issues/14871

nfehn:
Global coarsening multigrid calls `dealii::DoFTools::make_hanging_node_constraints()` as well as `dealii::VectorTools::interpolate_boundary_values()`, whereas the fine-level operator in `Poisson::Operator` only calls `dealii::DoFTools::make_zero_boundary_constraints()` (and not `dealii::DoFTools::make_hanging_node_constraints()` and also not `dealii::VectorTools::interpolate_boundary_values()`). I find this quite irritating. Why is this the case? Maybe @kronbichler @peterrum could help me here to better understand the code.

nfehn:
The problem described in issue #324 seems to be solved by commit https://github.com/exadg/exadg/pull/376/commits/7d166eb0ed3f0aba58dfe7b7e7217260bc244edf

peterrum:
> Global coarsening multigrid calls `dealii::DoFTools::make_hanging_node_constraints()` as well as `dealii::VectorTools::interpolate_boundary_values()`, whereas the fine-level operator in `Poisson::Operator` only calls `dealii::DoFTools::make_zero_boundary_constraints()` (and not `dealii::DoFTools::make_hanging_node_constraints()` and also not `dealii::VectorTools::interpolate_boundary_values()`). I find this quite irritating. Why is this the case? Maybe @kronbichler @peterrum could help me here to better understand the code.

`interpolate_boundary_values()` with zero function or `dealii::DoFTools::make_zero_boundary_constraints()` is the same. Regarding `dealii::DoFTools::make_hanging_node_constraints()`: in the case of LS this is done in https://www.dealii.org/developer/doxygen/deal.II/classMGConstrainedDoFs.html#a4e94f0431024e13f336d7ca620ba2aee.

include/exadg/poisson/user_interface/boundary_descriptor.h
```diff
@@ -44,14 +44,23 @@ enum class BoundaryType
 template<int rank, int dim>
 struct BoundaryDescriptor
 {
+  // Dirichlet
   std::map<dealii::types::boundary_id, std::shared_ptr<dealii::Function<dim>>> dirichlet_bc;
 
-  // dealii::ComponentMask is only used for continuous elements, and is ignored for DG
+  // ComponentMaskn (only used/relevant for continuous Galerkin, ignored for DG)
```
peterrum, line 50:
```suggestion
  // ComponentMask (only used/relevant for continuous Galerkin, ignored for DG)
```

peterrum (approved):

nfehn:
> interpolate_boundary_values() with zero function or dealii::DoFTools::make_zero_boundary_constraints() is the same.

What would you suggest to use in ExaDG, or should we use a mixture? Interpolating boundary values with a zero function sounds like unnecessary work.

It is still not clear to me where the actual PDE operator applies hanging node constraints (say, one uses no preconditioner at all to solve the Poisson problem), see my comment above https://github.com/exadg/exadg/pull/376#issuecomment-1463884288.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -517,20 +552,52 @@ MultigridPreconditionerBase<dim, Number>::do_initialize_dof_handler_and_constrai
...
+          std::set<dealii::types::boundary_id> dirichlet_boundary;
+          dirichlet_boundary.insert(it.first);
+
+          dealii::ComponentMask mask    = dealii::ComponentMask();
+          auto                  it_mask = dirichlet_bc_component_mask.find(it.first);
+          if(it_mask != dirichlet_bc_component_mask.end())
+            mask = it_mask->second;
+
+          constrained_dofs->make_zero_boundary_constraints(*dof_handler, dirichlet_boundary, mask);
+        }
```
kronbichler, line 599: I will comment on this also in the general section, but I feel that this might be a performance concern if we have cases with a high number of boundary ids. The reason is how the underlying implementation in deal.II works by a loop over all mesh elements. Thus, using a single `std::set` of boundary ids for all cases where the component mask is the same would be preferable. (I worked hard to make loops over mesh elements in deal.II faster during the last year, and I invite others to make further improvements, but the underlying complexity cannot be avoided.)

kronbichler:
> some tests were failing, which I fixed by commit https://github.com/exadg/exadg/commit/3d4e420d2181b6b0e9032023208471f20b10e256. I have no idea, though, why this is necessary. The function ConstraintUtil::add_constraints<dim>() called somewhat below immediately returns in the DG case, so where are the constrained dofs actually used? For me this looks like they are not really used, so I consider this an unexpected behavior. I would be happy if @kronbichler @peterrum had an idea where this comes from?

What kind of error did you get? As said by @peterrum, the `make_zero_boundary_constraints` function should have no effect. I could imagine that the line `constrained_dofs->initialize(*dof_handler);` is what is actually needed and not the rest, which would get indicated by an error message in deal.II's debug mode or segmentation fault in release mode. For example, it could be this line https://github.com/exadg/exadg/blob/5fdc4399e9feef2e080f4cc2f598caed341aad11/include/exadg/solvers_and_preconditioners/multigrid/transfers/mg_transfer_global_refinement.cpp#L102 but it is hard to explain based on the vague statement "some tests were failing".

kronbichler:
> Regarding dealii::DoFTools::make_hanging_node_constraints(): in the case of LS this is done in https://www.dealii.org/developer/doxygen/deal.II/classMGConstrainedDoFs.html#a4e94f0431024e13f336d7ca620ba2aee

I don't think that explanation is sufficient. Let me thus address the statements one-by-one:
> Global coarsening multigrid calls dealii::DoFTools::make_hanging_node_constraints() as well as dealii::VectorTools::interpolate_boundary_values(), whereas the fine-level operator in Poisson::Operator only calls dealii::DoFTools::make_zero_boundary_constraints() (and not dealii::DoFTools::make_hanging_node_constraints() 

It seems that we assume that the fine-level operator assumes that we work with continuous basis functions, since a missing `make_hanging_node_constraints` on the operator would be a bug. I did not inspect all code, but could this be simply forgotten?

> also not dealii::VectorTools::interpolate_boundary_values()

Here I see two aspects. On the one hand, @peterrum's explanation applies here in the sense that the multigrid `MGConstrainedDofs` constrained dofs take care of the boundary conditions (but only for local smoothing, not the global coarsening code paths). But to me `fine-level operator` to me indicates that the operator is the one entering the linear solver, where the constraints need to be transmitted by other mechanisms as all `MGConstrainedDofs` does is enter the multigrid preconditioner with separate numbers of the unknowns. So my assumption is that this was again assuming DG elements without degrees of freedom on the boundary.

kronbichler:
> > interpolate_boundary_values() with zero function or dealii::DoFTools::make_zero_boundary_constraints() is the same.
>
> What would you suggest to use in ExaDG, or should we use a mixture? Interpolating boundary values with a zero function sounds like unnecessary work.
> 
> It is still not clear to me where the actual PDE operator applies hanging node constraints (say, one uses no preconditioner at all to solve the Poisson problem), see my comment above https://github.com/exadg/exadg/pull/376#issuecomment-1463884288.

Technically, I consider `dealii::DoFTools::make_zero_boundary_constraints()` to be unnecessary in deal.II because its effect should be the same as `interpolate_boundary_conditions` with a zero function. I am not sure why it was added, and I would personally delete it or at least redirect the implementation to `interpolate_boundary_values`. I am simplifying things a bit. To also tell the whole story, it could also be that I overlook something, or, at least to me, a performance issue, because `interpolate_boundary_values` does quite some additional work to identify the position of the points, it sets up a possibly expensive `FEFaceValues`, and some more things. Also, I am not a friend of removing interfaces because of the cost in downstream projects.

So given the current status, I would tend to use use of a more uniform interface, which would suggest to use `make_zero_boundary_constraints` uniformly? I am asking here because I do not have a complete overview of all code at this stage.

kronbichler:
I do not like the fact that we need to pull the constraint mask through so many interfaces. The two questions I would have is:
- I assume this has its origins in the structure module where we have multiple components. Are there many anticipated uses outside the structure to motivate this in the Navier-Stokes module for example? If not, my reaction would be if solving it through two different setup functions, i.e., one with the old simpler interface, and the other one, would be possible as well. However, that suggestion might create ambiguity, and we need to look forward to say Hdiv.
- Another solution, which I would personally like more (but is a bit more work), would be to have the `Map` argument passed around also include a `ComponentMask`: This would mean that we use `std::map<dealii::boundary_id, BoundaryCondition>` with
```
struct BoundaryCondition
{
  std::shared_ptr<dealii::Function<dim>> boundary_function;
  ComponentMask component_mask;
};
```
We could leave the second field empty in most cases, but overall it would be safer because we do only set information that belongs together inside the same data structure. We could then possibly extend this to also collect boundary ids with the same component mask more easily, to address a possible performance problem mentioned above.

nfehn:
> but it is hard to explain based on the vague statement "some tests were failing"

When I have written this, I referred to the failing pipeline/checks run by github based on the first commit of this PR. At that time, some of the tests were failing. I explained in comment https://github.com/exadg/exadg/pull/376#issuecomment-1463421053 how I resolved this. I did not make notes which tests failed exactly, unfortunately.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -517,20 +552,52 @@ MultigridPreconditionerBase<dim, Number>::do_initialize_dof_handler_and_constrai
...
       constrained_dofs->clear();
-      this->initialize_constrained_dofs(*dof_handler, *constrained_dofs, dirichlet_bc);
+
+      if(level.is_dg)
+      {
+        std::set<dealii::types::boundary_id> dirichlet_boundary;
+        for(auto & it : dirichlet_bc)
+          dirichlet_boundary.insert(it.first);
+        constrained_dofs->initialize(*dof_handler);
+        constrained_dofs->make_zero_boundary_constraints(*dof_handler, dirichlet_boundary);
```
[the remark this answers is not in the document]
nfehn: I will try removing everything apart from the `initialize()` call.

peterrum:
You mean these failing tests: https://github.com/exadg/exadg/actions/runs/4377110963/jobs/7660153643

Did yo run these tests in debug mode on your machine?

nfehn:
> It seems that we assume that the fine-level operator assumes that we work with continuous basis functions, since a missing make_hanging_node_constraints on the operator would be a bug. I did not inspect all code, but could this be simply forgotten?

@kronbichler: Do you mean "discontinuous" instead of "continuous" here? For the SC21 paper, we tested DG with hanging nodes. I don't know if we tested CG with hanging nodes? (@peterrum @bergbauer)

nfehn:
@kronbichler Regarding the discussion here https://github.com/exadg/exadg/pull/376#issuecomment-1465976876: Intuitively, I would expect that multigrid uses `DoFTools::make_zero_boundary_constraints()` (multigrid operates on homogeneous operator) and the actual PDE operator uses `VectorTools::interpolate_boundary_values()` with the true boundary conditions (potentially inhomogeneous), but it is actually done the other way in the current implementation?!

I think we need to take into account what the following function does: https://github.com/exadg/exadg/blob/41f8752a19a667b64009fcd6f4325a087afe8baa/include/exadg/poisson/spatial_discretization/operator.cpp#L459

This raises the question where the implementation applies hanging node constraints (question already raised above), and why we initialize affine constraints with zero boundary conditions at all in the setup routine?

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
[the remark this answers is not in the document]
nfehn: tests pass with commit https://github.com/exadg/exadg/pull/376/commits/1d03f2754376b99c6365ff6f26e5a878a2f29e0b

kronbichler:
> Intuitively, I would expect that multigrid uses DoFTools::make_zero_boundary_constraints() (multigrid operates on homogeneous operator) and the actual PDE operator uses VectorTools::interpolate_boundary_values() with the true boundary conditions (potentially inhomogeneous), but it is actually done the other way in the current implementation?!

I agree, this sounds like the state we want to achieve. For using `interpolate_boundary_values`, we would need to have a way to recall the inhomogeneity, because we need to apply it to the vectors and possibly combine with other constraints such as periodicity or hanging nodes.

> This raises the question where the implementation applies hanging node constraints (question already raised above), and why we initialize affine constraints with zero boundary conditions at all in the setup routine?

Are there actual problems with the call to `make_hanging_node_constraints`, apart from the place discussed above? Regarding the zero boundary conditions and hanging nodes in general, we need to create them in all places that set up a `MatrixFree` object (or a matrix), because the operators (and matrices) need to resolve the constraints to ensure a compatible (say H1 conforming) operator that is non-singular. Of course, the `AffineConstraints` object then also needs to include periodic boundary conditions, and similar constraints (say no-normal flux). Does this general statement help, or do we need to look into specific places?

nfehn:
> Are there actual problems with the call to make_hanging_node_constraints, apart from the place discussed above? Regarding the zero boundary conditions and hanging nodes in general, we need to create them in all places that set up a MatrixFree object (or a matrix), because the operators (and matrices) need to resolve the constraints to ensure a compatible (say H1 conforming) operator that is non-singular. Of course, the AffineConstraints object then also needs to include periodic boundary conditions, and similar constraints (say no-normal flux). Does this general statement help, or do we need to look into specific places?

I think this is not done correctly in the code, so far. A zoom meeting would be helpful and probably be more effective than writing.

> Intuitively, I would expect that multigrid uses DoFTools::make_zero_boundary_constraints() (multigrid operates on homogeneous operator) and the actual PDE operator uses VectorTools::interpolate_boundary_values() with the true boundary conditions (potentially inhomogeneous), but it is actually done the other way in the current implementation?!

I changed my opinion on this topic. I think the actual PDE operator **has to use** `make_zero_boundary_constraints()` as well, since MatrixFree needs to be initialized with zero boundary data in order to correctly evaluate the homogeneous action of the operator in `vmult()` for the iterative solution of linear systems of equations. See also PR #384. Do you agree, @kronbichler?

kronbichler:
Yes, merging in the present state makes sense, thank you for the explanations. Regarding 
> What I actually want as well in ExaDG, is to constrain normal components (for potentially curved boundaries),

The solution is `VectorTools::compute_no_normal_flux_constraints` or `VectorTools::compute_nonzero_normal_flux_constraints`, https://www.dealii.org/current/doxygen/deal.II/group__constraints.html#gae3b53d69bd7afe08f0d033497833938e. This constrains one of the dofs on the boundary in terms of the other two, so as to ensure that the solution is tangential to the boundary. There is also a related multigrid function, `MGConstrainedDoFs::make_no_normal_flux_constraints`, for the homogeneous operator on the MG levels. These functions are a bit hard to organize, so having them integrated in a multigrid/structure module definitely makes sense.

kronbichler:
> I think the actual PDE operator has to use make_zero_boundary_constraints() as well, since MatrixFree needs to be initialized with zero boundary data in order to correctly evaluate the homogeneous action of the operator in vmult() for the iterative solution of linear systems of equations. See also PR https://github.com/exadg/exadg/pull/384. Do you agree, @kronbichler?

Yes, the matrix-free operators do not treat inhomogeneities. We might need to think about a solution longer-term, as this is also related to the `read_dof_values` vs `read_dof_values_plain`.

PR #377, opened 2023-03-09
adjust indent-all script to use new clang-format version

nfehn:
@btemuer is this what you did to solve it? Any concerns regarding my fix?

### DIFF
--- a/contrib/utilities/indent-all
+++ b/contrib/utilities/indent-all
@@ -26,8 +26,8 @@
 #
 #########################################################################
 
-if test -f "programs/clang-11/bin/clang-format"; then
-  cf=programs/clang-11/bin/clang-format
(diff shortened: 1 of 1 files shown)


btemuer:
No, I changed the condition and the first statement, not the second. I would suggest writing

```bash
if test -f "contrib/utilities/programs/clang-11/bin/clang-format"; then
  cf=contrib/utilities/programs/clang-11/bin/clang-format
else
  cf=clang-format
fi
```
and execute this script from the top-level directory so `clang-format` can find the source files. I think the `else` statement here makes sure to take the default `clang-format` executable in the path (version 10 in our Ubuntu) in case `v11.1` wasn't downloaded,

peterrum:
I have pushed parallel also the fix^^

peterrum (approved):

PR #378, opened 2023-03-10
Incorporate non-hypercube elements and every triangulation type in categorazation.h

### LINKED ISSUE #375: Dynamic cast of a triangulation to p:d:T
Maybe I am missing something here but why are we doing a dynamic cast to a `p:d:T`?
https://github.com/exadg/exadg/blob/d3c21325476fad9eed77e335ac78da17787ec207/include/exadg/poisson/preconditioners/multigrid_preconditioner.cpp#L103-L105
When I use a a fullydistributed or a serial triangulation the cast seems to fail and I get a segmentation fault from the `Categorization::do_cell_based_loops`, which tries to use `tria`.

necioglu:
This PR closes #375 and replaces `dim*2` (which is equal to `n_faces` for hypercubes) by dealii's `n_faces` in categorization.h.

### DIFF
--- a/include/exadg/matrix_free/categorization.h
+++ b/include/exadg/matrix_free/categorization.h
@@ -50,9 +50,9 @@ do_cell_based_loops(dealii::Triangulation<dim> const & tria,
   else
     data.cell_vectorization_category.resize(tria.n_active_cells());
 
-  unsigned int const n_faces_per_cell = tria.get_reference_cells()[0].n_faces();
   AssertThrow(tria.get_reference_cells().size() == 1,
               dealii::ExcMessage("No mixed meshes allowed."));
+  unsigned int const n_faces_per_cell = tria.get_reference_cells()[0].n_faces();
 
   // ... setup scaling factor
   std::vector<unsigned int> factors(n_faces_per_cell);
--- a/include/exadg/matrix_free/categorization.h
+++ b/include/exadg/matrix_free/categorization.h
@@ -50,8 +50,12 @@ do_cell_based_loops(dealii::Triangulation<dim> const & tria,
   else
     data.cell_vectorization_category.resize(tria.n_active_cells());
 
+  unsigned int const n_faces_per_cell = tria.get_reference_cells()[0].n_faces();
+  AssertThrow(tria.get_reference_cells().size() == 1,
+              dealii::ExcMessage("No mixed meshes allowed."));
+
   // ... setup scaling factor
-  std::vector<unsigned int> factors(dim * 2);
+  std::vector<unsigned int> factors(n_faces_per_cell);
 
   std::map<unsigned int, unsigned int> bid_map;
   for(unsigned int i = 0; i < tria.get_boundary_ids().size(); i++)
@@ -60,13 +64,13 @@ do_cell_based_loops(dealii::Triangulation<dim> const & tria,
   {
     unsigned int bids   = tria.get_boundary_ids().size() + 1;
     int          offset = 1;
-    for(unsigned int i = 0; i < dim * 2; i++, offset = offset * bids)
+    for(unsigned int i = 0; i < n_faces_per_cell; i++, offset = offset * bids)
       factors[i] = offset;
   }
 
   auto to_category = [&](auto & cell) {
     unsigned int c_num = 0;
-    for(unsigned int i = 0; i < dim * 2; i++)
+    for(unsigned int i = 0; i < n_faces_per_cell; i++)
     {
       auto & face = *cell->face(i);
       if(face.at_boundary())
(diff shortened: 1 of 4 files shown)


include/exadg/poisson/preconditioners/multigrid_preconditioner.cpp
```diff
@@ -100,8 +100,7 @@ MultigridPreconditioner<dim, Number, n_components>::fill_matrix_free_data(
 
   if(data.use_cell_based_loops && this->level_info[level].is_dg())
   {
-    auto tria = dynamic_cast<dealii::parallel::distributed::Triangulation<dim> const *>(
-      &this->dof_handlers[level]->get_triangulation());
+    auto tria = &this->dof_handlers[level]->get_triangulation();
```
peterrum, line 103: Thanks :+1:

include/exadg/matrix_free/categorization.h
```diff
@@ -50,8 +50,12 @@ do_cell_based_loops(dealii::Triangulation<dim> const & tria,
   else
     data.cell_vectorization_category.resize(tria.n_active_cells());
 
+  unsigned int const n_faces_per_cell = tria.get_reference_cells()[0].n_faces();
```
peterrum, line 53: The changes in this function look fine. But, I would like to point out that cell-centered face loops do not work yet in deal.II.

necioglu: I assume you mean for simplex? 

Yes, I have noticed that. As far as I remember, the code failed here for me:

https://github.com/dealii/dealii/blob/947199be8142d3fc7bde15d8f94f168573ccf5d2/include/deal.II/grid/tria_accessor.templates.h#L755-L764 

And once you mentioned that this Assert is mostly an indication of the function not being extended, yet.

nfehn (approved):

include/exadg/matrix_free/categorization.h
nfehn: I think we should assert in ExaDG (in categorization.h) that simplices are not supported for cell-based face loops. What do you think?

Principally, we should only rely on deal.II error messages if the error message is easy to interpret for ExaDG users and if this error message appears both in debug and release mode. To make an example, I encountered recently that the degree <=2 restriction of simplices in dealii does not give easy to interpret error messages in release mode, which is clearly sub-optimal for ExaDG users.

necioglu: > I think we should assert in ExaDG (in categorization.h) that simplices are not supported for cell-based face loops. What do you think?

I agree. I had different kind of errors from dealii in Debug mode, when I tried to enable cell-based face loops, see https://github.com/exadg/exadg/pull/380#issuecomment-1464139651. And I would say, they would be hard to interpret for someone who is not familiar with simplices.

However, I am not sure if we should for example assert degree <=2 in ExaDG, as well. Most of the dealii asserts I have encountered only appear in debug mode. I think if we add them on ExaDG, we would end up having too many. That is why I do development in debug, and calculations in release mode.

nfehn: Sure, but let's also keep in mind that you are the person most familiar with simplices in ExaDG. We should make sure that every other user of ExaDG is able to use simplices as well. I think if we place the asserts at a central place, its not too much work. Let's postpone the discussion on the "degree <=2 restriction", it is not related to this PR and was just an example. (The problem with degree <=2 is that we do not construct the finite element at a central place in ExaDG for all the modules, meaning that we would need this assert at several places, which is clearly sub-optimal.)

include/exadg/matrix_free/categorization.h
```diff
@@ -50,8 +50,12 @@ do_cell_based_loops(dealii::Triangulation<dim> const & tria,
...
+  AssertThrow(tria.get_reference_cells().size() == 1,
+              dealii::ExcMessage("No mixed meshes allowed."));
```
nfehn, line 55: I do not understand this error message. Didn't we say that pure simplex meshes do not allow cell-based face loops (which is the content of the present function)?

necioglu: No this is to check if we have multiple reference cells in our triangulation, i.e. mixed mesh. If we don't check this we might end up having a wrong `n_faces_per_cell`.

nfehn: Then I would write the Assert before the line with `n_faces_per_cell`.

include/exadg/matrix_free/categorization.h
nfehn: @peterrum what do you suggest here? An additional assert for simplex?

include/exadg/matrix_free/categorization.h
necioglu: done.

PR #380, opened 2023-03-10
Implement inverse mass operator for DG + simplex elements

### LINKED ISSUE #372: Cellwise inverse mass operator not available for simplices
The cellwise inverse mass operator of dealii is not available for non-tensor product elements, e.g. simplices:

https://github.com/dealii/dealii/blob/master/include/deal.II/matrix_free/evaluation_kernels.h#L5601-L5603

This is why, I am implementing a solution for this in ExaDG, using `BlockJacobiPreconditioner` or `SolverCG`. This solution can be removed, when dealii offers the functionality.

necioglu:
This PR closes #372.

For simplices, dealii::MatrixFreeOperators::CellwiseInverseMassMatrix is not available and won't be available in its current form, as described in https://github.com/exadg/exadg/issues/372#issuecomment-1460065247.

With this PR, we use the BlockJacobiPreconditioner initialized with the MassOperator, to compute the operation of applying the inverse mass matrix on a vector.  Internally, the BlockJacobiPreconditioner has two options on how to do this:

1. Building cell matrices and inverting them using a LU-Factorization.
2. Applying a cellwise, i.e., blockwise CG Solver on each block.

The latter can not be used with simplices , but I didn't place an Assert because this can be both seen as a dealii or an ExaDG issue.

On the one hand, for simplices "cell-centered face loops do not work yet in deal.II.", see https://github.com/exadg/exadg/pull/378#discussion_r1132161842. 

On the other hand, we actually don't need face loops for the Inverse mass operator, but the current implementation doesn't allow this: https://github.com/exadg/exadg/blob/f8e533cad5fdd2fb1c6845657d5fda59228590e4/include/exadg/incompressible_navier_stokes/user_interface/parameters.cpp#L530-L535

And, I needed to extend the instantiations of the mass_operator and operator_base, since compressible Navier-Stokes has `n_components=dim+2` and can use a Inverse mass operator.

### DIFF
--- a/include/exadg/compressible_navier_stokes/user_interface/parameters.h
+++ b/include/exadg/compressible_navier_stokes/user_interface/parameters.h
@@ -27,7 +27,6 @@
 #include <exadg/time_integration/restart_data.h>
 #include <exadg/time_integration/solver_info_data.h>
 #include <exadg/utilities/print_functions.h>
-#include "exadg/solvers_and_preconditioners/solvers/solver_data.h"
 
 namespace ExaDG
 {
@@ -193,18 +192,6 @@ public:
   // use combined operator for viscous term and convective term in order to improve run
   // time
   bool use_combined_operator;
-
-  /**************************************************************************************/
-  /*                                                                                    */
-  /*                            ELEMENTWISE INVERSE MASS                                */
-  /*                                                                                    */
-  /**************************************************************************************/
-  // Used when matrix-free inverse mass operator is not available and when the spatial
-  // discretization is DG, e.g. simplex.
-
-  bool solve_elementwise_mass_system_matrix_free;
-
-  SolverData solver_data_elementwise_inverse_mass;
 };
 
 } // namespace CompNS
--- a/include/exadg/compressible_navier_stokes/user_interface/parameters.h
+++ b/include/exadg/compressible_navier_stokes/user_interface/parameters.h
@@ -27,6 +27,7 @@
 #include <exadg/time_integration/restart_data.h>
 #include <exadg/time_integration/solver_info_data.h>
 #include <exadg/utilities/print_functions.h>
+#include "exadg/solvers_and_preconditioners/solvers/solver_data.h"
 
 namespace ExaDG
 {
@@ -192,6 +193,18 @@ public:
   // use combined operator for viscous term and convective term in order to improve run
   // time
   bool use_combined_operator;
+
+  /**************************************************************************************/
+  /*                                                                                    */
+  /*                            ELEMENTWISE INVERSE MASS                                */
+  /*                                                                                    */
+  /**************************************************************************************/
+  // Used when matrix-free inverse mass operator is not available and when the spatial
+  // discretization is DG, e.g. simplex.
+
+  bool solve_elementwise_mass_system_matrix_free;
+
+  SolverData solver_data_elementwise_inverse_mass;
 };
 
 } // namespace CompNS
--- a/include/exadg/convection_diffusion/user_interface/parameters.h
+++ b/include/exadg/convection_diffusion/user_interface/parameters.h
@@ -318,6 +318,18 @@ public:
 
   // use 3/2 overintegration rule for convective term
   bool use_overintegration;
+
+  /**************************************************************************************/
+  /*                                                                                    */
+  /*                            ELEMENTWISE INVERSE MASS                                */
+  /*                                                                                    */
+  /**************************************************************************************/
+  // Used when matrix-free inverse mass operator is not available and when the spatial
+  // discretization is DG, e.g. simplex.
+
+  bool solve_elementwise_mass_system_matrix_free;
+
+  SolverData solver_data_elementwise_inverse_mass;
 };
 
 } // namespace ConvDiff
--- a/include/exadg/convection_diffusion/user_interface/parameters.h
+++ b/include/exadg/convection_diffusion/user_interface/parameters.h
@@ -318,18 +318,6 @@ public:
 
   // use 3/2 overintegration rule for convective term
   bool use_overintegration;
-
-  /**************************************************************************************/
-  /*                                                                                    */
-  /*                            ELEMENTWISE INVERSE MASS                                */
-  /*                                                                                    */
-  /**************************************************************************************/
-  // Used when matrix-free inverse mass operator is not available and when the spatial
-  // discretization is DG, e.g. simplex.
-
-  bool solve_elementwise_mass_system_matrix_free;
-
-  SolverData solver_data_elementwise_inverse_mass;
 };
 
 } // namespace ConvDiff
--- a/include/exadg/operators/inverse_mass_operator.h
+++ b/include/exadg/operators/inverse_mass_operator.h
@@ -28,6 +28,8 @@
 
 // ExaDG
 #include <exadg/matrix_free/integrators.h>
+#include <exadg/operators/mass_operator.h>
+#include <exadg/solvers_and_preconditioners/preconditioners/block_jacobi_preconditioner.h>
 
 namespace ExaDG
 {
@@ -60,6 +62,35 @@ public:
     this->matrix_free = &matrix_free_in;
     dof_index         = dof_index_in;
     quad_index        = quad_index_in;
+
+    cellwise_inverse_mass_not_available = not(matrix_free->get_dof_handler(dof_index)
+                                                .get_triangulation()
+                                                .all_reference_cells_are_hyper_cube());
+
+    if(cellwise_inverse_mass_not_available)
+    {
+      initialize_inverse_mass_operator_with_block_jacobi();
+    }
+  }
+
+  void
+  initialize_inverse_mass_operator_with_block_jacobi()
+  {
+    // initialize mass operator
+    dealii::AffineConstraints<Number> const & constraint =
+      matrix_free->get_affine_constraints(dof_index);
+
+    MassOperatorData<dim> mass_operator_data;
+    mass_operator_data.dof_index  = dof_index;
+    mass_operator_data.quad_index = quad_index;
+
+    mass_operator.initialize(*matrix_free, constraint, mass_operator_data);
+
+    // build a BlockJacobiPreconditioner and use the vmult(dst,src) for applying the inverse mass
+    // operator on  source the vector
+    mass_preconditioner =
+      std::make_shared<BlockJacobiPreconditioner<MassOperator<dim, n_components, Number>>>(
+        mass_operator);
   }
 
   void
@@ -67,7 +98,14 @@ public:
   {
     dst.zero_out_ghost_values();
 
-    matrix_free->cell_loop(&This::cell_loop, this, dst, src);
+    if(cellwise_inverse_mass_not_available)
+    {
+      mass_preconditioner->vmult(dst, src);
+    }
+    else
+    {
+      matrix_free->cell_loop(&This::cell_loop, this, dst, src);
+    }
   }
 
 private:
@@ -94,6 +132,13 @@ private:
   dealii::MatrixFree<dim, Number> const * matrix_free;
 
   unsigned int dof_index, quad_index;
+
+  // BlockJacobi Preconditoner to be used when the CellwiseInverseMassMatrix is not available
+  bool cellwise_inverse_mass_not_available;
+
+  MassOperator<dim, n_components, Number> mass_operator;
+
+  std::shared_ptr<PreconditionerBase<Number>> mass_preconditioner;
 };
 
 } // namespace ExaDG
--- a/include/exadg/operators/inverse_mass_operator.h
+++ b/include/exadg/operators/inverse_mass_operator.h
@@ -74,26 +74,6 @@ public:
   }
 
   void
-  initialize_inverse_mass_operator_with_block_jacobi()
-  {
-    // initialize mass operator
-    dealii::AffineConstraints<Number> const & constraint =
-      matrix_free->get_affine_constraints(dof_index);
-
-    MassOperatorData<dim> mass_operator_data;
-    mass_operator_data.dof_index  = dof_index;
-    mass_operator_data.quad_index = quad_index;
-
-    mass_operator.initialize(*matrix_free, constraint, mass_operator_data);
-
-    // build a BlockJacobiPreconditioner and use the vmult(dst,src) for applying the inverse mass
-    // operator on  source the vector
-    mass_preconditioner =
-      std::make_shared<BlockJacobiPreconditioner<MassOperator<dim, n_components, Number>>>(
-        mass_operator);
-  }
-
-  void
   apply(VectorType & dst, VectorType const & src) const
   {
     dst.zero_out_ghost_values();
@@ -127,6 +107,26 @@ private:
 
       integrator.set_dof_values(dst, 0);
     }
+  }
+
+  void
+  initialize_inverse_mass_operator_with_block_jacobi()
+  {
+    // initialize mass operator
+    dealii::AffineConstraints<Number> const & constraint =
+      matrix_free->get_affine_constraints(dof_index);
+
+    MassOperatorData<dim> mass_operator_data;
+    mass_operator_data.dof_index  = dof_index;
+    mass_operator_data.quad_index = quad_index;
+
+    mass_operator.initialize(*matrix_free, constraint, mass_operator_data);
+
+    // build a BlockJacobiPreconditioner and use the vmult(dst,src) for applying the inverse mass
+    // operator on  source the vector
+    mass_preconditioner =
+      std::make_shared<BlockJacobiPreconditioner<MassOperator<dim, n_components, Number>>>(
+        mass_operator);
   }
 
   dealii::MatrixFree<dim, Number> const * matrix_free;
--- a/include/exadg/operators/inverse_mass_operator.h
+++ b/include/exadg/operators/inverse_mass_operator.h
@@ -155,8 +155,6 @@ private:
 
   unsigned int dof_index, quad_index;
 
-  // ExplicitMatrixFreeInverseMass is only available for tensor-product DG elements. For other DG
-  // elements, we use a BlockJacobiPreconditioner
   bool explicit_matrix_free_inverse_mass_available;
 
   MassOperator<dim, n_components, Number> mass_operator;
--- a/include/exadg/operators/inverse_mass_operator.h
+++ b/include/exadg/operators/inverse_mass_operator.h
@@ -45,7 +45,7 @@ private:
 
   // use a template parameter of -1 to select the precompiled version of this operator
   typedef dealii::MatrixFreeOperators::CellwiseInverseMassMatrix<dim, -1, n_components, Number>
-    CellwiseInverseMass;
+    ExplicitMatrixFreeInverseMass;
 
   typedef std::pair<unsigned int, unsigned int> Range;
 
@@ -63,11 +63,11 @@ public:
     dof_index         = dof_index_in;
     quad_index        = quad_index_in;
 
-    cellwise_inverse_mass_not_available = not(matrix_free->get_dof_handler(dof_index)
-                                                .get_triangulation()
-                                                .all_reference_cells_are_hyper_cube());
+    matrixfree_inverse_mass_not_available = not(matrix_free->get_dof_handler(dof_index)
+                                                  .get_triangulation()
+                                                  .all_reference_cells_are_hyper_cube());
 
-    if(cellwise_inverse_mass_not_available)
+    if(matrixfree_inverse_mass_not_available)
     {
       initialize_inverse_mass_operator_with_block_jacobi();
     }
@@ -78,7 +78,7 @@ public:
   {
     dst.zero_out_ghost_values();
 
-    if(cellwise_inverse_mass_not_available)
+    if(matrixfree_inverse_mass_not_available)
     {
       mass_preconditioner->vmult(dst, src);
     }
@@ -95,8 +95,8 @@ private:
             VectorType const & src,
             Range const &      cell_range) const
   {
-    Integrator          integrator(*matrix_free, dof_index, quad_index);
-    CellwiseInverseMass inverse(integrator);
+    Integrator                    integrator(*matrix_free, dof_index, quad_index);
+    ExplicitMatrixFreeInverseMass inverse(integrator);
 
     for(unsigned int cell = cell_range.first; cell < cell_range.second; ++cell)
     {
@@ -133,8 +133,8 @@ private:
 
   unsigned int dof_index, quad_index;
 
-  // BlockJacobi Preconditoner to be used when the CellwiseInverseMassMatrix is not available
-  bool cellwise_inverse_mass_not_available;
+  // BlockJacobi Preconditoner to be used when the MatrixFreeInverseMass is not available
+  bool matrixfree_inverse_mass_not_available;
 
   MassOperator<dim, n_components, Number> mass_operator;
 
--- a/include/exadg/operators/inverse_mass_operator.h
+++ b/include/exadg/operators/inverse_mass_operator.h
@@ -109,7 +109,7 @@ public:
 
       mass_operator.initialize(*matrix_free, constraint, mass_operator_data);
 
-      mass_preconditioner =
+      elementwise_inverse_mass =
         std::make_shared<BlockJacobiPreconditioner<MassOperator<dim, n_components, Number>>>(
           mass_operator);
     }
@@ -126,7 +126,7 @@ public:
     }
     else
     {
-      mass_preconditioner->vmult(dst, src);
+      elementwise_inverse_mass->vmult(dst, src);
     }
   }
 
@@ -160,7 +160,7 @@ private:
   MassOperator<dim, n_components, Number> mass_operator;
 
   std::shared_ptr<BlockJacobiPreconditioner<MassOperator<dim, n_components, Number>>>
-    mass_preconditioner;
+    elementwise_inverse_mass;
 };
 
 } // namespace ExaDG
--- a/include/exadg/operators/inverse_mass_operator.h
+++ b/include/exadg/operators/inverse_mass_operator.h
@@ -80,11 +80,11 @@ public:
     quad_index        = inverse_mass_operator_data.quad_index;
 
     dealii::FiniteElement<dim> const & fe = matrix_free->get_dof_handler(dof_index).get_fe();
-    // this checks if we have a tensor-product element
+    // this checks if we have a tensor-product element, e.g. simplex is out
     if(fe.base_element(0).dofs_per_cell == dealii::Utilities::pow(fe.degree + 1, dim))
     {
-      // this checks if we have a DG element
-      if(fe.dofs_per_vertex == 0)
+      // this checks if all unknows are on the interior of the face, e.g. continuous elements out
+      if((dim == 2 && fe.first_quad_index == 0) || (dim == 3 && fe.first_hex_index == 0))
         explicit_matrix_free_inverse_mass_available = true;
     }
 
--- a/include/exadg/operators/inverse_mass_operator.h
+++ b/include/exadg/operators/inverse_mass_operator.h
@@ -69,7 +69,20 @@ public:
 
     if(not(explicit_matrix_free_inverse_mass_available))
     {
-      initialize_inverse_mass_operator_with_block_jacobi();
+      // initialize mass operator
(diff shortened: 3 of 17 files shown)


nfehn:
At some places in `OperatorBase` one can find code like

```c++
  if(is_dg)
  {
    if(evaluate_face_integrals())
    {
    }
  }
```
It might be that we would have to place an `if(evaluate_face_integrals())` before the assert that you mention.

Other idea: I wonder whether the Assert that you mention could be placed somewhat deeper in the code, maybe closer to or in the face loop itself. If we do not have face loops (as for the mass operator), the assert will not be triggered (the current block jacobi preconditioner implementation probably just assumes that we always have face loops, which does not imply that we cannot relax this assert for your use case). Did you try something in this direction?

include/exadg/operators/inverse_mass_operator.h
```diff
@@ -60,14 +62,50 @@ class InverseMassOperator
...
+                                                .all_reference_cells_are_hyper_cube());
+
+    if(cellwise_inverse_mass_not_available)
+    {
+      initialize_inverse_mass_operator_with_block_jacobi();
+    }
+  }
+
+  void
+  initialize_inverse_mass_operator_with_block_jacobi()
```
nfehn, line 77: This looks like a public function. Why not just inserting the code implemented here in this function into the above if branch?

include/exadg/operators/inverse_mass_operator.h
```diff
@@ -60,14 +62,50 @@ class InverseMassOperator
...
 
-    matrix_free->cell_loop(&This::cell_loop, this, dst, src);
+    if(cellwise_inverse_mass_not_available)
+    {
+      mass_preconditioner->vmult(dst, src);
+    }
+    else
+    {
+      matrix_free->cell_loop(&This::cell_loop, this, dst, src);
+    }
```
nfehn, line 108: I think the variable name `cellwise_inverse_mass_not_available` is not optimal and hinders understanding the code. A cellwise inverse mass is available for all the branches here, otherwise we could not use a block Jacobi preconditioner to invert this operator.
The irritation might come from the naming in deal.II. Cellwise inverse mass actually means that the inverse mass can be applied in a matrix-free fashion using sum-factorization. The emphasis is probably not on "cellwise".

nfehn: I suggest to also rename `CellwiseInverseMass` into `MatrixfreeInverseMass` used in this class. The name was ok before the present PR, but is very irritating with the changes of this PR.

necioglu:
> Did you try something in this direction?

I have been working on this. To summarize, I have observed the two following situations:

1. If one sets `this->param.use_cell_based_face_loops= false;`, the code fails in this loop:
https://github.com/exadg/exadg/blob/9a8888704650f41ada6cdcb9252a5fb0991d7349/include/exadg/operators/operator_base.cpp#L634-L639
Clearly hardcoded for hypercubes. Inside this loop, the function `reinint_face_cell_based` is called and this is where the code fails. We could place an Assert in this for loop.

2.  If one sets `this->param.use_cell_based_face_loops= true;`, the code fails during setup. With setup I mean: `driver.setup()` followed by `matrix_free->reinit(...) `. And this is the dealii related part of the issue. 

And it is important to mention: If we set `this->param.implement_block_diagonal_preconditioner_matrix_free=true` this is done for every operator containing a block jacobi preconditioner. With the current design, we can not say do matrix-free for mass operator and matrix-based for e.g. viscous operator.

These are my observations only. I will think about a proper location for the Assert.

include/exadg/operators/inverse_mass_operator.h
```diff
@@ -60,14 +62,50 @@ class InverseMassOperator
     this->matrix_free = &matrix_free_in;
     dof_index         = dof_index_in;
     quad_index        = quad_index_in;
+
+    cellwise_inverse_mass_not_available = not(matrix_free->get_dof_handler(dof_index)
```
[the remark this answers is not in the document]
necioglu: But cellwise inverse mass is only not available for H(div) velocity elements, correct? The pressure elements are not effected.

include/exadg/operators/inverse_mass_operator.h
necioglu: > Why not just inserting the code implemented here in this function into the above if branch?

I didn't do this to keep the code easier to read. Would it be a good solution to make this function private?

btemuer: Member functions that are not part of the public interface (used to interact with the object from the outside) should always be private/protected.

include/exadg/operators/inverse_mass_operator.h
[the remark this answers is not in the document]
nfehn: yes. The implementation of inverse mass operator is generic and does not make a difference between velocity/pressure.

nfehn:
My intention was to set `implement_block_diagonal_preconditioner_matrix_free=false` so that all other operators are working correctly, in the hope that the block jacobi preconditioner still works because it never has to evaluate face integrals for the mass operator. One would probably need to adjust the current implementation in ExaDG. We could for example figure this out in person.

include/exadg/operators/inverse_mass_operator.h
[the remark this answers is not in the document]
necioglu: What do you think about using the exact Assert as in dealii: https://github.com/necioglu/dealii/blob/c1ef6fab04853405c037b3eb759482b4d30e5aac/include/deal.II/matrix_free/evaluation_kernels.h#L5601-L5603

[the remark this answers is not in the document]
nfehn: I do not understand the `<=` in this assert, and there is no error message apart from "not implemented". I think we need our own ExaDG version here.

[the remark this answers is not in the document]
kronbichler: The `<=` pertains to the fact that the implementation in deal.II detects different symmetries in the basis functions: https://github.com/necioglu/dealii/blob/c1ef6fab04853405c037b3eb759482b4d30e5aac/include/deal.II/matrix_free/shape_info.h#L60-L120 - such as collocation, support for even-odd decomposition, or additional shape functions besides the regular polynomial ones. In a sense `element_type` is not the best name, but it can be explained from the way it evolved.

include/exadg/operators/inverse_mass_operator.h
necioglu: made the function private in ad3d8b5b3b30e00d0a6242ef35c038ed8e4a8c2b.

include/exadg/operators/inverse_mass_operator.h
necioglu: I agree. Changed the name to `ExplicitMatrixfreeInverseMass`, to distinguish between matrix-free CG solver and the explicit matrix-free operator of dealii.

necioglu:
@nfehn I have moved the asssert from `parameters.cpp` to `operator_base.h`, as discussed. Additionally, I have added asserts for simplex elements, where a face loop is done.

include/exadg/operators/inverse_mass_operator.h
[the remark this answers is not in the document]
necioglu: This is the only topic I have not addressed yet. I couldn't find a good looking solution for checking if the element is FE_DGQ, rather then:

```
matrix_free->get_dof_handler(dof_index).get_fe() == 
dealii::FE_DGQ<dim>(matrix_free->get_dof_handler(dof_index).get_fe().degree);
```

And additionally we have the check for `.all_reference_cells_are_hyper_cube()`. This is why the statement doesn't look good. Do we have an easier way to check if the element is FE_DGQ?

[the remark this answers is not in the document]
btemuer: You could use `dynamic_cast` instead of creating a new object and comparing. It will be successful or fail accordingly.

include/exadg/operators/inverse_mass_operator.h
nfehn: I do not see a need to have a separate private member function, but we do it in a similar way at other places in ExaDG, so its ok. (From a software design perspective, some people might argue that it is needless complexity to introduce a function here; others might argue that the code is structured in a better way, more readable ...).

include/exadg/operators/inverse_mass_operator.h
```diff
@@ -60,14 +62,30 @@ class InverseMassOperator
     this->matrix_free = &matrix_free_in;
     dof_index         = dof_index_in;
     quad_index        = quad_index_in;
+
+    explicit_matrix_free_inverse_mass_available = matrix_free->get_dof_handler(dof_index)
```
nfehn, line 66: I think this is not enough? We only have the explicit matrix-free inverse mass for DG elements (L2 conforming), so that the mass matrix is block-diagonal.

include/exadg/operators/inverse_mass_operator.h
```diff
@@ -60,14 +62,30 @@ class InverseMassOperator
...
 
-    matrix_free->cell_loop(&This::cell_loop, this, dst, src);
+    if(not(explicit_matrix_free_inverse_mass_available))
+    {
+      mass_preconditioner->vmult(dst, src);
+    }
+    else
+    {
+      matrix_free->cell_loop(&This::cell_loop, this, dst, src);
+    }
```
nfehn, line 88: You can get rid of the `not` by changing the order of the two branches. This makes the code much easier to read.

include/exadg/operators/inverse_mass_operator.h
```diff
@@ -91,9 +109,36 @@ class InverseMassOperator
...
+    mass_preconditioner =
+      std::make_shared<BlockJacobiPreconditioner<MassOperator<dim, n_components, Number>>>(
+        mass_operator);
+  }
+
   dealii::MatrixFree<dim, Number> const * matrix_free;
 
   unsigned int dof_index, quad_index;
+
+  // BlockJacobi Preconditoner to be used when the MatrixFreeInverseMass is not available
```
nfehn, line 136: I think this comment belongs to line 141 (preconditioner) below. I would like to avoid the negation in "is not available" here, again for code readability. Make a positive statement like "For special cases, deal.II provides fast matrix-free routines to invert the mass operator in an explicit way".

include/exadg/operators/inverse_mass_operator.h
```diff
@@ -91,9 +109,36 @@ class InverseMassOperator
...
+  bool explicit_matrix_free_inverse_mass_available;
+
+  MassOperator<dim, n_components, Number> mass_operator;
+
+  std::shared_ptr<PreconditionerBase<Number>> mass_preconditioner;
```
nfehn, line 141: Here, you could use directly `BlockJacobiPreconditioner` instead of `PreconditionerBase` for clarity. We do not need dynamic polymorphism in this case.

include/exadg/operators/inverse_mass_operator.h
```diff
@@ -91,9 +109,36 @@ class InverseMassOperator
...
+      matrix_free->get_affine_constraints(dof_index);
+
+    MassOperatorData<dim> mass_operator_data;
+    mass_operator_data.dof_index  = dof_index;
+    mass_operator_data.quad_index = quad_index;
+
+    mass_operator.initialize(*matrix_free, constraint, mass_operator_data);
+
+    // build a BlockJacobiPreconditioner and use the vmult(dst,src) for applying the inverse mass
+    // operator on  source the vector
```
nfehn, line 126: I do not understand "on source the vector". I think we can just skip the whole comment here.

include/exadg/operators/inverse_mass_operator.h
```diff
@@ -91,9 +109,36 @@ class InverseMassOperator
     }
   }
 
+  void
+  initialize_inverse_mass_operator_with_block_jacobi()
+  {
+    // initialize mass operator
+    dealii::AffineConstraints<Number> const & constraint =
+      matrix_free->get_affine_constraints(dof_index);
```
nfehn, line 117: I am not sure whether we want constraints here, see the mass operator in the `Structure` module https://github.com/exadg/exadg/blob/41f8752a19a667b64009fcd6f4325a087afe8baa/include/exadg/structure/spatial_discretization/operator.cpp#L156-L158 for comparison.

include/exadg/operators/inverse_mass_operator.h
```diff
@@ -91,9 +109,36 @@ class InverseMassOperator
     }
   }
 
+  void
+  initialize_inverse_mass_operator_with_block_jacobi()
```
btemuer, line 113: I'm not sure this is a good name for this function. To me, this name sounds like _initialize variable with 0_, which is not what this function does, if you know what I mean.  I would suggest something like `construct_block_jacobi_preconditioner()`. What do you think?

btemuer:
While we are at it, is this a good place to start enforcing documentation for classes and public methods in ExaDG? Just asking.. `InverseMassOperator` is fairly small, it could be done in this PR.

include/exadg/operators/inverse_mass_operator.h
```diff
@@ -91,9 +109,36 @@ class InverseMassOperator
...
+  {
+    // initialize mass operator
+    dealii::AffineConstraints<Number> const & constraint =
+      matrix_free->get_affine_constraints(dof_index);
+
+    MassOperatorData<dim> mass_operator_data;
+    mass_operator_data.dof_index  = dof_index;
+    mass_operator_data.quad_index = quad_index;
+
+    mass_operator.initialize(*matrix_free, constraint, mass_operator_data);
```
nfehn, line 123: Why creating a new mass operator? Doesn't it already exist.

The variable `implement_block_diagonal_preconditioner_matrix_free` in operator data is not set here. Hence, I do not see how the present implementation of the inverse mass operator can support the matrix-free variant with elementwise iterative solver.

Note also that the solver tolerances in operator data are by default very coarse (reltol = 1e-2), since the block-jacobi preconditioner is typically used as a preconditioner. Here, however, the block Jacobi preconditioner is used as a solver, i.e. to invert the mass operator. With the default setting, I expect that spatial errors and convergence rates for simplices are severly limited, because solver tolerances for the inverse mass operator dominate over all other errors of the spatial discretization for sufficiently fine spatial resolutions.

Having said this, it could indeed make sense to create a new mass operator here, but the current implementation seems to be incomplete. Did you test the different variants?

include/exadg/operators/operator_base.cpp
```diff
@@ -637,6 +643,11 @@ OperatorBase<dim, Number, n_components>::apply_add_block_diagonal_elementwise(
     unsigned int const n_faces = dealii::ReferenceCells::template get_hypercube<dim>().n_faces();
     for(unsigned int face = 0; face < n_faces; ++face)
     {
+      AssertThrow(matrix_free->get_dof_handler(data.dof_index)
+                    .get_triangulation()
+                    .all_reference_cells_are_hyper_cube(),
+                  dealii::ExcMessage("Can't do cell based loop over faces for simplices."));
```
nfehn, line 649: I think we should not place these 3 Asserts inside the innermost loops over the faces (for performance reasons). Why not in `apply_inverse_block_diagonal()` where you have added the other assert.

include/exadg/operators/operator_base.cpp
```diff
@@ -541,6 +541,12 @@ OperatorBase<dim, Number, n_components>::apply_inverse_block_diagonal(VectorType
   // matrix-free
   if(this->data.implement_block_diagonal_preconditioner_matrix_free)
   {
+    AssertThrow((evaluate_face_integrals() and data.use_cell_based_loops) or
```
nfehn, line 544: It seems as if you could simply write

```c++
if(evaluate_face_integrals())
{
  AssertThrow(data.use_cell_based_loops, ...);
}
```
as already suggested here https://github.com/exadg/exadg/pull/380#issuecomment-1463802569. This would be much simpler to understand in my opinion.

include/exadg/operators/inverse_mass_operator.h
nfehn: I agree here. For me this is a sign, that the new member function introduced here is somewhat superfluous.

include/exadg/operators/inverse_mass_operator.h
necioglu: Yes, see the discussion above https://github.com/exadg/exadg/pull/380#discussion_r1134306730

include/exadg/operators/inverse_mass_operator.h
necioglu: To address both reviews at once, I will remove the function and add it to `initialize`.

include/exadg/operators/operator_base.cpp
necioglu: I wanted to add the Assert deeper in code. I didn't know it could lead to worse performance.

include/exadg/operators/inverse_mass_operator.h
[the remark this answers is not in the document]
kronbichler: > And additionally we have the check for .all_reference_cells_are_hyper_cube(). This is why the statement doesn't look good. Do we have an easier way to check if the element is FE_DGQ?

We don't need `FE_DGQ`, do we? In fact, all tensor-product DG elements work, and also systems of `FE_DGQ`. So the check is probably more complicated than that (on the positive side, you do not need `all_reference_cells_are_hyper_cube`):
```c++
  FiniteElement<dim> const &fe = matrix_free->get_dof_handler(dof_index).get_fe();
  if (((dim == 2 && fe.first_quad_index == 0) || (dim == 3 && fe.first_hex_index == 0))
      &&
      fe.base_element(0).dofs_per_cell == Utilities::pow(fe.degree + 1, dim))
```
The first check looks whether all unknowns are located on the interior of the cells (for which, unfortunately, I do not think there is a dimension-independent way to look up), whereas the second checks that we have a tensor product. Does that make sense to you? There might be simpler ways to achieve what I want (so feel free to improve on this), but I think that is the important features.

include/exadg/operators/inverse_mass_operator.h
necioglu: This makes me think about the original approach I had, because passing `implement_block_diagonal_preconditioner_matrix_free` doesn't seem easy. Maybe we need to construct the mass operator in spatial_operator_base and pass over to inverse_mass_operator, by extending the interface. But this would result in if-else branches in higher levels. This was the argument against this approach.

necioglu:
@nfehn @kronbichler  It seems to me like adding the matrix-free variant with elementwise iterative solver would complicate this PR, since this would require to change interface of `inverse_mass_operator`. We would need to initialize `inverse_mass_operator` with `implement_block_diagonal_preconditioner_matrix_free` and also the `SolverData`. I think the question of if we should do this, requires a separate discussion.

The variant with the explicit assembly of the matrices works fine and therefore I suggest we move forward only with this variant and add the elementwise iterative solver later in a separate PR, if needed, keeping in mind that I have limited time until the end of my Master's thesis. What do you think?

Of course, I will address the other reviews apart from this.

include/exadg/operators/inverse_mass_operator.h
nfehn: As I said above, it can make sense to create a new mass operator here, I am not per se against it but want to discuss it. With the arguments you provided (velocity vs. pressure), I think this is a reasonable decision for now.

nfehn:
> @nfehn @kronbichler It seems to me like adding the matrix-free variant with elementwise iterative solver would complicate this PR, since this would require to change interface of inverse_mass_operator. We would need to initialize inverse_mass_operator with implement_block_diagonal_preconditioner_matrix_free and also the SolverData. I think the question of if we should do this, requires a separate discussion.

My opinion on this is that the question which of the two variants is more efficient is not clear a priori, is interesting, and could be a relevant part of numerical experiments in your masters thesis (let's put it this way: I would do it in your situation). Did you check both variants in terms of performance? For this, you can simply hard-code the two variants in the inverse mass operator (and recompile the code when switching to the other variant), without any need to adjust interfaces in ExaDG. For such investigations, I recommend to study both 2d and 3d as well as different problems sizes and polynomial degrees, see e.g. figure 4.8 in my thesis (https://mediatum.ub.tum.de/doc/1601025/1601025.pdf). The executable to run such throughput studies for incompressible Navier-Stokes are already prepared for the inverse mass operator (see figure 4.10), you only need to switch to a simplex mesh.

From the perspective of software development, I would not merge the current state without further asserts, because we already have more information/knowledge which is currently not present in the code. The minimal solution would be to assert that `implement_block_diagonal_preconditioner_matrix_free = false` when constructing the inverse mass operator. Otherwise, users might set `implement_block_diagonal_preconditioner_matrix_free = true` and expect that a matrix-free variant is used internally, but never get to know that this is not the case (even though we already know this). Is it really that much effort to write a struct `InverseMassOperatorData` and pass this to the constructor/initialization of the inverse mass operator?

```c++
struct InverseMassOperatorData
{
  unsigned int dof_index;
  unsigned int quad_index;

  // only relevant if an explicit matrix-free inverse mass operator is not available
  bool implement_block_diagonal_preconditioner_matrix_free;

  // only relevant if elementwise mass operators are inverted by elementwise 
  // iterative solvers with matrix-free implementation
  SolverData solver_data_block_diagonal;
}
```

and
```c++
  void
  initialize(dealii::MatrixFree<dim, Number> const & matrix_free_in,
                  InverseMassOperatorData & inverse_mass_data_in)
  {
     // TODO fill MassOperatorData using InverseMassOperatorData
    // note that Elementwise::Solver and Elementwise::Preconditioner do not need to be part 
    // of InverseMassOperatorData since we know this a priori, namely CG solver and precondition 
    // identity. Hence, these elements of OperatorBaseData can be hard-coded here.
  }
```

necioglu:
> For this, you can simply hard-code the two variants in the inverse mass operator (and recompile the code when switching to the other variant), without any need to adjust interfaces in ExaDG.

This is what I did after seeing this comment https://github.com/exadg/exadg/pull/380#discussion_r1135323765. My initial tries didn't work well, also after changing the solver tolerances. Didn't work well, meaning I had convergence problem from a different part of the code. I will invest more time in this.

> Is it really that much effort to write a struct `InverseMassOperatorData` and pass this to the constructor/initialization of the inverse mass operator?

No, it is easy. I had a similar design already in the past. My concern was not about the difficulty of the implementation, but about the code design aspect.

include/exadg/operators/inverse_mass_operator.h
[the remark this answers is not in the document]
necioglu: I think I could simplify the first check with 
`fe.dofs_per_vertex == 0`
since it checks if we have DG elements, correct?

include/exadg/convection_diffusion/user_interface/parameters.h
```diff
@@ -318,6 +318,18 @@ class Parameters
 
   // use 3/2 overintegration rule for convective term
   bool use_overintegration;
+
+  /**************************************************************************************/
+  /*                                                                                    */
+  /*                            ELEMENTWISE INVERSE MASS                                */
+  /*                                                                                    */
+  /**************************************************************************************/
```
nfehn, line 326: I think we do not need a separate section for elementwise inverse mass. I think we could add these parameters to the SOLVER parameters above.

Another question is why you introduce these parameters here given that simplex meshes have not been realized so far for the ConvDiff module? Of course, one could use the elementwise iterative solver also for hex-elements, but there we know that the explicit matrix-free inverse mass operator is most efficient.

Generally, my perspective is to only introduce parameters in the user interface if they are relevant. So I would suggest to introduce these parameters once simplex is realized for ConvDiff. Since these parameters will then already exist in the IncNS module, it will be easy to copy that code to ConvDiff (once/if simplex will be realized for ConvDiff).

include/exadg/compressible_navier_stokes/user_interface/parameters.h
```diff
@@ -192,6 +193,18 @@ class Parameters
...
+  /*                                                                                    */
+  /*                            ELEMENTWISE INVERSE MASS                                */
+  /*                                                                                    */
+  /**************************************************************************************/
+  // Used when matrix-free inverse mass operator is not available and when the spatial
+  // discretization is DG, e.g. simplex.
+
+  bool solve_elementwise_mass_system_matrix_free;
+
+  SolverData solver_data_elementwise_inverse_mass;
```
nfehn, line 207: Same as for ConvDiff module, see comment https://github.com/exadg/exadg/pull/380/files#r1176622661

include/exadg/operators/operator_base.cpp
```diff
@@ -549,8 +561,7 @@ OperatorBase<dim, Number, n_components>::initialize_block_diagonal_preconditione
   {
     typedef Elementwise::PreconditionerIdentity<dealii::VectorizedArray<Number>> IDENTITY;
 
-    elementwise_preconditioner =
-      std::make_shared<IDENTITY>(elementwise_operator->get_problem_size());
+    elementwise_preconditioner = std::make_shared<IDENTITY>(integrator->dofs_per_cell);
```
nfehn, line 564: To which extent is this change related to the present PR, could you explain this?

include/exadg/operators/inverse_mass_operator.h
```diff
@@ -94,6 +154,13 @@ class InverseMassOperator
   dealii::MatrixFree<dim, Number> const * matrix_free;
 
   unsigned int dof_index, quad_index;
+
+  bool explicit_matrix_free_inverse_mass_available;
+
+  MassOperator<dim, n_components, Number> mass_operator;
+
+  std::shared_ptr<BlockJacobiPreconditioner<MassOperator<dim, n_components, Number>>>
+    mass_preconditioner;
```
nfehn, line 163: In the context of this class (`InverseMassOperator`), the name `elementwise_inverse_mass` would be more intuitive than `mass_preconditioner`.

include/exadg/operators/inverse_mass_operator.h
```diff
@@ -28,9 +28,24 @@
...
 #include <exadg/matrix_free/integrators.h>
+#include <exadg/operators/mass_operator.h>
+#include <exadg/solvers_and_preconditioners/preconditioners/block_jacobi_preconditioner.h>
 
 namespace ExaDG
 {
+struct InverseMassOperatorData
+{
+  unsigned int dof_index;
+  unsigned int quad_index;
```
nfehn, line 39: Let's also initialize these two parameters explicitly in this struct. The default value is `0`.

include/exadg/operators/inverse_mass_operator.h
```diff
@@ -43,31 +58,76 @@ class InverseMassOperator
...
   initialize(dealii::MatrixFree<dim, Number> const & matrix_free_in,
-             unsigned int const                      dof_index_in,
-             unsigned int const                      quad_index_in)
+             InverseMassOperatorData const           inverse_mass_operator_data)
   {
     this->matrix_free = &matrix_free_in;
-    dof_index         = dof_index_in;
-    quad_index        = quad_index_in;
+    dof_index         = inverse_mass_operator_data.dof_index;
+    quad_index        = inverse_mass_operator_data.quad_index;
```
nfehn, line 80: I am not sure if we should store both dof_index and quad_index as separate element variables in this class, while we could also store the inverse mass operator data only as an element variable.

nfehn:
The overall design is good now. To get this PR merged, I suggest to not introduce certain parameters for the inverse mass operator for those modules in ExaDG currently not implementing simplex elements (ConvDiff and CompNS).

necioglu:
> The overall design is good now. To get this PR merged, I suggest to not introduce certain parameters for the inverse mass operator for those modules in ExaDG currently not implementing simplex elements (ConvDiff and CompNS).

Thanks for the review. I will request a review once I do the changes.

include/exadg/operators/operator_base.cpp
necioglu: This is a bugfix.

In this line we initialize the `elementwise_preconditioner` but the `elementwise_operator` is not initialized yet. This is why the value `elementwise_operator->get_problem_size()` is just the default value and not correct. 

I couldn't find any use case of the identity elementwise preconditioner, apart from the use case introduced in this PR. This is probably why, this bug was not found before.

include/exadg/operators/inverse_mass_operator.h
necioglu: If we store the inverse mass operator data in this class, we store the bool variable  `implement_block_diagonal_preconditioner_matrix_free` and the solver data `solver_data_block_diagonal` as well, although they are only needed in the initialize function. Whereas, dof_index and quad_index are continuously needed in the cell_loop function.

nfehn (approved):

PR #384, opened 2023-03-14
Refactoring of affine constraints in OperatorBase

### LINKED ISSUE #383: Handling of constrained dofs in matrix-free and iterative solvers
The `vmult()` function in `OperatorBase` (applying the homogeneous operator) sets constrained dofs to zero before evaluating the actual matrix-vector product (in a matrix-free fashion), see https://github.com/exadg/exadg/blob/41f8752a19a667b64009fcd6f4325a087afe8baa/include/exadg/operators/operator_base.cpp#L282-L304

but the matrix free loops use `gather_evaluate()`, a combination of `read_dof_values()` and `evaluate()`, see https://github.com/exadg/exadg/blob/41f8752a19a667b64009fcd6f4325a087afe8baa/include/exadg/operators/operator_base.cpp#L1027-L1045

 The function `read_dof_values()` should actually use the constrained dofs passed to matrix-free via `AffineConstraints` (in contrast to `read_dof_values_plain()`), so there would be no need to set constrained dofs to zero in the src-vector in my opinion.
 
 For these reasons, I believe the current implementation is somewhat unclear or non-unique in terms of how to impose constraints for continuous Galerkin. I see two solutions to make the implementation clear or unique:

1. set constrained dofs to zero before `vmult()`, but then use `read_dof_values_plain()`.  Then, we can probably no longer write `gather_evaluate()`?
2. skip setting constrained dofs to zero before `vmult()`, and use `gather_evaluate()` (internally using `read_dof_values()`) as is done currently in `OperatorBase`. Further, add a comment to https://github.com/exadg/exadg/blob/41f8752a19a667b64009fcd6f4325a087afe8baa/include/exadg/poisson/spatial_discretization/operator.cpp#L147 and emphasize that `make_zero_boundary_constraints()` is essential, i.e. calling `interpolate_boundary_values()` for a problem with inhomogeneous boundary conditions would lead to wrong results (because `read_dofs_values()` uses these inhomogeneous boundary values, but we would actually need the homogeneous operator in `vmult()`).

@kronbichler @peterrum Do you agree? I would tend to realize option 2 since this simplifies the implementation in ExaDG and we should be able to just remove some lines of code in `OperatorBase` (making clear how contraints are applied in ExaDG and removing the ambiguity we have currently according to my understanding). Any concerns e.g. regarding performance?

nfehn:
closes #383

### DIFF
--- a/include/exadg/operators/operator_base.cpp
+++ b/include/exadg/operators/operator_base.cpp
@@ -260,6 +260,7 @@ OperatorBase<dim, Number, n_components>::apply(VectorType & dst, VectorType cons
   if(is_dg)
   {
     if(evaluate_face_integrals())
+    {
       matrix_free->loop(&This::cell_loop,
                         &This::face_loop,
                         &This::boundary_face_loop_hom_operator,
@@ -267,8 +268,11 @@ OperatorBase<dim, Number, n_components>::apply(VectorType & dst, VectorType cons
                         dst,
                         src,
                         true);
+    }
     else
+    {
       matrix_free->cell_loop(&This::cell_loop, this, dst, src, true);
+    }
   }
   else
   {
@@ -295,10 +299,14 @@ OperatorBase<dim, Number, n_components>::apply_add(VectorType & dst, VectorType 
   if(is_dg)
   {
     if(evaluate_face_integrals())
+    {
       matrix_free->loop(
         &This::cell_loop, &This::face_loop, &This::boundary_face_loop_hom_operator, this, dst, src);
+    }
     else
+    {
       matrix_free->cell_loop(&This::cell_loop, this, dst, src);
+    }
   }
   else
   {
--- a/include/exadg/operators/operator_base.cpp
+++ b/include/exadg/operators/operator_base.cpp
@@ -81,13 +81,6 @@ OperatorBase<dim, Number, n_components>::reinit(
                                                   this->data.dof_index,
                                                   this->data.quad_index);
 
-  if(not(is_dg))
-  {
-    constrained_indices.clear();
-    for(auto i : this->matrix_free->get_constrained_dofs(this->data.dof_index))
-      constrained_indices.push_back(i);
-  }
-
   // set multigrid level
   this->level = this->matrix_free->get_mg_level();
 
@@ -234,9 +227,10 @@ template<int dim, typename Number, int n_components>
 void
 OperatorBase<dim, Number, n_components>::set_constrained_values_to_zero(VectorType & vector) const
 {
-  for(unsigned int i = 0; i < constrained_indices.size(); ++i)
+  for(unsigned int const constrained_index :
+      matrix_free->get_constrained_dofs(this->data.dof_index))
   {
-    vector.local_element(constrained_indices[i]) = 0.0;
+    vector.local_element(constrained_index) = 0.0;
   }
 }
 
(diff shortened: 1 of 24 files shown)


include/exadg/operators/operator_base.cpp
```diff
@@ -262,45 +260,34 @@ OperatorBase<dim, Number, n_components>::apply(VectorType & dst, VectorType cons
...
-    // constrained values can be written back to the src vector, i.e. having completed
-    // the function apply(), the src vector should be the same as before calling this
-    // function.
+    // dst vector.
     for(unsigned int i = 0; i < constrained_indices.size(); ++i)
     {
-      const_cast<VectorType &>(src).local_element(constrained_indices[i]) =
-        constrained_values_src[i];
-      dst.local_element(constrained_indices[i]) = constrained_values_src[i];
+      dst.local_element(constrained_indices[i]) = src.local_element(constrained_indices[i]);
```
kronbichler, line 290: Do we still need this? Since we fill `constrained_indices` by `this->matrix_free->get_constrained_dofs(this->data.dof_index)`, we could also just do it here:
```c++
  for (const unsigned int constrained_index : matrix_free->get_constrained_dofs(this->data.dof_index)
    dst.local_element(constrained_index) = src.local_element(constrained_index);
```

include/exadg/poisson/spatial_discretization/operator.cpp
```diff
@@ -131,7 +131,11 @@ Operator<dim, n_components, Number>::distribute_dofs()
 
   dof_handler.distribute_dofs(*fe);
 
-  // affine constraints only relevant for continuous FE discretization
+  // Affine constraints are only relevant for continuous Galerin discretization.
+  // The AffineConstraints object is used to initialize MatrixFree. Here, we apply homogeneous
+  // boundary conditions as needed by vmult() in iterative solvers for linear systems of equations,
+  // implemented via dealii::MatrixFree and FEEvaluation::read_dof_values() (or gather_evaluate()).
+  // The actual inhomogeneous boundary data needs to be imposed separately and explicitly in ExaDG.
```
kronbichler, line 138: I think this is a good summary, but it sounds a bit like it would be ExaDG's fault. It is rather a design decision (or general limitation) in the implementation of `MatrixFree` and the related evaluators `FEEvaluation` in deal.II.

nfehn: Suggestion: "// The actual inhomogeneous boundary data needs to be imposed separately due to the implementation in deal.II that can not handle multiple AffineConstraints. Hence, we need to do this explicitly in ExaDG."

What do you think about this version, @kronbichler?

kronbichler: Yes, that is good.

kronbichler (approved):

include/exadg/poisson/spatial_discretization/operator.cpp
```diff
@@ -131,7 +131,11 @@ Operator<dim, n_components, Number>::distribute_dofs()
...
   if(param.spatial_discretization == SpatialDiscretization::CG)
   {
     affine_constraints.clear();
```
kronbichler, line 141: Just that I get your intentions right: In the code below, we would need to add `dealii::DoFTools::make_hanging_node_constraints` in the future to make it compatible with hanging nodes, but it is not added in the present PR?

nfehn: yes exactly, see PR #388

include/exadg/operators/operator_base.cpp
kronbichler: The function returns exactly a reference to `std::vector<unsigned int>`, https://github.com/dealii/dealii/blob/c9b6a28e288d5c9d955ee5a3befd181bcda3b8d9/include/deal.II/matrix_free/matrix_free.h#L2432-L2438 - so the underlying cost of the range-based for loop should be exactly the same (apart from one or two additional indirection per outer call to `vmult` to follow into the variables of `matrix_free`, which is essentially nothing in the grand scheme of things).

nfehn:
I think the discussion in issue #386 did not identify an optimal strategy to handle inhomogeneous boundary data in ExaDG. For the sake of a consistent code design, I think we should wait with the present PR until the picture is more clear, because for one option mentioned in #386 we would then need to revert the changes of the present PR again.

nfehn:
@kronbichler The options discussed in #386 seem to be a separate/second `AffineConstraints` object, or new functionality in `dealii::MatrixFree`. This would mean that we can proceed with the present PR, right?

kronbichler (approved):
Yes, this looks good as a first step to simplify interfaces.

PR #389, opened 2023-03-15
Enable periodic faces for global-coarsening multigrid

### LINKED ISSUE #387: GridUtilities::create_triangulation() does not take into account periodic faces
The newly introduced utility functions in `GridUtilities` do not take into account periodic faces. It seems as if we need to extend the interfaces such that `lambda_create_triangulation` can impose/set periodic faces.

This raises the question whether the `Grid` class (see https://github.com/exadg/exadg/blob/master/include/exadg/grid/grid.h) is well-designed at all? It currently looks like

```c++
class Grid
{
  ....

  std::shared_ptr<dealii::Triangulation<dim>> triangulation;

  std::vector<std::shared_ptr<dealii::Triangulation<dim> const>> coarse_triangulations;

  PeriodicFaces periodic_faces;

  std::shared_ptr<dealii::Mapping<dim>> mapping;
}
```

- For global coarsening multigrid, I assume we need a vector of `PeriodicFaces`, right? See also issue https://github.com/exadg/exadg/issues/282.
- Issue https://github.com/exadg/exadg/issues/322 points to similar problems in terms of mappings for global-coarsening multigrid.

SUGGESTION: Wouldn't it make more sense to design the `Grid` class the following way:

```c++
class Grid
{
  ....

  std::shared_ptr<dealii::Triangulation<dim>> triangulation;

  PeriodicFaces periodic_faces;

  std::shared_ptr<dealii::Mapping<dim>> mapping;
}
```
and then pass `Grid fine_grid` as well as `std::vector<Grid> coarse_grids` to multigrid (for global-coarsening multigrid) as well as to the grid utility functions and `lambda_create_triangulation()`. With this design, a `Grid` refers to a single multigrid level. It is essentially a struct grouping together `triangulation`, `periodic_faces`, `mapping`.

This is somewhat a prerequisite to proceed with #385 and #282, which is why I consider this of HIGH priority.

@kronbichler @peterrum @bergbauer @richardschu Please share your opinion.

nfehn:
closes #387

### DIFF
--- a/include/exadg/grid/grid.h
+++ b/include/exadg/grid/grid.h
@@ -47,7 +47,15 @@ public:
   /**
    * Constructor.
    */
-  Grid(GridData const & data, MPI_Comm const & mpi_comm);
+  Grid()
+  {
+  }
+
+  /**
+   * Initialize function
+   */
+  void
+  initialize(GridData const & data, MPI_Comm const & mpi_comm);
 
   /**
    * dealii::Triangulation.
--- a/include/exadg/grid/grid.h
+++ b/include/exadg/grid/grid.h
@@ -45,37 +45,9 @@ public:
     PeriodicFacePairs;
 
   /**
-   * dealii::Triangulation.
-   */
-  std::shared_ptr<dealii::Triangulation<dim>> triangulation;
-
-  /**
-   * dealii::GridTools::PeriodicFacePair's.
-   */
-  PeriodicFacePairs periodic_faces;
-
-  /**
-   * dealii::Mapping.
-   */
-  std::shared_ptr<dealii::Mapping<dim>> mapping;
-};
-
-/**
- * A grid manager class that we need in order to support global-coarsening multigrid and keep
- * interfaces lean.
- */
-template<int dim>
-class GridManager
-{
-public:
-  typedef typename std::vector<
-    dealii::GridTools::PeriodicFacePair<typename dealii::Triangulation<dim>::cell_iterator>>
-    PeriodicFaces;
-
-  /**
    * Constructor.
    */
-  GridManager()
+  Grid()
   {
   }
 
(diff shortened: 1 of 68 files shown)


include/exadg/grid/grid.h
```diff
@@ -68,6 +99,23 @@ class Grid
...
+   * a vector of dealii::GridTools::PeriodicFacePair's for the coarse triangulations required for
+   * global coarsening multigrid.
+   */
+  std::vector<PeriodicFaces> coarse_periodic_faces;
+
+  /**
+   * a vector of dealii::Mapping's for the coarse triangulations required for global coarsening
+   * multigrid.
+   */
+  std::vector<std::shared_ptr<dealii::Mapping<dim>>> coarse_mapping;
```
kronbichler, line 118: Do you plan to switch these three variables to
```c++
  std::vector<Grid<dim>> coarse_grids;
```
in the future?

nfehn: happening right now

nfehn: Things turned out much more complicated than I thought yesterday. "Hiding" mapping inside `Grid` inside `GridManager` is not compatible with the Moving Mesh (ALE) functionality in ExaDG, which would need some major refactoring that should not get mixed with the intention of the present PR, namely allowing periodic faces for global coarsening multigrid. I think we need to do things step-by-step and for now just insert a vector of PeriodicFaces for all the coarse triangulations into the class `Grid`. Five hours of making changes to the code indicate once more to work as much as possible on plain deal.II data types since this allows most stable interfaces, even though they might be growing over time (like the one for our multigrid implementation).

nfehn:
I think I have now included everything to enable periodic boundary conditions for global-coarsening multigrid.

include/exadg/grid/grid_utilities.h
```diff
@@ -90,6 +94,45 @@ get_mesh_smoothing(bool const use_local_smoothing_multigrid, ElementType const &
   return mesh_smoothing;
 }
 
+
+/**
```
nfehn, line 98: copied from PR #388 for the time being

kronbichler (approved):

nfehn:
It think we should first merge #388, because that PR originally introduces the utility function for periodic face pairs.

kronbichler (approved):

PR #394, opened 2023-03-23
[WIP] use target link library instead of macro

jh66637:


nfehn:
@jh66637 Could you explain how/whether this is related to https://github.com/dealii/dealii/pull/14962?

jh66637:
This is somewhat independent. With the fix in https://github.com/dealii/dealii/pull/14962 everything should work again. While we need the CMake macro for the tests, we do not necessarily need it for linking the executables. Just using ```target_link_libraries()``` is the modern way how to use CMake. This way we can also use the keywords under discussion independent of the macro in deal.II, so I would consider it to be more stable.

scripts/config_exadg.sh
```diff
@@ -37,6 +37,8 @@ FFTW=$WORKING_DIRECTORY/fftw/install
 LIKWID=$WORKING_DIRECTORY/likwid/install
 
 cmake \
+    -D CMAKE_BUILD_TYPE="DebugRelease" \
```
nfehn, line 40: What is the meaning of `DebugRelease` here? In the CMakeLists.txt file we have `ADD_CUSTOM_TARGET` where we can switch between debug and release?

CMakeLists.txt
```diff
@@ -25,7 +25,7 @@
 #
 #########################################################################
 
-CMAKE_MINIMUM_REQUIRED(VERSION 3.1.0)
+CMAKE_MINIMUM_REQUIRED(VERSION 3.16)
```
nfehn, line 28: Has anyone concerns regarding version 3.16? Is this decision final on the deal.II side?

cmake/macros/macro_exadg_pickup_exe.cmake
```diff
@@ -28,8 +28,6 @@
 MACRO(EXADG_PICKUP_EXE FILE_NAME TARGET_NAME EXE_NAME)
```
nfehn, line 28: It looks like `EXE_NAME` is no longer used?

nfehn:
Thanks for improving/simplying cmake content in ExaDG. For me it looks like the cmake flags and the changes in CMakeLists.txt (and the macro) are somewhat independent changes that would optimally be realized as separate PRs.

cmake/macros/macro_exadg_pickup_exe.cmake
jh66637: Yes, there is no need anymore. I have to remove it from the interface.

scripts/config_exadg.sh
```diff
@@ -37,6 +37,8 @@ FFTW=$WORKING_DIRECTORY/fftw/install
...
+    -D CMAKE_C_FLAGS="-march=native -Wno-array-bounds" \
```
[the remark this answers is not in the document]
jh66637: This relates to the deal.II setup script. I suppose CXX should work as well. These lines of code become relevant since the deal.II macro does not take care about it anymore.

jh66637:
@nfehn I will split this PR into separate PRs. As mentioned, this is not urgent anymore since the failing pipeline was fixed by deal.II. I still think this is the way to go. I am converting this PR to a draft in the meantime such that I have a reminder to do this. I hope I will find the time soon.

nfehn:
@jh66637 is this PR still work planned for the future, or is this topic already outdated?

jh66637:
I still think we should switch to the new interface. But since currently everything is working using the CMake macros from deal.II, this is not high priority for me.

nfehn:
Would it now make sense to continue with this PR?

jh66637:
Probably yes, since the changes in the build system are finalized in `deal.II 9.5`. There are some more pressing topics on my table but I will do this! I'll ping you when I found the time.

PR #395, opened 2023-03-24
introduce base class for time integrator constants

nfehn:
as a (potenial) prerequisite for PR #261

### DIFF
--- a/include/exadg/time_integration/time_integration_constants_base.h
+++ b/include/exadg/time_integration/time_integration_constants_base.h
@@ -1,0 +1,90 @@
+/*
+ * time_integration_constants_base.h
+ *
+ *  Created on: Mar 24, 2023
+ *      Author: fehn
+ */
+
+#ifndef INCLUDE_EXADG_TIME_INTEGRATION_TIME_INTEGRATION_CONSTANTS_BASE_H_
+#define INCLUDE_EXADG_TIME_INTEGRATION_TIME_INTEGRATION_CONSTANTS_BASE_H_
+
+
+namespace ExaDG
+{
+class TimeIntegratorConstantsBase
+{
+public:
+  TimeIntegratorConstantsBase(unsigned int const order_time_integrator,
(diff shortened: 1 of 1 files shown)


include/exadg/time_integration/time_integration_constants_base.h
```diff
@@ -0,0 +1,113 @@
...
+
+  /*
+   *  This function prints the time integrator constants
+   */
+  virtual void
+  print(dealii::ConditionalOStream & pcout) const = 0;
+
+protected:
+  // order of time integrator
+  unsigned int const order;
```
jh66637, line 90: This can be private. We already have a getter.

jh66637:
I am fine with this. 

A different way to avoid all of the virtual functions etc. would be to have a free (templated) function ```update_time_integrator_constants(BDF, current_order, time_steps)```  that works on the coefficient classes. 

It's ok for me If you want to stick with inheritance here.

include/exadg/time_integration/time_integration_constants_base.h
nfehn: Yes. The problem is that the previous implementation did not make use of this getter. If we change it now, it will imply more changes also in #261. Principally, I am open to changing it here or separately.

nfehn:
For templated functions, the required interface is in my opinion not well-documented and somewhat more implicit.

jh66637:
What I like about it is the loose coupling, but I am also fine with the inheritance :)

include/exadg/time_integration/time_integration_constants_base.h
jh66637: If we merge this first...
If you persist on this, I have to change https://github.com/exadg/exadg/pull/261 again anyway.

nfehn: It is not about "persisting". You do not see a value in the present PR (and doing it now before adding more code duplications) that removes more lines of code than it adds new lines of code?

PR #397, opened 2023-03-28
Picard solver as alternative to Newton for iNS

richardschu:
Implemented a Picard solver as an alternative to Newton for the incompressible flow problems, since it might be interesting to do only a fixed *single* nonlinear iteration in time-dependent (not quasi-static) problems. Within a Picard solver, this is a sensible linearization in time, whereas for Newton, it still is a single nonlinear step (possibly equally valid). The Poiseuille flow application was used for testing and changes to it will be removed before merging. I tested L2 errors in the velocity and pressure, linear-, and nonlinear iteration counts for: PressureInflow/ParabolicInflow, always with symmetry boundary, Picard or Newton, 'BDFCoupledSolution' or 'BDFPressureCorrection',  'DivergenceFormulation' or 'ConvectiveFormulation' for the convective term.
-All 32 cases give comparable errors (~1e-10 or less), Newton needs less nonlinear iterations, the linear solver iteration counts sometimes slightly increase for Picard, but the convergence criteria are identical in both cases (and the systems are vastly different, so that might in fact not be too easy to compare)

The structure module uses the nonlinear solver as well, but the operators are not implemented (assertions placed). I think Picard does not make too much sense for structure problems, at least it is unusual.

The filename newton_solver.h/_data.h is kept for comparison and will be changed.

A lot of the changes are only due to default 'NonlinearSolver::SolverType::Newton' (which I did not want to hide) and renaming in all the modules.

I changed the tests in the Poiseuille flow application - in a follow-up PR, I will try to test *more* combinations. We should think about how we want to unit-test!

### DIFF
--- a/include/exadg/solvers_and_preconditioners/nonlinear_solvers/newton_solver.h
+++ b/include/exadg/solvers_and_preconditioners/nonlinear_solvers/newton_solver.h
@@ -110,7 +110,7 @@ public:
       double             omega         = 1.0; // damping factor (begin with 1)
       double             norm_r_damp   = 1.0; // norm of residual using temporary solution
       unsigned int       n_iter_damp   = 0;   // counts iteration of damping scheme
-      unsigned int const max_iter_damp = 10;   // max iterations of damping scheme
+      unsigned int const max_iter_damp = 10;  // max iterations of damping scheme
       double const       tau           = 0.5; // a parameter (has to be smaller than 1)
       do
       {


applications/incompressible_navier_stokes/poiseuille/input.json
```diff
@@ -7,16 +7,19 @@
     "SpatialResolution": {
         "DegreeMin": "2",
         "DegreeMax": "2",
-        "RefineSpaceMin": "0",
-        "RefineSpaceMax": "0"
+        "RefineSpaceMin": "1",
```
richardschu, line 10: Changes to the input.json here will be removed completely - just to show how I tested.

applications/incompressible_navier_stokes/vortex/application.h
```diff
@@ -366,7 +366,8 @@ class Application : public ApplicationBase<dim, Number>
     // momentum step
 
     // Newton solver
-    this->param.newton_solver_data_momentum = Newton::SolverData(100, 1.e-12, 1.e-6); // TODO
+    this->param.nonlinear_solver_data_momentum =
+      NonlinearSolver::SolverData(100, 1.e-12, 1.e-6, NonlinearSolver::SolverType::Newton); // TODO
```
richardschu, line 370: This is not my TODO ... I do not see, why not being able to converge to an extremely low tolerance for some cases is a problem?

applications/structure/bar/application.h
```diff
@@ -250,9 +251,9 @@ class Application : public ApplicationBase<dim, Number>
 
     this->param.update_preconditioner                  = true;
     this->param.update_preconditioner_every_time_steps = 1;
-    this->param.update_preconditioner_every_newton_iterations =
-      this->param.newton_solver_data.max_iter;
-    this->param.update_preconditioner_once_newton_converged = true;
+    this->param.update_preconditioner_every_nonlinear_iterations =
```
richardschu, line 254: I renamed also the nonlinear solver variables in the structure module - maybe someone wants to implement some other nonlinear solution strategies in the future?

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_coupled.cpp
```diff
@@ -323,6 +328,45 @@ OperatorCoupled<dim, Number>::evaluate_nonlinear_residual(BlockVectorType &
   dst.block(1) *= -scaling_factor_continuity;
 }
 
+template<int dim, typename Number>
```
richardschu, line 331: This is very much similar to the rhs for the Stokes problem, which served as a basis here.

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_coupled.h
```diff
@@ -73,6 +73,16 @@ class NonlinearOperatorCoupled
     pde_operator->evaluate_nonlinear_residual(dst, src, rhs_vector, time, scaling_factor_mass);
   }
 
+  /*
```
richardschu, line 76: this is the function added for the 'operator_coupled' and 'operator_pressure_correction' and for the structure module (empty/asserted in the latter case).

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_pressure_correction.cpp
```diff
@@ -328,6 +330,34 @@ OperatorPressureCorrection<dim, Number>::evaluate_nonlinear_residual(
   dst.add(-1.0, *rhs_vector);
 }
 
+template<int dim, typename Number>
```
richardschu, line 333: rhs for Picard in the pressure correction scheme, similar to Stokes rhs once again.

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_pressure_correction.cpp
```diff
@@ -328,6 +330,34 @@ OperatorPressureCorrection<dim, Number>::evaluate_nonlinear_residual(
...
+
+  if(this->param.convective_problem())
+  {
+    this->convective_operator.set_time(time);
+    this->convective_operator.rhs_add(dst);
+  }
+
+  // constant right-hand side vector (body force vector, sum_alphai_ui term and extrapolated
+  // pressure gradient)
+  //  dst.add(1.0, *rhs_vector);
```
richardschu, line 356: TODO: remove this line.

include/exadg/incompressible_navier_stokes/spatial_discretization/operators/convective_operator.cpp
```diff
@@ -92,22 +92,77 @@ template<int dim, typename Number>
...
+void
+ConvectiveOperator<dim, Number>::face_loop_empty(dealii::MatrixFree<dim, Number> const &,
+                                                 VectorType &,
+                                                 VectorType const &,
+                                                 Range const &) const
+{
+}
+
+template<int dim, typename Number>
+void
```
richardschu, line 137: maybe choose a different name here? Any ideas?

include/exadg/incompressible_navier_stokes/spatial_discretization/operators/convective_operator.h
```diff
@@ -745,51 +828,118 @@ class ConvectiveKernel
...
 
     if(boundary_type == BoundaryTypeU::Dirichlet || boundary_type == BoundaryTypeU::DirichletCached)
     {
-      if(data.type_dirichlet_bc == TypeDirichletBCs::Mirror)
+      if(operator_type == OperatorType::full || operator_type == OperatorType::inhomogeneous)
       {
-        delta_uP = -delta_uM;
+        dealii::Tensor<1, dim, dealii::VectorizedArray<Number>> g;
+
+        if(data.linearization_type == LinearizationType::Newton)
```
richardschu, line 888: I did not write 2 different functions just because of Picard/Newton, since they only differ in how they treat the boundary data g. Here, Newton uses g=0 (homogeneous operator), whereas Picard uses the actual g.

include/exadg/incompressible_navier_stokes/spatial_discretization/operators/convective_operator.h
```diff
@@ -745,51 +828,118 @@ class ConvectiveKernel
...
+          {
+            AssertThrow(false, dealii::ExcMessage("Not implemented."));
+          }
+        }
+        else
+        {
+          AssertThrow(false, dealii::ExcMessage("Linearization type not implemented."));
+        }
+
+        if(data.type_dirichlet_bc == TypeDirichletBCs::Mirror)
```
richardschu, line 919: Note how the Newton implementation is recovered with g=0.

include/exadg/solvers_and_preconditioners/nonlinear_solvers/newton_solver.h
```diff
@@ -50,45 +50,63 @@ class Solver
...
   {
-    unsigned int newton_iterations = 0, linear_iterations = 0;
+    unsigned int nonlinear_iterations = 0, linear_iterations = 0;
 
-    VectorType residual, increment, temporary;
-    residual.reinit(solution);
-    increment.reinit(solution);
-    temporary.reinit(solution);
+    VectorType residual_rhs, nonlinear_solver_output, temporary_solution;
+    residual_rhs.reinit(solution);
```
richardschu, line 59: holds residual/rhs. I did not want to introduce yet another vector.

include/exadg/solvers_and_preconditioners/nonlinear_solvers/newton_solver.h
```diff
@@ -114,22 +145,24 @@ class Solver
...
       linear_iterations += n_iter_linear;
     }
 
     AssertThrow(norm_r <= this->solver_data.abs_tol || norm_r / norm_r_0 <= solver_data.rel_tol,
                 dealii::ExcMessage(
-                  "Newton solver failed to solve nonlinear problem to given tolerance. "
-                  "Maximum number of iterations exceeded!"));
+                  "Nonlinear solver failed to solve nonlinear problem to given tolerance. "
+                  "Maximum number of iterations (" +
+                  std::to_string(solver_data.max_iter) + ") exceeded!"));
```
richardschu, line 165: If we want to use linearized solvers, we have to give the nonlinear solvers additional parameters: 'enable_damping' (and more parameters potentially, these are hard-coded above), and some switch to activate the Assertion here, such that doing 5 nonlinear iters is ok, no matter the reduction in the residual.

include/exadg/structure/spatial_discretization/operator.cpp
```diff
@@ -304,6 +304,21 @@ Operator<dim, Number>::setup_operators()
   operator_data.material_descriptor = material_descriptor;
   operator_data.unsteady            = (param.problem_type == ProblemType::Unsteady);
   operator_data.density             = param.density;
+
```
richardschu, line 307: The structure module has no Picard yet. The nonlinear solver requires the according data structure and the rhs for Picard nonetheless. Both are asserted for 'NonlinearSolver::SolverType::Picard'.

include/exadg/structure/spatial_discretization/operator.h
```diff
@@ -91,6 +91,17 @@ class ResidualOperator
...
+  /*
+   * The implementation of the nonlinear (Picard) solver requires a function called
+   * 'rhs_picard_linearized_problem'.
+   */
+  void
+  evaluate_rhs_picard(VectorType & /*dst*/)
+  {
+    AssertThrow(
+      false, dealii::ExcMessage("Nonlinear Picard solver not implemented for structure module."));
+  }
```
richardschu, line 103: Picard for elasticity might be useless.

applications/incompressible_navier_stokes/poiseuille/application.h
```diff
@@ -264,12 +284,12 @@ class Application : public ApplicationBase<dim, Number>
 
     // pressure Poisson equation
     this->param.solver_pressure_poisson         = SolverPressurePoisson::CG;
-    this->param.solver_data_pressure_poisson    = SolverData(1000, 1.e-20, 1.e-6, 100);
+    this->param.solver_data_pressure_poisson    = SolverData(1000, 1.e-12, 1.e-4, 100);
```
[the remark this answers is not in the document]
nfehn: why didn't you increase the maximum number of iterations instead?

include/exadg/solvers_and_preconditioners/nonlinear_solvers/newton_solver.h
```diff
@@ -50,45 +50,63 @@ class Solver
...
+
+        // overwrite residual with right-hand side for Newton solver
+        // multiply by -1.0 since the linearized problem is
+        // "linear_operator * increment = - residual"
+        residual_rhs *= -1.0;
+      }
+      else if(solver_data.nonlinear_solver_type == SolverType::Picard)
+      {
+        // last solution as initial guess for Picard solver
+        nonlinear_solver_output = solution;
```
nfehn, line 86: I am not convinced that this "if-else-logic" is how we should implement a Picard solver. Generally, one should be able to introduce a Picard solver without touching every application. The current realization does not seem to be modular in this regard.

nfehn:
I understand the description of this PR as introducing a new numerical method whose benefit has not been demonstrated so far? Is the idea to have this type of solver in a kind of "experimental mode" in ExaDG? Against this background, touching 70 files is immense and I wonder whether this is the way to go. For the future, it would be better to first discuss and agree on a basic design/realization of a Picard solver (touching only a single file) before making changes with 1500 new lines of code in 70 files.

include/exadg/structure/user_interface/enum_types.h
```diff
@@ -131,6 +131,19 @@ enum class Preconditioner
...
 
+/*
+ *  Nonlinear solver
+ */
+
+enum class LinearizationType
+{
+  Undefined,
+  Newton
+};
```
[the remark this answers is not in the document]
richardschu: We might either i) use the Undefined as default, or ii) someone wants to define a new nonlinear solver. Not the way to go?

include/exadg/solvers_and_preconditioners/nonlinear_solvers/newton_solver.h
nfehn: If am of a different opinion, because making changes will become more and more difficult. The present solution is in my opinion a quick solution but not a clean design.

richardschu: So we should/could cook up a design for nonlinear solvers in general in base/derived classes. From my perspective, I can only think of solvers similar to Newton & Picard, since I am not familiar with other methods. I assume that other methods might even need other functionality than what we have in, say,  'convective_operator.h/.cpp'. I can generalize the nonlinear solver, but still targeting Newton or Picard.

nfehn:
Let me explain from a different perspective why the current design is problematic: 
- Because of the call `nonlinear_operator.evaluate_rhs_picard(residual_rhs);`, every nonlinear operator has to implement  now a function `evaluate_rhs_picard()`, even though some module might never want to use Picard! 
- Since you plug the nonlinear solver type into the `SolverData`, every high-level module using any type of nonlinear solver now  depends on this parameter, even though a certain module might never want to use Picard! 
- the circumstance that you renamed variables in the Newton solver indicates that combining the two cases in a single implementation in the current form is not intuitive and hinders understanding of the code.

In my opinion, this is not a **modular** design, because introducing a new feature for one module implies making changes to all other modules, even though they might never want to use this feature. It is not in line with the design that we want to follow in ExaDG in my opinion. In this respect, I would argue that we want to follow the library idea of deal.II also in ExaDG. Take as an example the implementation (or wrappers) for linear solvers. According to your design, implementing a new linear solver for non-symmetric problems would imply making changes to the Poisson problem with symmetric positive definite linear operator!? (Note that "making changes" could also mean "recompiling the code", since the nonlinear solver type parameter is included in the nonlinear solver implementation, which is then included by all modules).

If I remember correctly, this topic is also discussed under the name "dependency inversion principle" in software design ("high-level modules should depend on abstractions", abstraction meaning something like "interface", e.g. realized via a base class with inheritance and dynamic polymorphism). I do not want to imply that inheritance is the only option. The important aspect - according to my understanding - is to realize the very general goal in software development of minimizing dependencies. So far, I only thought of Newton as a nonlinear solver. So the argument that we need to do the changes only once and for all and never touch it in the future is not convincing in my opinion. By such an argument actually, other project copied the Newton solver implementation from ExaDG, arguing that a Newton solver is written once and then never changed. (For clarity, I would not change Newton because of a new Picard implementation, but my point is that there are multiple possibilities to extend the implementation of an existing Newton solver itself. So my point is one should probably avoid to argue with "never" in software development ... In most cases when I had taken this perspective in the past, I have proved myself wrong in this regard.) 

I hope these explanations help, but we can of course also discuss this topic in person.

richardschu:
Hi Niklas,
thank you for your time and effort checking out the PR. I see the points you are making, sounds reasonable and I even experienced that when I had to implement a `evaluate_rhs_picard` in the structure solver. The current state of the PR is pure *functionality*, so that we have a common starting ground and I could get familiar with that part of the code.
I will try to come up with a better design and keep you posted!

PR #402, opened 2023-04-13
Use as_dof_handler_iterator only for deal.II 9.5

peterrum:
as reported in https://groups.google.com/g/dealii/c/7zzMV86oB38


peterrum:
... `MatrixFree::get_affine_constraints()` is also missing...

PR #404, opened 2023-04-18
pass operator data into Structure::MultigridPreconditioner update/init

richardschu:
The scaling factor for the mass operator was not passed into the MG levels in the structure module. At init and update, the scaling factors (and the evaluation time for the operators) are now passed to all levels and kept up to date.
For the example applications, this barely made a difference ... but of course that depends on the parameters.

### DIFF
--- a/include/exadg/convection_diffusion/spatial_discretization/operator.cpp
+++ b/include/exadg/convection_diffusion/spatial_discretization/operator.cpp
@@ -818,16 +818,16 @@ Operator<dim, Number>::update_conv_diff_operator(double const       time,
 
 template<int dim, typename Number>
 void
-Operator<dim, Number>::move_grid(double const & time) const
+Operator<dim, Number>::move_grid(double const & time, bool const update_preconditioner) const
 {
-  grid_motion->update(time, false);
+  grid_motion->update(time, false /* print_solver_info */, update_preconditioner);
 }
 
 template<int dim, typename Number>
 void
-Operator<dim, Number>::move_grid_and_update_dependent_data_structures(double const & time)
+Operator<dim, Number>::move_grid_and_update_dependent_data_structures(double const & time, bool const update_preconditioner)
 {
-  grid_motion->update(time, false);
+  grid_motion->update(time, false /* print_solver_info */, update_preconditioner);
   matrix_free->update_mapping(*get_mapping());
   update_after_grid_motion();
 }
--- a/include/exadg/convection_diffusion/spatial_discretization/operator.cpp
+++ b/include/exadg/convection_diffusion/spatial_discretization/operator.cpp
@@ -818,18 +818,24 @@ Operator<dim, Number>::update_conv_diff_operator(double const       time,
 
 template<int dim, typename Number>
 void
-Operator<dim, Number>::move_grid(double const & time, bool const update_preconditioner) const
+Operator<dim, Number>::move_grid(double const & time) const
 {
-  grid_motion->update(time, false /* print_solver_info */, update_preconditioner);
+  // Driver::ale_update() is responsible for the preconditioner update
+  grid_motion->update(
+    time,
+    false /* print_solver_info */,
+    dealii::numbers::invalid_unsigned_int /* time_step_number used for preconditioner update */);
 }
 
 template<int dim, typename Number>
 void
-Operator<dim, Number>::move_grid_and_update_dependent_data_structures(
-  double const & time,
-  bool const     update_preconditioner)
+Operator<dim, Number>::move_grid_and_update_dependent_data_structures(double const & time)
 {
-  grid_motion->update(time, false /* print_solver_info */, update_preconditioner);
+  // Driver::ale_update() is responsible for the preconditioner update
+  grid_motion->update(
+    time,
+    false /* print_solver_info */,
+    dealii::numbers::invalid_unsigned_int /* time_step_number used for preconditioner update */);
   matrix_free->update_mapping(*get_mapping());
   update_after_grid_motion();
 }
--- a/include/exadg/convection_diffusion/spatial_discretization/operator.cpp
+++ b/include/exadg/convection_diffusion/spatial_discretization/operator.cpp
@@ -825,7 +825,9 @@ Operator<dim, Number>::move_grid(double const & time, bool const update_precondi
 
 template<int dim, typename Number>
 void
-Operator<dim, Number>::move_grid_and_update_dependent_data_structures(double const & time, bool const update_preconditioner)
+Operator<dim, Number>::move_grid_and_update_dependent_data_structures(
+  double const & time,
+  bool const     update_preconditioner)
 {
   grid_motion->update(time, false /* print_solver_info */, update_preconditioner);
   matrix_free->update_mapping(*get_mapping());
--- a/include/exadg/convection_diffusion/time_integration/time_int_bdf.cpp
+++ b/include/exadg/convection_diffusion/time_integration/time_int_bdf.cpp
@@ -68,6 +68,7 @@ TimeIntBDF<dim, Number>::setup_derived()
   {
     // compute the grid coordinates at start time (and at previous times in case of
     // start_with_low_order == false)
+
     pde_operator->move_grid(this->get_time());
     pde_operator->fill_grid_coordinates_vector(vec_grid_coordinates[0]);
 
--- a/include/exadg/convection_diffusion/time_integration/time_int_bdf.cpp
+++ b/include/exadg/convection_diffusion/time_integration/time_int_bdf.cpp
@@ -68,7 +68,8 @@ TimeIntBDF<dim, Number>::setup_derived()
   {
     // compute the grid coordinates at start time (and at previous times in case of
     // start_with_low_order == false)
-    pde_operator->move_grid(this->get_time(), false /* only update preconditioner prior to ALE update */);
+    pde_operator->move_grid(this->get_time(),
+                            false /* only update preconditioner prior to ALE update */);
     pde_operator->fill_grid_coordinates_vector(vec_grid_coordinates[0]);
 
     if(this->start_with_low_order == false)
@@ -76,7 +77,8 @@ TimeIntBDF<dim, Number>::setup_derived()
       // compute grid coordinates at previous times (start with 1!)
       for(unsigned int i = 1; i < this->order; ++i)
       {
-        pde_operator->move_grid(this->get_previous_time(i), false /* only update preconditioner prior to ALE update */);
+        pde_operator->move_grid(this->get_previous_time(i),
+                                false /* only update preconditioner prior to ALE update */);
         pde_operator->fill_grid_coordinates_vector(vec_grid_coordinates[i]);
       }
     }
@@ -137,7 +139,8 @@ void
 TimeIntBDF<dim, Number>::initialize_current_solution()
 {
   if(this->param.ale_formulation)
-    pde_operator->move_grid(this->get_time(), false /* only update preconditioner prior to ALE update */);
+    pde_operator->move_grid(this->get_time(),
+                            false /* only update preconditioner prior to ALE update */);
 
   pde_operator->prescribe_initial_conditions(solution[0], this->get_time());
 }
@@ -150,7 +153,8 @@ TimeIntBDF<dim, Number>::initialize_former_solutions()
   for(unsigned int i = 1; i < solution.size(); ++i)
   {
     if(this->param.ale_formulation)
-      pde_operator->move_grid(this->get_previous_time(i), false /* only update preconditioner prior to ALE update */);
+      pde_operator->move_grid(this->get_previous_time(i),
+                              false /* only update preconditioner prior to ALE update */);
 
     pde_operator->prescribe_initial_conditions(solution[i], this->get_previous_time(i));
   }
@@ -556,7 +560,8 @@ TimeIntBDF<dim, Number>::postprocessing() const
   if(this->param.ale_formulation && this->get_time_step_number() == 1 &&
      !this->param.restarted_simulation)
   {
-    pde_operator->move_grid_and_update_dependent_data_structures(this->get_time(), false /* only update preconditioner prior to ALE update */);
+    pde_operator->move_grid_and_update_dependent_data_structures(
+      this->get_time(), false /* only update preconditioner prior to ALE update */);
   }
 
   postprocessor->do_postprocessing(solution[0], this->get_time(), this->get_time_step_number());
--- a/include/exadg/convection_diffusion/time_integration/time_int_bdf.cpp
+++ b/include/exadg/convection_diffusion/time_integration/time_int_bdf.cpp
@@ -68,8 +68,7 @@ TimeIntBDF<dim, Number>::setup_derived()
   {
     // compute the grid coordinates at start time (and at previous times in case of
     // start_with_low_order == false)
-    pde_operator->move_grid(this->get_time(),
-                            false /* only update preconditioner prior to ALE update */);
+    pde_operator->move_grid(this->get_time());
     pde_operator->fill_grid_coordinates_vector(vec_grid_coordinates[0]);
 
     if(this->start_with_low_order == false)
@@ -77,8 +76,7 @@ TimeIntBDF<dim, Number>::setup_derived()
       // compute grid coordinates at previous times (start with 1!)
       for(unsigned int i = 1; i < this->order; ++i)
       {
-        pde_operator->move_grid(this->get_previous_time(i),
-                                false /* only update preconditioner prior to ALE update */);
+        pde_operator->move_grid(this->get_previous_time(i));
         pde_operator->fill_grid_coordinates_vector(vec_grid_coordinates[i]);
       }
     }
@@ -139,8 +137,7 @@ void
 TimeIntBDF<dim, Number>::initialize_current_solution()
 {
   if(this->param.ale_formulation)
-    pde_operator->move_grid(this->get_time(),
-                            false /* only update preconditioner prior to ALE update */);
+    pde_operator->move_grid(this->get_time());
 
   pde_operator->prescribe_initial_conditions(solution[0], this->get_time());
 }
@@ -153,8 +150,7 @@ TimeIntBDF<dim, Number>::initialize_former_solutions()
   for(unsigned int i = 1; i < solution.size(); ++i)
   {
     if(this->param.ale_formulation)
-      pde_operator->move_grid(this->get_previous_time(i),
-                              false /* only update preconditioner prior to ALE update */);
+      pde_operator->move_grid(this->get_previous_time(i));
 
     pde_operator->prescribe_initial_conditions(solution[i], this->get_previous_time(i));
   }
@@ -560,8 +556,7 @@ TimeIntBDF<dim, Number>::postprocessing() const
   if(this->param.ale_formulation && this->get_time_step_number() == 1 &&
      !this->param.restarted_simulation)
   {
-    pde_operator->move_grid_and_update_dependent_data_structures(
-      this->get_time(), false /* only update preconditioner prior to ALE update */);
+    pde_operator->move_grid_and_update_dependent_data_structures(this->get_time());
   }
 
   postprocessor->do_postprocessing(solution[0], this->get_time(), this->get_time_step_number());
--- a/include/exadg/convection_diffusion/time_integration/time_int_bdf.cpp
+++ b/include/exadg/convection_diffusion/time_integration/time_int_bdf.cpp
@@ -68,8 +68,7 @@ TimeIntBDF<dim, Number>::setup_derived()
   {
     // compute the grid coordinates at start time (and at previous times in case of
     // start_with_low_order == false)
-
-    pde_operator->move_grid(this->get_time());
+    pde_operator->move_grid(this->get_time(), false /* only update preconditioner prior to ALE update */);
     pde_operator->fill_grid_coordinates_vector(vec_grid_coordinates[0]);
 
     if(this->start_with_low_order == false)
@@ -77,7 +76,7 @@ TimeIntBDF<dim, Number>::setup_derived()
       // compute grid coordinates at previous times (start with 1!)
       for(unsigned int i = 1; i < this->order; ++i)
       {
-        pde_operator->move_grid(this->get_previous_time(i));
+        pde_operator->move_grid(this->get_previous_time(i), false /* only update preconditioner prior to ALE update */);
         pde_operator->fill_grid_coordinates_vector(vec_grid_coordinates[i]);
       }
     }
@@ -138,7 +137,7 @@ void
 TimeIntBDF<dim, Number>::initialize_current_solution()
 {
   if(this->param.ale_formulation)
-    pde_operator->move_grid(this->get_time());
+    pde_operator->move_grid(this->get_time(), false /* only update preconditioner prior to ALE update */);
 
   pde_operator->prescribe_initial_conditions(solution[0], this->get_time());
 }
@@ -151,7 +150,7 @@ TimeIntBDF<dim, Number>::initialize_former_solutions()
   for(unsigned int i = 1; i < solution.size(); ++i)
   {
     if(this->param.ale_formulation)
-      pde_operator->move_grid(this->get_previous_time(i));
+      pde_operator->move_grid(this->get_previous_time(i), false /* only update preconditioner prior to ALE update */);
 
     pde_operator->prescribe_initial_conditions(solution[i], this->get_previous_time(i));
   }
@@ -557,7 +556,7 @@ TimeIntBDF<dim, Number>::postprocessing() const
   if(this->param.ale_formulation && this->get_time_step_number() == 1 &&
      !this->param.restarted_simulation)
   {
-    pde_operator->move_grid_and_update_dependent_data_structures(this->get_time());
+    pde_operator->move_grid_and_update_dependent_data_structures(this->get_time(), false /* only update preconditioner prior to ALE update */);
   }
 
   postprocessor->do_postprocessing(solution[0], this->get_time(), this->get_time_step_number());
--- a/include/exadg/grid/grid_motion_elasticity.h
+++ b/include/exadg/grid/grid_motion_elasticity.h
@@ -56,27 +56,9 @@ public:
       pcout(std::cout,
             dealii::Utilities::MPI::this_mpi_process(
               structure_operator->get_dof_handler().get_communicator()) == 0),
-      iterations({0, {0, 0}}),
-      time_step_number_last_preconditioner_update(0)
+      iterations({0, {0, 0}})
   {
     pde_operator->initialize_dof_vector(displacement);
-  }
-
-  /**
-   * Check if the preconditioner should be updated or was updated at this time step already
-   */
-  bool
-  update_preconditioner(types::time_step const time_step_number)
-  {
-    bool const update_preconditioner =
-      this->param.update_preconditioner &&
-      time_step_number % this->param.update_preconditioner_every_time_steps == 0 &&
-      time_step_number > time_step_number_last_preconditioner_update;
-
-    if(update_preconditioner)
-      time_step_number_last_preconditioner_update = time_step_number;
-
-    return update_preconditioner;
   }
 
   /**
@@ -94,12 +76,16 @@ public:
     {
       VectorType const_vector;
 
+      bool const update_preconditioner =
+        this->param.update_preconditioner &&
+        time_step_number % this->param.update_preconditioner_every_time_steps == 0;
+
       auto const iter =
         pde_operator->solve_nonlinear(displacement,
                                       const_vector,
                                       0.0 /* no mass term */,
                                       time,
-                                      this->update_preconditioner(time_step_number));
+                                      update_preconditioner);
 
       iterations.first += 1;
       std::get<0>(iterations.second) += std::get<0>(iter);
@@ -188,8 +174,6 @@ private:
     unsigned int /* calls */,
     std::tuple<unsigned long long, unsigned long long> /* iteration counts {Newton, linear}*/>
     iterations;
-
-  types::time_step time_step_number_last_preconditioner_update;
 };
 
 } // namespace ExaDG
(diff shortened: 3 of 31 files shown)


include/exadg/structure/driver.cpp
```diff
@@ -111,7 +111,10 @@ Driver<dim, Number>::setup()
       AssertThrow(false, dealii::ExcMessage("Not implemented."));
     }
 
-    pde_operator->setup_solver();
+    if(application->get_parameters().problem_type == ProblemType::Unsteady)
+      pde_operator->setup_solver(time_integrator->get_scaling_factor_mass());
```
richardschu, line 115: here i needed to get the scaling factor from thetime integrator. hence I made it public, see `time_int_gen_alpha_base.h`

include/exadg/structure/preconditioners/multigrid_preconditioner.cpp
```diff
@@ -193,6 +234,14 @@ MultigridPreconditioner<dim, Number>::initialize_operator(unsigned int const lev
   if(nonlinear)
   {
     std::shared_ptr<PDEOperatorNonlinearMG> pde_operator_level(new PDEOperatorNonlinearMG());
+
+    if(data.unsteady)
```
richardschu, line 238: scaling factors are correct at first init with this

include/exadg/structure/spatial_discretization/operator.cpp
```diff
@@ -710,11 +718,14 @@ Operator<dim, Number>::compute_initial_acceleration(VectorType &       accelerat
   VectorType rhs(acceleration);
   rhs = 0.0;
 
+  double scaling_factor_mass;
+
```
richardschu, line 722: we set the factor to zero, so store the value we had before and reset it after computing the initial acceleration

include/exadg/structure/spatial_discretization/operator.cpp
```diff
@@ -904,12 +922,15 @@ unsigned int
...
-                                    double const       time) const
+                                    double const       time,
+                                    bool const         update_preconditioner) const
 {
   // unsteady problems
   elasticity_operator_linear.set_scaling_factor_mass_operator(factor);
   elasticity_operator_linear.set_time(time);
 
+  linear_solver->update_preconditioner(update_preconditioner);
+
```
richardschu, line 933: note that we update the preconditioner before the actual solve!

include/exadg/structure/spatial_discretization/operators/linear_operator.cpp
```diff
@@ -45,9 +45,11 @@ LinearOperator<dim, Number>::do_cell_integral(IntegratorCell & integrator) const
     integrator.submit_gradient(sigma, q);
 
     if(this->operator_data.unsteady)
+    {
       integrator.submit_value(this->scaling_factor_mass * this->operator_data.density *
                                 integrator.get_value(q),
                               q);
+    }
   }
```
richardschu, line 53: i can remove these, I had some trusty std::cout here

include/exadg/structure/spatial_discretization/operators/nonlinear_operator.cpp
```diff
@@ -288,9 +288,11 @@ NonLinearOperator<dim, Number>::do_cell_integral(IntegratorCell & integrator) co
     integrator.submit_gradient(delta_P, q);
 
     if(this->operator_data.unsteady)
+    {
       integrator.submit_value(this->scaling_factor_mass * this->operator_data.density *
                                 integrator.get_value(q),
                               q);
+    }
```
richardschu, line 295: same here

include/exadg/structure/time_integration/time_int_gen_alpha.cpp
```diff
@@ -146,12 +146,12 @@ TimeIntGenAlpha<dim, Number>::do_timestep_solve()
   else
     displacement_np = displacement_last_iter;
 
+  bool const update_preconditioner =
+    this->param.update_preconditioner &&
+    ((this->time_step_number - 1) % this->param.update_preconditioner_every_time_steps == 0);
+
```
richardschu, line 152: preconditioner might be updated for linear (instat) solve as well -> scaling for adaptive timestepping and large dt variations?

include/exadg/time_integration/time_int_gen_alpha_base.h
```diff
@@ -52,10 +52,10 @@ class TimeIntGenAlphaBase : public TimeIntBase
   void
   set_current_time_step_size(double const & time_step_size) final;
 
-protected:
   double
```
richardschu, line 55: made this public to call it in the structure driver

richardschu (commented):
what do you think @nfehn ?

include/exadg/structure/preconditioners/multigrid_preconditioner.cpp
```diff
@@ -128,25 +149,34 @@ template<int dim, typename Number>
...
-  VectorType const & vector_linearization = pde_operator_nonlinear->get_solution_linearization();
-
-  // convert Number --> MultigridNumber, e.g., double --> float, but only if necessary
-  VectorTypeMG         vector_multigrid_type_copy;
-  VectorTypeMG const * vector_multigrid_type_ptr;
-  if(std::is_same<MultigridNumber, Number>::value)
+  if(data.unsteady)
   {
-    vector_multigrid_type_ptr = reinterpret_cast<VectorTypeMG const *>(&vector_linearization);
+    set_time(pde_operator->get_time());
```
nfehn, line 154: Did you figure out the precise reason for `set_time()`? This would be important because if time does not have an impact on the linear system of equations, we should avoid updating the preconditioner (for reasons of computational costs).

include/exadg/structure/preconditioners/multigrid_preconditioner.cpp
```diff
@@ -70,7 +70,7 @@ template<int dim, typename Number>
 void
 MultigridPreconditioner<dim, Number>::update()
 {
-  if(nonlinear)
+  if(nonlinear or data.unsteady)
```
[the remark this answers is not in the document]
nfehn: I have just seen that we do it exactly the same way in the momentum operator for incompressible flows. So should be fine as is. I think the responsibility is here in the hands of the user, who has to specify how often to update the preconditioner in the application file. If no update is needed (for certain model problems), the user has to deactivate this in the application to achieve optimal performance.

include/exadg/structure/spatial_discretization/operator.cpp
```diff
@@ -428,10 +428,18 @@ Operator<dim, Number>::setup_operators()
...
   pcout << std::endl << "Setup elasticity solver ..." << std::endl;
 
+  double active_scaling_factor_mass =
+    param.problem_type == ProblemType::Unsteady ? scaling_factor_mass : 0.0;
+
+  if(param.large_deformation)
+    elasticity_operator_nonlinear.set_scaling_factor_mass_operator(active_scaling_factor_mass);
+  else
+    elasticity_operator_linear.set_scaling_factor_mass_operator(active_scaling_factor_mass);
+
```
nfehn, line 442: I assume this is in analogy to how it is done in other modules such as IncNS? I think there is no need to compute an active scaling factor, because it will never be used for a steady problem (because the operator itself contains an if-statement `if(unsteady)`)? If so, I suggest to skip the first two lines.

BTW: The circumstance that a setup of the operator is done in `setup_solvers()` is exactly the Henne-Ei-Problem we have discussed in person.

include/exadg/structure/spatial_discretization/operator.cpp
```diff
@@ -754,6 +766,12 @@ Operator<dim, Number>::compute_initial_acceleration(VectorType &       accelerat
 
   // invert mass operator to get acceleration
   mass_solver->solve(acceleration, rhs);
+
+  // revert scaling factor to initialized value
+  if(param.large_deformation)
+    elasticity_operator_nonlinear.set_scaling_factor_mass_operator(scaling_factor_mass);
```
nfehn, line 772: I would write each of these lines directly after evaluating the elasticity operator in the respective if/else-branches above so that it is more clear for which parts of the code this is relevant.

include/exadg/structure/spatial_discretization/operator.cpp
```diff
@@ -710,11 +718,14 @@ Operator<dim, Number>::compute_initial_acceleration(VectorType &       accelerat
   VectorType rhs(acceleration);
   rhs = 0.0;
 
+  double scaling_factor_mass;
```
nfehn, line 721: No need to define this outside. We can simply write `double const` inside the if-branch and the else-branch. Makes the code easier to read in my opinion (and in the sense of @btemuer, the use of `scaling_factor_mass` is here again in a `const` way).

include/exadg/structure/spatial_discretization/operators/linear_operator.cpp
nfehn: I actually like it to place brackets here (should also be in line with our coding conventions in the wiki).

include/exadg/structure/time_integration/driver_steady_problems.cpp
```diff
@@ -124,8 +124,8 @@ DriverSteady<dim, Number>::do_solve()
     // calculate right-hand side vector
     pde_operator->compute_rhs_linear(rhs_vector, 0.0 /* time */);
 
-    unsigned int const N_iter_linear =
-      pde_operator->solve_linear(solution, rhs_vector, 0.0 /* no mass term */, 0.0 /* time */);
+    unsigned int const N_iter_linear = pde_operator->solve_linear(
+      solution, rhs_vector, 0.0 /* no mass term */, 0.0 /* time */, param.update_preconditioner);
```
nfehn, line 128: We have to rethink this critically. In my opinion, this means we initialize the multigrid preconditioner (or, better, the smoothers, coarse grid solver) twice or a steady linear problem: once in the setup-routines, and once within the solver call. However, this is unnecessary work. So we should check if this is the case, and if yes, pass `false` to this function here with a comment that the preconditioner has already been initialized.

Note that this is different for the Newton solver above, because there the preconditioner can be updated every n-th Newton iteration.

include/exadg/structure/preconditioners/multigrid_preconditioner.cpp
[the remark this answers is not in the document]
richardschu: This way, the steady linear solver does not do the update - as it would otherwise do the update twice (in a row, basically)!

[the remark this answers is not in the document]
richardschu: The cases where we do additional inits are now:
Steady nonlinear solver (+1 initially)
Unsteady nonlinear solver (+ 1 initially)
QuasiStatic(+1 initially)
One can also not skip the update in the first (quasi-)time step, since the update might be triggered every *n* Newton iterations or once the solver has converged.
One solution would be to add another flag to the `update` in the Newton solver, which checks if we are in the very first step after init! What do you think? ... changes some other files as well though ... then we could also think about checking the case `update_once_newton_converged` and the other flags (doing two head-to-head updates basically, with updated iterate/solution vector potentially?). Sounds more like a separate PR to me ... updating that Newton flags and (quasi-)time steppers?

include/exadg/structure/preconditioners/multigrid_preconditioner.cpp
richardschu: So fine? This is only triggered if `unsteady` evaluates to true ... and the time is a double ... so probably thats not the problem is it? the time is never used to trigger a preconditioner update? and updating the time variable is not costly, is it?
Anyways I have not found out what the `time` changes. I can think of changing a stiffness coefficient over time (initial damping and initially stiffer springs help a ton in cardiovasc scenarios!). For now, I do not see a reason this is needed -  maybe it is motivated by someone expecting `update` to update the time of the operators as well, which in itself sounds reasonable to me :)

include/exadg/structure/time_integration/driver_steady_problems.cpp
richardschu: well actually this evaluates to false here (steady, linear case). so I just put false. `param.update_preconditioner` seemed like here we have some control over that which we do not ...

include/exadg/structure/time_integration/time_int_gen_alpha.cpp
```diff
@@ -174,7 +174,8 @@ TimeIntGenAlpha<dim, Number>::do_timestep_solve()
     unsigned int const iter = pde_operator->solve_linear(displacement_np,
                                                          rhs,
                                                          this->get_scaling_factor_mass(),
-                                                         this->get_mid_time());
+                                                         this->get_mid_time(),
+                                                         update_preconditioner);
```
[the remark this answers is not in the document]
richardschu: I added (this->time_step_number > 1) so it is exactly what you would expect. User has full control via the timestep control of prec update and the initial update directly after init is skipped ---> no additional updates!

richardschu (commented):
The update is only triggered for unsteady or nonlinear cases (as before), skipping now one initial update for linear solver in the time dep case. For the nonlinear solver this is not possible, since we might want some updates in the first solution step (every n Newton iterations or once Newton is converged).

include/exadg/structure/time_integration/driver_steady_problems.cpp
```diff
@@ -124,8 +124,8 @@ DriverSteady<dim, Number>::do_solve()
     // calculate right-hand side vector
```
richardschu, line 124: Note that in the nonlinear case above, we have to pass `param.update_preconditioner` to even allow updates, as the Newton solver checks with an && every *n* iterations or at the end, if the preconditioner should be updated. We could do a follow-up PR to get rid of all the head-to-head calls of the preconditioner update...
For my target applications (nonlinear time-dependent problems + FSI), the effect is probably not sooo big.

include/exadg/structure/preconditioners/multigrid_preconditioner.cpp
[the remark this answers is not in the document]
nfehn: Currently, I think we should not let the Newton solver take the responsibility for this. I think this aspect is a convention between the spatial operator (initialization of solvers/preconditioners) and the time integrator (where we have the modulo operator, including a convention regarding the first time step for which we want to initiate an update).

include/exadg/structure/preconditioners/multigrid_preconditioner.cpp
nfehn: Then, I suggest to keep it to be prepared for the future / prevent foreseeable bugs.

include/exadg/structure/spatial_discretization/operator.cpp
nfehn: NB means "nota bene" (something like "Please note that ...")

include/exadg/structure/spatial_discretization/operator.cpp
```diff
@@ -726,12 +735,18 @@ Operator<dim, Number>::compute_initial_acceleration(VectorType &       accelerat
     {
       body_force_operator.evaluate_add(rhs, displacement, time);
     }
+
+    // revert scaling factor to initialized value
+    elasticity_operator_nonlinear.set_scaling_factor_mass_operator(scaling_factor_mass);
```
nfehn, line 740: I think one could even shift this line even before evaluation the body force operator. This would make the "manipulation" more local in the code.

include/exadg/structure/spatial_discretization/operator.cpp
```diff
@@ -750,6 +765,9 @@ Operator<dim, Number>::compute_initial_acceleration(VectorType &       accelerat
       // pull_back_body_force = false in this case.
       body_force_operator.evaluate_add(rhs, displacement, time);
     }
+
+    // revert scaling factor to initialized value
+    elasticity_operator_linear.set_scaling_factor_mass_operator(scaling_factor_mass);
```
nfehn, line 770: same as here https://github.com/exadg/exadg/pull/404/files#r1175037799

include/exadg/structure/time_integration/driver_steady_problems.cpp
```diff
@@ -124,8 +124,8 @@ DriverSteady<dim, Number>::do_solve()
...
     pde_operator->compute_rhs_linear(rhs_vector, 0.0 /* time */);
 
-    unsigned int const N_iter_linear =
-      pde_operator->solve_linear(solution, rhs_vector, 0.0 /* no mass term */, 0.0 /* time */);
+    unsigned int const N_iter_linear = pde_operator->solve_linear(
+      solution, rhs_vector, 0.0 /* no mass term */, 0.0 /* time */, false);
```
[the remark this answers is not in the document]
nfehn: Could you write `false /* update of preconditioner */` to be line with the previous two arguments? In case we do not pass variables with a concrete name but "numbers", it helps to understand the code or interpret the parameters passed to a function (no need to open the declaration of that function).

include/exadg/structure/time_integration/time_int_gen_alpha.cpp
```diff
@@ -171,10 +171,12 @@ TimeIntGenAlpha<dim, Number>::do_timestep_solve()
...
-    unsigned int const iter = pde_operator->solve_linear(displacement_np,
-                                                         rhs,
-                                                         this->get_scaling_factor_mass(),
-                                                         this->get_mid_time());
+    unsigned int const iter =
+      pde_operator->solve_linear(displacement_np,
+                                 rhs,
+                                 this->get_scaling_factor_mass(),
+                                 this->get_mid_time(),
+                                 update_preconditioner && (this->time_step_number > 1));
```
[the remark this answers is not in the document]
nfehn: I would skip `&& (this->time_step_number > 1)` here and do it identically to the nonlinear case. My argument is that the number of updates over a whole unsteady simulation is `N_dt / N_update_interval >> 1`, so the increase in costs is minimal. (Note that the currrent code is potentially problematic because there might come other reasons for preconditioner updates in the future (apart from scaling_factor_mass). Strictly avoiding the preconditioner update in the first call could then cause robustness issues, a bug that might be hard to find once we have forgotton that we have added  `&& (this->time_step_number > 1)` in the present PR.)

include/exadg/grid/grid_motion_elasticity.h
```diff
@@ -94,7 +94,8 @@ class GridMotionElasticity : public GridMotionBase<dim, Number>
       pde_operator->initialize_dof_vector(rhs);
       pde_operator->compute_rhs_linear(rhs, time);
 
-      auto const iter = pde_operator->solve_linear(displacement, rhs, 0.0 /* no mass term */, time);
+      auto const iter = pde_operator->solve_linear(
+        displacement, rhs, 0.0 /* no mass term */, time, param.update_preconditioner);
```
nfehn, line 98: I am not sure if this is what we want to have (of course, it would then also be a bug in the nonlinear case above): I think we want to trigger the update of the preconditioner for the ALE mesh motion solver only every `N_update_interval`th time step (using a logic similar as in the single-field time integration schemes). This logic is not implemented inside `pde_operator->solve_(non)linear()`, right? Does this imply that we update the preconditioner for every ALE mesh motion solve (in every FSI iteration) if `param.update_preconditioner == true`? This could become very expensive.

include/exadg/structure/time_integration/time_int_gen_alpha.cpp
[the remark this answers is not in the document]
nfehn: see here: https://github.com/exadg/exadg/pull/404#discussion_r1175059454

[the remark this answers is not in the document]
richardschu: Perfectly fine with that, as this was my initial thought :+1: ... that one saved update is probably not making the kraut fett!

include/exadg/structure/time_integration/driver_steady_problems.cpp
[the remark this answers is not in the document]
richardschu: I added the comment!

include/exadg/structure/spatial_discretization/operator.cpp
richardschu: Ok, now it is as close as possible - even though it fit on one screen before ; )

include/exadg/grid/grid_motion_elasticity.h
richardschu: the grid mapping is constructed from the initial configuration (Lagrangian frame) right?
There are several aspects to this:
1) is the "mesh stiffness" updated depending on the actual displacement or are the parameters computed once initially and kept constant?
2) time step size might vary
3) nonlinear mesh motion model might be used

`pde_opereator->solve_(non)linear()` updates if the passed flag is `true` and in the nonlinear case additionally at Newton convergence or every _n_th Newton step.
As the mapping is a critical ingredient in practical applications, I would actually leave the choice to the user and allow setting the latter two flags in the nonlinear case + a time step logic like you mentioned. So I would  recommend using the same logic as in `time_int_gen_alpha.cpp`, line 149:

`  bool const update_preconditioner =
    ((this->time_step_number - 1) % this->param.update_preconditioner_every_time_steps == 0);`

`pde_operator->solve_nonlinear()` takes care of the rest :100: 

Agree?

richardschu: Thinking about this again, I realized that the current grid motion techniques are (i) all solving stationary problems, (ii) the Lagrangian reference frame is considered as base and (iii) the parameters for "grid stiffening" do not depend on the solution itself, so for poisson and linear elasticity operaetors, we should deactivate the preconditioner update completely, shouldn't we? For the nonlinear elasticity case it makes sense though ...
Can you confirm (i)-(iii) @nfehn ?

include/exadg/convection_diffusion/time_integration/time_int_bdf.cpp
```diff
@@ -68,16 +68,17 @@ TimeIntBDF<dim, Number>::setup_derived()
   {
     // compute the grid coordinates at start time (and at previous times in case of
     // start_with_low_order == false)
-
-    pde_operator->move_grid(this->get_time());
+    pde_operator->move_grid(this->get_time(),
+                            false /* only update preconditioner prior to ALE update */);
```
nfehn, line 72: I have problems understanding this comment, probably the notion "ALE update" is not fully clear to me at this point.

include/exadg/grid/grid_motion_elasticity.h
```diff
@@ -56,16 +56,35 @@ class GridMotionElasticity : public GridMotionBase<dim, Number>
...
+  update_preconditioner(types::time_step const time_step_number)
+  {
+    bool const update_preconditioner =
+      this->param.update_preconditioner &&
+      time_step_number % this->param.update_preconditioner_every_time_steps == 0 &&
+      time_step_number > time_step_number_last_preconditioner_update;
+
+    time_step_number_last_preconditioner_update = time_step_number;
+
+    return update_preconditioner;
```
richardschu, line 79: the update is only done once per time step (if the update is desired every _n_ time steps).

include/exadg/grid/grid_motion_elasticity.h
```diff
@@ -94,7 +117,8 @@ class GridMotionElasticity : public GridMotionBase<dim, Number>
       pde_operator->initialize_dof_vector(rhs);
       pde_operator->compute_rhs_linear(rhs, time);
 
-      auto const iter = pde_operator->solve_linear(displacement, rhs, 0.0 /* no mass term */, time);
+      auto const iter = pde_operator->solve_linear(
```
richardschu, line 121: linear elasticity leads to a constant matrix, so do not update.

include/exadg/grid/grid_motion_interface.h
```diff
@@ -49,7 +52,7 @@ class GridMotionInterface
    * Updates the mapping, i.e., moves the grid.
    */
   virtual void
-  update(double const time, bool const print_solver_info) = 0;
+  update(double const time, bool const print_solver_info, types::time_step time_step_number) = 0;
```
richardschu, line 55: interface is now done with time step number

include/exadg/incompressible_flow_with_transport/driver.cpp
```diff
@@ -507,7 +507,9 @@ Driver<dim, Number>::ale_update() const
   dealii::Timer sub_timer;
 
   sub_timer.restart();
-  grid_motion->update(fluid_time_integrator->get_next_time(), false);
+  grid_motion->update(fluid_time_integrator->get_next_time(),
```
richardschu, line 510: similar to the other drivers, we give the actual time step number (in case one uses a `GridMotionElasticity` replacing the currently used `GridMotionFunction` or `GridMotionPoisson`), see `Driver::setup()`. For the currently possible `GridMotion` objects, this would not result in an update. Alternatively we would pass `dealii::numbers::invalid_unsigned_int` or simply 0 in the drivers for iNS and ConvDiff.

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
```diff
@@ -1465,15 +1465,23 @@ template<int dim, typename Number>
 void
 SpatialOperatorBase<dim, Number>::move_grid(double const & time) const
 {
-  grid_motion->update(time, false);
+  // Driver::ale_update() is responsible for the preconditioner update
+  grid_motion->update(
+    time,
+    false /* print_solver_info */,
+    dealii::numbers::invalid_unsigned_int /* time_step_number used for preconditioner update */);
```
richardschu, line 1472: The `move_grid()` and `move_grid_and_update_dependent_data_structures()` are used during initialization/postprocessing as far as I can tell?
`grep -rnw -e 'move_grid'` in `exadg/include` gives
"
exadg/convection_diffusion/spatial_discretization/interface.h:108:  move_grid(double const & time) const = 0;
exadg/convection_diffusion/spatial_discretization/operator.h:202:  move_grid(double const & time) const;
exadg/convection_diffusion/spatial_discretization/operator.cpp:821:Operator<dim, Number>::move_grid(double const & time) const
exadg/convection_diffusion/time_integration/time_int_bdf.cpp:72:    pde_operator->move_grid(this->get_time());
exadg/convection_diffusion/time_integration/time_int_bdf.cpp:80:        pde_operator->move_grid(this->get_previous_time(i));
exadg/convection_diffusion/time_integration/time_int_bdf.cpp:141:    pde_operator->move_grid(this->get_time());
exadg/convection_diffusion/time_integration/time_int_bdf.cpp:154:      pde_operator->move_grid(this->get_previous_time(i));
exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp:1466:SpatialOperatorBase<dim, Number>::move_grid(double const & time) const
exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.h:389:  move_grid(double const & time) const;
exadg/incompressible_navier_stokes/time_integration/time_int_bdf_coupled_solver.cpp:67:    pde_operator->move_grid(this->get_time());
exadg/incompressible_navier_stokes/time_integration/time_int_bdf_coupled_solver.cpp:82:      pde_operator->move_grid(this->get_previous_time(i));
exadg/incompressible_navier_stokes/time_integration/time_int_bdf_pressure_correction.cpp:147:    pde_operator->move_grid(this->get_time());
exadg/incompressible_navier_stokes/time_integration/time_int_bdf_pressure_correction.cpp:160:      pde_operator->move_grid(this->get_previous_time(i));
exadg/incompressible_navier_stokes/time_integration/time_int_bdf.cpp:99:    operator_base->move_grid(this->get_time());
exadg/incompressible_navier_stokes/time_integration/time_int_bdf.cpp:107:        operator_base->move_grid(this->get_previous_time(i));
exadg/incompressible_navier_stokes/time_integration/time_int_bdf_dual_splitting.cpp:139:    pde_operator->move_grid(this->get_time());
exadg/incompressible_navier_stokes/time_integration/time_int_bdf_dual_splitting.cpp:152:      pde_operator->move_grid(this->get_previous_time(i));

"

So I would rather stick to the convention of triggering the update only in `Driver::ale_update()`.

include/exadg/grid/grid_motion_function.h
```diff
@@ -48,16 +48,19 @@ class GridMotionFunction : public GridMotionBase<dim, Number>
       mesh_movement_function(mesh_movement_function),
       triangulation(triangulation)
   {
-    update(start_time, false);
+    update(start_time, false, dealii::numbers::invalid_unsigned_int);
```
richardschu, line 51: you prefer 0 whenever I put `dealii::numbers::invalid_unsigned int`? it is `static_cast<unsigned int>(-1)`, this way one would now immediately get that the `time_step_number` is fishy when debugging (and having an enormous unsigned int unexpectedly)!

include/exadg/convection_diffusion/spatial_discretization/operator.cpp
```diff
@@ -820,14 +820,22 @@ template<int dim, typename Number>
 void
 Operator<dim, Number>::move_grid(double const & time) const
 {
-  grid_motion->update(time, false);
+  // Driver::ale_update() is responsible for the preconditioner update
+  grid_motion->update(
+    time,
+    false /* print_solver_info */,
+    dealii::numbers::invalid_unsigned_int /* time_step_number used for preconditioner update */);
```
nfehn, line 827: The problem here is that the function `move_grid()` is public and can be used in a different context that our current `Driver`. In this sense, `Operator` is a low level module, whose implementation should not depend on our current high-level implementation in `Driver`. As a consequence, there should actually also not appear a comment on `Driver` in this class (in an ideal world).

I think the origin of this problem is that the two functions `move_grid()` and `move_grid_and_update_dependent_data_structures()` are a sub-optimal design. I think we should remove `move_grid()` and let the time integrator directly operate on `grid_motion`. The function `move_grid_and_update_dependent_data_structures()` needs to be changed to `update_spatial_discretization_after_grid_motion()` (i.e. the first part `grid_motion->update()` should no longer be part of that function). This way, there is only one possibility to move the grid in the future.

In order to proceed with the present PR, I suggest that I could address this in the future once the present PR is merged.

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
nfehn: see comment https://github.com/exadg/exadg/pull/404/files#r1177448382

nfehn (approved):

include/exadg/convection_diffusion/spatial_discretization/operator.cpp
nfehn: I created issue #412 and will therefore resolve the conversation here

nfehn:
I will merge once the checks have passed

PR #406, opened 2023-04-20
Refactor and extend VariableCoefficients

### LINKED ISSUE #405: Extend VariableCoefficients with cell-based face loops
Currently, the `VariableCoefficients` class does not support cell-based face loops, and respective parts in the code are marked `TODO`. However, this is necessary in many cases using cell-based face loops. The class should be extended accordingly.

btemuer:
- Remove separate VariableCoefficientsCell class,

- Add cell-based face loops functionality,

- Add bool parameters to turn on or off functionality like face loops, cell-based face loops, and/or differing neighbor coefficients.

- Add documentation + little bit of cosmetics

Closes #405

### DIFF
--- a/include/exadg/operators/variable_coefficients.h
+++ b/include/exadg/operators/variable_coefficients.h
@@ -44,52 +44,40 @@ class VariableCoefficients
 public:
   /**
    * Initializes the variable coefficients object based on @p matrix_free and @p quad_index
-   * according to the desired properties @p use_face_loops_in, @p use_neighbor_faces_in,
-   * and @p use_cell_based_face_loops_in, i.e. resizes the desired tables to the required number of
-   * cells, faces, and quadrature points.
+   * according to the desired properties @p use_faces_in, and @p use_cell_based_face_loops_in,
+   * i.e. resizes the desired tables to the required number of cells, faces, and quadrature points.
    *
    * @param matrix_free Underlying matrix-free description
    * @param quad_index Quadrature index in the matrix-free description to create the
    * coefficient tables according to the quadrature points
-   * @param use_face_loops_in Boolean switch to use a coefficient table for faces
-   * @param use_neighbor_faces_in Boolean switch to use a coefficient table for neighbors, in case
-   * the coefficients can differ on the two sides of a face
-   * @param use_cell_based_face_loops_in Boolean switch to use a coefficient table for cell-based
+   * @param use_faces_in Boolean switch to use a coefficient table for faces (and neighbors)
+   * @param use_cell_based_faces_in Boolean switch to use a coefficient table for cell-based
    * face access
    */
   template<int dim, typename Number>
   void
   initialize(dealii::MatrixFree<dim, Number> const & matrix_free,
              unsigned int const                      quad_index,
-             bool const                              use_face_loops_in,
-             bool const                              use_neighbor_faces_in,
-             bool const                              use_cell_based_face_loops_in)
+             bool const                              use_faces_in,
+             bool const                              use_cell_based_faces_in)
   {
-    use_face_loops            = use_face_loops_in;
-    use_neighbor_faces        = use_neighbor_faces_in;
-    use_cell_based_face_loops = use_cell_based_face_loops_in;
+    if(not use_faces_in)
+      AssertThrow(not use_cell_based_faces_in,
+                  dealii::ExcMessage("Using cell-based faces does not make sense if using faces"
+                                     " is disabled."));
+
+    use_faces            = use_faces_in;
+    use_cell_based_faces = use_cell_based_faces_in;
 
     reinit(matrix_free, quad_index);
   }
 
   /**
-   * Same as above, but fills the coefficient tables with the given @p constant_coefficient.
+   * Sets the coefficient @p constant_coefficient everywhere.
    */
-  template<int dim, typename Number>
   void
-  initialize(dealii::MatrixFree<dim, Number> const & matrix_free,
-             unsigned int const                      quad_index,
-             bool const                              use_face_loops_in,
-             bool const                              use_neighbor_faces_in,
-             bool const                              use_cell_based_face_loops_in,
-             coefficient_type const &                constant_coefficient)
+  set_coefficients(coefficient_type const & constant_coefficient)
   {
-    initialize(matrix_free,
-               quad_index,
-               use_face_loops_in,
-               use_neighbor_faces_in,
-               use_cell_based_face_loops_in);
-
     fill(constant_coefficient);
   }
 
@@ -188,22 +176,22 @@ private:
   {
     coefficients_cell.reinit(matrix_free.n_cell_batches(), matrix_free.get_n_q_points(quad_index));
 
-    if(use_face_loops)
+    if(use_faces)
+    {
       coefficients_face.reinit(matrix_free.n_inner_face_batches() +
                                  matrix_free.n_boundary_face_batches(),
                                matrix_free.get_n_q_points_face(quad_index));
-
-    if(use_neighbor_faces)
       coefficients_face_neighbor.reinit(matrix_free.n_inner_face_batches(),
                                         matrix_free.get_n_q_points_face(quad_index));
 
-    if(use_cell_based_face_loops)
-    {
-      unsigned int const n_faces_per_cell =
-        matrix_free.get_dof_handler().get_triangulation().get_reference_cells()[0].n_faces();
+      if(use_cell_based_faces)
+      {
+        unsigned int const n_faces_per_cell =
+          matrix_free.get_dof_handler().get_triangulation().get_reference_cells()[0].n_faces();
 
-      coefficients_face_cell_based.reinit(matrix_free.n_cell_batches() * n_faces_per_cell,
-                                          matrix_free.get_n_q_points_face(quad_index));
+        coefficients_face_cell_based.reinit(matrix_free.n_cell_batches() * n_faces_per_cell,
+                                            matrix_free.get_n_q_points_face(quad_index));
+      }
     }
   }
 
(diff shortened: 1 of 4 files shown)


include/exadg/operators/variable_coefficients.h
```diff
@@ -23,168 +23,231 @@
...
+  dealii::Table<2, coefficient_type> coefficients_face_cell_based;
+
+  //! Boolean switch to use a coefficient table for faces
+  bool use_face_loops{false};
+
+  /**
+   * @brief Boolean switch to use a coefficient table for neighbors, in case the coefficients
+   * can differ on the two sides of a face.
+   */
+  bool use_neighbor_faces{false};
```
nfehn, line 247: I can not see a use case in the current code where face loops are used, but no neighboring data. So I wonder whether we should introduce this flexibility now. Similarly, it looks like the current code allows the use of neighboring faces, but `use_face_loops` being deactivated. From a logical perspective, I believe this boolean should be nested into the other boolean `use_face_loops`.

include/exadg/operators/variable_coefficients.h
```diff
@@ -23,168 +23,231 @@
...
+
+  //! Boolean switch to use a coefficient table for cell-based face access
+  bool use_cell_based_face_loops{false};
```
nfehn, line 250: This boolean should also be nested into `use_face_loops`. If `use_cell_based_face_loops` is activated, the code has to assert that that `use_face_loops` is active. In order to avoid asserts in the code, I suggest to "nest" the booleans, in the sense that `cell_based_face_loops` is only relevant if `face_loops` is activated. I think this logic would then be consistent with the general idea of cell-based face loops in ExaDG.

include/exadg/operators/variable_coefficients.h
```diff
@@ -23,168 +23,231 @@
...
+  /**
+   * Same as above, but fills the coefficient tables with the given @p constant_coefficient.
+   */
   template<int dim, typename Number>
   void
   initialize(dealii::MatrixFree<dim, Number> const & matrix_free,
              unsigned int const                      quad_index,
+             bool const                              use_face_loops_in,
+             bool const                              use_neighbor_faces_in,
+             bool const                              use_cell_based_face_loops_in,
```
nfehn, line 84: I am not sure if duplicating the above interface here is desirable. Shouldn't we rather have only one initialize function and then call a second function in user code to set/fill the coefficient tables?

include/exadg/operators/variable_coefficients.h
```diff
@@ -23,168 +23,231 @@
...
   dealii::Table<2, coefficient_type> coefficients_face_neighbor;
 
-  // TODO
-  //  // cell-based face loops
-  //  dealii::Table<2, coefficient_type> coefficients_face_cell_based;
+  //! Coefficient table for faces with cell-based access
+  dealii::Table<2, coefficient_type> coefficients_face_cell_based;
+
+  //! Boolean switch to use a coefficient table for faces
+  bool use_face_loops{false};
```
nfehn, line 241: In terms of the naming, I think `loops` should not appear here. Just `use_faces` would be better.

include/exadg/operators/variable_coefficients.h
btemuer: One use case is if the coefficients are computed from an at least C0-continuous function of just space and time, like the field functions. In that case the coefficients are the same on both sides and it wouldn't be necessary to store them twice. Isn't that right?

include/exadg/operators/variable_coefficients.h
btemuer: Right, I'll look into how I can implement this logic 👍

include/exadg/operators/variable_coefficients.h
btemuer: Both make sense to me. I can happily make it more explicit with a separate function.

include/exadg/operators/variable_coefficients.h
nfehn: Cell-based face loops always use neighboring data, right? This dependency should also be reflected in the implementation. I am thinking of the following logic:

```c++
if(face_loops)
{
  if(use_cell_based_face_loops)
  {
    
  }
  else // when using a separate loop over faces in MatrixFree implementation
  {
    ...

    if(use_neighbor_faces)
    {
    }
  }
}
```
As said above, I would not implement `use_neighbor_faces` for now because of missing use cases. The implementation of a spatial PDE operator decides very naturally whether to read neighboring data or not. Hence, in terms of data transfer from main memory (as a main factor for computational efficiency) it should not be relevant to introduce the optimization `use_neighbor_faces=false`. One could see a benefit in storage requirements, but this is typically less of a concern in our applications with dealii's matrix-free implementation. Or do I miss something?

btemuer: I'd be fine with removing the optimization here and in the operators just have:

```
auto const coefficient = get_coefficient_face(face, q);
```

and omit the additional call to `get_coefficient_face_neighbor(face,q)` if it is not needed, i.e. if it is a priori known that the coefficients are continuous. Same is true calculating and setting the coefficients at every time step for every discretization.

This is what you mean, right? Because then exactly we would primarily benefit from storage.

nfehn: Yes exactly.

btemuer:
New changes are ready for review :) @nfehn

include/exadg/operators/variable_coefficients.h
```diff
@@ -23,168 +23,213 @@
...
-    // matrix_free.get_n_q_points_face(quad_index));
+    if(use_faces)
+    {
+      coefficients_face.reinit(matrix_free.n_inner_face_batches() +
+                                 matrix_free.n_boundary_face_batches(),
+                               matrix_free.get_n_q_points_face(quad_index));
+      coefficients_face_neighbor.reinit(matrix_free.n_inner_face_batches(),
+                                        matrix_free.get_n_q_points_face(quad_index));
+
+      if(use_cell_based_faces)
```
nfehn, line 187: @btemuer Do you think it makes sense to add a comment here why we need all coefficient tables in case of cell-based face loops (I mean also the `coefficients_face` and `coefficients_face_neighbor` tables above)?

include/exadg/operators/variable_coefficients.h
```diff
@@ -23,168 +23,213 @@
...
+  //! Coefficient table for faces
   dealii::Table<2, coefficient_type> coefficients_face;
+
+  //! Coefficient table for neighbor faces
   dealii::Table<2, coefficient_type> coefficients_face_neighbor;
 
-  // TODO
-  //  // cell-based face loops
-  //  dealii::Table<2, coefficient_type> coefficients_face_cell_based;
+  //! Coefficient table for faces with cell-based access
```
nfehn, line 225: Maybe add a comment that this is an "additional" data structure (in addition to the face coefficients above), because `use_cell_based_face_loops` is additional/optional functionality. See also comment https://github.com/exadg/exadg/pull/406/files#r1176381729.

include/exadg/operators/variable_coefficients.h
```diff
@@ -23,168 +23,213 @@
...
+  dealii::Table<2, coefficient_type> coefficients_face_cell_based;
+
+  //! Boolean switch to use a coefficient table for faces
+  bool use_faces{false};
+
+  //! Boolean switch to use a coefficient table for cell-based face access
+  bool use_cell_based_faces{false};
```
nfehn, line 232: I think the naming is still not 100% clear. We could think whether a naming like "enable" or "store" instead of "use"  (and "face_data" instead of "faces") is clearer in the context of this class (which is about the data structure and not so much about the functionality of filling the data or performing loops). What do you think, @btemuer? Of course, these changes are not mandatory for the present PR.

include/exadg/operators/variable_coefficients.h
```diff
@@ -23,168 +23,213 @@
 
 namespace ExaDG
 {
+/**
+ * This class serves as a manager of several table objects, which hold quadrature-point-level
+ * coefficient information for the different matrix-free loop types and access methods.
+ *
+ * The class provides functionality for cell loops, separate face loops, and cell-based face loops.
+ * For the separate face loops, coefficient information from the neighbor cell can also be stored
+ * and used.
```
nfehn, line 32: This sentence can now be removed.

Maybe replace "functionality" by "data structures" (or similar) in the sentence above, or is "functionality" intended here?

include/exadg/operators/variable_coefficients.h
btemuer: I don't think it would be better to have that explanation here. The variable coefficients class is better encapsulated, when it just "does what it's told" without the need to know the reasoning behind. Such an explanation would be much better off in a place where this parameter is defined and explained. This will also be much clearer, when we refactor our parameter interface with nested parameters... Then this boolean will for example occur inside the preconditioner parameters.

However, I agree the name `use_cell_based_loops` somehow implies here that it is the only one to be used. Also respecting your comment from below, I will change the variable names to `store_*`.

include/exadg/operators/variable_coefficients.h
btemuer: Sure!

include/exadg/operators/variable_coefficients.h
btemuer: I will replace "functionality" by "data structures", better than removing in my opinion.

include/exadg/operators/variable_coefficients.h
btemuer: Much better naming, I'm on it.

include/exadg/operators/variable_coefficients.h
nfehn: with removing I only mean the second sentence, because we skipped one boolean parameter.

btemuer: Oh yes, I see 👍

nfehn (approved):

btemuer:
Can we merge?

Issue #407, opened 2023-04-25
Periodic Boundary Conditions and Local Smoothing Multigrid


jh66637:
Running the fda_benchmark problem with precursor leads to a segmentation fault with the current master. As far as I can see, this is related to changes in the multigrid implementation. It seems that periodic boundary conditions are not covered anymore. 

@necioglu @nfehn

nfehn:
There was this PR #389 on periodic faces for multigrid. Do you use the "global coarsening" multigrid variant?

jh66637:
Thanks for pointing this out. The FDA benchmark is using the default local smoothing multigrid variant. So I think it is unrelated to the PR.

jh66637:
You are right. The turbulent channel test case is probably the better example. I checked it and the segfault happens there as well. 
I moved to the code version before https://github.com/exadg/exadg/pull/389 got merged. The problem happens there already. I will try to find the last version where it worked and keep you posted.

PR #408, opened 2023-04-26
only access boundary indices if they exist

jh66637:
Fixes https://github.com/exadg/exadg/issues/407

@nfehn I don't know if this becomes problematic if a DBC and a Periodic BC is present. What do you think?


nfehn:
Maybe. I think the old code in ExaDG has been used before to solve problems with periodic boundaries. Maybe something has changed in deal.II? @peterrum Maybe you could help here.

nfehn:
@jh66637 does the call `mg_constrained_dofs.get_boundary_indices()` cause the segmentation fault?

> @nfehn I don't know if this becomes problematic if a DBC and a Periodic BC is present. What do you think?

Just to make sure that I understand it correctly: Do you think the old code or the new code might be problematic?

PR #381, opened 2023-03-11
Generalized newtonian fluids

richardschu:
extension of the incompressible Navier-Stokes solver towards generalized Newtonian fluids. I did convergence tests, giving the same convergence rates as their Newtonian counterparts (`BDFCoupledSolution`, `BDFPressureCorrection`, `BDFDualSplittingScheme`).

Some bullet points regarding the implementation:

parameters are unified in nested structs defined in `user_interface/viscosity_model_data`
thus i modified the applications, which had `use_turbulence_model == true`

the viscosity update is triggered at the same point were it was triggered for the turbulence models, i.e., prior to momentum update steps and in addition to that, I introduced the nonlinear coupling in `evaluate_nonlinear_residual` called within Newton solver. So that is a fixed-point scheme for the viscosity, but Newton for velocity and pressure. I did not see any effects on the line-search technique and the current implementation does the job for hemodynamics.

removed `iNS::spatial_operator_base::get_viscosity()` removed, since it is misleading -> TODO's inserted in the preconditioners. for now, variable viscosity seemed not to be necessary.

two restrictions remain : i) overintegration remains limited to `Overintegration::Standard`, ii) cell_based_face_loops == false required 

renaming to be done: turbulence_model.h/.cpp -> viscosity_model.h/.cpp for better comparison on github

PR #414, opened 2023-04-28
Create Generalized Laplace Operator

### LINKED ISSUE #357: Poisson problem with variable coefficients
It would be really nice to be able to solve the variable coefficient Poisson equation with ExaDG. Apart from being an interesting PDE in and of itself, it is for example relevant for an efficient preconditioning of the Darcy equations (analogous to block preconditioning for the coupled IncNS solver, just with a tensor-valued variable coefficient).

I wrote down a quick derivation of the DG discretization [here](https://github.com/exadg/exadg/files/10841309/DG_Discretization_of_the_Variable_Coefficient_Poisson_Equation.3.pdf), which I would like to implement into ExaDG. Please be welcome to look at it.

There are however some points that have to be discussed about the concrete implementation. I would use this issue to collect these points and discuss with the community to refer back to during the implementation.

Right now, work is ongoing for general support of tensor-valued variable coefficients.
 
Blocked by:
- #128 
- #289 
- #354

bugrahantemur:
This PR introduces a new Laplace operator implementation that can be used by any module in ExaDG.

The requirements that this new operator fulfills were listed in #357. A document with the math behind this operator is also linked in that issue.

In short, the operator is templated for the scalar and vector-valued problems, as well as coefficients that couple or do not couple between the components of the gradient.

For a proof-of-concept, this PR also removes the diffusive operator in the convection-diffusion module and replaces it with the new generalized Laplace operator. Other than this, as few places as possible are touched in this PR to keep it manageable for reviews. 

---
Here are some details about the implementation:

- Operators that use the implemented kernel (these are right now the operator itself, and the combined operator in the convection-diffusion module) are responsible for implementing their own cell-loop/face-loop functions to decide how to calculate coefficients, how to store them (like having cell-based face loops or not), and how and when to update them.

- `do_boundary_integral_cell_based()` is introduced in the base class operator `OperatorBase` fully analogous to the already present `do_face_int_integral_cell_based()` as these two functions need to access the cell-based face coefficients unlike their non-cell-based counterparts.

- `fill_matrix_free_data()` in `IncNS::SpatialOperatorBase` is made virtual and overriden in `IncNS::OperatorCoupled`, since the generalized Laplace operator requires to set additional flags if it is used as a preconditioner.

---
Note: The convection-diffusion module is currently the only place where the operator is used. However, the applications that test this module do not use variable coefficients as the diffusivity is assumed constant, i.e. the variable coefficient functionality is disabled for optimization. During development, these tests were also successfully run with variable coefficients, but this is not part of the PR to keep changes to a minimum.  

Closes #357


nfehn:
Yes, I have it on my agenda, but I am currently struggling to get ExaDG running on my new machine, get tests passing, etc.

include/exadg/convection_diffusion/spatial_discretization/operator.cpp
```diff
@@ -855,7 +863,10 @@ Operator<dim, Number>::update_after_grid_motion()
   // of elements
   if(param.diffusive_problem())
   {
-    diffusive_kernel->calculate_penalty_parameter(*matrix_free, get_dof_index());
+    if(param.use_combined_operator)
+      combined_operator.update_after_grid_motion();
+    else
+      diffusive_operator.update_after_grid_motion();
```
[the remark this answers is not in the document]
btemuer: That's a good point. The reason for this is that the operator calculates the coefficient and puts it into the kernel, i.e. the coefficient calculation is not part of the kernel functions.

In other words, the kernel describes how the coefficients are used for the integrals, but does not know how to calculate these coefficients. This is delegated to the operator which uses the kernel. So, how the coefficients are calculated can differ for different operators. This is in contrast to the penalty parameter, where the calculation is always the same. 

Take for example the viscous operator which might have a completely different way of calculating the coefficients via generalized Newtonian fluid models, or another operator which has a continuous field function for the coefficients. They can however use the same kernel.

This results in this change you marked that we must make sure that the operators using this kernel recalculate the coefficients in new time steps or after grid motion. 

Finally, as the `GeneralizedLaplace::Operator` provides all the functionality needed as an actual "operator", we can for example create new operators that inherit from it and only override the coefficient calculation routines.

I hope this makes sense.

nfehn:
@btemuer Now I think I got a first overview of this PR. I very much appreciate the effort that you made so far.

In my opinion, the current design has the following problem
- we have duplicated code between the `GeneralizedLaplace::Operator` and the `CombinedOperator` (all the functions calculating the variable coefficient). The code duplication is somewhat substantial, so I believe we really have to avoid this.
- the implementation seems to be somewhat limited to variable coefficient fields provided via `Functions`. Extending this to other types of models will probably require changes at several places in the code, because of the code duplication item above.

These problems are kind of solved in the `IncNS` module, where we have variable viscosity models in the form of turbulence models and generalized Newtonian models (see the recent PR #381). The solution is to use the design principle "composition", i.e. a class that operates on / manipulates the variable coefficient data structure living in `Kernel`. It is important to not realize this functionality inside the `Operator`, because this limits flexibility and causes code duplication. In principle, we want to be able to use variable coefficient fields based on an arbitrary underlying function or model, which might depend on other dof-vectors or originate from any "multiphysics" context. This modularity can in my opinion be achieved in an elegant way by composition, see the class `IncNS::SpatialOperatorBase` with member variables `TurbulenceModel` and `GeneralizedNewtonianModel`. The viscous `Kernel` or `Operator`, however, should not know where the variable viscosity comes from (analytical function vs. a "complex" model). In principle, one could add another model like `AnalyticalViscosityModel` derived from `ViscosityModelBase` and implementing a variable viscosity field provided by the user via a `dealii::Function`. Note also that in order to be able to combine different models (assuming they are "additive"), the variable viscosity models in the `IncNS` module implement an `add_viscosity()` (not a `calculate_viscosity()`) logic.

If `not(coefficient_is_variable)`, we need to make sure in my opinion that the `VariableCoefficient` data structures storing coefficients for all q-points is not used, but is instead by-passed using the constant coefficient that is a member variable of `OperatorData`. Otherwise, I have the fear that we make the standard case with constant coefficient too slow in terms of computational efficiency, increasing the data transfer from main memory significantly when reading q-point data unnecessarily in the constant coefficient case. This logic with "optimized code path" for the constant coefficient case should be handled by `Kernel` (owning the `VariableCoefficient` data structure) according to the single-responsibility  principle. In this sense, the `VariableCoefficient` object is only a feature of the `Kernel` that is only used if `coefficient_is_variable`.

include/exadg/convection_diffusion/spatial_discretization/operator.cpp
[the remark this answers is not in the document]
nfehn: I think one should consider the `Kernel` as the central object to be used 
for updating data, see the `IncNS` case for comparison: 
`IncNS::SpatialOperatorBase<dim, Number>::update_after_grid_motion()`.

Regarding this aspect: 
>we can for example create new operators that inherit from it and only override the coefficient calculation routines.

This serves as a good example in ExaDG where we intentionally decided 
against inheritance. The responsibility of the `Operator` is only to 
implement the cell / face integrals of the DG discretization of a PDE 
operator. The physics should happen as much as possible in the `Kernel`.

btemuer:
Your proposed approach sounds quite good, and a separate object implementing the coefficient calculation can be done easily, but I think there is an important point which may have slipped your attention, which is the operators inside the multigrid preconditioners.

Namely, the preconditioners operate on a number of _actual_ operators, like the `IncNS::MomentumOperator` or the `ConvDiff::CombinedOperator`  which exist simultaneously on multiple levels. These operators each __create__ and own their own kernel objects (this works fine as long as every needed information is present in the kernel). If the kernel has mesh-dependent _physics_, like calculating the penalty parameter in the viscous or the diffusive kernels, `update_after_grid_motion` is called on the operator, see:

https://github.com/exadg/exadg/blob/9aba12677bb4025a2608eeff5b5f0c396932527c/include/exadg/incompressible_navier_stokes/preconditioners/multigrid_preconditioner_momentum.cpp#L234-L240

which in turn might do something like:

https://github.com/exadg/exadg/blob/9aba12677bb4025a2608eeff5b5f0c396932527c/include/exadg/incompressible_navier_stokes/spatial_discretization/operators/momentum_operator.cpp#L132-L137

However, this cannot include updating the coefficients if the operator or the kernel is not supposed to own the coefficient calculation object. Was this clear, else we can maybe discuss in person?

btemuer:
I pushed a new design trying to fit into our constraints. Here are some points:

- I implemented a new generic `AnalyticalTensorCoefficientFunctionModel` which is templated for the kernels. This object can be used to evaluate a `dealii::Function`  and set the values at the quadrature points for any kernel that supports the operations `set_coefficients_cell()` etc.

- In the kernel operator I added an enum to select different coefficient models. Right now only `ConstantFunction` and `VariableFunction` are supported. This was the only way I found which allows the kernel data to stay portable between different instances of the kernel, which is very important due to the MultiGrid implementation (see my comment from above). Other models, e.g. computing the coefficients from a solution vector, may add the desired dof_index to the kernel data (analogous to the convective kernel design) and implement these models independently from the kernel.

This design allowed an elegant way of avoiding code duplication, keeping the FE operators clean (as @nfehn highlighted), and also keeping the kernels as small as possible in terms of implemented algorithms but also as detailed as necessary in the current framework.

nfehn:
> However, this cannot include updating the coefficients if the operator or the kernel is not supposed to own the coefficient calculation object. Was this clear, else we can maybe discuss in person?

As a general answer to this question: You are right that multigrid has its own design challenges. In ExaDG, the general/current strategy is to ask the fine-level operator for the data, and then interpolate the data to coarser multigrid levels. We do this currently e.g. to interpolate the velocity vector (IncNS) or displacement vector (Structure) to coarser levels. These vectors originate from the linearization inside the Newton solver. Generally, I would argue that one does not even have the information to recompute the variable coefficient on the coarser levels as it is done on the fine level. The variable coefficient might dependent on the solution vector, which is only known on the fine level and not on the coarser levels. If I take the example of a turbulence SGS model: Interpolating the fine-level turbulent viscosity to coarser levels is something different than interpolating the velocity to coarser levels and recomputing the turbulent viscosity on the coarser levels (using the viscosity model). However, I think the former is what one wants to have for multigrid? 
In the `Structure` module, the operator owns the `MaterialHandler`, which could also be seen similar to the variable coefficient model. However, I think this operator is not used in a "combined" fashion with other operators. Since we want to write a generalized Laplace operator, I believe we can not follow the design chosen in the `Structure` module.

 So in my opinion the procedure to enable multigrid should be:

1. Ask the fine-level operator for the data
2. Interpolate fine-level data to coarser levels e.g. in `MyModule::MultigridPreconditioner::update()` (this is an open problem in ExaDG, where implementations still have to be developed; questions: q-point vs dof-data layout for interpolation; which type of interpolation for structure-preservation)
3. Inject the coarse data into the coarse-level operators/kernels.
4. Ready for multigrid

The philosophy for our geometric multigrid algorithm implementation in ExaDG is: 

*The fine-level operator must be able to provide all the data to construct the coarse-level multigrid operators.*

The general design of the PDE operator, however, should not be driven by the multigrid preconditioner. To solve the current problem with an analytical variable coefficient model regarding multigrid: I see no harm if the multigrid preconditioner for a certain module depends on the spatial operator of this module, which is the case for most multigrid preconditioners in ExaDG. In your case and to avoid the problem of interpolating fine-level data to coarse levels, one could easily pass the analytical variable coefficient model to this particular multigrid preconditioner during initialization. Just as the variable coefficient model is "module-dependent", the multigrid preconditioners may also be module-dependent. I see this as the burden/price of geometric multigrid.

include/exadg/operators/generalized_laplace_operator.h
```diff
@@ -0,0 +1,849 @@
...
+template<typename T, int coefficient_rank, int dim, typename Number>
+static inline DEAL_II_ALWAYS_INLINE //
+  T
+  coeff_mult(dealii::Tensor<coefficient_rank, dim, Number> const & coeff, T const & x)
+{
+  if constexpr(coefficient_rank == 4)
+    return dealii::double_contract<2, 0, 3, 1>(coeff, x);
+  else
+    return coeff * x;
+}
```
nfehn, line 50: "Multiplication **of what** with coefficients of different ranks". The word "different" is misunderstandable. I would write "... from the left with a coefficient tensor of rank `coefficient rank`".

In my opinion, the variable names `T`, `coeff_mult`, `x` are too short. Also `2,0,3,1` is somewhat hard to understand.

include/exadg/operators/generalized_laplace_operator.h
```diff
@@ -0,0 +1,849 @@
...
+  double IP_factor{1.0};
+
+  //!
+  bool calculate_and_store_cell_based_face_coefficients{false};
+
+  //!
+  CoefficientModel coefficient_model;
+
+  //!
+  std::shared_ptr<dealii::Function<dim>> coefficient_function;
```
nfehn, line 74: See comment https://github.com/exadg/exadg/pull/414#issuecomment-1563993006

include/exadg/operators/generalized_laplace_operator.h
```diff
@@ -0,0 +1,849 @@
...
+};
+
+/**
+ * Generalized Laplace kernel. This class is responsible for providing the core pieces of
+ * information required to evaluate the Laplace operator, i.e. the fluxes, numerical parameters
+ * (like the internal penalty parameter), and the coefficients.
```
nfehn, line 80: "interior penalty" should be explained for this generalized Laplace operator. Does the operator support both continuous and discontinuous Galerkin discretizations?

include/exadg/operators/generalized_laplace_operator.h
```diff
@@ -0,0 +1,849 @@
...
+ *
+ * @tparam n_components Number of components of the solution (vector) field. Solution is scalar if
+ * this is equal to 1.
+ * @tparam coupling_coefficient Boolean switch to differentiate if the coefficient type couples
+ * the solution vector components or not.
+ *  - `false` always results in a scalar coefficient,
```
nfehn, line 86: I do not fully understand this. What about a 1-component solution field (gradient has dim components) with a diagonal 2-tensor as variable coefficient? The result of `coeff_mult()` would again be a vector with dim components. Would one call this a "coupling"?

include/exadg/operators/generalized_laplace_operator.h
```diff
@@ -0,0 +1,849 @@
...
+ *  - `true` results in a 2-rank tensor coefficient if `n_components = 1` and in a 4-rank tensor
```
nfehn, line 87: Regarding the spelling, I think I would write "2-tensor" or "rank-2 tensor" or "tensor of rank 2". But maybe your writing is common as well.

include/exadg/operators/generalized_laplace_operator.h
```diff
@@ -0,0 +1,849 @@
...
+ *  coefficient if `n_components > 1`.
+ */
+template<int dim, typename Number, int n_components = 1, bool coupling_coefficient = false>
```
nfehn, line 90: I find the name `coupling_coefficient` irritating. From the name, I would expect a coefficient, not a boolean. What about `coefficient_couples_components`?

bugrahantemur: Yeah it's a bit irritating 🤣. `coefficient_couples_components` sounds better. Regarding your question from above https://github.com/exadg/exadg/pull/414#discussion_r1206431481 maybe we should even call it `bool coefficient_is_scalar` (with reversed true/false of course), because the case you described with a diagonal rank-2 tensor would have to fall into `coupling_coefficient=true` with the current implementation.

include/exadg/operators/generalized_laplace_operator.h
bugrahantemur: See answer https://github.com/exadg/exadg/pull/414#discussion_r1206458052

include/exadg/operators/generalized_laplace_operator.h
bugrahantemur: You can find literally anything regarding this 😂. Let's go with `rank-X tensor`.

include/exadg/operators/generalized_laplace_operator.h
bugrahantemur: I'll change the sentence, rename the variables and put a comment about the double contraction👍

include/exadg/operators/generalized_laplace_operator.h
bugrahantemur: I've done the implementation with only DG in my head. So I'm not sure if there are any intricacies apart from the interior penalty that I missed about continuous Galerkin. What do you think?

include/exadg/operators/variable_coefficients.h
```diff
@@ -235,6 +237,147 @@ class VariableCoefficients
   bool store_cell_based_face_data{false};
 };
 
+template<typename kernel_type>
```
nfehn, line 240: I would like to avoid this dependency of the analytical coefficient model on the kernel. The connection is actually the `VariableCoefficients` class and not the kernel in my opinion, just as you placed this functionality in the file variable_coefficients.h. I would not template `VariableCoefficients` here, but let the analytical model explicitly use the class `VariableCoefficients`.

include/exadg/operators/generalized_laplace_operator.h
```diff
@@ -0,0 +1,849 @@
...
+
+  using IntegratorCell = CellIntegrator<dim, n_components, Number>;
+  using IntegratorFace = FaceIntegrator<dim, n_components, Number>;
+
+  using This = Kernel<dim, Number, n_components, coupling_coefficient>;
+
+public:
+  static constexpr int DIM              = dim;
+  static constexpr int COEFFICIENT_RANK = coefficient_rank;
+  using NUMBER                          = Number;
```
nfehn, line 116: hopefully no longer necessary with comment https://github.com/exadg/exadg/pull/414/files?diff=unified&w=0#r1206535671

include/exadg/operators/generalized_laplace_operator.h
```diff
@@ -0,0 +1,849 @@
...
+  }
+
+private:
+  //! Generalized Laplace kernel data
+  KernelData<dim> data{};
+
+  //! Element polynomial degree
+  unsigned int degree{1};
+
+  //! Internal penalty parameter
```
nfehn, line 486: Internal -> interior 

Please add "... of symmetric interior penalty (SIP) method"

include/exadg/operators/generalized_laplace_operator.h
```diff
@@ -0,0 +1,849 @@
...
+  using vector = dealii::Tensor<1, dim, scalar>;
+
+  static constexpr unsigned int value_rank = (n_components > 1) ? 1 : 0;
+  static constexpr unsigned int coefficient_rank =
+    (coupling_coefficient) ? ((n_components > 1) ? 4 : 2) : 0;
+
+  using value_type    = dealii::Tensor<value_rank, dim, scalar>;
+  using gradient_type = dealii::Tensor<value_rank + 1, dim, scalar>;
+
+  using coefficient_type = dealii::Tensor<coefficient_rank, dim, scalar>;
```
nfehn, line 770: can one ask the kernel for these types (to avoid this duplication of code)?

include/exadg/operators/generalized_laplace_operator.h
```diff
@@ -0,0 +1,849 @@
...
+ * This class provides the loop structures for the integrals that are evaluated for a generalized
+ * Laplace term.
+ *
+ * @tparam n_components Number of components of the solution (vector) field. Solution is scalar if
+ * this is equal to 1.
+ * @tparam coupling_coefficient Boolean switch to differentiate if the coefficient type couples
+ * the solution vector components or not.
+ *  - `false` always results in a scalar coefficient,
+ *  - `true` results in a 2-rank tensor coefficient if `n_components = 1` and in a 4-rank tensor
+ *  coefficient if `n_components > 1`.
```
nfehn, line 753: We should find a way to make this information unique, e.g. by referring to the kernel.

include/exadg/operators/generalized_laplace_operator.h
```diff
@@ -0,0 +1,849 @@
...
+  bool coefficient_is_variable{};
+
+  //! Constant coefficient
+  mutable coefficient_type constant_coefficient{};
+
+  //! Variable coefficients
+  mutable VariableCoefficients<coefficient_type> variable_coefficients{};
+};
+
+namespace Boundary
```
nfehn, line 502: This is superfluous and just leads to a duplication of names without further information, e.g. `Boundary::BoundaryType`, `Boundary::BoundaryDescriptor`.

include/exadg/operators/generalized_laplace_operator.h
bugrahantemur: I guess I could define them in the namespace as alias templates, so outside the classes instead of inside them 👍

include/exadg/operators/generalized_laplace_operator.h
```diff
@@ -0,0 +1,781 @@
...
+    {
+      constant_coefficient =
+        FunctionEvaluator<CoefficientType::rank, dim, Number>::value(*data.coefficient_function,
+                                                                     {},
+                                                                     {});
+    }
+    else if(data.coefficient_model == CoefficientModel::VariableFunction)
+    {
+      AnalyticalTensorCoefficientFunctionEvaluator<dim, Number, CoefficientType::rank>{
+        matrix_free, quad_index, data.coefficient_function, variable_coefficients}();
```
nfehn, line 323: Syntax: I find this syntax very complicated. Writing `analytical_coefficient_function.evaluate(variable_coefficients)` would be clearer / easier to understand (without need for documentation). Extracting that this line of code modifies `variable_coefficients` took me several minutes. Hence, this style of programming requires much more documentation in my opinion. I doubt this additional effort in documentation justifies the one line of code saved by this programming style.

Design: More importantly, this design does not solve my main criticsm from the last review, namely that `GeneralizedLaplace::Kernel` should in my opinion not depend on `AnalyticalTensorCoefficientFunctionEvaluator`. 

Example: When adding a new type to `CoefficientModel`, the kernel/operator has to be recompiled. I consider this an unnecessary dependency.

If we do not want to provide access to `Kernel::variable_coefficients` via a public getter returning a reference to `VariableCoefficients` (in order to fill the variable coefficients from "outside"), a lambda function with argument `VariableCoefficients &` could be passed to the kernel during construction/initialization, which allows to compute the coefficient field.

What to do if one wants to realize an analytical coefficient function that is time dependent? I believe one has to touch again `Kernel` with the current design, but this is unnecessary. Would you adjust the constructor of `AnalyticalTensorCoefficientFunctionEvaluator` and make `double time` a member variable of that class? This is probably not a good design / unnecessary, because one rather wants to call the `evaluate()` function suggested above with a second argument `time`.

include/exadg/operators/generalized_laplace_operator.h
```diff
@@ -0,0 +1,781 @@
...
+                               "sense with scalar coefficients."))}
+
+    ValueType const jump_value = value_m - value_p;
+
+    return -0.5 * coefficient * jump_value;
+  }
+
+  /**
+   * Calculates and returns the `flux` on the faces which should be weighted with the shape
+   * function values.
```
nfehn, line 256: It is not clear to me why the documentation of these functions should explain "weighted by ...". Implementation of the weak form is in my opinion the responsibility of the `Operator`.

include/exadg/convection_diffusion/spatial_discretization/operators/combined_operator.cpp
```diff
@@ -290,28 +299,37 @@ CombinedOperator<dim, Number>::do_face_integral(IntegratorFace & integrator_m,
...
+      scalar const value_flux =
+        diffusive_kernel->calculate_value_flux(diffusivity * normal_gradient_m,
+                                               diffusivity * normal_gradient_p,
+                                               value_m,
+                                               value_p,
+                                               normal_m,
+                                               diffusivity);
+
+      value_flux_m += value_flux;
+      value_flux_p += -value_flux; // - sign since n⁺ = -n⁻
```
nfehn, line 321: I do not like that the present PR makes such subtle sign changes.

include/exadg/operators/generalized_laplace_operator.h
nfehn: see comment https://github.com/exadg/exadg/pull/414#discussion_r1219851578

include/exadg/operators/generalized_laplace_operator.h
```diff
@@ -0,0 +1,781 @@
...
+{
+namespace GeneralizedLaplace
+{
+// constant expression parameters
+template<int n_components>
+constexpr unsigned int value_rank = (n_components > 1) ? 1 : 0;
+
+template<int n_components, bool coefficient_is_scalar>
+constexpr unsigned int coefficient_rank = (coefficient_is_scalar) ? 0 :
+                                                                    ((n_components > 1) ? 4 : 2);
```
nfehn, line 42: I think explanations/motivation for these numbers would be helpful here.

include/exadg/operators/generalized_laplace_operator.h
nfehn: Let's also make the function clearer by using full names.

include/exadg/operators/generalized_laplace_operator.h
bugrahantemur: I wrote the documentation like this to make it clear where the name of the function comes from. Because that is our motivation to name this function `get_value_flux()`. The caller shall multiply the return value with the shape function values. The naming of the function is not obvious from its body, which for example also works with gradients here. I just wanted to prevent any confusion caused by the function name.

include/exadg/operators/generalized_laplace_operator.h
```diff
@@ -0,0 +1,781 @@
...
+                         ValueType const &       value_p,
+                         Vector const &          normal,
+                         CoefficientType const & coefficient)
+  {
+    ValueType const    jump_value  = value_m - value_p;
+    GradientType const jump_tensor = outer_product(jump_value, normal);
+
+    GradientType const average_gradient = 0.5 * (gradient_m + gradient_p);
+
+    return -coeff_mult(coefficient, (average_gradient - tau * jump_tensor)) * normal;
```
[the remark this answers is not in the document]
bugrahantemur: That's because I didn't want to implement operator as the negative Laplace operator. But of course I can change it to be the negative Laplace as in other parts of ExaDG.

bugrahantemur:
I'm wondering if `std::visit` to get coefficients on `std::variant<VariableCoefficients<Coeff>, Coeff>`  in the innermost loop is going to ruin performance as it won't be inlined...

include/exadg/convection_diffusion/preconditioners/multigrid_preconditioner.cpp
```diff
@@ -50,12 +51,14 @@ MultigridPreconditioner<dim, Number>::initialize(
   PDEOperator const &                         pde_operator,
   MultigridOperatorType const &               mg_operator_type,
   bool const                                  mesh_is_moving,
+  Number const                                mg_diffusivity,
```
nfehn, line 54: I think one should ask the `pde_operator` for the diffusivity instead of extending the interface.

include/exadg/convection_diffusion/preconditioners/multigrid_preconditioner.cpp
```diff
@@ -236,6 +239,18 @@ MultigridPreconditioner<dim, Number>::initialize_operator(unsigned int const lev
...
+  /*
+   * Set the diffusivity on the operator to the provided constant diffusivity value.
+   * Although the operator can deal with variable coefficients, for now, we use a constant
+   * diffusivity in the multigrid. This can and porobably will be changed in the future to be able
+   * to employ multigrid with variable diffusivity.
+   */
+  if(data.diffusive_problem)
+  {
+    auto & diffusivity = pde_operator_level->get_diffusivity();
+    diffusivity        = mg_diffusivity;
```
nfehn, line 251: Why returning a non-const reference and then overwrite the value? I think I would then prefer a set-function as we usually do in ExaDG.

include/exadg/convection_diffusion/spatial_discretization/operator.cpp
```diff
@@ -826,11 +830,76 @@ template<int dim, typename Number>
...
   {
+    // update SIPG penalty parameter of diffusive operator which depends on the deformation
+    // of elements
     diffusive_kernel->calculate_penalty_parameter(*matrix_free, get_dof_index());
+
+    // update the diffusivity coefficients which depend on the quadrature point coordinates
+    // and may depend on the time, e.g. in case of an analytical coefficient function.
+    // However, the ALE interface does not support passing in the time here, so the coefficients
+    // are calculated with the simulation start time.
+    calculate_diffusivity(this->param.start_time);
```
nfehn, line 843: The problem is that this code is not placed correctly in `update_after_grid_motion()`. Also as a user, I can not expect that for a time-dependent problem on a non-moving mesh, the calculation of a time-dependent diffusivity will be done in a function that explicitly relates to grid-motion (ALE) problems.

include/exadg/convection_diffusion/spatial_discretization/operator.cpp
```diff
@@ -826,11 +830,76 @@ template<int dim, typename Number>
...
+  // Initialization necessary for variable diffusivity models (right now only
+  // AnalyticalFieldFunction)
+  if(param.diffusivity_model == DiffusivityModel::AnalyticalSpaceAndTimeFunction)
+  {
+    using VariableDiffusivity =
+      VariableCoefficients<typename GeneralizedLaplace::Kernel<dim, Number>::CoefficientType>;
+
+    diffusivity = VariableDiffusivity{};
+
+    std::get<VariableDiffusivity>(diffusivity)
```
nfehn, line 862: what is `get` doing here? Could you explain and/or add a comment?

nfehn:
Let me raise a general question: I see the novelty of the present PR in the introduction of a generalized Laplace operator. However, it does not really seem to fit optimally that this is new implementation is then used in the present PR to replace the diffusive operator in the `ConvDiff` module. So far, the `ConvDiff` module does not support variable diffusivities, while we have variable diffusivities/viscosities in the `IncNS` module, which is therefore much more suitable as an application example for a generalized, variable coefficient Laplace operator. As a consequence, the present PR has to introduce new functionality for the `ConvDiff` module (functions like `calculate_viscosisty()` in the spatial operator, changes to the multigrid preconditioner), even though there seems to be no motivation for such functionality currently from an application perspective. As a consequence, I believe the present PR mixes too many topics, that make code review unnecessarily complex and somewhat hinder the progress in the sense of merging new functionality step by step. So I want to raise the question whether it would be better to first introduce the generalized Laplace operator before using it in a concrete module in ExaDG.

nfehn:
Related to the comment https://github.com/exadg/exadg/issues/357#issuecomment-1649350939, I believe the present PR needs to decide whether to support for now variable coefficients for the generalized Laplace operator or not. I mainly see the following two options:
- If not, using the new generalized Laplace operator for the diffusive operator in the `ConvDiff` already in the present PR is reasonable since this operator does so far not support variable coefficients. Then, no redundancies are introduced by the present PR. The present PR replaces `ConvDiff::DiffusiveOperator` by a more general implementation. The present PR does, however, not extend the functionality of the `ConvDiff` module.
- If yes, the implementation of the generalized Laplace operator introduced here probably needs to care more about the DG methodology for the generalized Laplace operator, and this should probably be introduced stand-alone, i.e. the present PR should really focus on such aspects and not on changes in the `ConvDiff` module. This would imply that we have for now some redundancy with `IncNS::ViscousOperator`, but this is acceptable since we have the plan to replace this functionality later by the new variant. If no one wants to do this in the future, one would simply remove the new generalized Laplace operator again from ExaDG, which is easy since it is not used so far. Of course, this is not intended, but I just want to describe the general procedure to avoid code duplications in ExaDG over a longer period. In any case, I think this is better than introducing another variable coefficient operator in `ConvDiff` actually not needed right now.

include/exadg/operators/variable_coefficients.h
```diff
@@ -235,6 +243,225 @@ class VariableCoefficients
...
+      }
+    }
+  };
+
+  dealii::MatrixFree<dim, Number> const & matrix_free;
+  unsigned int const                      quad_index;
+  dealii::Function<dim> &                 coefficient_function;
+  double const                            time;
+  VariableCoefficients<CoefficientType> & variable_coefficients;
+};
```
nfehn, line 391: This should not be realized inside `variable_coefficients.h`.

include/exadg/operators/generalized_laplace_operator.h
```diff
@@ -0,0 +1,744 @@
...
+   *
+   * If the coefficient is not variable, returns the constant coefficient.
+   */
+  CoefficientType
+  get_coefficient_face_cell_based(unsigned int const cell_based_face, unsigned int const q) const
+  {
+    return std::visit(VarCoeffUtils::GetCoefficientFaceCellBased<CoefficientType>{cell_based_face,
+                                                                                  q},
+                      coefficients);
+  }
```
nfehn, line 319: To be honest, I do not see the advantages of the variant/visitor design patter here. In my opinion, one needs to write more lines of code, and the code is more difficult to understand than a very simple if-else logic. Since we are here in the generalized Laplace operator, there is no concern right now that this if-else logic needs to be duplicated at multiple places. And even in such a case, one could also introduce free functions in the future, but only once we need this complexity. So where is the advantage of the present design? Defining two functions `operator()` with the same name is essentially a complicated way of implementing an if-else logic.

include/exadg/operators/variable_coefficients.h
```diff
@@ -235,6 +243,225 @@ class VariableCoefficients
...
+ * Visitor to access the coefficient on cells to use with variants either holding a single constant
+ * coefficient value or a VariableCoefficients object,
+ *
+ * i.e. visitor for std::variant<CoefficientType, VariableCoefficients<CoefficientType>>
+ */
+template<typename CoefficientType>
+struct GetCoefficientCell
+{
+  unsigned int const cell;
+  unsigned int const q;
```
nfehn, line 403: In terms of code design, I am not convinced that it is a good idea that the most complex case (with face, q variables) dictates the "interface" also for the simplest case of a constant coefficient (not even mentioning that it is quite hard to identify an "interface" from this struct definition). While I do not see many other variants to be realized in the future, I also do not understand the motivation for the present design if we had more variants in the future. If all these variants have different interfaces, the interface of this struct will grow according to the number of all variants, just as the simple if-elseif logic.

bugrahantemur:
I think this PR is stuck right now between trying to have a use for the generalized Laplace operator, and the features it needs to have due to its original motivation, which was to be used in the context of the Darcy equation. I am strongly against merging such a big contribution without using it anywhere. On the other hand, trying to use it with existing modules, as pointed out, then causes problems, for example by introducing features that are not application motivated. The interfaces with the modules become a problem, as well. 

In light of this, I think it is wiser that I further develop this operator together with the Darcy module in my own repository, use and test it there. Also concerning external constraints like development time, I think this is the best way to go. After the operator has been used there successfully with the Darcy module, we can later decide what parts of it we want to have in ExaDG and how we want it specifically. We can the import parts here step by step. This would be easier for everybody, I guess.

So, I would actually suggest closing this PR for now. What do you think @nfehn ?

bugrahantemur:
Yes, we have come a long way due to our continuous efforts and that's great. What I'm talking about is just going to be transient (I'm planning something like 2-3 months maximum). In the meantime, I will keep up with the developments in the main repo, but I consider this a necessary step for the speed of development I need right now (conference mid-September 😅) As soon as I have things working and got some results, I want to put it all into the main repo, just as was our plan all along. Of course, it will all build on the state (of this PR) we have achieved until now.

PR #417, opened 2023-05-03
Provide configure script and installation description for HDF5

### LINKED ISSUE #410: HDF5 is enabled by default, but no installation description is provided
The default setting in `config_dealii.sh` is `DEAL_II_WITH_HDF5:BOOL="ON"`, but the wiki page https://github.com/exadg/exadg/wiki/Installation does not provide an installation description for HDF5.

nfehn:
closes #410 

Together with the configure script introduced here, let us discuss the installation description for HDF5 in the wiki https://github.com/exadg/exadg/wiki/Installation during this PR. @jh66637 I did not understand whether `sudo apt install ...` and the variant with `.configure ...` are two different alternatives or not. In particular, I did not know what exactly you mean by `download hdf5`. Could you explain this?

### DIFF
--- a/scripts/config_hdf5.sh
+++ b/scripts/config_hdf5.sh
@@ -1,0 +1,30 @@
+#!/bin/sh
+#########################################################################
+# 
+#                 #######               ######  #######
+#                 ##                    ##   ## ##
+#                 #####   ##  ## #####  ##   ## ## ####
+#                 ##       ####  ## ##  ##   ## ##   ##
+#                 ####### ##  ## ###### ######  #######
+#
+#  ExaDG - High-Order Discontinuous Galerkin for the Exa-Scale
+#
+#  Copyright (C) 2021 by the ExaDG authors
+#
+#  This program is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation, either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+#########################################################################
+
+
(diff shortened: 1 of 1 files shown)


scripts/config_hdf5.sh
```diff
@@ -0,0 +1,30 @@
...
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+#########################################################################
+
+
+./configure --prefix=$WORKING_DIRECTORY/hdf5/install/ --enable-unsupported --enable-symbols=yes --disable-threadsafe --disable-cxx --enable-hl --without-szlib --enable-parallel --enable-shared CC=mpicc
```
nfehn, line 30: I think we should use separate rows for every option for readability, agree? See for comparison e.g. the configure script for PETSc.

jh66637: I agree, I just copied the line from a script I had.

jh66637:
Download HDF5 would be 
```
wget https://github.com/HDFGroup/hdf5/archive/refs/tags/hdf5-1_12_2.tar.gz
tar -xzf  hdf5-1_12_2.tar.gz
cd hdf5-hdf5-1_12_2/
```

jh66637:
I think the wiki entry looks good. Just some minor comments:
1) "Next, go to your working directory" -> "If you want to install HDF5 from sources"
2) ```make``` -> ```make -j<N>```
3) I am pretty sure that one has to provide the HDF5 library directory and include directory in the deal.ii configure script if installation is done from sources. Since I am running on a system with environment modules, this is not a problem since ```module load``` exports the corresponding path. I am currently not able to check it without environment modules, though.

nfehn:
So the manual installation is an alternative to `sudo apt ...`. Or do I still have to run the configure-script/make command?

nfehn:
could you check again the wiki page?

jh66637:
> So the manual installation is an alternative to `sudo apt ...`. Or do I still have to run the configure-script/make command?

It's an alternative on systems where you can not use a package manager.

jh66637:
Regarding the Wiki:
If you could change the HDF5 section to HDF5 (optional) I think it is more consistent to other optional packages. Before doing ```wget``` I would probably ```cd hdf5```, but both should work; it's probably a matter of taste.

nfehn:
I made these changes as suggested

jh66637:
It would be best if someone else can confirm the installation procedure works. In general, I am confident since it worked for me some month ago. So this is OK from my side once the line breaks in the configure script are added :)

jh66637 (approved):

PR #418, opened 2023-05-03
fix config error for trilinos by setting a GlobalOrdinal type

### LINKED ISSUE #409: Configuration error with trilinos
With the standard configuration script `config_trilinos.sh`, I get a configuration error for trilinos:

```bash
Processing enabled package: Tpetra (Classic, TSQR, Core)
-- Tpetra: Enabling deprecated code
-- Determine whether Tpetra will assume that MPI is CUDA aware:
--   - Tpetra_ENABLE_CUDA is OFF, so Tpetra will assume that MPI is not CUDA aware.
-- Tpetra execution space availability (ON means available): 
--   - Serial:  ON
--   - Threads: OFF
--   - OpenMP:  OFF
--   - Cuda:    OFF
-- Tpetra: Tpetra_INST_INT_LONG_LONG is enabled by default.
-- Tpetra: Tpetra_INST_INT_UNSIGNED is disabled by default.
-- Tpetra: Tpetra_INST_INT_UNSIGNED_LONG is disabled by default.
-- Tpetra: Tpetra_INST_INT_INT is disabled by default.
-- Tpetra: Tpetra_INST_INT_LONG is disabled by default.
-- 
-- Tpetra: Validate global ordinal setting ...
-- Tpetra: ERROR Tpetra requires a GlobalOrdinal type, but none were specified.
-- This can happens when the default is overridden by user options without enabling
-- another GlobalOrdinal type in its place. Please check your configuration to ensure
-- that the default is not being specifically disabled without enabling one of the others:
--    Tpetra_INST_INT_INT
--    Tpetra_INST_INT_LONG
--    Tpetra_INST_INT_LONG_LONG (Default, but was overridden)
--    Tpetra_INST_INT_UNSIGNED
--    Tpetra_INST_INT_UNSIGNED_LONG
CMake Error at packages/tpetra/CMakeLists.txt:1145 (MESSAGE):
  Tpetra: Tpetra requires a global ordinal type, but none were specified.
  See above message for options.
```

I suggest to set `Tpetra_INST_INT_LONG_LONG=ON` to solve this. Any opinions?

nfehn:
closes #409

### DIFF
--- a/scripts/config_trilinos.sh
+++ b/scripts/config_trilinos.sh
@@ -127,7 +127,7 @@ cmake \
     -D Trilinos_ENABLE_Teko:BOOL=OFF \
     -D Teko_ENABLE_TESTS:BOOL=OFF \
     -D Trilinos_ENABLE_MueLu:BOOL=ON \
-    -D Tpetra_INST_INT_LONG_LONG:BOOL=OFF \
+    -D Tpetra_INST_INT_LONG_LONG:BOOL=ON \
     -D Tpetra_INST_INT_LONG:BOOL=ON \
     -D MueLu_ENABLE_TESTS:BOOL=OFF \
     -D MueLu_ENABLE_EXAMPLES:BOOL=OFF \
--- a/scripts/config_trilinos.sh
+++ b/scripts/config_trilinos.sh
@@ -128,7 +128,6 @@ cmake \
     -D Teko_ENABLE_TESTS:BOOL=OFF \
     -D Trilinos_ENABLE_MueLu:BOOL=ON \
     -D Tpetra_INST_INT_LONG_LONG:BOOL=ON \
-    -D Tpetra_INST_INT_LONG:BOOL=ON \
     -D MueLu_ENABLE_TESTS:BOOL=OFF \
     -D MueLu_ENABLE_EXAMPLES:BOOL=OFF \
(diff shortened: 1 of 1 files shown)


scripts/config_trilinos.sh
```diff
@@ -127,7 +127,7 @@ cmake \
     -D Trilinos_ENABLE_Teko:BOOL=OFF \
     -D Teko_ENABLE_TESTS:BOOL=OFF \
     -D Trilinos_ENABLE_MueLu:BOOL=ON \
-    -D Tpetra_INST_INT_LONG_LONG:BOOL=OFF \
+    -D Tpetra_INST_INT_LONG_LONG:BOOL=ON \
     -D Tpetra_INST_INT_LONG:BOOL=ON \
```
jh66637, line 131: I think this should be off if the other option is turned on? This is at least what I did.

kronbichler: It has been a while since I last tried, but if both `LONG` and `LONG_LONG` can be on at the same time (I believe on Linux 64 bit they map to the same type), we can go with this.

nfehn: I think the second line (`...INT_LONG:BOOL`) is a leftover, so I removed it. However, I do not remember exactly.

PR #420, opened 2023-05-03
remove function IncNS::Parameters::linear_problem_has_to_be_solved()

nfehn:
The changes of this PR might appear somewhat unnecessary, so let me explain them in detail:

The name `IncNS::Parameters::linear_problem_has_to_be_solved()` is actually hard to understand in the context of `IncNS` (with monolithic vs. splitting-type solvers), because the name is actually meant in the sense of "no nonlinear problem". Moreover, what is the meaning of this function in case of splitting methods (pressure-correction, dual splitting), where we always have a linear problem to solve (pressure Poisson equation)? 

For this reason, I believe it is better to remove this function entirely. Since the function is mostly used in if-else constructions, we do not even need the negation `not nonlinear_problem_has_to_be_solved()` in the code, but can instead simply change the order of the `if` and `else` branch.

@richardschu This is code you are currently working on. I would therefore suggest you as a reviewer.

### DIFF
--- a/include/exadg/incompressible_navier_stokes/time_integration/driver_steady_problems.cpp
+++ b/include/exadg/incompressible_navier_stokes/time_integration/driver_steady_problems.cpp
@@ -156,23 +156,7 @@ DriverSteadyProblems<dim, Number>::do_solve(double const time, bool unsteady_pro
     }
   }
 
-  // linear problem
-  if(this->param.linear_problem_has_to_be_solved())
-  {
-    // calculate rhs vector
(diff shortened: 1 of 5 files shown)


richardschu (commented):
looking good! if the block after `if(this->param.nonlinear_problem_has_to_be_solved())` is large, you _could_ add a comment. We can remove one assertion now, asserting`this->param.nonlinear_problem_has_to_be_solved()` inside the if clause :laughing: 
Thank you for removing this redundancy.

PR #421, opened 2023-05-03
add override/final keywords to avoid warnings

### LINKED ISSUE #399: suggest-override compiler warnings
I started to get a lot of compiler warnings with `-Wsuggest-override` for functions that actually do override but are not marked `override`. Just curious why I wasn't getting these warnings before. Does anybody else have this issue? I think we would want to sort these warnings out.

nfehn:
closes #399

### DIFF
--- a/applications/compressible_navier_stokes/couette/application.h
+++ b/applications/compressible_navier_stokes/couette/application.h
@@ -50,7 +50,7 @@ public:
   }
 
   double
-  value(dealii::Point<dim> const & p, unsigned int const component = 0) const
+  value(dealii::Point<dim> const & p, unsigned int const component = 0) const final
   {
     double const T =
       T_0 + DYN_VISCOSITY * U_0 * U_0 / (2.0 * LAMBDA) * (p[1] * p[1] / (H * H) - 1.0);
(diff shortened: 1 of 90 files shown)


nfehn:
anyone interested to provide a review?

applications/compressible_navier_stokes/couette/application.h
```diff
@@ -104,7 +104,7 @@ class DensityBC : public dealii::Function<dim>
   }
 
   double
-  value(dealii::Point<dim> const & p, unsigned int const component = 0) const
+  value(dealii::Point<dim> const & p, unsigned int const component = 0) const final
   {
     (void)component;
```
btemuer, line 109: Nothing to do with this PR and also dunno if we want to discuss this at all, but something I want to point out is how we deal with unused function parameters:

- Sometimes we use the `(void) variable` style like here, and we even specify here a default value for an unused variable, which does not make much sense. I know that using default parameters in virtual functions is generally not a good idea, but if the base class from deal.II already has it, there is nothing we can do about it,

- and sometimes we comment out the variable name like `unsigned int /*component*/` for example couple lines down below.

This is just something to think about. Feel free to resolve the comment right away if it is unimportant.

nfehn: It worth talking about this. Nevertheless, I suggest to not mix this with the present PR.

kronbichler (approved):

nfehn:
@richardschu Martin was faster ;). Given that I will briefly look over the changes before merging, I think you do not have to review

PR #423, opened 2023-05-04
redesign SpatialOperator::move_grid_and_update_dependent_data_structures()

nfehn:
This PR is an attempt to make the usage of ALE in time integrators and multigrid preconditioners more intuitive / easier to understand.

This PR originated from PR #404 and is one step towards realizing what is described in comment https://github.com/exadg/exadg/pull/404#discussion_r1177448382.

### DIFF
--- a/include/exadg/convection_diffusion/preconditioners/multigrid_preconditioner.cpp
+++ b/include/exadg/convection_diffusion/preconditioners/multigrid_preconditioner.cpp
@@ -116,9 +116,36 @@ MultigridPreconditioner<dim, Number>::update()
     this->initialize_mapping();
 
     this->update_matrix_free();
+
+    update_operators_after_grid_motion();
   }
 
-  update_operators();
+  set_time(pde_operator->get_time());
+  set_scaling_factor_mass_operator(pde_operator->get_scaling_factor_mass_operator());
+
+  if(data.convective_problem and
+     data.convective_kernel_data.velocity_type == TypeVelocityField::DoFVector)
+  {
+    VectorType const & velocity = pde_operator->get_velocity();
+
+    // convert Number --> MultigridNumber, e.g., double --> float, but only if necessary
+    VectorTypeMG         vector_multigrid_type_copy;
+    VectorTypeMG const * vector_multigrid_type_ptr;
+    if(std::is_same<MultigridNumber, Number>::value)
+    {
+      vector_multigrid_type_ptr = reinterpret_cast<VectorTypeMG const *>(&velocity);
+    }
+    else
+    {
+      vector_multigrid_type_copy = velocity;
+      vector_multigrid_type_ptr  = &vector_multigrid_type_copy;
+    }
+
+    set_velocity(*vector_multigrid_type_ptr);
+  }
+
+  // Once the operators are updated, the update of smoothers and the coarse grid solver is generic
+  // functionality implemented in the base class
 
   update_smoothers();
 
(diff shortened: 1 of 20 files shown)


include/exadg/convection_diffusion/preconditioners/multigrid_preconditioner.cpp
```diff
@@ -116,9 +116,36 @@ MultigridPreconditioner<dim, Number>::update()
     this->initialize_mapping();
 
     this->update_matrix_free();
+
+    update_operators_after_grid_motion();
```
nfehn, line 120: now we only have one `if(mesh_is_moving)` statement in the multigrid preconditioner (and not spread over two functions as done previously).

include/exadg/convection_diffusion/time_integration/time_int_bdf.cpp
```diff
@@ -557,7 +557,9 @@ TimeIntBDF<dim, Number>::postprocessing() const
   if(this->param.ale_formulation and this->get_time_step_number() == 1 and
      not this->param.restarted_simulation)
   {
-    pde_operator->move_grid_and_update_dependent_data_structures(this->get_time());
+    pde_operator->move_grid(this->get_time());
```
nfehn, line 560: In comment https://github.com/exadg/exadg/pull/404#discussion_r1177448382, it has been proposed to let the time integrator directly operate on the grid motion object, which can now be addressed easily without affecting the update of MatrixFree or the update of spatial operators (such as the interior penalty parameter).

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.cpp
```diff
@@ -460,7 +460,9 @@ TimeIntBDF<dim, Number>::postprocessing() const
   if(this->param.ale_formulation and this->get_time_step_number() == 1 and
      not this->param.restarted_simulation)
   {
-    operator_base->move_grid_and_update_dependent_data_structures(this->get_time());
+    operator_base->move_grid(this->get_time());
+    operator_base->update_matrix_free_after_grid_motion();
+    operator_base->update_spatial_operators_after_grid_motion();
```
richardschu, line 465: the sequence of these steps is important ... do we or _should we_ enforce the correct sequence of steps? One could work with `bool`s and assert invalid combinations. On the other side it is highlighted in the comments in `spatial_operator_base`.

richardschu (commented):
Looks great - do we want to enforce the sequence of steps? that is the only thing I thought about! - probably the reason for the design chosen before!

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.cpp
nfehn: Yes, good point. The problem in my opinion is that the sequence of these three steps is neither the responsibility of the `SpatialOperatorBase` (because the `GridMotion` and `MatrixFree` objects are not owned by the spatial operator) nor of the `TimeIntegrator` (because the `GridMotion` and `MatrixFree` objects are not owned by the time integrator). For this reason, I suggest to keep it as is for now and address this in upcoming PRs in combination with other refactorings. One could of course introduce an additional function here or there, but this means that the interface is duplicated with the function `move_grid()` and needs to be adapated at two places when introducing new parameters (such as the update of a preconditioner, which was the origin of the present PR ...).

Overall, we are somewhat in conflict in terms of a separation of concerns between `SpatialOperator`/`Driver`/`TimeIntegrator` regarding data structures such as `Grid`/`GridMotion`/`Mapping`/`MatrixFree`. This is a somewhat everlasting topic in ExaDG where we still need to improve the design.

I think it makes sense that the `Driver` owns most of these data structures, because this allows us great flexibility in terms of coupled or multiphysics problems. One solution could be to pass a lambda with the sequence of three steps from the `Driver` (where all the relevant data structures live) to the `TimeIntegrator`, which can then call this lambda without any knowledge of `GridMotion`/`Mapping`/`MatrixFree` (data structures that should actually not occur in the time integration class).

richardschu: I will dig into the grid mapping now anyways, then I can better understnad the problem. For now, I think we can proceed with the three consecutive calls! Once I have the full picture, we can go for some refactoring :100: .

PR #425, opened 2023-05-05
IncNS::SpatialOperatorBase: use member function get_mapping()

nfehn:
The long term goal behind this change would be to remove `GridMotion` from `SpatialOperatorBase`.


richardschu (commented):
I would like to review, but I am short on time now! Maybe I can look at that change over the weekend! I will keep you posted. :laughing: 
Jokes aside ... go for it

PR #428, opened 2023-05-05
Redesign grid motion calls in time integrator

### LINKED ISSUE #426: Make GridMotion and MatrixFree const in spatial operator
The `Driver` owns `GridMotion` and `MatrixFree`. Hence, the pointers to these objects in `SpatialOperator` should be const

```C++
  std::shared_ptr<GridMotionInterface<dim, Number> const> grid_motion;
  std::shared_ptr<dealii::MatrixFree<dim, Number> const> matrix_free;
```

The affected modules are those that support ALE, i.e. `ConvDiff` and `IncNS`.

nfehn:
closes #426

The PR in its current incomplete state shows the principal idea with lambda's used in the time integrator. The alternative is to pass `GridMotion` and `MatrixFree` to the `TimeIntegrator` class.

Feel free to discuss this approach. The main motivation is to avoid calling non-const functions of objects of type `GridMotion` and `MatrixFree` in classes where these objects should actually be `const`, such as `TimeIntegrator` and `SpatialOperator`, because `GridMotion` and `MatrixFree` are owned by `Driver`. The lambda functions (passed to the time integrator) would be defined in the `Driver` class. This still needs to be implemented.

@jh66637 @richardschu I ping you explicitly because of your interest in ALE methods.


include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.h
```diff
@@ -40,6 +40,16 @@ class SpatialOperatorBase;
 template<typename Number>
 class PostProcessorInterface;
 
+template<typename Number>
+class HelpersALE
+{
+public:
+  std::function<void(double const & time)> const move_grid;
```
jh66637, line 47: Can you point me to the point where you initialize the ```std::functions```?
I'm asking because I think ```fill_grid_ccordinates(dst, dof_handler)``` and maybe also the other functions could be something which can be defined in a general way, such that we could have a collection of free functions under the namespace ```ALEUtilities``` instead.

nfehn: As written in the introductory statement https://github.com/exadg/exadg/pull/428#issue-1697494278, this has not been implemented so far. The background/reason for the struct/class design is that I only want to pass one object to the time integration class, not a separate object for every lambda function (because we will probably need more than one lambda).

jh66637: Ah, I missed that. 

I agree if we have to use lambdas because the functions are not general enough for free functions. But I currently don't see any reason why free functions are not sufficient. I think we should discuss it when the initialization of the lambdas is implemented :)

nfehn: > But I currently don't see any reason why free functions are not sufficient.

const-correctness !?

The operations to be done on certain objects are not of type `const`. The objects are living in the `Driver`. So if we define the lambda's inside the driver, it is easier to see which parts are able to manipulate objects in a non-const way. In some sense, the lambda functions are "member functions" of `Driver`, manipulating data structures owned by the `Driver`. There is of course some "magic" in the sense that these functions are finally called by the `TimeIntegrator`.

Realizing these functions as true member functions of the `Driver` has the problem that it is not clear who owns the dof vectors to operate on (`Driver` vs. `TimeIntegrator`), as we discussed some time ago ;)

At least this was my idea/motivation behind. Is this understandable?

nfehn: @jh66637 I have now realized the implementation of the lambda functions in the `Driver`. Note that this functionality has (more or less) already been implemented in `Driver::ale_update()` before, and somewhat redundantly in `SpatialOperator` in the previous code version. Zwei Fliegen mit einer Klappe ...

nfehn:
@jh66637 @richardschu `IncNS::SpatialOperatorBase` now no longer depends on `GridMotionBase` (i.e. there is no longer the need to call `get_dynamic_mapping()` in the spatial operator, returning the mapping of the reference configuration or the deformed mapping). In the `Driver`, the spatial operator is just initialized with a standard `dealii::Mapping`. All ALE-specific data structures now live and are used in the `Driver` class. These changes hopefully make ALE-related code easy to understand in **ExaDG**.

peterrum (approved):
Makes sense :+1:  Let's merge!

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.h
jh66637: After looking at the new commits, I get it :)

PR #433, opened 2023-05-11
add plotting in interface_coupling before assertion

richardschu:
In case the setup phase of `RemotePointEvaluation` in `interface_coupling.cpp` fails, this add an output of the grid and points that were used. This way, one can visually check if the points match or not. Output is only triggered if the assertion is triggered.

### DIFF
--- a/include/exadg/functions_and_boundary_conditions/interface_coupling.cpp
+++ b/include/exadg/functions_and_boundary_conditions/interface_coupling.cpp
@@ -61,11 +61,9 @@ InterfaceCoupling<rank, dim, Number>::setup(
                               return marked_vertices_src_;
                             }));
 
-    auto const * points = & interface_data_dst->get_array_q_points(quad_index);
+    auto const * points = &interface_data_dst->get_array_q_points(quad_index);
 
-    map_evaluator[quad_index].reinit(*points,
-                                     dof_handler_src_.get_triangulation(),
-                                     mapping_src_);
+    map_evaluator[quad_index].reinit(*points, dof_handler_src_.get_triangulation(), mapping_src_);
 
     if(not map_evaluator[quad_index].all_points_found())
     {
@@ -75,25 +73,32 @@ InterfaceCoupling<rank, dim, Number>::setup(
       unsigned int n_points_not_found = 0;
       for(unsigned int i = 0; i < points->size(); ++i)
       {
-         if(not map_evaluator[quad_index].point_found(i))
-         {
-        	 n_points_not_found += 1;
-        	 points_not_found.push_back((*points)[i]);
-         }
+        if(not map_evaluator[quad_index].point_found(i))
+        {
+          n_points_not_found += 1;
+          points_not_found.push_back((*points)[i]);
+        }
       }
-      n_points_not_found = dealii::Utilities::MPI::sum(n_points_not_found, dof_handler_src->get_communicator());
+      n_points_not_found =
+        dealii::Utilities::MPI::sum(n_points_not_found, dof_handler_src->get_communicator());
 
       std::string const file_name =
         "interface_coupling_quad_index_" + dealii::Utilities::to_string(quad_index);
 
-      write_grid(dof_handler_src->get_triangulation(), mapping_src_, 4, "./", file_name, 0, dof_handler_src->get_communicator());
+      write_grid(dof_handler_src->get_triangulation(),
+                 mapping_src_,
+                 4,
+                 "./",
+                 file_name,
+                 0,
+                 dof_handler_src->get_communicator());
 
       write_points(points_not_found, "./", file_name, 0, dof_handler_src->get_communicator());
 
-      AssertThrow(
-        map_evaluator[quad_index].all_points_found(),
-        dealii::ExcMessage(std::string("Setup of InterfaceCoupling was not successful: "
-        		+ std::to_string(n_points_not_found) + " points have not been found.")));
+      AssertThrow(map_evaluator[quad_index].all_points_found(),
+                  dealii::ExcMessage(std::string("Setup of InterfaceCoupling was not successful: " +
+                                                 std::to_string(n_points_not_found) +
+                                                 " points have not been found.")));
     }
   }
 }
--- a/include/exadg/functions_and_boundary_conditions/interface_coupling.cpp
+++ b/include/exadg/functions_and_boundary_conditions/interface_coupling.cpp
@@ -19,14 +19,9 @@
  *  ______________________________________________________________________
  */
 
-// deal.II
-#include <deal.II/grid/grid_generator.h>
-#include <deal.II/numerics/data_out.h>
-#include <deal.II/particles/data_out.h>
-#include <deal.II/particles/particle_handler.h>
-
 // ExaDG
 #include <exadg/functions_and_boundary_conditions/interface_coupling.h>
+#include <exadg/postprocessor/write_output.h>
 
 namespace ExaDG
 {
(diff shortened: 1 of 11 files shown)


include/exadg/functions_and_boundary_conditions/interface_coupling.cpp
```diff
@@ -64,11 +70,67 @@ InterfaceCoupling<rank, dim, Number>::setup(
...
+    for(unsigned int d = 0; d < dim; ++d)
+    {
+      min_coord = p[d] < min_coord ? p[d] : min_coord;
+      max_coord = p[d] > max_coord ? p[d] : max_coord;
+    }
   }
+  min_coord = dealii::Utilities::MPI::min(min_coord, dof_handler.get_communicator());
+  max_coord = dealii::Utilities::MPI::max(max_coord, dof_handler.get_communicator());
+  dealii::GridGenerator::hyper_cube(particle_dummy_tria, min_coord, max_coord);
+  dealii::MappingQGeneric<dim>                 particle_dummy_mapping(1 /* mapping_degree */);
```
richardschu, line 122: dummy mapping with order 1, is identity anyways.

include/exadg/functions_and_boundary_conditions/interface_coupling.cpp
```diff
@@ -64,11 +70,67 @@ InterfaceCoupling<rank, dim, Number>::setup(
...
+  // point output via particle handler
+  dealii::Triangulation<dim> particle_dummy_tria;
+  double                     min_coord = points[0][0];
+  double                     max_coord = points[0][0];
+  for(auto const & p : points)
+  {
+    for(unsigned int d = 0; d < dim; ++d)
+    {
+      min_coord = p[d] < min_coord ? p[d] : min_coord;
+      max_coord = p[d] > max_coord ? p[d] : max_coord;
```
nfehn, line 116: I think here we should make use of `std::min/max` functionality.

include/exadg/functions_and_boundary_conditions/interface_coupling.cpp
```diff
@@ -64,11 +70,67 @@ InterfaceCoupling<rank, dim, Number>::setup(
...
+    if(not map_evaluator[quad_index].all_points_found())
+    {
+      std::string const file_name =
+        "interface_coupling_quad_index_" + dealii::Utilities::to_string(quad_index);
+      plot_points_and_triangulation(interface_data_dst->get_array_q_points(quad_index),
+                                    dof_handler_src_,
+                                    mapping_src_,
+                                    file_name);
+      AssertThrow(
+        false,
```
nfehn, line 82: I think I would repeat here the condition that fails, because this contains more information than just `false`. In terms of performance, this does not hurt since the simulation will terminate soon.

nfehn:
Thanks for sharing this functionality!

As a general comment, I think we should discuss the design in order to maximize the impact of this code for ExaDG, i.e. we should aim at utility functions rather than a private member function in `InterfaceCoupling`:

- It looks as if this function could be realized as a free function as well, making the usage of this function more flexibie
- It looks as if the two parts of writing the triangulation and the points could be separated, which makes usage of this functionality again more flexible
- When splitting the function into two parts, the first part with the triangulation should already be available in ExaDG? If not, the currently existing functionality should be extended.
- This design has also the advantage that the filename endings `grid` and `points` can be added where the function is called. One could then even add the extensions `_dst` and `_src` according to the nomenclature in `InterfaceCoupling` to make the meaning of grid and points more clear (i.e., to which side (dst/src) each item belongs to).

nfehn:
To make the third point more explicit, I was thinking of this file https://github.com/exadg/exadg/blob/master/include/exadg/postprocessor/write_output.h regarding the output of a triangulation. If the function `write_grid()` is not sufficient, one could add another function of this style.

include/exadg/functions_and_boundary_conditions/interface_coupling.cpp
```diff
@@ -64,11 +70,67 @@ InterfaceCoupling<rank, dim, Number>::setup(
...
 
-    AssertThrow(
-      map_evaluator[quad_index].all_points_found() == true,
-      dealii::ExcMessage(
-        "Setup of InterfaceCoupling was not successful. Not all points have been found."));
+    if(not map_evaluator[quad_index].all_points_found())
+    {
+      std::string const file_name =
+        "interface_coupling_quad_index_" + dealii::Utilities::to_string(quad_index);
+      plot_points_and_triangulation(interface_data_dst->get_array_q_points(quad_index),
```
peterrum, line 77: Does one actually want to output all points and not only the few not found ones?

peterrum: https://www.dealii.org/developer/doxygen/deal.II/classUtilities_1_1MPI_1_1RemotePointEvaluation.html#af500b74c1b221de1168dabaabb885a6a

richardschu:
Nice ideas everyone! :+1: 
I think it goes towards having general functionality for higher-order grid (no solution) output - potentially adapting the `write_grid` function or adding an additional one and a general point output function. In the InterfaceCoupling I can then do a private member function which passes only points not found to the general particle output function! Bravo!

richardschu:
Please ignore the changes in the `pressure_wave` application, it is just so you see how I tested (and for me its easier to undo those changes in the end).
I moved the point output to `exadg/postprocessor/write_output`, where I adapted the `write_grid` function, which previously generated lower-oder grids only (no subdivision and no mappong!). Thus I changed the arguments to match those of `write_surface_mesh` and changed the calls in iNS and cNS `output_generator.cpp` accordingly.
These functions are now just called in `interface_coupling.h` whenever one cannot match the points on the interface!
The general point plotting is quite nice to have for debugging.

include/exadg/postprocessor/write_output.h
```diff
@@ -78,13 +81,60 @@ write_boundary_IDs(dealii::Triangulation<dim> const & triangulation,
...
-  dealii::GridOut grid_out;
-  grid_out.write_mesh_per_processor_as_vtu(triangulation, filename);
+  dealii::DataOut<dim> data_out;
+
+  dealii::DataOutBase::VtkFlags flags;
+  flags.write_higher_order_cells = n_subdivisions > 1;
+  data_out.set_flags(flags);
+
+  data_out.attach_triangulation(triangulation);
+  data_out.build_patches(mapping, 4, dealii::DataOut<dim>::curved_inner_cells);
```
nfehn, line 100: This number of "4" seems to be hard-coded?

include/exadg/postprocessor/write_output.h
```diff
@@ -78,13 +81,60 @@ write_boundary_IDs(dealii::Triangulation<dim> const & triangulation,
...
+           dealii::Mapping<dim> const &       mapping,
+           unsigned int const                 n_subdivisions,
            std::string const &                folder,
-           std::string const &                file)
+           std::string const &                file,
+           unsigned int const &               counter,
+           MPI_Comm const &                   mpi_comm)
 {
-  std::string filename = folder + file + "_grid";
+  std::string filename = file + "_grid";
```
nfehn, line 91: I don't see the reason to append the filename here in these functions `_grid` (and `_points` below). Should we maybe change this in the future? It is somewhat independent of the present PR.

include/exadg/postprocessor/write_output.h
```diff
@@ -78,13 +81,60 @@ write_boundary_IDs(dealii::Triangulation<dim> const & triangulation,
...
+  for(auto const & p : points)
+  {
+    for(unsigned int d = 0; d < dim; ++d)
+    {
+      min_coord = std::min(p[d], min_coord);
+      max_coord = std::max(p[d], max_coord);
+    }
+  }
+  min_coord = dealii::Utilities::MPI::min(min_coord, mpi_comm);
+  max_coord = dealii::Utilities::MPI::max(max_coord, mpi_comm);
```
nfehn, line 125: This looks like some generic functionality that could be realized independently. I wonder whether this is already available in deal.II as some (particle) utility function? @peterrum @kronbichler Do you know something? If not, I suggest to go with the present code version.

include/exadg/functions_and_boundary_conditions/interface_coupling.cpp
```diff
@@ -60,14 +61,45 @@ InterfaceCoupling<rank, dim, Number>::setup(
...
+      for(unsigned int i = 0; i < points->size(); ++i)
+      {
+        if(not map_evaluator[quad_index].point_found(i))
+        {
+          n_points_not_found += 1;
+          points_not_found.push_back((*points)[i]);
+        }
+      }
+      n_points_not_found =
+        dealii::Utilities::MPI::sum(n_points_not_found, dof_handler_src->get_communicator());
```
nfehn, line 83: `points_not_found()` and `n_points_not_found()` could be two useful additional member functions of `dealii::RPE` for the future.

nfehn:
Looks good (apart from the changes in the application). Some of the comments are just ideas for improvements for the future.

include/exadg/postprocessor/write_output.h
nfehn: If one wants to provide min/max as utility functions, I think one should then provide independent functions for min and for max. One could also think of some norms or min/max values of `dealii::Point`. Then, one only needs to loop over the vector (plus MPI afterwards).

kronbichler, line 125: Let me add two comments to the code:
- I would compute the `min/max` for each coordinate, not all coordinates, and then `GridGenerator::hyper_rectangle`. The question is then if this should be somehow combined with `dealii::BoundingBox` - today, there is a function https://github.com/dealii/dealii/blob/d47537fa246886e5308d5aecbf42358d73a00774/include/deal.II/grid/grid_tools.h#L380-L382 but not the other way around. (Would that be useful?)
- If performance is important, one should compute `Utilities::MPI::max({-min_coord, max_coord}, mpi_comM);` in a single go (not sure of this syntax works, but you get what I mean). The reason is that global communication is expensive, so doing it once on 2 elements (or `2 * dim` elements when computing a hyper rectangle) lowers compute time by a factor of 2 in this particular function. I am not saying we should do that here, but the MPI person in my considers this the better programming style with MPI, so I should mention it.

peterrum:
My preference would be to introduce a new version of `Particles::DataOut::build_patches` that does not take a `ParticleHandler` but a list of points and maybe a list of properties. I have written lines similar to the ones in this PR already so often. One comment: by introducing a dummy triangulation with a single cell, all particles will be sent  to the same rank.

include/exadg/postprocessor/write_output.h
richardschu: I see the point, but this makes sure the files are not overwritten calling the output with identical file names ... otherwise we shift that to the calling functions ... I would be ok with both!

include/exadg/postprocessor/write_output.h
richardschu: thanks for the hint! in the future if I can't avoid `Utilities::MPI::max` or the like, I will consider such a strategy!
the `BoundingBox` and `Hyperrectangle` removed some code here as well!

richardschu:
> My preference would be to introduce a new version of `Particles::DataOut::build_patches` that does not take a `ParticleHandler` but a list of points and maybe a list of properties. I have written lines similar to the ones in this PR already so often. One comment: by introducing a dummy triangulation with a single cell, all particles will be sent to the same rank.

Regarding that issue with the single cell triangulation: I introduced one function that uses an input tria and mapping (might not work if any point is not in the tria) and another one generating a dummy tria!

include/exadg/functions_and_boundary_conditions/interface_coupling.cpp
nfehn: As a reminder, I opened an issue in dealii: https://github.com/dealii/dealii/issues/15248

PR #434, opened 2023-05-11
fix failing Poisseuille test case

### LINKED ISSUE #431: Test on Poiseuille with parabolic inflow is failing
Could this be due to recent changes in dealii/master?

Any ideas on this?

nfehn:
closes #431 

I am not sure if changing the test output is a robust solution for the future, @kronbichler ?

I suspect the origin of the failing test in the calculation of the time step size (CFL condition) due to a slightly different velocity value.

### DIFF
--- a/applications/incompressible_navier_stokes/poiseuille/application.h
+++ b/applications/incompressible_navier_stokes/poiseuille/application.h
@@ -207,17 +207,17 @@ private:
 
 
     // TEMPORAL DISCRETIZATION
-    this->param.solver_type                   = SolverType::Unsteady;
-    this->param.temporal_discretization       = TemporalDiscretization::BDFDualSplittingScheme;
-    this->param.treatment_of_convective_term  = TreatmentOfConvectiveTerm::Explicit;
-    this->param.calculation_of_time_step_size = TimeStepCalculation::CFL;
-    this->param.adaptive_time_stepping        = true;
-    this->param.max_velocity                  = max_velocity;
-    this->param.cfl                           = 2.0e-1;
+    this->param.solver_type                     = SolverType::Unsteady;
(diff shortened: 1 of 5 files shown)


nfehn:
the test failed again, so it does not seem to be robust anymore.

kronbichler:
> the test failed again, so it does not seem to be robust anymore.

Looking at the test, I think we were simply lucky before, and it never was stable. It's also reflected in the expected output where we take 101 steps to go from 0.1 to 0.2, but 100 time steps for all other 0.1 increments.

Note: We might also think of a different strategy in the truncation, so that factors like 0.1, 0.01 etc get stably "truncated", e.g. by saying
```
  new_time_step = dealii::Utilities::truncate_to_n_digits(new_time_step * 0.99995, 4);
```
or any related action. We always will have cases that are near the threshold of rounding, but having it exactly at 1e-4 means that we have a high probability of hitting it with regular numbers.

nfehn:
> Looking at the test, I think we were simply lucky before, and it never was stable. It's also reflected in the expected output where we take 101 steps to go from 0.1 to 0.2, but 100 time steps for all other 0.1 increments.

I think this is a different topic. We should probably introduce tolerances here
https://github.com/exadg/exadg/blob/284831c03bc24f7c85ebc5fb76d5a85d69ddd37e/include/exadg/time_integration/solver_info_data.h#L72-L74

as we have already in other parts of the code, see e.g. `TimeControl`, https://github.com/exadg/exadg/blob/284831c03bc24f7c85ebc5fb76d5a85d69ddd37e/include/exadg/postprocessor/time_control.cpp#L88-L152

nfehn:
I slightly changed a parameter as suggested by @kronbichler. Let's see if this makes the test more robust and whether the test is now passing on github.

PR #440, opened 2023-05-15
Pass in references to functions instead of shared_ptr

### LINKED ISSUE #439: `shared_ptr` to function is passed into the FunctionEvaluator
- incurs overhead for reference counting,

- the `FunctionEvaluator` does not manipulate the pointer (in terms of ownership), so the object should actually be passed in, not the pointer.

btemuer:
Refactoring/clean code.

Closes #439, see issue description.

### DIFF
--- a/include/exadg/functions_and_boundary_conditions/evaluate_functions.h
+++ b/include/exadg/functions_and_boundary_conditions/evaluate_functions.h
@@ -40,7 +40,7 @@ struct FunctionEvaluator
 {
   static inline DEAL_II_ALWAYS_INLINE //
     dealii::Tensor<rank, dim, dealii::VectorizedArray<Number>>
-    value(std::shared_ptr<dealii::Function<dim>> const                function,
+    value(dealii::Function<dim> &                                     function,
           dealii::Point<dim, dealii::VectorizedArray<Number>> const & q_points,
           double const &                                              time)
   {
@@ -55,10 +55,10 @@ struct FunctionEvaluator
 
   static inline DEAL_II_ALWAYS_INLINE //
     dealii::Tensor<rank, dim, dealii::VectorizedArray<Number>>
-    value(std::shared_ptr<FunctionCached<rank, dim>> const function,
-          unsigned int const                               face,
-          unsigned int const                               q,
-          unsigned int const                               quad_index)
+    value(FunctionCached<rank, dim> const & function,
+          unsigned int const                face,
+          unsigned int const                q,
+          unsigned int const                quad_index)
   {
     (void)function;
     (void)face;
@@ -72,12 +72,12 @@ struct FunctionEvaluator
 
   static inline DEAL_II_ALWAYS_INLINE //
     dealii::Tensor<rank, dim, dealii::VectorizedArray<Number>>
-    value(std::shared_ptr<dealii::Function<dim>> const                    function,
+    value(FunctionWithNormal<dim> &                                       function_with_normal,
           dealii::Point<dim, dealii::VectorizedArray<Number>> const &     q_points,
           dealii::Tensor<1, dim, dealii::VectorizedArray<Number>> const & normals,
           double const &                                                  time)
   {
-    (void)function;
+    (void)function_with_normal;
     (void)q_points;
     (void)normals;
     (void)time;
@@ -89,7 +89,7 @@ struct FunctionEvaluator
 
   static inline DEAL_II_ALWAYS_INLINE //
     dealii::SymmetricTensor<rank, dim, dealii::VectorizedArray<Number>>
-    value_symmetric(std::shared_ptr<dealii::Function<dim>> const                function,
+    value_symmetric(dealii::Function<dim> &                                     function,
                     dealii::Point<dim, dealii::VectorizedArray<Number>> const & q_points,
                     double const &                                              time)
   {
@@ -104,10 +104,10 @@ struct FunctionEvaluator
 
   static inline DEAL_II_ALWAYS_INLINE //
     dealii::SymmetricTensor<rank, dim, dealii::VectorizedArray<Number>>
-    value_symmetric(std::shared_ptr<FunctionCached<rank, dim>> const function,
-                    unsigned int const                               face,
-                    unsigned int const                               q,
-                    unsigned int const                               quad_index)
+    value_symmetric(FunctionCached<rank, dim> const & function,
+                    unsigned int const                face,
+                    unsigned int const                q,
+                    unsigned int const                quad_index)
   {
     (void)function;
     (void)face;
@@ -125,7 +125,7 @@ struct FunctionEvaluator<0, dim, Number>
 {
   static inline DEAL_II_ALWAYS_INLINE //
     dealii::Tensor<0, dim, dealii::VectorizedArray<Number>>
-    value(std::shared_ptr<dealii::Function<dim>> const                function,
+    value(dealii::Function<dim> &                                     function,
           dealii::Point<dim, dealii::VectorizedArray<Number>> const & q_points,
           double const &                                              time)
   {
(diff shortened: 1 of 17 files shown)


include/exadg/incompressible_navier_stokes/spatial_discretization/operators/weak_boundary_conditions.h
```diff
@@ -460,12 +460,13 @@ inline DEAL_II_ALWAYS_INLINE //
...
       {
-        h = FunctionEvaluator<1, dim, Number>::value(bc, q_points, time);
+        h = FunctionEvaluator<1, dim, Number>::value(*bc, q_points, time);
       }
       else
       {
         auto normals_m = integrator.get_normal_vector(q);
-        h              = FunctionEvaluator<1, dim, Number>::value(bc, q_points, normals_m, time);
+        h              = FunctionEvaluator<1, dim, Number>::value(
+          *(std::dynamic_pointer_cast<FunctionWithNormal<dim>>(bc)), q_points, normals_m, time);
```
nfehn, line 469: This will be required more often in the code if `FunctionWithNormal` is used more often.

nfehn:
Overall, I can live with these changes, mainly because `FunctionEvaluator` is one of the lowest level implementations in ExaDG and because `FunctionWithNormal` is not used often in the code. I want to discuss in more detail the pros/cons:

The concrete downsides of the changes of this PR are (i) many `*` for dereferencing, (ii) dynamic pointer casts on the calling side (so potentially at many places in the code), see https://github.com/exadg/exadg/pull/440/files#r1194717112.

With reference to the comment here https://github.com/exadg/exadg/issues/439#issuecomment-1549109429, I think it is difficult to realize `Functions` consistently with value semantics in ExaDG. The question is, therefore, whether we want to introduce a mixed value/pointer semantics style. This argument should in my opinion count as much as the overhead/performance argument.

include/exadg/functions_and_boundary_conditions/evaluate_functions.h
```diff
@@ -248,7 +246,7 @@ struct FunctionEvaluator<2, dim, Number>
 {
   static inline DEAL_II_ALWAYS_INLINE //
     dealii::Tensor<2, dim, dealii::VectorizedArray<Number>>
-    value(std::shared_ptr<dealii::Function<dim>> const                function,
+    value(dealii::Function<dim> &                                     function,
```
peterrum, line 249: One could also write:
```suggestion
    value(std::shared_ptr<dealii::Function<dim>> const &                                   function,
```

include/exadg/incompressible_navier_stokes/spatial_discretization/operators/weak_boundary_conditions.h
btemuer: Yes, but that should be fine as the function call is specialized for `FunctionWithNormal` with the `normals` parameter. So no actual polymorphism is present here and it is better to be explicit about this. It should be the caller's responisiblity to call a function with the correct parameters (in which case it cannot fail), instead of the function throwing an exception if a condition is not met inside it, in this case if the dynamic cast failed.

Looking at it now, I think we should even make this a `static_pointer_cast`, as the actual type is known to be a `FunctionWithNormal` at compile time so that this function can be called at all.

include/exadg/functions_and_boundary_conditions/evaluate_functions.h
btemuer: Yes but why would we pass in a pointer if we're not doing anything with the pointer but just need the object it points at, right?

btemuer:
Thanks @nfehn  for your comment. I don't agree with your first point that many `*` for pointer dereferencing are bad. I see the opposite, it is better to have interfaces that work with the object instead of the pointers, as the pointers disclose much more information than the object. Also, this cannot be a cause of mistakes in the future, as forgetting to dereference etc. is impossible. So I see no negatives on many `*`.

And it still wouldn't count as value semantics but still as pointer/reference semantics, as we are not passing by value but by reference and we are modifying the object passed-in. In this way, we are just keeping information away from the callee that it doesn't need.

For your second point, I think `static_pointer_cast` is safer and better (see my comment from above). You are right that many `dynamic_cast` would not have been good.

nfehn:
For clarity, I use the term "value semantics" for call-by-value and call-by-reference, and "pointer semantics" if the variable is a pointer/address (with access to member variables via `->`).

Generally speaking, I see more flexibility with the shared pointers. I think over the last years we had to introduce shared pointers at several places when we needed more flexibility due to new functionality. In such a case, the many `*` are disadvantageous, because many lines of code changed now have to be touched again. Whether usage of `*` is nice code is more subjective.

Overall, I am not against the present changes, also because right now I cannot argue concretely how the `Function` functionality in ExaDG might be extended in the future. But I also cannot exclude this, and then we would potentially have to revert these changes again.

nfehn (approved):

btemuer:
Great :+1:, thanks for the nice discussion :raised_hands: . I am sure this is a good step towards best practices in the use of C++ pointers. It's actually one of the main principles of smart pointers which I got recently reminded of. I don't think we would ever run into a case where our function evaluator will need the pointer to the function. If that becomes the case it would mean that we have a design problem that we can address at that point.

I will very soon replace the `dynamic_pointer_cast` with `static_pointer_cast` and push the changes.

include/exadg/incompressible_navier_stokes/spatial_discretization/operators/weak_boundary_conditions.h
nfehn: I don't understand why one should know at compile time that we have a `FunctionWithNormal`. Looking at the `BoundaryDescriptor`, we only know that we have a pointer to `dealii::Function`, don't we? So I would argue we have a downcast here. 

Hence, if the user makes a mistake in the application.h file, the current code with `static_pointer_cast` could lead to undefined behavior, while a dynamic cast should give a seg-fault, right?

So I think we have to keep the dynamic_pointer_cast here.

include/exadg/incompressible_navier_stokes/spatial_discretization/operators/weak_boundary_conditions.h
```diff
@@ -460,12 +460,13 @@ inline DEAL_II_ALWAYS_INLINE //
...
       {
-        h = FunctionEvaluator<1, dim, Number>::value(bc, q_points, time);
+        h = FunctionEvaluator<1, dim, Number>::value(*bc, q_points, time);
       }
       else
       {
         auto normals_m = integrator.get_normal_vector(q);
-        h              = FunctionEvaluator<1, dim, Number>::value(bc, q_points, normals_m, time);
+        h              = FunctionEvaluator<1, dim, Number>::value(
+          *(std::static_pointer_cast<FunctionWithNormal<dim>>(bc)), q_points, normals_m, time);
```
nfehn, line 469: dynamic_pointer_cast

nfehn (changes requested):
the static/dynamic_pointer_cast problem needs to be addressed

include/exadg/incompressible_navier_stokes/spatial_discretization/operators/weak_boundary_conditions.h
btemuer: Yeah, I meant it because of the `if-else` condition that we have here. If that boolean variable can be set independently from the function's type, such a mistake can of course occur. That isn't safe in and of itself, but you're right that a `dynamic_pointer_cast` will catch that error eventually. So sure, I'll make that a dynamic cast.

include/exadg/incompressible_navier_stokes/spatial_discretization/operators/weak_boundary_conditions.h
btemuer: Done

nfehn (approved):

PR #442, opened 2023-05-19
Remove FunctionCached

nfehn:
The previous implementation was unnecessarily complex. 150 removed lines of code say it all.

This is a follow-up to PR #441 (see comment https://github.com/exadg/exadg/pull/441#issuecomment-1551465494).


include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
```diff
@@ -195,23 +195,19 @@ SpatialOperatorBase<dim, Number>::setup(
...
-    std::set<dealii::types::boundary_id> bids;
-    for(auto const & bc : boundary_descriptor->velocity->dirichlet_cached_bc)
-      bids.insert(bc.first);
-
-    interface_data_dirichlet_cached->setup(matrix_free,
-                                           get_dof_index_velocity(),
-                                           quad_indices,
-                                           bids);
+    interface_data_dirichlet_cached = std::make_shared<ContainerInterfaceData<1, dim, double>>();
+    interface_data_dirichlet_cached->template setup<Number>(
```
richardschu, line 204: interesting syntax here - looked it up and learned something, thanks ; )

richardschu (approved):
Looking good overall, thank you for decluttering!

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
nfehn: see e6393c33de4851c38486cf319fbb604a10e66d39

nfehn:
@peterrum @kronbichler please let me know if you want to discuss this PR before merging.

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_dual_splitting.cpp
```diff
@@ -839,8 +839,7 @@ OperatorDualSplitting<dim, Number>::local_interpolate_velocity_dirichlet_bc_boun
         }
         else if(boundary_type == BoundaryTypeU::DirichletCached)
         {
-          auto bc =
-            this->boundary_descriptor->velocity->dirichlet_cached_bc.find(boundary_id)->second;
+          auto bc = this->boundary_descriptor->velocity->get_dirichlet_cached_data();
```
peterrum, line 842: bid is not needed anymore?

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
```diff
@@ -195,23 +195,18 @@ SpatialOperatorBase<dim, Number>::setup(
...
-    for(auto const & bc : boundary_descriptor->velocity->dirichlet_cached_bc)
-      bids.insert(bc.first);
-
-    interface_data_dirichlet_cached->setup(matrix_free,
+    interface_data_dirichlet_cached = std::make_shared<ContainerInterfaceData<1, dim, double>>();
+    interface_data_dirichlet_cached->setup(*matrix_free,
                                            get_dof_index_velocity(),
                                            quad_indices,
-                                           bids);
+                                           boundary_descriptor->velocity->dirichlet_cached_bc);
```
peterrum, line 207: I guess you allocate the data structures for the complete boundary?

include/exadg/structure/user_interface/boundary_descriptor.h
```diff
@@ -127,6 +127,42 @@ struct BoundaryDescriptor
     AssertThrow(counter == 1,
                 dealii::ExcMessage("Boundary face with non-unique boundary type found."));
   }
+
+  void
+  set_dirichlet_cached_data(
```
peterrum, line 132: What about adding some Doxygen comments here?

peterrum (approved):

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_dual_splitting.cpp
nfehn: There is only one data structure for all boundary IDs. This data is then accessed by face, quadrature point, and vectorized array indices (for all boundary IDs belonging to the cached-type BC). Before the present PR, all `FunctionCached` objects (for the different boundary IDs) were pointing to the same data structure. (The background is that we collect all boundaries and setup dealii::RPE once).

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
nfehn: Yes, the "boundary condition" is no longer a map with bid+function, but only a set of boundary ids (= dirichlet_cached_bc), since the underlying "Function" to evaluate is a purely internal data structure.

include/exadg/structure/user_interface/boundary_descriptor.h
nfehn: So far, I hesitated because we have the same set-functions in multiple boundary descriptors (which are different for the different modules in ExaDG).

I am also not sure if this is the final design. Actually, these functions should not be public, but only accessible for the respective `MyModule::SpatialOperator` class, which sets up the data structures for cached-type boundary conditions. Maybe this internal use is why I considered documentation not that important for this function. Another solution could be to not pass this data structure back to the `BoundaryDescriptor` (via a shared pointer), but only to the implementation of operators and kernels, with the need extend the interfaces of such classes (!).

Lange Rede, kurzer Sinn. I could principally add some explanation, but this text will be copied for multiple boundary descriptors :(

peterrum (approved):
Good to go!

include/exadg/functions_and_boundary_conditions/function_cached.h
```diff
@@ -1,66 +0,0 @@
...
-
-  // read data
-  inline data_type
-  tensor_value(unsigned int const face,
-               unsigned int const q,
-               unsigned int const v,
-               unsigned int const quad_index) const
-  {
-    return interface_data->get_data(quad_index, face, q, v);
-  }
```
nfehn, line 53: This is where the bug reported in #482 comes from. `get_data()` and `tensor_value()` use a different order of arguments. Since all parameters are of the same type, no compilation error showed up ...

include/exadg/functions_and_boundary_conditions/evaluate_functions.h
```diff
@@ -189,18 +189,18 @@ struct FunctionEvaluator<1, dim, Number>
...
+          unsigned int const                             quad_index)
   {
     dealii::Tensor<1, dim, dealii::VectorizedArray<Number>> value;
 
     std::array<dealii::Tensor<1, dim, Number>, dealii::VectorizedArray<Number>::size()>
       tensor_array;
 
     for(unsigned int v = 0; v < dealii::VectorizedArray<Number>::size(); ++v)
-      tensor_array[v] = function.tensor_value(face, q, v, quad_index);
+      tensor_array[v] = function.get_data(face, q, v, quad_index);
```
nfehn, line 203: Here, the order of arguments has to be changed when changing the function name. The same for the two other calls below.

PR #443, opened 2023-05-22
refactor/generalize coarse grid mappings in MultigridPreconditionerBase

nfehn:
To generalize the implementation of mappings in the multigrid preconditioner, the following changes are made here:

- the member variable `coarse_grid_mappings` is now a vector of shared ptrs to `const dealii::Mapping<dim>` (instead of `ExaDG::MappingDoFVector<dim, Number>`)
- we always create a vector of coarse triangulations if h-MG is involved (and not just if the dynamic cast to `dealii::MappingQCache` is successful, encoding somewhat implicitly that simplices will not work with this code path in the present code version ...). If a plain `dealii::Mapping<dim>` is used on the fine level, the coarse grid mappings simply point to this mapping
- refactor `MultigridPreconditionerBase` so that the member function `get_mapping()` is generic (i.e. works for any types of mapping, multigrid coarsening strategy, hypercube vs simplex)

The overall goal is to shift mapping functionality outside of the multigrid preconditioner and closer to the application, in order to have more flexibility regarding how to construct mappings and/or the element type (hypercube vs. simplex). It is probably better to split this overall effort into more than one PR (I am mentioning this so that the motivation for the changes of the present PR is more clear).


include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.h
```diff
@@ -313,13 +319,9 @@ class MultigridPreconditionerBase : public PreconditionerBase<Number>
...
   std::shared_ptr<dealii::Mapping<dim> const> mapping;
 
-  // Only relevant for global coarsening,where this vector contains coarse level mappings,
+  // Only relevant for global coarsening, where this vector contains coarse level mappings,
   // and the fine level mapping as the last element of the vector.
-  std::vector<std::shared_ptr<MappingDoFVector<dim, Number>>> coarse_grid_mappings;
-
-  // Only relevant for local-smoothing-multigrid and in case that a mapping of type MappingQCache is
-  // used.
-  std::shared_ptr<MappingDoFVector<dim, Number>> mapping_dof_vector;
```
peterrum, line 322: Why has this been removed?

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.h
```diff
@@ -313,13 +319,9 @@ class MultigridPreconditionerBase : public PreconditionerBase<Number>
   // In case of global coarsening, this is the mapping associated to the fine level triangulation.
   std::shared_ptr<dealii::Mapping<dim> const> mapping;
 
-  // Only relevant for global coarsening,where this vector contains coarse level mappings,
+  // Only relevant for global coarsening, where this vector contains coarse level mappings,
```
peterrum, line 322: ... or is this comment out of date?

peterrum (approved):
Looks reasonable.

peterrum:
@nfehn You need to to add your new email address to you github account. GitHub is not picking up your commits...

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.h
nfehn: I think for the use in the multigrid algorithm, we only need a `dealii::Mapping`. So my goal is to make the implementation more generic. If I see it correctly, we only need `MappingDoFVector` during construction, and could then let a `std::shared_ptr<dealii::Mapping>` point to this object. Following this strategy, the "construction code" can later be shifted to a lambda function implemented outside multigrid, e.g. in the application.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.h
nfehn: yes, out of date.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -354,30 +350,46 @@ template<int dim, typename Number>
...
+        MappingTools::initialize_multigrid<dim, Number>(coarse_grid_mappings,
+                                                        mapping_q_cache,
+                                                        coarse_triangulations);
+      }
+      else if(multigrid_variant == MultigridVariant::LocalSmoothing)
+      {
+        std::shared_ptr<MappingDoFVector<dim, Number>> mapping_dof_vector =
+          std::make_shared<MappingDoFVector<dim, Number>>(mapping_q_cache->get_degree());
+
+        MappingTools::initialize_multigrid(mapping_dof_vector, mapping_q_cache, *triangulation);
```
kronbichler, line 375: Seeing this code, I'm wondering if we should let `initialize_multigrid` return `std::shared_ptr<MappingDoFVector<dim, Number>>` instead of passing it in as first argument (also for the other function). To me, it would look more intuitive if the computed quantity is returned, and in fact both cases seem to throw away the previous content anyway. This must not necessarily happen in this PR, though.

kronbichler (approved):
This looks great.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
kronbichler: Ah, that's again one of those cases. I believe it will not work automatically. We could of course return `std::vector<std::shared_ptr<dealii::Mapping>>` directly, but then it is maybe less clear how it gets computed.

nfehn: It makes sense in my opinion to unify the interfaces already in this PR. I talked to  @peterrum and we discussed to consider the case with a vector of mappings for coarse h-levels as the default case/interface and to consider dealii implementations with a single triangulation/mapping (for all h-levels) as legacy code supported internally.

nfehn:
Thanks for your comments!

@peterrum @kronbichler could you briefly look at the newest commits before merging?

kronbichler (approved):

PR #448, opened 2023-05-24
Make access to MultigridPreconditionerBase::n_levels safe

nfehn:
Currently, one could access the integer variables `coarse_level, fine_level, n_levels` before setup/initialization.

I suggest to remove all those member variables and only use the member function `get_number_of_levels()` instead. This function asserts that `level_info.size()>0`, which would give an error when called in an uninitialized state.

### DIFF
--- a/include/exadg/convection_diffusion/preconditioners/multigrid_preconditioner.cpp
+++ b/include/exadg/convection_diffusion/preconditioners/multigrid_preconditioner.cpp
@@ -327,7 +327,7 @@ void
 MultigridPreconditioner<dim, Number>::update_smoothers()
 {
   // Skip coarsest level
-  for(unsigned int level = 0; level <= this->get_number_of_levels() - 1; ++level)
+  for(unsigned int level = 1; level <= this->get_number_of_levels() - 1; ++level)
     this->update_smoother(level);
 }
 
--- a/include/exadg/convection_diffusion/preconditioners/multigrid_preconditioner.cpp
+++ b/include/exadg/convection_diffusion/preconditioners/multigrid_preconditioner.cpp
@@ -281,11 +281,14 @@ template<int dim, typename Number>
 void
 MultigridPreconditioner<dim, Number>::set_velocity(VectorTypeMG const & velocity)
 {
+  unsigned int const fine_level   = this->get_number_of_levels() - 1;
+  unsigned int const coarse_level = 0;
+
   // copy velocity to finest level
-  this->get_operator(this->fine_level)->set_velocity_copy(velocity);
+  this->get_operator(fine_level)->set_velocity_copy(velocity);
 
   // interpolate velocity from fine to coarse level
-  for(unsigned int level = this->fine_level; level > this->coarse_level; --level)
+  for(unsigned int level = fine_level; level > coarse_level; --level)
   {
     auto & vector_fine_level   = this->get_operator(level - 0)->get_velocity();
     auto   vector_coarse_level = this->get_operator(level - 1)->get_velocity();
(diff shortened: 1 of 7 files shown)


include/exadg/convection_diffusion/preconditioners/multigrid_preconditioner.cpp
```diff
@@ -281,11 +281,14 @@ template<int dim, typename Number>
 void
 MultigridPreconditioner<dim, Number>::set_velocity(VectorTypeMG const & velocity)
 {
+  unsigned int const fine_level   = this->get_number_of_levels() - 1;
+  unsigned int const coarse_level = 0;
```
peterrum, line 285: I guess `coarse_level = 0` is per convention in ExaDG? In deal.II we can terminate at any level.

include/exadg/convection_diffusion/preconditioners/multigrid_preconditioner.cpp
```diff
@@ -281,11 +281,14 @@ template<int dim, typename Number>
 void
 MultigridPreconditioner<dim, Number>::set_velocity(VectorTypeMG const & velocity)
 {
+  unsigned int const fine_level   = this->get_number_of_levels() - 1;
```
peterrum, line 284: I `get_number_of_levels` checking that the returned value is `>0`?

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -658,10 +657,10 @@ template<int dim, typename Number>
 void
 MultigridPreconditionerBase<dim, Number>::initialize_smoothers()
 {
-  this->smoothers.resize(0, this->n_levels - 1);
+  this->smoothers.resize(0, get_number_of_levels() - 1);
```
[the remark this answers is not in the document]
peterrum: We should be able to skip that one. In practice, I sometimes use the smoother on coarsest level for preconditioning the coarse Krylov solver. But I guess this is not the case of ExaDG?

include/exadg/convection_diffusion/preconditioners/multigrid_preconditioner.cpp
nfehn: Yes, it was hard-coded to 0 before. So I suggest to introduce this again once that functionality is really implemented.

include/exadg/convection_diffusion/preconditioners/multigrid_preconditioner.cpp
nfehn: I can add this.

nfehn:
> We should be able to skip that one. In practice, I sometimes use the smoother on coarsest level for preconditioning the coarse Krylov solver. But I guess this is not the case of ExaDG?


Indeed, I found `smoothers[0]` in `initialize_coarse_solver()`, so I should at least add a comment where we resize the smoothers-vector.

Actually, I would prefer an implementation that is more explicit about this. What do you think @peterrum @kronbichler ?

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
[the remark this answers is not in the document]
nfehn: see https://github.com/exadg/exadg/pull/448#issuecomment-1561113407

include/exadg/convection_diffusion/preconditioners/multigrid_preconditioner.cpp
nfehn: Has already been done. see implementation of function.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -793,7 +794,7 @@ template<int dim, typename Number>
 void
 MultigridPreconditionerBase<dim, Number>::update_smoother(unsigned int level)
 {
-  AssertThrow(level > 0,
+  AssertThrow(level > 0 and level < this->get_number_of_levels(),
```
nfehn, line 797: This additional check caused a test to fail.

include/exadg/convection_diffusion/preconditioners/multigrid_preconditioner.cpp
```diff
@@ -326,7 +327,7 @@ void
 MultigridPreconditioner<dim, Number>::update_smoothers()
 {
   // Skip coarsest level
-  for(unsigned int level = this->coarse_level + 1; level <= this->fine_level; ++level)
+  for(unsigned int level = 0; level <= this->get_number_of_levels() - 1; ++level)
```
nfehn, line 330: Here is the mistake causing the failing test described here https://github.com/exadg/exadg/pull/448/files#r1205088650

nfehn: fixed here https://github.com/exadg/exadg/pull/448/commits/b0d63fffe82ddc9dff2a54a1d4861b88db2ecbe5

PR #451, opened 2023-05-24
remove last entry of vector coarse_triangulations/coarse_mappings

### LINKED ISSUE #449: Remove finest triangulation from vector Grid::coarse_triangulations or rename
See https://github.com/exadg/exadg/blob/3bc6d4178089288615107f36d74e25872251d9fc/include/exadg/grid/grid.h#L71-L75

It is not clear from the name why this variable should include the finest triangulation as the last element of the vector. However, interfaces in ExaDG currently expect this implicitly. I suggest to change the logic and only extend the vector inside the multigrid preconditioner if required by interfaces of deal.II.

The same holds for the coarse_mappings in `MultigridPreconditionerBase`. If I see it correctly, we even initialize the `MappingQCache` for the finest level even though we already have a fine level mapping, see https://github.com/exadg/exadg/blob/3bc6d4178089288615107f36d74e25872251d9fc/include/exadg/grid/mapping_dof_vector.h#L476-L490

nfehn:
closes #449

### DIFF
--- a/include/exadg/grid/grid_utilities.h
+++ b/include/exadg/grid/grid_utilities.h
@@ -442,7 +442,7 @@ create_fine_and_coarse_triangulations(
 
   // Depending on the type of geometric multigrid coarsening, we need to explicitly create the
   // coarse triangulations.
-  if(data.multigrid == MultigridVariant::GlobalCoarsening and involves_h_multigrid)
+  if(data.multigrid == MultigridVariant::GlobalCoarsening)
   {
     if(involves_h_multigrid)
     {
--- a/include/exadg/grid/grid_utilities.h
+++ b/include/exadg/grid/grid_utilities.h
@@ -241,10 +241,14 @@ create_triangulation(
     else if(data.element_type == ElementType::Hypercube)
     {
       if(data.multigrid == MultigridVariant::LocalSmoothing)
+      {
         triangulation_description_setting =
           dealii::TriangulationDescription::construct_multigrid_hierarchy;
+      }
       else
+      {
         triangulation_description_setting = dealii::TriangulationDescription::default_setting;
+      }
     }
     else
     {
@@ -273,7 +277,10 @@ create_triangulation(
 
 /**
  * Given a fine_triangulation, this function creates all the coarse triangulations required for
- * global-coarsening multigrid
+ * multigrid implementations that expect a vector of triangulations.
+ *
+ * The vector coarse_triangulations only includes the levels coarser than the fine triangulation,
+ * where the first entry of the vector corresponds to the coarsest level.
  */
 template<int dim>
 inline void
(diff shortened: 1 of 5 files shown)


nfehn:
The code should now be more explicit about coarse levels, the fine level, and all levels.

include/exadg/grid/grid_utilities.h
```diff
@@ -290,118 +297,129 @@ create_coarse_triangulations(
...
-
-    // we start with the finest level
-    unsigned int              level        = fine_triangulation.n_global_levels() - 1;
-    std::vector<unsigned int> refine_local = vector_local_refinements;
-
-    // make the last entry of the coarse_triangulations point to the fine_triangulation
-    coarse_triangulations[level].reset(&fine_triangulation, [](auto *) {
-      // empty deleter, since fine_triangulation is an external field
-      // and its destructor is called somewhere else
-    });
```
nfehn, line 363: I think we no longer need this "hard to understand" code (also not somewhere else). In my opinion this also shows that appending the vector of coarse triangulations by the fine triangulation was somewhat artificial in the previous code version.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -148,27 +148,9 @@ MultigridPreconditionerBase<dim, Number>::initialize_levels(unsigned int const d
...
-        dealii::ExcMessage(
-          "You are using global coarsening multigrid, but the vector coarse_triangulations is empty. "
-          "Most likely, you forgot to set the parameter GridData::create_coarse_triangulations."));
-    }
-
-    unsigned int const n_h_levels = this->get_number_of_h_levels();
-
-    for(unsigned int h = 0; h < n_h_levels; h++)
-      h_levels.push_back(h);
-  }
```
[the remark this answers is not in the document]
nfehn: The point is that `get_number_of_h_levels()` should return 1 if no h-multigrid is involved. Then, the loop below only inserts a value of "0", while the value has been `n_global_levels-1` before for the case without h-multigrid involved. However, I think the concrete number should not be relevant.

include/exadg/grid/grid_utilities.h
```diff
@@ -290,118 +297,129 @@ create_coarse_triangulations(
...
+            level--;
+        }
       }
+
+      AssertThrow(level == 0,
+                  dealii::ExcMessage(
+                    "There occurred a logical error when creating geometric coarsening sequence."));
     }
   }
+  else
```
nfehn, line 413: There was an else-branch missing in the previous code version

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -370,9 +352,19 @@ MultigridPreconditionerBase<dim, Number>::get_number_of_h_levels() const
...
+    }
+    else if(multigrid_variant == MultigridVariant::GlobalCoarsening)
+    {
+      return grid->coarse_triangulations.size() + 1;
+    }
+    else
+    {
+      AssertThrow(false, dealii::ExcMessage("Not implemented."));
+      return 0;
+    }
```
nfehn, line 367: Preparation for new multigrid coarsening sequences like "NonMatching/NonNested".

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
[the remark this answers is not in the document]
nfehn: This change introduced a bug, as turned out now. Going to be fixed in #464

PR #456, opened 2023-05-25
Multigrid: Use more generic loop over levels

nfehn:
This PR introduces generic loops over all multigrid levels and the smoothing levels, where the operation to be performed on a particular level is provided by a lambda with `unsigned int level` as argument.

This code can especially be used in derived multigrid classes, where the for-loop over the multigrid levels is written out explicitly many times, which is error prone (see for example https://github.com/exadg/exadg/pull/448/commits/b0d63fffe82ddc9dff2a54a1d4861b88db2ecbe5).

I consider the many member functions in derived multigrid preconditioner classes as needless complexity, which I therefore removed in the present PR.

### DIFF
--- a/include/exadg/convection_diffusion/preconditioners/multigrid_preconditioner.cpp
+++ b/include/exadg/convection_diffusion/preconditioners/multigrid_preconditioner.cpp
@@ -119,8 +119,8 @@ MultigridPreconditioner<dim, Number>::update()
   this->for_all_levels([&](unsigned int const level) {
     // the velocity field of the convective term is a function of the time
     this->get_operator(level)->set_time(pde_operator->get_time());
-    // in case of adaptive time stepping, the scaling factor of the derivative term changes over
-    // time
+    // in case of adaptive time stepping, the scaling factor of the time derivative term changes
+    // over time
     this->get_operator(level)->set_scaling_factor_mass_operator(
       pde_operator->get_scaling_factor_mass_operator());
   });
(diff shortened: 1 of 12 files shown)


kronbichler (approved):
Looks helpful to me. I would personally think that the for loop from `0` to `get_number_of_levels()` (exclusive) is easier to understand and maintain compared to the cost of introducing another abstraction that one needs to understand as a developer and thus implies a threshold to get productive. Secondly, I also think that lambda functions are easy to mis-use and lead to hard-to-understand and hard-to-debug code because one can jump around without seeing the names. Still, I can see some value of this here, so if you think it makes sense, we can go for it.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.h
```diff
@@ -147,14 +147,14 @@ class MultigridPreconditionerBase : public PreconditionerBase<Number>
    * This function initializes the matrix-free objects for all multigrid levels.
    */
   virtual void
-  initialize_matrix_free();
+  initialize_matrix_free_objects();
```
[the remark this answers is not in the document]
nfehn: same for update function below

include/exadg/convection_diffusion/preconditioners/multigrid_preconditioner.cpp
```diff
@@ -135,14 +146,25 @@ MultigridPreconditioner<dim, Number>::update()
...
+    unsigned int const coarse_level = 0;
 
-  // Once the operators are updated, the update of smoothers and the coarse grid solver is generic
-  // functionality implemented in the base class
+    // copy velocity to finest level
+    this->get_operator(fine_level)->set_velocity_copy(*vector_multigrid_type_ptr);
 
-  update_smoothers();
+    // interpolate velocity from fine to coarse level
+    for(unsigned int level = fine_level; level > coarse_level; --level)
```
peterrum, line 156: What about introducing `for_all_levels_reversed()`?

peterrum (approved):

include/exadg/convection_diffusion/preconditioners/multigrid_preconditioner.cpp
nfehn: The problem is that we have the condition `level > coarse_level`, because the functionality here is to "compare" two levels with each other. Writing `level-0` and `level-1` versus `level+1` and `level+0` is a convention. Hence, I do not consider this a classical "for-loop" over the levels. In this case, I consider the use of lambda error prone (and would agree here with the general comment by @kronbichler).

A generic function as suggested above would need to include an implementation of the index logic inside the function. Any suggestions regarding name and interface of such a generic function? `(transfer_)from_fine_to_all_coarser_levels([](uint, uint){})` with a lambda taking two parameters `unsigned int fine_level`/`unsigned int coarse_level` as arguments?

nfehn: You are definitely right that this is another type of functionality currently appearing many times in the code.

nfehn: see https://github.com/exadg/exadg/pull/456/commits/789280dc6864931adfa3470d5bd4d0256700ce1e

peterrum (approved):

PR #457, opened 2023-06-01
dont store mean velocity calculator data as const reference

jh66637:
Classes should always store a copy of `data` and not a `const` reference.


nfehn:
makes sense, all items of `MeanVelocityCalculatorData` should be copieable without need for a copy constructor.

nfehn (approved):

nfehn (approved):

PR #458, opened 2023-06-01
expose data needed for pressure time derivatives

jh66637:
I want to compute pressure time derivatives of the INSE from outside. To be able to do so some getters are required and `compute_bdf_time_derivative` has to be able to work on vectors of pointers.

Motivation: This is needed for aero-acoustic source term computation.

### DIFF
--- a/include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.cpp
+++ b/include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.cpp
@@ -177,6 +177,59 @@ TimeIntBDF<dim, Number>::advance_one_timestep_partitioned_solve(bool const use_e
 
 template<int dim, typename Number>
 void
+TimeIntBDF<dim, Number>::get_quantities_and_times(
+  std::vector<VectorType const *> &                             quantities,
+  std::vector<double> &                                         times,
+  std::function<VectorType const *(unsigned int const)> const & get_quantity) const
+{
+  /*
+   *
+   *   time t
+   *  -------->   t_{n-2}        t_{n-1}         t_{n}         t_{n+1}
+   *  _______________|______________|______________|______________|___________\
+   *                 |              |              |              |           /
+   *           quantities[2]  quantities[1]  quantities[0]
+   *              times[2]       times[1]       times[0]
+   */
+
+  unsigned int const current_order = this->get_current_order();
+
+  quantities.resize(current_order);
+  times.resize(current_order);
+
+  for(unsigned int i = 0; i < current_order; ++i)
+  {
+    quantities[i] = get_quantity(i);
+    times[i]      = this->get_previous_time(i);
+  }
+}
+
+template<int dim, typename Number>
+void
+TimeIntBDF<dim, Number>::get_quantities_and_times_np(
+  std::vector<VectorType const *> &                             quantities,
+  std::vector<double> &                                         times,
+  std::function<VectorType const *(unsigned int const)> const & get_quantity,
+  std::function<VectorType const *()> const &                   get_quantity_np) const
+{
+  /*
+   *
+   *   time t
+   *  -------->   t_{n-2}        t_{n-1}         t_{n}         t_{n+1}
+   *  _______________|______________|______________|______________|___________\
+   *                 |              |              |              |           /
+   *           quantities[3]  quantities[2]  quantities[1]  quantities[0]
+   *              times[3]       times[2]       times[1]       times[0]
+   */
+
+  get_quantities_and_times(quantities, times, get_quantity);
+
+  quantities.insert(quantities.begin(), get_quantity_np());
+  times.insert(times.begin(), this->get_next_time());
+}
+
+template<int dim, typename Number>
+void
 TimeIntBDF<dim, Number>::initialize_vec_convective_term()
 {
   this->operator_base->evaluate_convective_term(vec_convective_term[0],
@@ -383,34 +436,7 @@ void
 TimeIntBDF<dim, Number>::get_velocities_and_times(std::vector<VectorType const *> & velocities,
                                                   std::vector<double> &             times) const
 {
-  /*
-   * the convective term is nonlinear, so we have to initialize the transport velocity
-   * and the discrete time instants that can be used for interpolation
-   *
-   *   time t
-   *  -------->   t_{n-2}   t_{n-1}   t_{n}     t_{n+1}
-   *  _______________|_________|________|___________|___________\
-   *                 |         |        |           |           /
-   *               sol[2]    sol[1]   sol[0]
-   *             times[2]  times[1]  times[0]
-   */
-  unsigned int current_order = this->order;
-  if(this->time_step_number <= this->order and this->param.start_with_low_order == true)
-  {
-    current_order = this->time_step_number;
-  }
-
-  AssertThrow(current_order > 0 and current_order <= this->order,
-              dealii::ExcMessage("Invalid parameter current_order"));
-
-  velocities.resize(current_order);
-  times.resize(current_order);
-
-  for(unsigned int i = 0; i < current_order; ++i)
-  {
-    velocities.at(i) = &get_velocity(i);
-    times.at(i)      = this->get_previous_time(i);
-  }
+  get_quantities_and_times(velocities, times, [&](const auto i) { return &get_velocity(i); });
 }
 
 template<int dim, typename Number>
@@ -418,37 +444,33 @@ void
 TimeIntBDF<dim, Number>::get_velocities_and_times_np(std::vector<VectorType const *> & velocities,
                                                      std::vector<double> &             times) const
 {
-  /*
-   * the convective term is nonlinear, so we have to initialize the transport velocity
-   * and the discrete time instants that can be used for interpolation
-   *
-   *   time t
-   *  -------->     t_{n-2}   t_{n-1}   t_{n}     t_{n+1}
-   *  _______________|_________|________|___________|___________\
-   *                 |         |        |           |           /
-   *               sol[3]   sol[2]    sol[1]     sol[0]
-   *              times[3] times[2]  times[1]   times[0]
-   */
-  unsigned int current_order = this->order;
-  if(this->time_step_number <= this->order and this->param.start_with_low_order == true)
-  {
-    current_order = this->time_step_number;
-  }
-
-  AssertThrow(current_order > 0 and current_order <= this->order,
-              dealii::ExcMessage("Invalid parameter current_order"));
-
-  velocities.resize(current_order + 1);
-  times.resize(current_order + 1);
-
-  velocities.at(0) = &get_velocity_np();
-  times.at(0)      = this->get_next_time();
-  for(unsigned int i = 0; i < current_order; ++i)
-  {
-    velocities.at(i + 1) = &get_velocity(i);
-    times.at(i + 1)      = this->get_previous_time(i);
-  }
+  get_quantities_and_times_np(
+    velocities,
+    times,
+    [&](const auto i) { return &get_velocity(i); },
+    [&]() { return &get_velocity_np(); });
 }
+
+template<int dim, typename Number>
+void
+TimeIntBDF<dim, Number>::get_pressures_and_times(std::vector<VectorType const *> & pressures,
+                                                 std::vector<double> &             times) const
+{
+  get_quantities_and_times(pressures, times, [&](const auto i) { return &get_pressure(i); });
+}
+
+template<int dim, typename Number>
+void
+TimeIntBDF<dim, Number>::get_pressures_and_times_np(std::vector<VectorType const *> & pressures,
+                                                    std::vector<double> &             times) const
+{
+  get_quantities_and_times_np(
+    pressures,
+    times,
+    [&](const auto i) { return &get_pressure(i); },
+    [&]() { return &get_pressure_np(); });
+}
+
 
 template<int dim, typename Number>
 void
--- a/include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.h
+++ b/include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.h
@@ -81,6 +81,14 @@ public:
                               std::vector<double> &             times) const;
 
   void
+  get_pressures_and_times(std::vector<VectorType const *> & pressures,
+                          std::vector<double> &             times) const;
+
+  void
+  get_pressures_and_times_np(std::vector<VectorType const *> & pressures,
+                             std::vector<double> &             times) const;
+
+  void
   ale_update();
 
   void
@@ -132,6 +140,19 @@ protected:
   std::shared_ptr<HelpersALE<Number> const> helpers_ale;
 
 private:
+  void
+  get_quantities_and_times(
+    std::vector<VectorType const *> &                             quantities,
+    std::vector<double> &                                         times,
+    std::function<VectorType const *(unsigned int const)> const & get_quantity) const;
+
+  void
+  get_quantities_and_times_np(
+    std::vector<VectorType const *> &                             quantities,
+    std::vector<double> &                                         times,
+    std::function<VectorType const *(unsigned int const)> const & get_quantity,
+    std::function<VectorType const *()> const &                   get_quantity_np) const;
+
   void
   initialize_vec_convective_term();
 
--- a/include/exadg/time_integration/time_int_bdf_base.cpp
+++ b/include/exadg/time_integration/time_int_bdf_base.cpp
@@ -140,6 +140,14 @@ TimeIntBDFBase::get_previous_time(int const i /* t_{n-i} */) const
   return t;
 }
 
+unsigned int
+TimeIntBDFBase::get_current_order() const
+{
+  if(start_with_low_order == true and time_step_number <= order)
+    return this->time_step_number;
+  return order;
+}
+
 double
 TimeIntBDFBase::get_time_step_size() const
 {
--- a/include/exadg/time_integration/time_int_bdf_base.h
+++ b/include/exadg/time_integration/time_int_bdf_base.h
@@ -97,6 +97,12 @@ public:
 
 protected:
   /*
+   * Get current order of time integrator
+   */
+  unsigned int
+  get_current_order() const;
+
+  /*
    * Do one time step including different updates before and after the actual solution of the
    * current time step.
    */
--- a/tests/time_integration/bdf_time_derivative_01.cc
+++ b/tests/time_integration/bdf_time_derivative_01.cc
@@ -1,0 +1,93 @@
+/*  ______________________________________________________________________
+ *
+ *  ExaDG - High-Order Discontinuous Galerkin for the Exa-Scale
+ *
+ *  Copyright (C) 2021 by the ExaDG authors
+ *
+ *  This program is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *  ______________________________________________________________________
+ */
+
+#include <iostream>
+#include <numeric>
+
+#include <deal.II/lac/la_vector.h>
+
+#include <exadg/time_integration/bdf_constants.h>
+
+// Check compute_bdf_time_derivative() from quantities and times
+
+using namespace ExaDG;
+using VectorType = dealii::LinearAlgebra::Vector<double>;
+
+VectorType
+get_vector_with_value(double const value)
+{
+  VectorType vec(1);
+  vec[0] = value;
+  return vec;
+}
+
+void
+test(unsigned int const order, double const target_slope)
+{
+  std::cout << "Check derivative of slope " << target_slope << std::endl;
+
+  // construct quantites and times
+  std::vector<VectorType> quantities_np_values(order + 1);
+  std::vector<double>     times_np(order + 1);
+
+  // current time
+  times_np[0] = 10.0;
+
+  // start value
+  quantities_np_values[0] = get_vector_with_value(100.0);
+
+  double const dt = 1.0;
+  for(unsigned int i = 1; i < order + 1; ++i)
+  {
+    times_np[i] = times_np[i - 1] - dt;
+
+    quantities_np_values[i] =
+      get_vector_with_value(-target_slope * dt + quantities_np_values[i - 1][0]);
+  }
+
+  // transfrom to vector of pointers
+  std::vector<VectorType const *> quantities_np(quantities_np_values.size());
+  std::transform(quantities_np_values.begin(),
+                 quantities_np_values.end(),
+                 quantities_np.begin(),
+                 [](VectorType const & t) { return &t; });
+
+  // compute temporal derivative
+  VectorType derivative(1);
+  derivative[0] = -99.0;
+  compute_bdf_time_derivative(derivative, quantities_np, times_np);
+
+  // check if computed slope equals target slope
+  assert(std::abs(derivative[0] - target_slope) < 1e-12);
+  std::cout << "OK" << std::endl;
+}
+
+int
+main()
+{
+  for(unsigned int order = 1; order <= 4; ++order)
+  {
+    std::cout << "Order " << order << std::endl;
+    test(order, 0.0);
+    test(order, -1.0);
+  }
+  return 0;
+}
--- a/include/exadg/time_integration/bdf_constants.h
+++ b/include/exadg/time_integration/bdf_constants.h
@@ -23,10 +23,12 @@
 #define INCLUDE_EXADG_TIME_INTEGRATION_BDF_CONSTANTS_H_
 
 // C/C++
+#include <algorithm>
 #include <vector>
 
 // deal.II
 #include <deal.II/base/conditional_ostream.h>
+#include <deal.II/base/exceptions.h>
 
 // ExaDG
 #include <exadg/time_integration/time_integration_constants_base.h>
@@ -76,17 +78,67 @@ private:
  */
 template<typename VectorType>
 void
+compute_bdf_time_derivative(VectorType &                            derivative,
+                            VectorType const &                      solution_np,
+                            std::vector<VectorType const *> const & previous_solutions,
+                            BDFTimeIntegratorConstants const &      bdf,
+                            double const &                          time_step_size)
+{
+  derivative.equ(bdf.get_gamma0() / time_step_size, solution_np);
+
+  for(unsigned int i = 0; i < previous_solutions.size(); ++i)
+    derivative.add(-bdf.get_alpha(i) / time_step_size, *previous_solutions[i]);
+}
+
+template<typename VectorType>
+void
 compute_bdf_time_derivative(VectorType &                       derivative,
                             VectorType const &                 solution_np,
                             std::vector<VectorType> const &    previous_solutions,
                             BDFTimeIntegratorConstants const & bdf,
                             double const &                     time_step_size)
 {
-  derivative.equ(bdf.get_gamma0() / time_step_size, solution_np);
+  std::vector<VectorType const *> previous_solutions_ptrs(previous_solutions.size());
 
-  for(unsigned int i = 0; i < previous_solutions.size(); ++i)
-    derivative.add(-bdf.get_alpha(i) / time_step_size, previous_solutions[i]);
+  std::transform(previous_solutions.begin(),
+                 previous_solutions.end(),
+                 previous_solutions_ptrs.begin(),
+                 [](VectorType const & t) { return &t; });
+
+  compute_bdf_time_derivative(
+    derivative, solution_np, previous_solutions_ptrs, bdf, time_step_size);
 }
+
+template<typename VectorType>
+void
+compute_bdf_time_derivative(VectorType &                            derivative,
+                            std::vector<VectorType const *> const & previous_solutions_np,
+                            std::vector<double> const &             times_np)
+{
+  AssertThrow(
+    previous_solutions_np.size() == times_np.size() and times_np.size() > 0,
+    dealii::ExcMessage(
+      "times and previous_solutions_np handed to compute_bdf_time_derivative() have different sizes or size 0."));
+
+  std::vector<VectorType const *> previous_solutions(previous_solutions_np.begin() + 1,
+                                                     previous_solutions_np.end());
+
+  // compute time step sizes from times
+  std::vector<double> time_steps(times_np.size() - 1);
+  for(unsigned int i = 0; i < time_steps.size(); ++i)
+    time_steps[i] = times_np[i] - times_np[i + 1];
+
+
+  // construct BDF constants
+  unsigned int const         order = previous_solutions.size();
+  BDFTimeIntegratorConstants bdf(order, true);
+  bdf.update(order, true, time_steps);
+
+  // compute temporal derivative
+  compute_bdf_time_derivative(
+    derivative, *previous_solutions_np[0], previous_solutions, bdf, time_steps[0]);
+}
+
 
 } // namespace ExaDG
 
--- a/tests/time_integration/bdf_time_derivative_01.output
+++ b/tests/time_integration/bdf_time_derivative_01.output
@@ -1,0 +1,20 @@
+Order 1
+Check derivative of slope 0
+OK
+Check derivative of slope -1
+OK
+Order 2
+Check derivative of slope 0
+OK
+Check derivative of slope -1
+OK
+Order 3
+Check derivative of slope 0
+OK
+Check derivative of slope -1
+OK
+Order 4
+Check derivative of slope 0
+OK
+Check derivative of slope -1
+OK


include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.cpp
```diff
@@ -450,6 +450,79 @@ TimeIntBDF<dim, Number>::get_velocities_and_times_np(std::vector<VectorType cons
...
 }
 
+template<int dim, typename Number>
+void
+TimeIntBDF<dim, Number>::get_pressures_and_times(std::vector<VectorType const *> & pressures,
+                                                 std::vector<double> &             times) const
+{
+  /*
+   * the convective term is nonlinear, so we have to initialize the transport velocity
+   * and the discrete time instants that can be used for interpolation
```
nfehn, line 460: This comment does not seem to fit here.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.h
```diff
@@ -80,6 +80,14 @@ class TimeIntBDF : public TimeIntBDFBase<Number>
...
                               std::vector<double> &             times) const;
 
+  void
+  get_pressures_and_times(std::vector<VectorType const *> & pressures,
+                          std::vector<double> &             times) const;
+
+  void
+  get_pressures_and_times_np(std::vector<VectorType const *> & pressures,
+                             std::vector<double> &             times) const;
+
```
nfehn, line 90: It could still make sense to have these new functions in the public interface. However, the same code should be used for velocity and pressure in the internal implementation.

jh66637: Yes, I would like to keep the interface, but I am restructuring the internal code :)

jh66637: Yes, I would like to keep the interface, but I restructured the internal code :)

include/exadg/time_integration/time_int_bdf_base.h
```diff
@@ -101,6 +104,49 @@ class TimeIntBDFBase : public TimeIntBase
...
+   */
+  void
+  fill_with_previous_times(std::vector<double>::iterator const & begin,
+                           std::vector<double>::iterator const & end) const
+  {
+    fill_at_previous_times(begin, end, [&](auto const i) { return get_previous_time(i); });
+  }
+
+  BDFTimeIntegratorConstants const &
+  get_bdf_constants() const;
```
jh66637, line 148: TODO (for myself): It would be better to place this in `TimeIntBDFBase` :)

include/exadg/time_integration/time_int_bdf_base.h
```diff
@@ -101,6 +104,49 @@ class TimeIntBDFBase : public TimeIntBase
...
+     *                end       ...     begin
+     */
+
+    AssertThrow(std::distance(begin, end) <= get_current_order(),
+                dealii::ExcMessage("Iterator range exceeds available time steps."));
+
+    unsigned int i = 0;
+    for(auto it = begin; it != end; ++it, ++i)
+      *it = f(i);
+  }
```
nfehn, line 135: This function makes the code cryptic in my opinion. Note that "previous times" in the documentation above does not occur in the implementation of the function.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.h
```diff
@@ -132,6 +140,39 @@ class TimeIntBDF : public TimeIntBDFBase<Number>
...
+  template<typename Lambda>
+  void
+  get_quantity_and_times(std::vector<VectorType const *> & quantities,
+                         std::vector<double> &             times,
+                         Lambda const &                    get_quantity) const
+  {
+    unsigned int const current_order = this->get_current_order();
+
+    quantities.resize(current_order);
+    this->fill_at_previous_times(quantities.begin(), quantities.end(), get_quantity);
```
nfehn, line 152: I do not really see a value in calling this function over running a for-loop explicitly. Having a template `Lambda` here makes the code unnecessarily complex to understand in my opinion.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.h
```diff
@@ -132,6 +140,39 @@ class TimeIntBDF : public TimeIntBDFBase<Number>
...
+
+    times.resize(current_order);
+    this->fill_with_previous_times(times.begin(), times.end());
```
nfehn, line 155: The naming `fill_at_...` and `fill_with_...` is not intuitive in my opinion.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.h
```diff
@@ -132,6 +140,39 @@ class TimeIntBDF : public TimeIntBDFBase<Number>
...
+  get_quantities_and_times_np(std::vector<VectorType const *> & quantities,
+                              std::vector<double> &             times,
+                              Lambda1 const &                   get_quantity,
+                              Lambda2 const &                   get_quantity_np) const
+  {
+    unsigned int const current_order = this->get_current_order();
+
+    quantities.resize(current_order + 1);
+    quantities[0] = get_quantity_np();
+    this->fill_at_previous_times(quantities.begin() + 1, quantities.end(), get_quantity);
```
nfehn, line 169: Again, I do not see the advantage over an explicit loop.

It would be more desirable in my opinion to design `get_quantities_and_times_np()` as a combination of `get_quantities_and_times` plus some additional operations for time t_{n+1}. However, I did not think about all the details.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.h
```diff
@@ -132,6 +140,39 @@ class TimeIntBDF : public TimeIntBDFBase<Number>
   std::shared_ptr<HelpersALE<Number> const> helpers_ale;
 
 private:
+  template<typename Lambda>
+  void
+  get_quantity_and_times(std::vector<VectorType const *> & quantities,
```
nfehn, line 145: should also be `quantities` in the name since we have a  vector of DoF vectors.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.h
```diff
@@ -132,6 +140,39 @@ class TimeIntBDF : public TimeIntBDFBase<Number>
   std::shared_ptr<HelpersALE<Number> const> helpers_ale;
 
 private:
+  template<typename Lambda>
```
nfehn, line 143: I think one should generally introduce a template parameter only if the type is not known (in the sense of generic programming). I think this is not given here, so that the template parameter should be avoided here.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.cpp
```diff
@@ -383,73 +383,42 @@ void
...
-              dealii::ExcMessage("Invalid parameter current_order"));
-
-  velocities.resize(current_order);
-  times.resize(current_order);
-
-  for(unsigned int i = 0; i < current_order; ++i)
-  {
-    velocities.at(i) = &get_velocity(i);
-    times.at(i)      = this->get_previous_time(i);
-  }
```
nfehn, line 413: This code is much more readable in my opinion than the current suggestion. I suggest to generalize the above code by just replacing `velocity` by a more general `vector` and a corresponding lambda `get_vector(i)`.

nfehn:
I like the restructuring in parts. However, the code should not be more generic than what we need in the context of the present time integration routines in my opinion. If one has to pass begin() and end() iterators to a function doing not more than a for-loop, one can instead write the for loop explicitly in my opinion. This makes the code much easier to understand.

This is different in PR https://github.com/exadg/exadg/pull/456, because there the generic functions take care of the begin and end of the for loop (which was exactly the motivation for introducing generic functions in that PR).

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.h
jh66637: That's why I introduced the iterator version. This way, the same function can be used to fill the vector. I can also use `.insert()` and get rid of the parts you don't like. I am just used to avoid this because it implies more iterations than needed. Since in this part of the code It doesn't matter, I will restructure accordingly.

jh66637:
I restructured as you suggested :)

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.cpp
jh66637: [1f39cef](https://github.com/exadg/exadg/pull/458/commits/1f39cef8d688232b73b239620a45f409a169e862)

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.h
jh66637: The only way to avoid this is by using a `std::function` as I did in [1f39cef](https://github.com/exadg/exadg/pull/458/commits/1f39cef8d688232b73b239620a45f409a169e862). The disadvantage is performance as far as I remember. Anyway, in this part of the code it does not matter.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.h
jh66637: [1f39cef](https://github.com/exadg/exadg/pull/458/commits/1f39cef8d688232b73b239620a45f409a169e862)

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.h
jh66637: Not used anymore
[1f39cef](https://github.com/exadg/exadg/pull/458/commits/1f39cef8d688232b73b239620a45f409a169e862)

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.h
jh66637: [1f39cef](https://github.com/exadg/exadg/pull/458/commits/1f39cef8d688232b73b239620a45f409a169e862)

include/exadg/time_integration/time_int_bdf_base.h
jh66637: Removed in [1f39cef](https://github.com/exadg/exadg/pull/458/commits/1f39cef8d688232b73b239620a45f409a169e862)

include/exadg/time_integration/time_int_bdf_base.cpp
```diff
@@ -146,6 +146,22 @@ TimeIntBDFBase<Number>::get_previous_time(int const i /* t_{n-i} */) const
...
 }
 
+template<typename Number>
+unsigned int
+TimeIntBDFBase<Number>::get_current_order() const
+{
+  if(start_with_low_order == true && time_step_number <= order)
+    return this->time_step_number;
+  return order;
+}
```
nfehn, line 156: This is a good refactoring!

include/exadg/time_integration/time_int_bdf_base.h
```diff
@@ -101,6 +101,15 @@ class TimeIntBDFBase : public TimeIntBase
...
   get_previous_time(int const i /* t_{n-i} */) const;
 
+  /*
+   * Get current order of time integrator
+   */
+  unsigned int
+  get_current_order() const;
+
+  BDFTimeIntegratorConstants const &
+  get_bdf_constants() const;
```
nfehn, line 111: From the current PR, I do not see that these should be public.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.cpp
```diff
@@ -175,6 +175,59 @@ TimeIntBDF<dim, Number>::advance_one_timestep_partitioned_solve(bool const use_e
...
+  std::vector<double> &                                         times,
+  std::function<VectorType const *(unsigned int const)> const & get_quantity) const
+{
+  /*
+   *
+   *   time t
+   *  -------->   t_{n-2}   t_{n-1}   t_{n}     t_{n+1}
+   *  _______________|_________|________|___________|___________\
+   *                 |         |        |           |           /
+   *               sol[2]    sol[1]   sol[0]
```
nfehn, line 191: you could think about renaming `sol` -> `quantities`.

include/exadg/time_integration/time_int_bdf_base.h
jh66637: I agree on `get_current_order()`. This should be private. As mention in the description, I want to use this infrastructure to compute the temporal derivative of the pressure outside the `INSE` module. So I need `get_bdf_constants()` to stay public :)

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.cpp
jh66637: [54e19d5](https://github.com/exadg/exadg/pull/458/commits/54e19d55e77b5520afe808311d28e9403135757f)

include/exadg/time_integration/bdf_time_integration.h
```diff
@@ -79,10 +94,15 @@ compute_bdf_time_derivative(VectorType &                       derivative,
...
 {
-  derivative.equ(bdf.get_gamma0() / time_step_size, solution_np);
+  std::vector<VectorType const *> previous_solutions_ptrs(previous_solutions.size());
 
-  for(unsigned int i = 0; i < previous_solutions.size(); ++i)
-    derivative.add(-bdf.get_alpha(i) / time_step_size, previous_solutions[i]);
+  std::transform(previous_solutions.begin(),
+                 previous_solutions.end(),
+                 previous_solutions_ptrs.begin(),
+                 [](VectorType const & t) { return &t; });
```
nfehn, line 102: I wonder whether one could do the transform in the other direction to receive `std::vector<VectorType const &>` by looping over the ptrs vector and calling `previous_solution.push_back(...)`. I am thinking about this because I would prefer the function containing the actual implementation to use value-semantics. This would require that the assigment/type conversion `std::vector<VectorType const &> const & <-> std::vector<VectorType> const &` can be done automatically (I think so ...).

include/exadg/time_integration/time_int_bdf_base.h
nfehn: Hm, I feel this should not be public information. A main problem that I see is that outside the class one can not know whether the BDF constants are correct, because they are updated in case of adaptive time step sizes. Things are more encapsulated if this happens within the time integrator.

include/exadg/time_integration/bdf_time_integration.h
jh66637: It is not possible to have a `std::vector<T const&>`, see https://godbolt.org/z/8T4xx8Tdj. This is probably also the reason get_velocies_and_times returns `std::vector<VectorType const*>`.

include/exadg/time_integration/time_int_bdf_base.h
jh66637: Or is it OK to have `compute_dp_dt()` in the BDF time integrator of the INSE?

nfehn: Yes, this is what I would like to discuss: In terms of encapsulation, is it better (i) that the time integrator computes a temporal derivative when requested, or (ii) that the time integrator returns time integration constants (which is somewhat "private" information)?

I tend to prefer the first option. What do you think?

jh66637: I also prefer the first one.
Just one note: To always get the temporal derivatives correct, we have to do some more work. This is because pressure_np might already be pushed back or not. Therefore, I would like to keep track of the times in the same way we keep track of the solution vectors and check the order before computing the derivative. Is this OK for you?

nfehn: Yes, I think such a check is the advantage of the first option over the second one. Right now, it is not fully clear to me what you mean by "check the order".

I think one should pass the time t at which to evaluate the temporal derivative to that public member function of the time integrator, and then check inside that function whether this time t is equivalent to time t_np stored by the time integrator (up to some tolerance).
The assertion that time instants and solution vectors of the time integrator are in sync is an independent responsibility of the time integrator (not directly related to the functionality discussed here).

nfehn: So I am principally OK with this. If this requires fundamental changes of the time integrators, we should discuss beforehand whether to address these basic changes in a separate PR.

jh66637: > Yes, I think such a check is the advantage of the first option over the second one. Right now, it is not fully clear to me what you mean by "check the order".

With order, I mean sequence of solutions in the vector.

> I think one should pass the time t at which to evaluate the temporal derivative to that public member function of the time integrator, and then check inside that function whether this time t is equivalent to time t_np stored by the time integrator (up 
to some tolerance). 

Fully agree.

The assertion that time instants and solution vectors of the time integrator are in sync is an independent responsibility of the time integrator (not directly related to the functionality discussed here).

I would still like to do it in the PR if it is OK. The whole point, I need introduced code, is just to compute temporal pressure derivatives. Is it OK for you if this is also done in this PR?

jh66637: > So I am principally OK with this. If this requires fundamental changes of the time integrators, we should discuss beforehand whether to address these basic changes in a separate PR.

I think this only requires minor changes in the `INSE::BDFTimeIntegrator`.

jh66637: @nfehn To achieve this, I have to keep track of the time the same way we do it for the vector of solutions. Since `solution_np` is set in `do_timestep_solve()` also `time_history_np` should be set in this function. I added a draft in [385a46b](https://github.com/exadg/exadg/pull/458/commits/385a46b2d47f6f81a9ca3077fbeb4546c736b654). Can you relate to these changes? If yes, I will rebase and split the functions to header and source files.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_pressure_correction.cpp
```diff
@@ -315,6 +315,8 @@ template<int dim, typename Number>
 void
 TimeIntBDFPressureCorrection<dim, Number>::do_timestep_solve()
 {
+  this->time_history_np = this->get_next_time();
```
nfehn, line 318: 3 times the same change! See also comment https://github.com/exadg/exadg/pull/458/files#r1219198819.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.h
```diff
@@ -80,6 +80,53 @@ class TimeIntBDF : public TimeIntBDFBase<Number>
...
+      return times[i] > times[j];
+    });
+
+    // apply permutation to vector of quantities
+    std::vector<VectorType const *> unsorted_quantities = quantities;
+
+    std::transform(permutation.begin(), permutation.end(), quantities.begin(), [&](auto const i) {
+      return unsorted_quantities[i];
+    });
+  }
```
nfehn, line 114: This should not be public functionality of the time integrator. It should rather appear in the `Driver` class if this is needed (even though I still do not know the concrete reason for this ...).

The function `prepare_vectors_for_next_timestep()` is done in `do_timestep_post_solve()` or `advance_one_timestep_post_solve()`, right? Hence, one should in the `Driver` extract the pressure time derivative at the point of the algorithm where it is available. 

I don't like implementations where we kind of "post-correct" things because we do not want to think about a proper design.

nfehn:
I am so far not happy with the current design. I believe we have to rethink again ...

If one has the vector of `times`, one has everything to set up the `BDFConstants` and to compute a temporal derivative. So one option I consider quite elegant is to write a utility function `compute_time_derivative(derivative, solutions, times)` that internally sets up a `BDFConstants` object and then calls the already existing function. The computational overhead for recomputing the BDF constants is probably not visible at all in the multiphysics context.

jh66637:
In general, I am fine with having a utility function which does that. But firstly one needs the correct vectors which can be passed to the utility function. If you consider keeping track of the vector of times like I proposed, I would argue keeping track of the quantities as it is done is already a bug. This is because `get_times_and_quantities_np()` is only correct if called before pushing back the vectors (and the times and quantities are out of sync additionally). Therefore, to get the correct vectors and times (in the correct order) independent on the time of calling `get_times_and_quantities_np()`, we somehow have to keep track of the times as well.

nfehn:
What we need in my opinion is a discussion to which extent the time integrator interface with pre-solve, solve, and post-solve steps is sufficient. The circumstance that changing the order of certain instructions will invalidate the results is to be expected. What you describe above could call for additional checks in `get_quantities_and_times_np()`. However, this does not imply that one needs to - in my opinion redundantly - store time `t_np` in the time integrator.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.h
```diff
@@ -159,6 +219,10 @@ class TimeIntBDF : public TimeIntBDFBase<Number>
   // postprocessor
   std::shared_ptr<PostProcessorInterface<Number>> postprocessor;
 
+  // time history
+  double              time_history_np;
```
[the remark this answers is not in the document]
nfehn: note that I adjusted the above pseudo-code.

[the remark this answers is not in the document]
jh66637: Computing `get_quantities_and_times_np()` before do_timestep solve will pass the test and give wrong results (see below). In my opinion, the user should be able to obtain at least a vector of quantities and corresponding correct times, no matter when the function is called. 
```
[q^end, q^n, q^n-1, ..., q^end-1]
[t^n+1,  t^n,  t^n-1, ...,  t^end-1]
```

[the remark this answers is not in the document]
nfehn: I see it a bit differently. The function name `_np()` indicates that this function accesses post-solve information. There is no guarantee to have correct results before calculating the time step. Before solving, you can only call `get_quantities_and_times()` and expect correct results. We can also rename the function

`get_quantities_and_times_np() -> get_quantities_and_times_post_solve()`,

but it will remain a getter function, i.e. extracting information from a time integrator class that is in a certain state currently. I would argue time integration implementations generally suffer from the problem that you describe. However, I do not see stringent solutions that do not rely on any form of conventions.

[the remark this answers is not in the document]
jh66637: So you are saying exposing `BDFCoefficients` is not OK because it can be misused, but it is OK to have a function which might give information that is not just the information from the old time-step (which I would expect) but simply wrong?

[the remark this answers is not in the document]
nfehn: Ensuring that `solutions` and `times` are in sync is in my opinion clearer from a user perspective (and easier from an implementation perspective) than `solutions`, `times`, and `bdf_constants`. The reason is that `bdf_constants` are updated in `do_timestep_pre_solve()` 
https://github.com/exadg/exadg/blob/290dd00a2fb1a4cec698eae55ec3680379167262/include/exadg/time_integration/time_int_bdf_base.cpp#L208 

while `solutions` and `times` are updated in `do_timestep_post_solve()` 
https://github.com/exadg/exadg/blob/290dd00a2fb1a4cec698eae55ec3680379167262/include/exadg/time_integration/time_int_bdf_base.cpp#L217

Moreover, the function `get_solutions_and_times()` returns both solutions and times, while `get_bdf_constants()` would be an independent function for which it is somewhat difficult to ensure that this is consistent with `get_solutions_and_times()`.

Finally, the BDF coefficients live in the IncNS time integrator to compute the **velocity** time derivative. There is per se no guarantee in my opinion that the vector of pressure solutions fits to such coefficients (e.g. regarding the size of the vector / order of the finite difference approximation of the derivative). This is another argument to only return solution vectors and times (and compute the time derivative based on this data only, as suggested here https://github.com/exadg/exadg/pull/458#discussion_r1219424100). Mathematically, the BDF constants are redundant information in addition to the `times` vector. I think one can safely argue that this is indeed more dangerous than providing minimal / unique information.

Regarding your question: Yes, using a function not in a way as described by the name and the documentation is principally "dangerous", where "dangerous" means it does not do what it is described to do. One cannot expect a function with name `_np()` to provide correct data at time `np`, if `np` data has not been computed.

Overall, I feel the discussion got heated and therefore suggest a pause for now.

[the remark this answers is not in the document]
jh66637: @nfehn I still think the data returned should be at least in sync, but let's proceed as you described and discuss this at a different point. The implemented functionality works, as long as times and solutions are in sync (independent who cares for it). Just let me add a test because the functionality is currently not used in the code. I will ping you when I am ready.

nfehn:
Looks like very clean code now!

jh66637:
@nfehn I added tests and am ready if the pipeline doesn't fail :)

tests/time_integration/bdf_time_derivative_01.cc
```diff
@@ -0,0 +1,119 @@
...
+
+  // check if derivative is 0
+  assert(derivative[0] < 1e-12);
+  std::cout << "OK" << std::endl;
+}
+
+void
+test_linear_changing_quantites(unsigned int const order)
+{
+  std::cout << "Check derivative of linear changing quantitites" << std::endl;
```
nfehn, line 74: -> quantities

tests/time_integration/bdf_time_derivative_01.cc
```diff
@@ -0,0 +1,119 @@
...
+
+// Check BDF time derivative given times and quantities
+
+using namespace ExaDG;
+using VectorType = dealii::LinearAlgebra::Vector<double>;
+
+void
+test_constant_quantites(unsigned int const order)
+{
+  std::cout << "Check derivative of constant quantitites" << std::endl;
```
nfehn, line 37: -> quantities

tests/time_integration/bdf_time_derivative_01.cc
```diff
@@ -0,0 +1,119 @@
...
+  assert(derivative[0] < 1e-12);
+  std::cout << "OK" << std::endl;
+}
+
+void
+test_linear_changing_quantites(unsigned int const order)
+{
+  std::cout << "Check derivative of linear changing quantitites" << std::endl;
+
+  // construct qunatities with slope -1
```
nfehn, line 76: -> quantities

Now, we have all I think :)

tests/time_integration/bdf_time_derivative_01.cc
```diff
@@ -0,0 +1,119 @@
...
+  std::transform(constant_quantities_np_values.begin(),
+                 constant_quantities_np_values.end(),
+                 constant_quantities_np.begin(),
+                 [](VectorType const & t) { return &t; });
+
+
+  // construct times
+  std::vector<double> times_np(order + 1);
+  std::iota(times_np.begin(), times_np.end(), 0.0);
+  std::sort(times_np.begin(), times_np.end(), [](auto const a, auto const b) { return a > b; });
```
nfehn, line 95: I do not understand sorting the times, but not sorting the vectors. The above comment "construct qunatities with slope -1" does not make much sense without further information if one later sorts in my opinion, because the slope does not only depend on the function values, but also on the time values.

This goes in a similar direction as the other discussions that we had. I do not see much value over constructing the times vector from the very beginning as intended.

tests/time_integration/bdf_time_derivative_01.cc
```diff
@@ -0,0 +1,119 @@
...
+
+#include <exadg/time_integration/bdf_constants.h>
+
+// Check BDF time derivative given times and quantities
+
+using namespace ExaDG;
+using VectorType = dealii::LinearAlgebra::Vector<double>;
+
+void
+test_constant_quantites(unsigned int const order)
```
nfehn, line 35: -> quantities

tests/time_integration/bdf_time_derivative_01.cc
```diff
@@ -0,0 +1,119 @@
...
+  std::transform(constant_quantities_np_values.begin(),
+                 constant_quantities_np_values.end(),
+                 constant_quantities_np.begin(),
+                 [](VectorType const & t) { return &t; });
+
+
+  // construct times
+  std::vector<double> times_np(order + 1);
+  std::iota(times_np.begin(), times_np.end(), 0.0);
+  std::sort(times_np.begin(), times_np.end(), [](auto const a, auto const b) { return a > b; });
```
nfehn, line 58: same as other comment

tests/time_integration/bdf_time_derivative_01.cc
```diff
@@ -0,0 +1,119 @@
...
+  std::cout << "OK" << std::endl;
+}
+
+int
+main()
+{
+  for(unsigned int order = 1; order <= 4; ++order)
+  {
+    std::cout << "Order " << order << std::endl;
+    test_constant_quantites(order);
```
nfehn, line 115: This tests in not more than checking that the sum of all bdf-constants is zero, as introduced in #474. Since the function call `compute_bdf_derivative()` is tested with the second test below, I question whether we need this test. I also do not like that the two functions copy so many lines of code. This implies unnecessary maintenance effort for the future.

nfehn (changes requested):

tests/time_integration/bdf_time_derivative_01.cc
jh66637: [e7c9fe3](https://github.com/exadg/exadg/pull/458/commits/e7c9fe31d7f63ff8ab425c4b392fd36f4dd4f637)

tests/time_integration/bdf_time_derivative_01.cc
```diff
@@ -0,0 +1,119 @@
...
+
+  compute_bdf_time_derivative(derivative, constant_quantities_np, times_np);
+
+  // check if derivative is 0
+  assert(derivative[0] < 1e-12);
+  std::cout << "OK" << std::endl;
+}
+
+void
+test_linear_changing_quantites(unsigned int const order)
```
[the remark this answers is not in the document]
jh66637: [e7c9fe3](https://github.com/exadg/exadg/pull/458/commits/e7c9fe31d7f63ff8ab425c4b392fd36f4dd4f637)

tests/time_integration/bdf_time_derivative_01.cc
jh66637: [e7c9fe3](https://github.com/exadg/exadg/pull/458/commits/e7c9fe31d7f63ff8ab425c4b392fd36f4dd4f637)

tests/time_integration/bdf_time_derivative_01.cc
jh66637: It uses the different interface which constructs the BDF object internally. I think we should test it. In my opinion, the more we test, the better.

It is a test and should not be touched, so maintenance will not be a problem. I will adapt it, but I really think we should keep both tests.

tests/time_integration/bdf_time_derivative_01.cc
```diff
@@ -0,0 +1,107 @@
...
+
+  double target_slope = -99.0;
+  if(shape == Shape::Linear)
+  {
+    // construct quantities with slope -1
+    target_slope = -1.0;
+    for(unsigned int i = 0; i < order + 1; ++i)
+    {
+      VectorType vec(1);
+      vec[0] = i;
```
nfehn, line 57: In my opinion, this line has to make use of `target_slope` for a reasonable test, e.g.

```c++
vec[0] = 1 + double(i)*target_slope;
````

Then, all that remains in the if-else branch is setting the `target_slope`.

File: tests/time_integration/bdf_time_derivative_01.cc [hunk 27]
```diff
@@ -0,0 +1,107 @@
...
+  std::transform(quantities_np_values.begin(),
+                 quantities_np_values.end(),
+                 quantities_np.begin(),
+                 [](VectorType const & t) { return &t; });
+
+
+  // construct times which are in reverse order in realistic applications
+  std::vector<double> times_np(order + 1);
+  std::iota(times_np.begin(), times_np.end(), 0.0);
+  std::sort(times_np.begin(), times_np.end(), [](auto const a, auto const b) { return a > b; });
```
nfehn, line 84: For me it looks like you want to  use `std.:iota/sort` functionality. Filling `times_np` in the same loop as `quantities_np` like `times_np[i] = (1-i)*dt;` with a comment "// times vector containing [dt, 0, -dt, ..., (order-1)*dt]" would be better suited for the purpose of testing. By looking at the code, I cannot see what the code does or should do.

tests/time_integration/bdf_time_derivative_01.cc
```diff
@@ -0,0 +1,107 @@
...
+
+
+  // compute temporal derivative
+  VectorType derivative(1);
+  derivative[0] = -99.0;
+  compute_bdf_time_derivative(derivative, quantities_np, times_np);
+
+  // check if computed slope equals target slope
+  assert(std::abs(derivative[0] - target_slope) < 1e-12);
```
nfehn, line 93: Since `target_slope` is not used in the computations above, I would argue this test does not ensure that the implementation is correct, but only that the same results are obtained as when writing the test.

nfehn (changes requested):

tests/time_integration/bdf_time_derivative_01.cc
nfehn: > It is a test and should not be touched, so maintenance will not be a problem.

I don't think this is correct. I will remember having you as a volunteer to adjust tests for whatever reasons due to code changes in the future :). 
But seriously: A software needs continuous updates and refactoring. If one has to adjust 10 tests rather than 2, the motivation to make changes to a code will decrease a lot. I am convinced that one should not treat writing of tests shabbily for such reasons.

tests/time_integration/bdf_time_derivative_01.cc
jh66637: I think that is a matter of taste. I prefer to avoid raw for loops and use the STL instead. Anyway, I restructured again.

nfehn: Regarding the philosophical discussion, it seems as if you want to use `std::generate` instead of iota + sort (with non-default ordering). Regarding my personal preferences: I am not a fan of glorification ...

nfehn (approved):

PR #461, opened 2023-06-01
allow to compute spatially weighted errors

jh66637:
This adds the possibility to restrict error computation only to parts of the domain


include/exadg/postprocessor/error_calculation.h
```diff
@@ -78,6 +79,10 @@ struct ErrorCalculationData
   // write errors to file?
   bool write_errors_to_file;
 
+  // If set a spatially weighted norm is computed. This can be used to compute the error only over
```
nfehn, line 82: Grammatically, I think one would need a comma here: "If set, ...". 

However, I am against such implicit logic (i.e. doing things depending on whether a point is set or not). In ExaDG, the general procedure is / should be to explicitly activate postprocessing functionality, e.g. in order to easily switch between weighted and not-weighted error computation (not matter whether dealii is explicit or implicit about this). Moreover, one should not forget the documentation-purpose of such boolean variables. It is in my opinion much clearer or more intuitive to see what a code does if the user has to set such a boolean variable. One can then also assert in the code that in case of `true`, there must be a valid pointer. Note that `false` and an invalid pointer is not the same in practice. Namely, an invalid pointer might not originate from the intention to not use weighting, but from bugs in the implementation. Imagine writing a function with the intention to initialize a shared pointer to `dealii::Function`, where the user writes this function as receiving the shared_ptr as value and not as reference. We had this type of bugs in ExaDG ... My suggestion with boolean + asserts in the code is safer in my opinion and gives hints to a user where the mistake might have happened.

nfehn:
This PR has currently failing checks and conflicting files.

jh66637:
@nfehn I added an additional `bool` as suggested. The pipeline will fail due to https://github.com/exadg/exadg/pull/584. I will trigger another pipeline run when this problem is resolved.

nfehn (approved):

PR #463, opened 2023-06-01
add pointwise output generator for incompressible navier stokes

jh66637:
I added the `PointwiseOutputGenerator` (`POG`) to the `IncNS`. Since in the post-processor the solution is not present as a `Vector` or `BlockVector` I added initialization for `POGBase` with `std::vector<VectorType const *>` to ensure it can be used in every situation.

### DIFF
--- a/applications/incompressible_navier_stokes/turbulent_channel/application.h
+++ b/applications/incompressible_navier_stokes/turbulent_channel/application.h
@@ -499,6 +499,24 @@ private:
     pp_data.mass_data.filename               = this->output_parameters.filename;
     pp_data.mass_data.reference_length_scale = 1.0;
 
+    // pointwise output
+    pp_data.pointwise_output_data.time_control_data.is_active  = false;
+    pp_data.pointwise_output_data.time_control_data.start_time = START_TIME;
+    pp_data.pointwise_output_data.time_control_data.end_time   = END_TIME;
+    pp_data.pointwise_output_data.time_control_data.trigger_interval =
+      (END_TIME - START_TIME) / 1000.0;
+    pp_data.pointwise_output_data.directory =
+      this->output_parameters.directory + "pointwise_output/";
+    pp_data.pointwise_output_data.filename       = this->output_parameters.filename;
+    pp_data.pointwise_output_data.write_velocity = true;
+    pp_data.pointwise_output_data.write_pressure = true;
+    pp_data.pointwise_output_data.update_points_before_evaluation = false;
+
+    dealii::Point<dim> evaluation_point;
+    evaluation_point[0] = 0.5 * DIMENSIONS_X1;
+    evaluation_point[1] = 0.5 * DIMENSIONS_X2;
+    pp_data.pointwise_output_data.evaluation_points.push_back(evaluation_point);
+
     MyPostProcessorData<dim> pp_data_turb_ch;
     pp_data_turb_ch.pp_data = pp_data;
 
--- a/include/exadg/incompressible_navier_stokes/postprocessor/pointwise_output_generator.cpp
+++ b/include/exadg/incompressible_navier_stokes/postprocessor/pointwise_output_generator.cpp
@@ -2,7 +2,7 @@
  *
  *  ExaDG - High-Order Discontinuous Galerkin for the Exa-Scale
  *
- *  Copyright (C) 2021 by the ExaDG authors
+ *  Copyright (C) 2023 by the ExaDG authors
  *
  *  This program is free software: you can redistribute it and/or modify
  *  it under the terms of the GNU General Public License as published by
@@ -19,18 +19,15 @@
  *  ______________________________________________________________________
  */
 
-
-// ExaDG
-#include <exadg/compressible_navier_stokes/postprocessor/pointwise_output_generator.h>
+#include <exadg/incompressible_navier_stokes/postprocessor/pointwise_output_generator.h>
 #include <exadg/utilities/print_functions.h>
 
 namespace ExaDG
 {
-namespace CompNS
+namespace IncNS
 {
 template<int dim>
-PointwiseOutputData<dim>::PointwiseOutputData()
-  : write_rho(false), write_rho_u(false), write_rho_E(false)
+PointwiseOutputData<dim>::PointwiseOutputData() : write_velocity(false), write_pressure(false)
 {
 }
 
@@ -40,11 +37,10 @@ PointwiseOutputData<dim>::print(dealii::ConditionalOStream & pcout) const
 {
   PointwiseOutputDataBase<dim>::print(pcout);
 
-  if(this->time_control_data.is_active and this->evaluation_points.size() > 0)
+  if(this->time_control_data.is_active && this->evaluation_points.size() > 0)
   {
-    print_parameter(pcout, "Write rho", write_rho);
-    print_parameter(pcout, "Write rho_u", write_rho_u);
-    print_parameter(pcout, "Write rho_E", write_rho_E);
+    print_parameter(pcout, "Write velocity", write_velocity);
+    print_parameter(pcout, "Write pressure", write_pressure);
   }
 }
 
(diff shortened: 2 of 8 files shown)


jh66637:
I accidentally closed this during rebase.

richardschu:
> I accidentally closed this during rebase.

happens to the best of us! :smile:

applications/incompressible_navier_stokes/turbulent_channel/application.h
```diff
@@ -485,6 +485,25 @@ class Application : public ApplicationBase<dim, Number>
...
+    pp_data.pointwise_output_data.filename       = this->output_parameters.filename;
+    pp_data.pointwise_output_data.write_velocity = true;
+    pp_data.pointwise_output_data.write_pressure = true;
+    pp_data.pointwise_output_data.update_points_before_evaluation = false;
+    if constexpr(dim == 2)
+      pp_data.pointwise_output_data.evaluation_points.emplace_back(
+        dealii::Point<dim>{0.5 * DIMENSIONS_X1, 0.5 * DIMENSIONS_X2});
+    if constexpr(dim == 3)
+      pp_data.pointwise_output_data.evaluation_points.emplace_back(
+        dealii::Point<dim>{0.5 * DIMENSIONS_X1, 0.5 * DIMENSIONS_X2, 0.0});
```
nfehn, line 505: Why do you write `constexpr`? I think one should fill `dealii::Point<dim>` depending on `dim`, but then call `emplace_back()` only once.

include/exadg/postprocessor/pointwise_output_generator_base.cpp
```diff
@@ -58,11 +58,11 @@ PointwiseOutputDataBase<dim>::print(dealii::ConditionalOStream & pcout) const
 template struct PointwiseOutputDataBase<2>;
 template struct PointwiseOutputDataBase<3>;
 
-template<int dim, typename VectorType>
+template<int dim, typename VectorType, typename Number>
```
nfehn, line 61: why do yo introduce another template parameter here?

Considering the instantiations below, it looks like `VectorType::value_type` and `Number` are always the same.

include/exadg/postprocessor/pointwise_output_generator_base.cpp
```diff
@@ -280,21 +282,28 @@ PointwiseOutputGeneratorBase<dim, VectorType>::write_time(double time)
...
-template class PointwiseOutputGeneratorBase<3, BlockVectorType<double>>;
+template class PointwiseOutputGeneratorBase<3, BlockVectorType<float>, float>;
+template class PointwiseOutputGeneratorBase<3, BlockVectorType<double>, double>;
+
+// Instantiations for arbitrary collection of pointers to VectorTypes
+template class PointwiseOutputGeneratorBase<2, std::vector<VectorType<float> const *>, float>;
+template class PointwiseOutputGeneratorBase<2, std::vector<VectorType<double> const *>, double>;
+
+template class PointwiseOutputGeneratorBase<3, std::vector<VectorType<float> const *>, float>;
+template class PointwiseOutputGeneratorBase<3, std::vector<VectorType<double> const *>, double>;
```
nfehn, line 306: I consider it impractical that different vector versions show up as different instantiations of this class. It looks as if only the `evaluate()` member function depends on `VectorType`, so only this member function should be templated.

include/exadg/postprocessor/pointwise_output_generator_base.cpp
```diff
@@ -58,11 +58,11 @@ PointwiseOutputDataBase<dim>::print(dealii::ConditionalOStream & pcout) const
 template struct PointwiseOutputDataBase<2>;
 template struct PointwiseOutputDataBase<3>;
 
-template<int dim, typename VectorType>
+template<int dim, typename VectorType, typename Number>
 void
-PointwiseOutputGeneratorBase<dim, VectorType>::evaluate(VectorType const & solution,
-                                                        double const       time,
-                                                        bool const         unsteady)
+PointwiseOutputGeneratorBase<dim, VectorType, Number>::evaluate(VectorType const & solution,
```
nfehn, line 63: It looks like `PointwiseOutputGeneratorBase` does not need the `VectorType` at all, since the vector is only passed to `do_evaluate()`, but this function is purely virtual in the base class.

The circumstance that we want to be flexible regarding the VectorType and that the vectors might not always be available/accessible (see description of PR) calls for another design in  my opinion. Passing a lambda without any arguments or templates for dof-vector to the `evaluate()` function seems more natural.

The function `POGBase::evaluate()` would become protected, and concrete derived classes would implement a public function `evaluate()` for concrete vector types, which then first define the lambda and afterwards call `POGBase::evaluate(lambda)`.

applications/incompressible_navier_stokes/turbulent_channel/application.h
```diff
@@ -485,6 +485,25 @@ class Application : public ApplicationBase<dim, Number>
     pp_data.mass_data.filename               = this->output_parameters.filename;
     pp_data.mass_data.reference_length_scale = 1.0;
 
+    // pointwise output
+    pp_data.pointwise_output_data.time_control_data.is_active  = true;
```
nfehn, line 489: I think this should not be active by default.

nfehn:
any other opinions on this topic apart from my comments?

include/exadg/postprocessor/pointwise_output_generator_base.cpp
jh66637: It's sadly not possible to template `evaluate()` with the current design, since `evaluate()` is virtual.

nfehn: ok. I still believe we should strive for a solution where the vector type is not a template. I'll make another comment elsewhere.

include/exadg/postprocessor/pointwise_output_generator_base.cpp
nfehn: In `StatisticsManager` (https://github.com/exadg/exadg/blob/265375b649a615ffea5e4652168b801145f72c09/include/exadg/postprocessor/statistics_manager.h#L114-L121) we have defined different `evaluate_functions()` differing in the argument. Maybe such a design could be realized here?

nfehn: The suggestion with the lambda refers to this function: https://github.com/exadg/exadg/blob/9227a66efec494a3a32ccde479e7305a00849fe8/include/exadg/postprocessor/pointwise_output_generator_base.h#L125

Instead of using the vector as argument in the different functions, we could use the function itself as an argument in order to hide the type of the vector. I think we only want to be able to call an "action", while it is not that important which arguments are needed for this action. Or do I misunderstand something?

jh66637:
@nfehn I will restructure as suggested :+1:

jh66637:
@nfehn In principle this is ready from my side. During restructuring I came across some errors which I fixed in https://github.com/exadg/exadg/pull/583.

Before reviewing this PR https://github.com/exadg/exadg/pull/583 has to be merged and I will rebase this PR to keep the diff short.

jh66637:
I rebased :)

include/exadg/incompressible_navier_stokes/postprocessor/pointwise_output_generator.cpp
```diff
@@ -0,0 +1,115 @@
...
+                                                double const       time,
+                                                bool const         unsteady)
+{
+  this->do_evaluate(
+    [&]() {
+      if(pointwise_output_data.write_velocity)
+      {
+        auto const values =
+          this->template compute_point_values<dim>(velocity, *dof_handler_velocity);
+        this->write_quantity("Velocity", values, 0);
```
nfehn, line 96: What is the meaning of 0 here? Does the number here have to be consisting with another number written somewhere else in the code? There is no number for the "pressure-call" below.

In similar cases, I often add a comment to the function call for clarity/documentation.

nfehn:
looks good overall

nfehn (approved):

include/exadg/incompressible_navier_stokes/postprocessor/pointwise_output_generator.cpp
jh66637: It is the first selected component. For the overload of `write_quantity` for scalar quantities, a first selected component is not considered.

I will add a comment.

PR #465, opened 2023-06-01
add rotation as mesh movement function

jh66637:
In some cases it is helpful to have a mesh movement function which just rotates the mesh.

### DIFF
--- a/include/exadg/grid/mesh_movement_functions.h
+++ b/include/exadg/grid/mesh_movement_functions.h
@@ -297,6 +297,48 @@ protected:
   double const                time_period;
 };
 
+
+template<int dim>
+struct RotationMeshMovementData
+{
+  RotationMeshMovementData() : counter_clockwise_degrees_per_s(0.0)
+  {
+  }
+
+  double counter_clockwise_degrees_per_s;
+};
+
+template<int dim>
+class RotationMeshMovementFunctions : public dealii::Function<dim>
+{
+public:
+  RotationMeshMovementFunctions(RotationMeshMovementData<dim> const & data)
+    : dealii::Function<dim>(dim),
+      angular_velocity{dealii::numbers::PI / 180. * data.counter_clockwise_degrees_per_s}
+  {
+  }
+
+  double
+  value(dealii::Point<dim> const & x, unsigned int const coordinate_direction = 0) const override
+  {
+    double displacement = 0.0;
+
+    double const phi = std::atan2(x(1), x(0)) + this->get_time() * angular_velocity;
+    double const r   = std::sqrt(x(0) * x(0) + x(1) * x(1));
+    if(coordinate_direction == 0)
+      displacement = -r * std::cos(phi) - x(0);
+    else if(coordinate_direction == 1)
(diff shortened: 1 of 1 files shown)


include/exadg/grid/mesh_movement_functions.h
```diff
@@ -297,6 +299,48 @@ class RectangleMeshMovementFunctions : public dealii::Function<dim>
...
 
+
+template<int dim>
+struct RotationMeshMovementData
+{
+  RotationMeshMovementData() : counter_clockwise_degrees_per_s(0.0)
+  {
+  }
+
+  double counter_clockwise_degrees_per_s;
```
nfehn, line 310: counter_clockwise is not a good name. If the rotation vector points in positive z/x3-direction, the value should be positive (and else negative).

Please add some documentation.

To extend the usability of this function, the axis around which we rotate should become an additional parameter. It is reasonable in my opinion to limit this code to rotations around one of the dim coordinate axis through the origin.

include/exadg/grid/mesh_movement_functions.h
```diff
@@ -297,6 +299,48 @@ class RectangleMeshMovementFunctions : public dealii::Function<dim>
...
+};
+
+template<int dim>
```
nfehn, line 313: As a minimal documentation, we should briefly describe the motion like "rotation around positive z/x3-axis" (in the current code version). As written above, it would make sense to extend the functionality towards rotation around any of the coordinate axis.

include/exadg/grid/mesh_movement_functions.h
nfehn: for clarity, I would explicitly write `per_second`.

include/exadg/grid/mesh_movement_functions.h
```diff
@@ -297,6 +299,48 @@ class RectangleMeshMovementFunctions : public dealii::Function<dim>
...
+class RotationMeshMovementFunctions : public dealii::Function<dim>
```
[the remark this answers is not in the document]
nfehn: `GridTools::rotate()` only works on the `dealii::triangulation` I think. Hence, I fear we can not use this for our ALE applications, where we need to manipulate a (high-order) mapping.

I consider it useful to have such utility functions in ExaDG.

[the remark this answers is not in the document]
peterrum: I did not say that one would use `GridTools::rotate()` directly to move vertices; but all the logic for rotating around any axis is implemented internally there.

nfehn:
This PR has currently failing checks.

jh66637:
I think this is only needed by me and for me, it's not high priority. I will ping you when I found the time to work on this and mark the PR as a draft in the mean time.

nfehn:
Personally, I feel we do not necessarily need PRs of this priority/relevance as (active/inactive) PRs to the master of ExaDG, in order to keep the focus and keep the list of TODOs short.

jh66637:
OK, I will close this PR due to its low priority.

PR #382, opened 2023-03-12
Enum parameter handling

jh66637:
As discussed in https://github.com/exadg/exadg/pull/264 I am filing a new PR for parameter handling with enums. Personally, I like the proposed interface, which automatically generates the ```dealii::Selections```. Thus, possible parameters can be printed, and it is checked if a given parameter is valid.

This PR is dependent on https://github.com/exadg/exadg/pull/264 and can only be merged after.

If a parameter is spelled wrong in, e.g., a `*.json` file the error Message shows which parameters can be used. For the test the output looks as follows:
```
38: set EnumType  =   Any one of Undefined, AScoped, BScoped, CScoped
38: 
38: # Listing of Parameters
38: # ---------------------
38: set EnumType = Undefined
38: 
38: terminate called after throwing an instance of 'dealii::ParameterHandler::ExcValueDoesNotMatchPattern'
38:   what():  
38: --------------------------------------------------------
38: An error occurred in line <1202> of file </home/johannes/workspace/dealii/source/base/parameter_handler.cc> in function
38:     void dealii::ParameterHandler::set(const string&, const string&)
38: The violated condition was: 
38:     patterns[pattern_index]->match(new_value)
38: Additional information: 
38:     The string <Undefineda> does not match the given pattern <[Selection
38:     Undefined|AScoped|BScoped|CScoped ]>.
```

PR #467, opened 2023-06-02
use scopes for better readability

jh66637:
https://github.com/exadg/exadg/pull/382#discussion_r1214035302
Instead of fixing the indentation manually I think it would be better to avoid switching clang-format off. In my opinion, adding scopes as done in this PR has two advantages:

1) No need to manually care for indentation
2) Better readability


nfehn:
We can do this. However, this will probably cause conflicts regarding the enum-patterns changes, doesn't it? So I suggest to make this PR a draft for now and fix the indentation manually in PR #382 for now.

jh66637:
Since https://github.com/exadg/exadg/pull/382 is merged, I rebased and reopened this PR.

nfehn (approved):

PR #468, opened 2023-06-02
Fix bug in MG c-transfer

### LINKED ISSUE #453: LocalSmoothing multigrid broken for c-MG?
I realized that for the poisson/sine test case, `LocalSmoothing` multigrid for hex elements (distributed or fully-distributed tria) requires significantly more iterations for cph-coarsening than in Table 5.3 of my thesis (https://mediatum.ub.tum.de/doc/1601025/1601025.pdf), with now around 11 iterations instead of 5 (for polynomial degrees 2-5).  Interestingly, `GlobalCoarsening` MG instead requires only around 7 iterations, which is much better but still noticably higher than 5. It is further strange that `GlobalCoarsening` performs better here for interior penalty factor tau=1, given that issue #452 reports a lack of robustness for global-coarsening MG w.r.t. the interior penalty parameter.

Iteration counts seem plausible for e.g. h-MG or p-MG with iterations around 12-14 (for polynomial degrees 2-5) in case of `LocalSmoothing` (and slightly higher around 13-15 for `GlobalCoarsening`). Also for pure h-MG, `LocalSmoothing` and `GlobalCoarsening` should actually yield the same results, right?

@peterrum @kronbichler

kronbichler:
This fixes #453: We switched to `prolongate_and_add` some time ago, but forgot to add the contribution when using the auxiliary vector (that part is unnecessary, but we should optimize for the global-coarsening algorithm anyway).

### DIFF
--- a/include/exadg/solvers_and_preconditioners/multigrid/transfers/mg_transfer_c.cpp
+++ b/include/exadg/solvers_and_preconditioners/multigrid/transfers/mg_transfer_c.cpp
@@ -233,7 +233,8 @@ MGTransferC<dim, Number, VectorType, components>::prolongate_and_add(unsigned in
     case 14: do_prolongate_and_add<14>(dst, src); break;
     case 15: do_prolongate_and_add<15>(dst, src); break;
     default:
-      AssertThrow(false, dealii::ExcMessage("MGTransferC::prolongate() not implemented for this degree!"));
+      AssertThrow(false, dealii::ExcMessage("MGTransferC::prolongate_and_add() not "
+                                            "implemented for this degree!"));
       // clang-format on
   }
 }
--- a/include/exadg/solvers_and_preconditioners/multigrid/transfers/mg_transfer_c.cpp
+++ b/include/exadg/solvers_and_preconditioners/multigrid/transfers/mg_transfer_c.cpp
@@ -117,8 +117,9 @@ MGTransferC<dim, Number, VectorType, components>::do_restrict_and_add(VectorType
 template<int dim, typename Number, typename VectorType, int components>
 template<int degree>
 void
-MGTransferC<dim, Number, VectorType, components>::do_prolongate(VectorType &       dst,
-                                                                VectorType const & src) const
+MGTransferC<dim, Number, VectorType, components>::do_prolongate_and_add(
+  VectorType &       dst,
+  VectorType const & src) const
 {
   src.update_ghost_values();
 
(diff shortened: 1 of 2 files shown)


peterrum (approved):

nfehn:
Nice!

> that part is unnecessary, but we should optimize for the global-coarsening algorithm anyway

I did not fully understand the implications of this last sentence. Should we change something in the code (in addition to the present PR)?

kronbichler:
> I did not fully understand the implications of this last sentence.

Sorry, I was very imprecise. The point is that we use a copy of the vector that we write into, and then use `copy_locally_owned_range_from` (or `+=` now). That is unnecessary from a computational point of view, but we need it for technical reasons, because `FEEvaluation::distribute_local_to_global` checks that the vector partitioner is equivalent to the one for the (auxiliary) `MatrixFree` object. That is typically not the case as the vector will obey the partitioner of a `MatrixFree` object with face integrals (face ghosts), whereas the one for transfer does not have face ghosts. So we have an assertion that is too strict. If one wanted to improve this, one would avoid setting up a separate `MatrixFree` object and instead make sure to link the indices of cells in the two cases. At this point, the second part of my statement gets relevant, we should direct work into improving global coarsening and not local smoothing.

include/exadg/solvers_and_preconditioners/multigrid/transfers/mg_transfer_c.cpp
```diff
@@ -216,21 +217,21 @@ MGTransferC<dim, Number, VectorType, components>::prolongate_and_add(unsigned in
...
+    case  8: do_prolongate_and_add< 8>(dst, src); break;
+    case  9: do_prolongate_and_add< 9>(dst, src); break;
+    case 10: do_prolongate_and_add<10>(dst, src); break;
+    case 11: do_prolongate_and_add<11>(dst, src); break;
+    case 12: do_prolongate_and_add<12>(dst, src); break;
+    case 13: do_prolongate_and_add<13>(dst, src); break;
+    case 14: do_prolongate_and_add<14>(dst, src); break;
+    case 15: do_prolongate_and_add<15>(dst, src); break;
     default:
       AssertThrow(false, dealii::ExcMessage("MGTransferC::prolongate() not implemented for this degree!"));
```
nfehn, line 236: Just noticed that the message also does not use `_and_add`.

nfehn (approved):

PR #469, opened 2023-06-02
add time integration constants for adams methods

jh66637:
This PR only introduces the Time Stepping Constants needed for Adams methods and is a split-off of https://github.com/exadg/exadg/pull/261.
Note that I renamed `bdf_time_integration.h` to `bdf_constants.h`, since for me, it was hard to distinguish between `time_int_bdf` and `bdf_time_integration`.


nfehn:
> Note that I renamed bdf_time_integration.h to bdf_constants.h, since for me, it was hard to distinguish between time_int_bdf and bdf_time_integration.

Full agreement.

include/exadg/time_integration/ab_constants.h
```diff
@@ -0,0 +1,65 @@
...
+
+// deal.II
+#include <deal.II/base/conditional_ostream.h>
+
+// ExaDG
+#include <exadg/time_integration/time_integration_constants_base.h>
+
+namespace ExaDG
+{
+class ABTimeIntegratorConstants : public TimeIntegratorConstantsBase
```
nfehn, line 36: Here, we should explain what "AB" means (similar for AM (and BDF)).

include/exadg/time_integration/ab_constants.cpp
```diff
@@ -0,0 +1,158 @@
...
+// ExaDG
+#include <exadg/time_integration/ab_constants.h>
+
+namespace ExaDG
+{
+ABTimeIntegratorConstants::ABTimeIntegratorConstants(unsigned int const order,
+                                                     bool const         start_with_low_order)
+  : TimeIntegratorConstantsBase(order, start_with_low_order), alpha(order)
+{
+  // In case of  ABM it is possible to have AB of order 0
```
nfehn, line 34: Meaning of "ABM" is unclear. 

What do you mean by "order 0"?

include/exadg/time_integration/time_integration_constants_base.h
```diff
@@ -70,6 +71,19 @@ class TimeIntegratorConstantsBase
   print(dealii::ConditionalOStream & pcout) const = 0;
 
 protected:
+  /**
+   * Can be used to fill the unused components of the vector with zeros.
```
nfehn, line 75: Here, one could add the information that the last entries of the vector are zeroed out corresponding to time instants lagging behind most.

include/exadg/time_integration/time_integration_constants_base.h
```diff
@@ -70,6 +71,19 @@ class TimeIntegratorConstantsBase
...
+   * This is needed since current_order might be smaller than order, e.g.,
+   * when using start_with_low_order = true
+   */
+  void
+  zero_out_unused_constants(unsigned int const first_idx, std::vector<double> & constants)
```
nfehn, line 80: I do not like the name `first_idx`. The previous name `current_order` was more intuitive in my opinion, also regarding comment https://github.com/exadg/exadg/pull/469/files#r1217691459.

The naming `..._unused_constants()` is somewhat misleading. If I remember correctly, we zero out exactly for the reason that the constants are indeed "used", but that the respective entries of the vector should not have an effect in order to get correct results (which we realize by multiplication of a vector by 0). Hence, a naming like `switch_off_high_order_constants()` or `disable_high_order_constants()` would be much more intuitive in my opinion.

include/exadg/time_integration/am_constants.cpp
```diff
@@ -0,0 +1,159 @@
...
+    alpha[0] = 8.0 / 12.0;
+    alpha[1] = -1.0 / 12.0;
+  }
+  else if(current_order == 4) // AM 4
+  {
+    gamma0   = 9.0 / 24.0;
+    alpha[0] = 19.0 / 24.0;
+    alpha[1] = -5.0 / 24.0;
+    alpha[2] = 1.0 / 24.0;
+  }
```
nfehn, line 97: This way of implementation is not very nice I think (which I probably introduced myself ...). I think the motivation was to write the assert only once and not in several functions (like the `set_...()` functions). I think a `switch` with `default` branch (calling a private member function `assert_order()`) would be better.

I make this comment here as a reminder for future work. No need to address this in the present PR.

nfehn:
As a common effort, we should work on introducing tests for general/basic functionality such as time integrator constants.

Regarding such tests, I am thinking of checking consistency (like the property that certain time integrator constants should sum up to zero) or that the temporal derivative / integral is computed exactly for an analytical polynomial function of a certain order.

include/exadg/time_integration/ab_constants.cpp
jh66637: Adams--Bashforth--Moulton predictor-corrector methods are also called ABM. I will add a comment.

Order 0 means, that the prediction (usually obtained from AB) is the solution of the last time step. To be able to write generic code, I would like to be able to initialize the constants with `oder=0` (which means, that no coefficients are managed). It is hard to explain this with just the constants. If it is OK for you, I would keep it this way and comment on this in a follow-up PR in which the actual time integrators are introduced.

include/exadg/time_integration/am_constants.cpp
jh66637: I would also be fine to change it within the PR. Just let me know which way you prefer :)

nfehn: I am certainly happy if you want to care about such a refactoring!

include/exadg/time_integration/ab_constants.h
jh66637: [0e3277c](https://github.com/exadg/exadg/pull/469/commits/0e3277c19ced8ac6f66980f5bbfc840be8bdf672)

include/exadg/time_integration/time_integration_constants_base.h
jh66637: [a63b519](https://github.com/exadg/exadg/pull/469/commits/a63b51958000e5041f91ccbb314b126ec95e8b48)

include/exadg/time_integration/time_integration_constants_base.h
jh66637: [a63b519](https://github.com/exadg/exadg/pull/469/commits/a63b51958000e5041f91ccbb314b126ec95e8b48)

include/exadg/time_integration/am_constants.cpp
jh66637: [ac5015c](https://github.com/exadg/exadg/pull/469/commits/ac5015c1a7c94f70ddf90a8d4c9a9bbc7f72acf0)

jh66637: I don't think we need `assert_order()`. Arriving at the default `assert_order()` would always trigger. I just placed a standard assert in the default branch.

jh66637:
@nfehn I added tests for AB and AM which sum up over the time integration constants. I think it would be great to have additional tests which check if a polynomial can be represented exactly and have tests for BDF. Is it OK to add those later?

tests/time_integration/ab_constants_02.cc
```diff
@@ -0,0 +1,59 @@
...
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *  ______________________________________________________________________
+ */
+
+#include <iostream>
+
+#include <exadg/time_integration/ab_constants.h>
+
+// Check if AB time integration constants in case of adaptive timestepping
```
nfehn, line 26: "if" is too much?

tests/time_integration/am_constants_01.cc
```diff
@@ -0,0 +1,58 @@
...
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *  ______________________________________________________________________
+ */
+
+#include <iostream>
+
+#include <exadg/time_integration/am_constants.h>
+
+// Check if AM time integration constants in case of adaptive timestepping
```
nfehn, line 26: "if"

nfehn: adaptive?

tests/time_integration/am_constants_02.cc
```diff
@@ -0,0 +1,59 @@
...
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *  ______________________________________________________________________
+ */
+
+#include <iostream>
+
+#include <exadg/time_integration/am_constants.h>
+
+// Check if AM time integration constants in case of adaptive timestepping
```
nfehn, line 26: see above comments

tests/time_integration/ab_constants_01.cc
```diff
@@ -0,0 +1,58 @@
...
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *  ______________________________________________________________________
+ */
+
+#include <iostream>
+
+#include <exadg/time_integration/ab_constants.h>
+
+// Check if AB time integration constants in case of adaptive timestepping
```
nfehn, line 26: adaptive?

nfehn:
Thanks for adding the tests!

As a general comment: I do not understand why git doesn't check that one file is just renamed. For the future, we should try to avoid this.

nfehn:
The problem is that, formally, we changed the implementation of BDF methods in this PR, so we might have introduced a bug during "renaming the file" that his hard to find.

tests/time_integration/ab_constants_01.cc
jh66637: [0ff6c74](https://github.com/exadg/exadg/pull/469/commits/0ff6c7473ea0718357e4456919e541b6eb516f7c)

tests/time_integration/am_constants_02.cc
jh66637: [0ff6c74](https://github.com/exadg/exadg/pull/469/commits/0ff6c7473ea0718357e4456919e541b6eb516f7c)

tests/time_integration/am_constants_01.cc
jh66637: [0ff6c74](https://github.com/exadg/exadg/pull/469/commits/0ff6c7473ea0718357e4456919e541b6eb516f7c)

tests/time_integration/ab_constants_02.cc
jh66637: [0ff6c74](https://github.com/exadg/exadg/pull/469/commits/0ff6c7473ea0718357e4456919e541b6eb516f7c)

jh66637:
I agree. This is strange. My local git client displays the file as renamed...

tests/time_integration/am_constants_01.cc
```diff
@@ -0,0 +1,58 @@
...
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *  ______________________________________________________________________
+ */
+
+#include <iostream>
+
+#include <exadg/time_integration/am_constants.h>
+
+// Check AM time integration constants in case of adaptive timestepping
```
nfehn, line 26: adaptive?

jh66637: Sorry. [9e24f5b](https://github.com/exadg/exadg/pull/469/commits/9e24f5bc344aeb90c4ccf005defab3480521079e)

nfehn:
did you rename the .cpp file into the .h file? (It looks like this when looking at the commits.)

jh66637:
I am not quite sure. If so, it was by mistake while trying to fix the git displaying the file as renamed.

jh66637:
I don't know how to fix it, I tried undo renaming and using again using `git mv ` on the command line... Locally, everything is displayed correctly.
For what it's worth: Only the #include changed in the file, which is marked as deleted...

nfehn:
Could you go back to the old name and let PR #473 do the rest?

jh66637:
Done. Note that compilation will now fail because the #include can not be found.

nfehn:
I thought we merge the present PR first?

jh66637:
Yes, but If the follow-up PR is applied on to as is, it has to be merged like this. Otherwise, the include has to be adapted in the follow-up PR.

nfehn:
Yes, I will adapt this later.

include/exadg/time_integration/bdf_time_integration.cpp
```diff
@@ -61,127 +61,111 @@ BDFTimeIntegratorConstants::get_alpha(unsigned int const i) const
...
+                 ((time_steps[0] + time_steps[1] + time_steps[2]) *
+                  (time_steps[1] + time_steps[2]) * time_steps[2]);
+      break;
+    }
+    case 4:
+    {
+      gamma0 = 1.0 + time_steps[0] / (time_steps[0] + time_steps[1]) +
+               time_steps[0] / (time_steps[0] + time_steps[1] + time_steps[2]) +
+               time_steps[0] / (time_steps[0] + time_steps[1] + time_steps[2] + time_steps[3]);
+      alpha[0] = (time_steps[0] + time_steps[1]) * (time_steps[0] + time_steps[1] + time_steps[2]) *
```
nfehn, line 142: Why did you remove this line for example? (and others)

I am not happy with how we do this currently. I am investing way to much effort into the code review of such a PR. We need to find a better way between refactoring and new content. I thought it would be possible to make these refactorings by only highlighting what has been changed, not all the old implementation showing up as new code.

jh66637: I am also not happy how this is displayed, but there is nothing I can do about it. If you want to, I can reset the file to its old state.

PR #471, opened 2023-06-02
remove template parameter from BDF Time Integrator

jh66637:
The template Parameter in `BDFTimeIntegratorBase` was only used to define `VectorType`. The concepts in the class do not change independent of the `VectorType`, so I think we should not have the template parameter in the base class.


nfehn:
Sounds reasonable. Thanks for finding this out!

nfehn (approved):

PR #473, opened 2023-06-05
rename file for BDF constants

nfehn:
follow up to #469


jh66637:
Should `extrapolation_scheme.h` also be renamed to `extrapolation_constants.h` to be consistent? I did not think about that before.

nfehn:
yes, I will rename these files as well

PR #475, opened 2023-06-05
Weak linear damping structure

richardschu:
This first PR adds mass-proportional or so-called weak (linear) damping to the structure balance of linear momentum. We add `integral( test_function * coefficient * D/Dt(displacement) )`. Drawing from the getters and setters for the mass operator scaling, it suffices to provide a rhs vector accounting for the constant remainder added to the rhs. Hence, I renamed the `get_scaling_factor_mass` and related functions, such that it is clear if it refers to the remainder from the acceleration or velocity term.


applications/structure/bar/tests/weak_damping.json
```diff
@@ -0,0 +1,34 @@
...
+    "TemporalResolution": {
+        "RefineTimeMin": "0",
+        "RefineTimeMax": "0"
+    },
+    "Application": {
+        "Length": "100.0",
+        "Height": "10.0",
+        "Width": "10.0",
+        "ProblemType": "Unsteady",
+        "WeakDamping": "10000.0",
```
richardschu, line 22: note that this is an absurdly high damping coefficient. One would use 0.01 insted. But we (i) test preconditioner performance wrt damping coefficient and (ii) it gives some visible effect using the Neumann BC (= visible lag in the displacement).

include/exadg/structure/time_integration/time_int_gen_alpha.cpp
```diff
@@ -124,7 +124,24 @@ TimeIntGenAlpha<dim, Number>::do_timestep_solve()
...
+                                                    velocity_n,
+                                                    acceleration_n);
+  if(this->param.weak_damping_coefficient > 1e-20 * this->param.density)
+  {
+    this->compute_const_vector_velocity_remainder(const_vector /* placeholder */,
+                                                  displacement_n,
+                                                  velocity_n,
+                                                  acceleration_n);
+    rhs.add(param.weak_damping_coefficient, const_vector);
+    const_vector = 0; // set to 0 since we apply_mass_operator
```
richardschu, line 141: `const_vector` would be zero if we do not add damping, so we do not need `const_vector=0;`
but with weak damping,  `const_vector` is filled (serves as temp container), then we add to the `rhs`. Thus, `const_vector=0;` is needed, right?

include/exadg/time_integration/time_int_gen_alpha_base.cpp
```diff
@@ -122,6 +130,23 @@ TimeIntGenAlphaBase<Number>::compute_const_vector(VectorType &       const_vecto
...
+  VectorType const & acceleration_n) const
+{
+  double const factor_dis = -get_scaling_factor_mass_from_velocity();
+  double const factor_vel = alpha_f - (1.0 - alpha_f) * (gamma - beta) / beta;
+  double const factor_acc = -(1.0 - alpha_f) * (gamma - 2.0 * beta) / (2.0 * beta) * time_step;
+
+  const_vector.equ(factor_dis, displacement_n);
+  const_vector.add(factor_vel, velocity_n);
+  const_vector.add(factor_acc, acceleration_n);
+}
```
richardschu, line 148: we rewrite the velocity term from the gen.-alpha scheme in terms of the velocity. d^{n+1} is our unknown and the remaining terms are collected on the rhs. This can be compared to the remaining terms from the acceleration rewritten in terms of the displacement d.

applications/structure/bar/tests/weak_damping.output
```diff
@@ -0,0 +1,625 @@
...
+
+________________________________________________________________________________
+
+ Time step number = 198     t = 9.85000e+01 -> t + dt = 9.90000e+01
+________________________________________________________________________________
+
+________________________________________________________________________________
+
+ Time step number = 200     t = 9.95000e+01 -> t + dt = 1.00000e+02
+________________________________________________________________________________
```
nfehn, line 622: Can you adjust the parameters so that this output information is not printed all the time?

nfehn:
You write two times "linear" in the issue description. I just want to make sure that nonlinear formulations (large displacements) are supported as well for the changes / new functionality proposed here?

include/exadg/structure/time_integration/time_int_gen_alpha.cpp
```diff
@@ -124,7 +124,24 @@ TimeIntGenAlpha<dim, Number>::do_timestep_solve()
   VectorType const_vector, rhs;
   const_vector.reinit(displacement_n);
   rhs.reinit(displacement_n);
-  this->compute_const_vector(rhs, displacement_n, velocity_n, acceleration_n);
+
+  // add const_vector resulting from acceleration and weak damping terms
+  double total_scaling_factor_mass = this->get_scaling_factor_mass_from_acceleration();
+  this->compute_const_vector_acceleration_remainder(rhs /* placeholder */,
```
nfehn, line 130: I do not understand the "placeholder" comment here (and below). The vector is changed inside the function call, so I would not say it is a "placeholder".

include/exadg/structure/time_integration/time_int_gen_alpha.cpp
```diff
@@ -124,7 +124,24 @@ TimeIntGenAlpha<dim, Number>::do_timestep_solve()
...
+  if(this->param.weak_damping_coefficient > 1e-20 * this->param.density)
+  {
+    this->compute_const_vector_velocity_remainder(const_vector /* placeholder */,
+                                                  displacement_n,
+                                                  velocity_n,
+                                                  acceleration_n);
+    rhs.add(param.weak_damping_coefficient, const_vector);
+    const_vector = 0; // set to 0 since we apply_mass_operator
+    total_scaling_factor_mass +=
+      param.weak_damping_coefficient * this->get_scaling_factor_mass_from_velocity();
```
nfehn, line 143: I would shift these lines upwards, because above you also first compute the scaling factor and then the vector.

include/exadg/time_integration/time_int_gen_alpha_base.h
```diff
@@ -53,17 +53,26 @@ class TimeIntGenAlphaBase : public TimeIntBase
   set_current_time_step_size(double const & time_step_size) final;
 
   double
-  get_scaling_factor_mass() const;
+  get_scaling_factor_mass_from_acceleration() const;
+
+  double
+  get_scaling_factor_mass_from_velocity() const;
```
nfehn, line 59: I do not see why to add a new function? Adding the two scaling factors appears twice in this PR, so this is duplicated code which would not occur when adding the two factors inside the pre-existing function `get_scaling_factor_mass()`.

include/exadg/time_integration/time_int_gen_alpha_base.h
```diff
@@ -53,17 +53,26 @@ class TimeIntGenAlphaBase : public TimeIntBase
...
+  compute_const_vector_acceleration_remainder(VectorType &       const_vector,
+                                              VectorType const & displacement_n,
+                                              VectorType const & velocity_n,
+                                              VectorType const & acceleration_n) const;
+
+  void
+  compute_const_vector_velocity_remainder(VectorType &       const_vector,
+                                          VectorType const & displacement_n,
+                                          VectorType const & velocity_n,
+                                          VectorType const & acceleration_n) const;
```
nfehn, line 75: Also here, I do not see the motivation to introduce a new function, which has the same interface as the pre-existing function. From a design perspective, it currently looks like combining the two functions into a single one would be more natural.

include/exadg/structure/user_interface/parameters.cpp
```diff
@@ -159,6 +160,7 @@ Parameters::print_parameters_physical_quantities(dealii::ConditionalOStream cons
   if(problem_type == ProblemType::Unsteady)
   {
     print_parameter(pcout, "Density", density);
+    print_parameter(pcout, "Weak damping coefficient", weak_damping_coefficient);
```
nfehn, line 163: In case nonlinear problems should not be supported right now, this parameter should only be printed in cases where it is relevant.

include/exadg/structure/user_interface/parameters.h
```diff
@@ -130,6 +130,9 @@ class Parameters
   // density rho_0 in initial configuration (only relevant for unsteady problems)
   double density;
 
+  // linear weak damping coefficient (mass proportional) for unsteady problems
+  double weak_damping_coefficient;
```
nfehn, line 134: In case nonlinear problems should not be supported right now, this needs to be documented here.

Further, an assert would need to be added to Parameters::check().

include/exadg/grid/mapping_deformation_structure.h
```diff
@@ -121,8 +121,12 @@ class DeformedMapping : public DeformedMappingBase<dim, Number>
         this->param.update_preconditioner &&
         time_step_number % this->param.update_preconditioner_every_time_steps == 0;
 
-      auto const iter = pde_operator->solve_nonlinear(
-        displacement, const_vector, 0.0 /* no mass term */, time, update_preconditioner);
+      auto const iter = pde_operator->solve_nonlinear(displacement,
+                                                      const_vector,
+                                                      0.0 /* no mass terms */,
+                                                      0.0 /* no mass terms */,
```
nfehn, line 127: why do you write two times "no mass terms"? I thought one of them refers to damping, so shouldn't one then write "no damping term" for one of the parameters?

include/exadg/structure/spatial_discretization/operator.cpp
```diff
@@ -272,6 +272,20 @@ Operator<dim, Number>::get_quad_index_gauss_lobatto() const
   return matrix_free_data->get_quad_index(get_quad_gauss_lobatto_name());
 }
 
+template<int dim, typename Number>
+double
+Operator<dim, Number>::compute_scaling_factor_mass(
+  double const scaling_factor_mass_from_acceleration,
+  double const scaling_factor_mass_from_velocity) const
```
nfehn, line 279: `scaling_factor_mass_from_velocity` gets multiplied by `weak_damping_coeficient`, so I am not sure if `scaling_factor_mass` is a good name. The product `param.weak_damping_coefficient * scaling_factor_mass_from_velocity` is in my opinion a scaling factor of the mass operator.

include/exadg/time_integration/time_int_gen_alpha_base.cpp
```diff
@@ -94,26 +94,34 @@ TimeIntGenAlphaBase<Number>::set_current_time_step_size(double const & time_step
...
-TimeIntGenAlphaBase<Number>::get_scaling_factor_mass() const
+TimeIntGenAlphaBase<Number>::get_scaling_factor_mass_from_acceleration() const
 {
-  return (1.0 - alpha_m) / ((1.0 - alpha_f) * beta * std::pow(time_step, 2.0));
+  return ((1.0 - alpha_m) / ((1.0 - alpha_f) * beta * std::pow(time_step, 2.0)));
+}
+
+template<typename Number>
+double
+TimeIntGenAlphaBase<Number>::get_scaling_factor_mass_from_velocity() const
```
nfehn, line 104: I think the time integrator should use the wording `mass_operator` and `damping_operator`. In case the damping operator is very simple, it is the `SpatialOperator` or `ElasticityOperatorBase` (where there is a method called `apply_damping_operator()`) that translates this into a simple mass scaling.

include/exadg/structure/time_integration/time_int_gen_alpha.cpp
```diff
@@ -120,13 +120,27 @@ TimeIntGenAlpha<dim, Number>::do_timestep_solve()
...
   const_vector.reinit(displacement_n);
-  rhs.reinit(displacement_n);
-  this->compute_const_vector(rhs, displacement_n, velocity_n, acceleration_n);
-  pde_operator->apply_mass_operator(const_vector, rhs);
+  pde_operator->apply_mass_operator(const_vector, tmp);
 
+  // add weak damping term remainder to rhs
+  if(this->param.weak_damping_active)
+  {
+    this->compute_const_vector_velocity_remainder(tmp, displacement_n, velocity_n, acceleration_n);
```
nfehn, line 138: Maybe also here one should change the naming towards `const_vector_acceleration_term` and `const_vector_damping_term`?

include/exadg/time_integration/time_int_gen_alpha_base.h
```diff
@@ -53,17 +53,26 @@ class TimeIntGenAlphaBase : public TimeIntBase
...
 
   void
-  compute_const_vector(VectorType &       const_vector,
-                       VectorType const & displacement_n,
-                       VectorType const & velocity_n,
-                       VectorType const & acceleration_n) const;
+  compute_const_vector_acceleration_remainder(VectorType &       const_vector,
+                                              VectorType const & displacement_n,
+                                              VectorType const & velocity_n,
+                                              VectorType const & acceleration_n) const;
```
nfehn, line 69: I think we should add comments here (and below) that these "const vectors" do not appear directly as vectors in the residual (nonlinear) or rhs (linear) of the system of equations, but that the mass operator (or damping operator below) still need to be applied to these vectors.

include/exadg/grid/mapping_deformation_structure.h
richardschu: It is because both acceleration and damping terms lead to a scaled mass matrix. I could put `/* scaling_factor_mass_from_acceleration */` and /* scaling_factor_mass_from_velocity */ to make it more clear? For the current implementation and usage, `factor acceleration` and `factor damping` in my opinion would also work. Which one do you prefer?

include/exadg/structure/spatial_discretization/operator.cpp
```diff
@@ -768,6 +786,15 @@ Operator<dim, Number>::apply_mass_operator(VectorType & dst, VectorType const &
   mass_operator.apply(dst, src);
 }
 
+template<int dim, typename Number>
+void
+Operator<dim, Number>::apply_add_weak_damping_operator(VectorType & dst, VectorType & src) const
+{
+  src *= param.weak_damping_coefficient;
```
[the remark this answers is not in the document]
richardschu: Adapted that! Never change `src`, I will keep that in mind :) ... makes sense tbh.

include/exadg/structure/spatial_discretization/operator.cpp
richardschu: I am not sure I get this comment -- you mean (i): the argument `scaling_factor_mass_from_velocity` has a misleading name since it is scaled with `weak_damping_coefficient` OR (ii) the local variable `scaling_factor_mass` should be renamed? My intention was to have one common scaling factor for the mass operator (hence `scaling_factor_mass` or maybe `total_...` or `final_...` or `summed_...`, which in my opinion is bit long and not really required).

At the same time, the arguments passed here originate from time integration only, they do not include any "physics" or any information regarding the operator. We get the scaling factors from the `time_integrator` via `time_integrator->get_scaling_factor_mass_from_acceleration()` and `time_integrator->get_scaling_factor_mass_from_velocity()`. From my current viewpoint, this split makes sense.

So I think the naming here actually makes sense, knowing the arguments are scaling factors from the time integration scheme only, and the total scaling factor of the mass operator _has_ to consider information on the actual operator, leading to the final `scaling_factor_mass`.

include/exadg/time_integration/time_int_gen_alpha_base.cpp
richardschu: It feels like this comment is more general and does not only affect the 3 highlighted lines, does it? If so, we should discuss this in person please, otherwise please consider the following comments:

I think this would mix the time integration constants (which we are actually computing) with the scaling factors in the operators. I would rather focus on the "pure" time integration factors stemming from the different sources here and keep out the physics, otherwise it gets misleading outside in my opinion.

If we take `get_scaling_factor_damping_operator()`, it depends on the actual damping used (strong or weak damping) and one would expect `parameters.weak_damping_coefficient` to contribute. I am in favour of the strict separation of time integration constants and operator scaling, which was prior to this PR not an issue.

include/exadg/structure/time_integration/time_int_gen_alpha.cpp
richardschu: The current naming is rather general and also explains exactly where it comes from. This was not the case before, but previously this was not an issue, since there was no confusion with other contributions to the mass operator scaling. Now, however, we have multiple contributions and may even have more in the future. I am not sure that there is no other weak form (maybe from viscoelasticity?) that could not introduce the same or very similar term with a different scaling, and the current naming (as "chunky" as it is) in my opinion reflects pretty well where terms and vectors are coming from and what they are. Hence, I would actually stick to the current naming.

include/exadg/time_integration/time_int_gen_alpha_base.h
richardschu: Good suggesstion, makes their intended use a lot clearer! I will add some comments.

include/exadg/time_integration/time_int_gen_alpha_base.h
```diff
@@ -53,17 +53,39 @@ class TimeIntGenAlphaBase : public TimeIntBase
...
+
+  /*
+   * computes the finite element vector corresponding to the *remainder* of the velocity term,
+   * int( phi_u, density * weak_damping_coefficient * D/Dt(d) )_Omega, which does *not* depend on
+   * the displacement and is shifted to the right hand side. Applying the mass operator to this
+   * vector allows computing the contribution to the rhs once before up front before solving the
```
nfehn, line 81: "before up front before" unclear

include/exadg/time_integration/time_int_gen_alpha_base.h
```diff
@@ -53,17 +53,39 @@ class TimeIntGenAlphaBase : public TimeIntBase
...
+   */
+  void
+  compute_const_vector_acceleration_remainder(VectorType &       const_vector,
+                                              VectorType const & displacement_n,
+                                              VectorType const & velocity_n,
+                                              VectorType const & acceleration_n) const;
+
+  /*
+   * computes the finite element vector corresponding to the *remainder* of the velocity term,
+   * int( phi_u, density * weak_damping_coefficient * D/Dt(d) )_Omega, which does *not* depend on
```
nfehn, line 79: I think "int" can be skipped, since this is indicated through the brackets. "phi_u" is somewhat unclear.

include/exadg/time_integration/time_int_gen_alpha_base.h
```diff
@@ -53,17 +53,39 @@ class TimeIntGenAlphaBase : public TimeIntBase
...
+   * the displacement and is shifted to the right hand side. Applying the mass operator to this
```
nfehn, line 80: "is shifted to the right hand side" -> from this comment it is not clear to me whether the functions takes into account the minus sign (originating from shifting terms to the rhs) or not?

nfehn:
Overall, I am not convinced regarding the naming `mass_from_acceleration` and `mass_from_velocity` in the time integrator. When reading the name `mass_from_velocity`, I expect this term to originate from the acceleration term of the PDE, which is not the case, right? I think the naming should instead distinguish between `mass/acceleration` on the one hand, and `damping` on the other hand. The circumstance that the damping term may be written as a scaled mass operator is something that should be the responsibility of the spatial discretization in my opinion. The time integrator should communicate to the spatial operator only the constants that originate from translating a second `a = d^2/dt^2 (d)` and first `v = d/dt (d)` temporal derivative into a time-discrete setting.

Moreover, it is somewhat inconsequent that we assemble the `const_vector` with contributions from two PDE terms (mass + damping) in the time integrator, while we assemble the `scaling_factors` with contributions from two PDE terms (mass + damping) not in the time integrator but in the spatial operator. This code design is hard to understand in my opinion. From a design perspective, it does not make a difference in my opinion whether we have `const_vectors` or `scaling_factors`. I think the question helping in the code design should be: should the time integrator know the `weak_damping_coefficient`? Then, enforcing this answer consistently will lead to a more clean design in my opinion.  Regarding the first paragraph above, it would be consistent to pass two vectors from the time integration scheme to the spatial operator, but then apply the mass operator and the damping operator inside the spatial discretization class prior to solving the linear/nonlinear problem. So I would personally answer the question with "no" (implying that the time integrator only  has the notion `mass/acceleration` and `damping`, but the concrete realization of `damping` and `weak_damping_coefficient` is somewhat invisible in the time integrator), but one might also argue for "yes" leading to another design. In the end, the design should be consistent and easy to understand.

nfehn:
Something else that does in my opinion not become clear from the present PR is the wording "weak". Could you explain what you mean by this? "linear damping" seems to be intuitive, if I understand it correctly.

include/exadg/time_integration/time_int_gen_alpha_base.h
richardschu: I changed the comment removing the questionable parts. There is no integral form needed here, which removes the "phi_u" part. Let me know what you think of the comment, once I have pushed an update!

richardschu:
> Something else that does in my opinion not become clear from the present PR is the wording "weak". Could you explain what you mean by this? "linear damping" seems to be intuitive, if I understand it correctly.
I followed the naming in [Wick](https://link.springer.com/article/10.1007/s00466-013-0890-3) (see Eqn. 9) , who referred to c*D/Dt(d) as `weak linear damping`. When considering for Rayleigh damping, a matrix D=a*M+b*K is added. Weak linear damping refers to the mass-proportional part, whereas strong damping refers to the stiffness-proportional part. One usually goes for a>0 and b=0, since the latter damping term might be nonlinear. You can find "strongly damped wave equation" in Richter's book on Fluid-Structure Interaction as well (Theorem 5 on page 28 of the preliminary version I have). I think the naming comes from the damping effect increasing or decreasing with frequency, see, e.g. [link](https://wiki.tum.de/pages/viewpage.action?pageId=630294659)). I am not the greatest expert on these damping terms, but in FSI, the linear weak damping is popular (see, e.g., the work by [Radtke et al.](https://link.springer.com/article/10.1007/s00466-016-1268-0)).

richardschu:
> Overall, I am not convinced regarding the naming `mass_from_acceleration` and `mass_from_velocity` in the time integrator. When reading the name `mass_from_velocity`, I expect this term to originate from the acceleration term of the PDE, which is not the case, right? I think the naming should instead distinguish between `mass/acceleration` on the one hand, and `damping` on the other hand. The circumstance that the damping term may be written as a scaled mass operator is something that should be the responsibility of the spatial discretization in my opinion. The time integrator should communicate to the spatial operator only the constants that originate from translating a second `a = d^2/dt^2 (d)` and first `v = d/dt (d)` temporal derivative into a time-discrete setting.
> 
> Moreover, it is somewhat inconsequent that we assemble the `const_vector` with contributions from two PDE terms (mass + damping) in the time integrator, while we assemble the `scaling_factors` with contributions from two PDE terms (mass + damping) not in the time integrator but in the spatial operator. This code design is hard to understand in my opinion. From a design perspective, it does not make a difference in my opinion whether we have `const_vectors` or `scaling_factors`. I think the question helping in the code design should be: should the time integrator know the `weak_damping_coefficient`? Then, enforcing this answer consistently will lead to a more clean design in my opinion. Regarding the first paragraph above, it would be consistent to pass two vectors from the time integration scheme to the spatial operator, but then apply the mass operator and the damping operator inside the spatial discretization class prior to solving the linear/nonlinear problem. So I would personally answer the question with "no" (implying that the time integrator only has the notion `mass/acceleration` and `damping`, but the concrete realization of `damping` and `weak_damping_coefficient` is somewhat invisible in the time integrator), but one might also argue for "yes" leading to another design. In the end, the design should be consistent and easy to understand.

thank you for sour comments, I will think about this some more and come back once I have made up my mind again - I believe I had some reasons why I was going this way but cannot remember rn ... I will comment again, once I push!

richardschu:
> Overall, I am not convinced regarding the naming `mass_from_acceleration` and `mass_from_velocity` in the time integrator. When reading the name `mass_from_velocity`, I expect this term to originate from the acceleration term of the PDE, which is not the case, right?

No this is not the case, you are right. The term is (testfunction, coefficient * D/Dt(displacement)).

> I think the naming should instead distinguish between `mass/acceleration` on the one hand, and `damping` on the other hand. 

I see your point. In my understanding, the mass and acceleration terms happen to refer to the same term, that is, the acceleration term that in a displacement-based formulation leads to a mass matrix scaled by some coefficient. I view anything that contributes to that mass matrix a mass term (coming from different sources). In that sense the naming I suggested made clear, *where* the term came from (`from_acceleration`) and what it lead to (`mass`). In the time integrator that targets second order ODEs, this gives quite a hint at what its intended use is. Before this PR, the `get_scaling_factor_mass` was clear, since we had only the acceleration term in the linear momentum balance leading to scaled mass matrix contribution. *Now*, I suggest going for `get_scaling_factor_acceleration` and `get_scaling_factor_velocity`. As we have just seen, `mass` is not clear enough (from the misunderstanding between the two of us), but acceleration and velocity should be fine. We still assume the primary variable be the displacement, but that is true for the whole integrator.   

> The circumstance that the damping term may be written as a scaled mass operator is something that should be the responsibility of the spatial discretization in my opinion. The time integrator should communicate to the spatial operator only the constants that originate from translating a second `a = d^2/dt^2 (d)` and first `v = d/dt (d)` temporal derivative into a time-discrete setting.

On this, we fully agree, apart from the naming just discussed, would you require any changes in this regard?

richardschu:
> Moreover, it is somewhat inconsequent that we assemble the `const_vector` with contributions from two PDE terms (mass + damping) in the time integrator, while we assemble the `scaling_factors` with contributions from two PDE terms (mass + damping) not in the time integrator but in the spatial operator. This code design is hard to understand in my opinion. From a design perspective, it does not make a difference in my opinion whether we have `const_vectors` or `scaling_factors`. I think the question helping in the code design should be: should the time integrator know the `weak_damping_coefficient`? Then, enforcing this answer consistently will lead to a more clean design in my opinion. Regarding the first paragraph above, it would be consistent to pass two vectors from the time integration scheme to the spatial operator, but then apply the mass operator and the damping operator inside the spatial discretization class prior to solving the linear/nonlinear problem. So I would personally answer the question with "no" (implying that the time integrator only has the notion `mass/acceleration` and `damping`, but the concrete realization of `damping` and `weak_damping_coefficient` is somewhat invisible in the time integrator), but one might also argue for "yes" leading to another design. In the end, the design should be consistent and easy to understand.
Yep, I would also argue for no. I simply followed the design that was there before, 

I now pass the `const_vector_acceleration_remainder` and `const_vectorVelocity_remainder` to the operator, which returns the `const_vector` applying the mass operator(s) inside the spatial operator. The damping operator is defined in the spatial operator in `apply_add_weak_damping_operator`. Now, the interface does not need to include the `apply_mass_operator` and `apply_add_weak_damping_operator`. Let me know what you think @nfehn !

include/exadg/structure/spatial_discretization/operator.cpp
```diff
@@ -761,11 +780,39 @@ Operator<dim, Number>::compute_initial_acceleration(VectorType &       accelerat
   mass_solver->solve(acceleration, rhs);
 }
 
+// template<int dim, typename Number>
+// void
+// Operator<dim, Number>::apply_mass_operator(VectorType & dst, VectorType const & src) const
+//{
+//  mass_operator.apply(dst, src);
+//}
```
nfehn, line 788: Let's make sure that we do not forget this in the end.

include/exadg/structure/spatial_discretization/operator.cpp
```diff
@@ -871,13 +918,17 @@ template<int dim, typename Number>
 std::tuple<unsigned int, unsigned int>
 Operator<dim, Number>::solve_nonlinear(VectorType &       sol,
                                        VectorType const & rhs,
-                                       double const       factor,
+                                       double const       scaling_factor_mass_from_acceleration,
+                                       double const       scaling_factor_mass_from_velocity,
```
nfehn, line 922: I would skip `mass_from` in the name (in both cases). The reason is that regarding this function interface, it does not matter whether the damping term is realized as a scaled mass matrix. Imagine we would add another type of damping, then we would need to also adjust the interface of the function, while we only have to change the internal implementation if we use the name `scaling_factor_damping` or `scaling_factor_velocity`.

include/exadg/grid/mapping_deformation_structure.h
```diff
@@ -141,8 +145,12 @@ class DeformedMapping : public DeformedMappingBase<dim, Number>
       pde_operator->initialize_dof_vector(rhs);
       pde_operator->compute_rhs_linear(rhs, time);
 
-      auto const iter = pde_operator->solve_linear(
-        displacement, rhs, 0.0 /* no mass term */, time, false /* do not update preconditioner */);
+      auto const iter = pde_operator->solve_linear(displacement,
+                                                   rhs,
+                                                   0.0 /* no mass terms */,
+                                                   0.0 /* no mass terms */,
```
nfehn, line 151: I think these comments should be changed for the same reason as described here https://github.com/exadg/exadg/pull/475/files#r1282039088. The comments should be something like "no acceleration term" and "no damping term". Note that there are several other occurrences of this type in the code.

include/exadg/structure/spatial_discretization/operator.h
```diff
@@ -331,6 +342,10 @@ class Operator : public dealii::Subscriptor, public Interface::Operator<Number>
   unsigned int
   get_quad_index_gauss_lobatto() const;
 
+  double
```
nfehn, line 345: Here, a comment would be useful saying that this function assumes a damping operator that is realized as a scaled mass matrix.

nfehn:
I think we have a good design now. So apart from some minor comments, this is then ready-to-merge from my side.

richardschu:
... we really have to look into that bug concerning unsorted constraints at operator setup. The current ExaDG master does not pass tests on my machine.

PR #477, opened 2023-06-06
shift coarse-grid Chebyshev smoother object to MGCoarseChebyshev

nfehn:
As a consequence, the number of smoother objects in `MultigridPreconditionerBase` can be reduced by one and is now consistent to the number of actual smoothing levels. So far, `smoothers[0]` was somewhat unclear and only used in case of a Chebyshev coarse-grid solver. The logic should now be clearer and more explicit.

### DIFF
--- a/include/exadg/incompressible_navier_stokes/preconditioners/multigrid_preconditioner_momentum.cpp
+++ b/include/exadg/incompressible_navier_stokes/preconditioners/multigrid_preconditioner_momentum.cpp
@@ -150,12 +150,14 @@ MultigridPreconditioner<dim, Number>::update()
       });
   }
 
-  // Once the operators are updated, the update of smoothers and the coarse grid solver is generic
-  // functionality implemented in the base class.
-  this->update_smoothers();
-
-  // singular operators do not occur for this operator
-  this->update_coarse_solver(false /* operator_is_singular */);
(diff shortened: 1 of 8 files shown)


include/exadg/incompressible_navier_stokes/preconditioners/multigrid_preconditioner_momentum.cpp
```diff
@@ -150,12 +150,14 @@ MultigridPreconditioner<dim, Number>::update()
...
-  // Once the operators are updated, the update of smoothers and the coarse grid solver is generic
-  // functionality implemented in the base class.
-  this->update_smoothers();
-
-  // singular operators do not occur for this operator
-  this->update_coarse_solver(false /* operator_is_singular */);
+  // In case the operators have been updated, we also need to update the smoothers and the coarse
+  // grid solver. This is generic functionality implemented in the base class.
+  if(mesh_is_moving or data.unsteady_problem or
+     (mg_operator_type == MultigridOperatorType::ReactionConvectionDiffusion))
```
nfehn, line 156: We now only update this preconditioner if the operator has been updated before. In some cases, this makes the code more efficient.

kronbichler (approved):
It's good that we can still use the `level` variable to access fields, even though the range is from `1` to `n_levels`. Looks good in my opinion.

PR #479, opened 2023-06-06
let nonlinear viscosity trigger a Newton solve in Navier-Stokes module

richardschu:
Follow-up from #381, where we did not allow for a nonlinear viscous term triggering a nonlinear solve. Up to this point, we allowed for the viscous term being implicit if the convective term was treated implicit as well.
Now, a nonlinear viscous term can trigger a nonlinear solver.
For incomrpessible Navier-Stokes we have (i) steady driver, (ii) `BDFCoupledSolution`, (iii) `BDFPressureCorrection` and (iv) `BDFDualSplittingScheme` as possible options. (i) and (iv) do not require any changes, since already before we demanded the convective term either being implicit (in case of (i) ) or explicit (in case of (iv) ). Therefore, the `BDFCoupledSolution` and `BDFPressureCorrection` schemes remain.
The main problem is that up to now `nonlinear_problem_has_to_be_solved()` and `implicit_convective_problem()` defined in `incompressible_navier_stokes/user_interface/parameters.cpp` and similar functions could be used interchangeably, which is now no longer the case and leads to incorrect behavior!
I will add comments in the code down below.


include/exadg/incompressible_navier_stokes/spatial_discretization/operator_pressure_correction.cpp
```diff
@@ -330,8 +330,11 @@ OperatorPressureCorrection<dim, Number>::evaluate_nonlinear_residual(
 
   this->mass_operator.apply_scale(dst, scaling_factor_mass, src);
 
-  if(this->param.convective_problem())
+  // implicitly treated convective term
+  if(this->param.implicit_convective_problem())
+  {
     this->convective_operator.evaluate_nonlinear_operator_add(dst, src, time);
```
richardschu, line 336: otherwise it is added via the `rhs_vector` added down below.

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
```diff
@@ -554,9 +554,13 @@ SpatialOperatorBase<dim, Number>::initialize_operators(std::string const & dof_i
 
   data.unsteady_problem = unsteady_problem_has_to_be_solved();
   if(param.temporal_discretization == TemporalDiscretization::BDFDualSplittingScheme)
```
richardschu, line 556: this is just so you see that I thought about those lines!

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_coupled_solver.cpp
```diff
@@ -241,6 +226,28 @@ TimeIntBDFCoupled<dim, Number>::do_timestep_solve()
     if(this->param.right_hand_side)
       pde_operator->evaluate_add_body_force_term(rhs, this->get_next_time());
 
+    // Add the convective term to the right-hand side of the equations
```
richardschu, line 229: adds convective term on the rhs with a minus sign.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_pressure_correction.cpp
```diff
@@ -353,10 +353,9 @@ TimeIntBDFPressureCorrection<dim, Number>::momentum_step()
   }
 
   /*
-   *  if a variable viscosity is used: update
-   *  viscosity model before calculating rhs_momentum
+   *  explicit variable viscosity update executed prior to calculation of rhs_momentum
```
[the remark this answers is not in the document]
richardschu: The condition shoudl be `if(this->param.viscosity_is_variable() and not this->param.viscous_term_is_nonlinear())`

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_coupled_solver.cpp
```diff
@@ -176,8 +176,8 @@ TimeIntBDFCoupled<dim, Number>::do_timestep_solve()
     solution_np = solution_last_iter;
   }
 
-  // update viscosity model
-  if(this->param.viscosity_is_variable())
+  // explicit viscosity update
+  if(this->param.viscosity_is_variable() and not this->param.viscous_term_is_nonlinear())
   {
```
richardschu, line 181: note that if we go for `TreatmentOfVariableViscosity::Implicit` and hence `this->param.viscous_term_is_nonlinear()` is `true`, the residual update in the Newton solver updates the viscosity prior to solving.

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_coupled.cpp
```diff
@@ -1176,8 +1176,10 @@ OperatorCoupled<dim, Number>::apply_preconditioner_pressure_block(VectorType &
         this->momentum_operator.get_scaling_factor_mass_operator());
     }
 
-    if(this->param.nonlinear_problem_has_to_be_solved())
```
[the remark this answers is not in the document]
nfehn: yes

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_coupled_solver.cpp
```diff
@@ -176,8 +176,8 @@ TimeIntBDFCoupled<dim, Number>::do_timestep_solve()
     solution_np = solution_last_iter;
   }
 
-  // update viscosity model
-  if(this->param.viscosity_is_variable())
+  // explicit viscosity update
+  if(this->param.viscosity_is_variable() and not this->param.viscous_term_is_nonlinear())
```
nfehn, line 180: Here, I do not like the name `viscous_term_is_nonlinear()`. For me, the question is not whether the term is nonlinear or linear, but whether the user decides for an implicit vs. explicit treatment of this term, or do I misunderstand something?

To avoid the negation, a statement like `if TreatmentOfVariableViscosity::Explicit` would be easier to read in my opinion.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_pressure_correction.cpp
```diff
@@ -353,10 +353,9 @@ TimeIntBDFPressureCorrection<dim, Number>::momentum_step()
...
    */
-  if(this->param.viscosity_is_variable())
+  if(this->param.viscosity_is_variable() and not this->param.viscous_term_is_nonlinear())
```
nfehn, line 358: same as here https://github.com/exadg/exadg/pull/479/files#r1293216935

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_coupled_solver.cpp
nfehn: Note that the function `viscous_term_is_nonlinear()` already includes `viscosity_is_variable()` internally, so this is something to be corrected.

include/exadg/incompressible_navier_stokes/user_interface/parameters.cpp
```diff
@@ -634,14 +618,13 @@ Parameters::implicit_convective_problem() const
 bool
 Parameters::nonlinear_viscous_problem() const
 {
-  return viscous_problem() and viscous_term_is_nonlinear();
+  return (viscous_problem() and viscous_term_is_nonlinear());
```
nfehn, line 621: We should check whether we really need the function `viscous_term_is_nonlinear()`, or whether we could just insert the two lines of this function here. The use of `viscous_term_is_nonlinear()` does not seem to be intuitive in the temporal/spatial discretization routines.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_coupled_solver.cpp
```diff
@@ -241,6 +241,28 @@ TimeIntBDFCoupled<dim, Number>::do_timestep_solve()
...
+      {
+        // evaluate convective term for all previous times since the mesh has been updated
+        for(unsigned int i = 0; i < this->vec_convective_term.size(); ++i)
+        {
+          // in a general setting, we only know the boundary conditions at time t_{n+1}
+          pde_operator->evaluate_convective_term(this->vec_convective_term[i],
+                                                 solution[i].block(0),
+                                                 this->get_next_time());
+        }
+      }
```
nfehn, line 260: Where does this code come from? This should only be implemented once. Lines 245-260 (with another closing bracket) can be done before the `if(nonlinear)-else` statement in this function.

Only the two lines below, adding the convective terms to the `rhs` vector may be written redundantly for the linear and nonlinear case.

I would suggest to do this change in a separate PR, since it is a pure refactoring of the existing code, but might potentially break the current code version (if we introduce a mistake here).

include/exadg/incompressible_navier_stokes/user_interface/parameters.h
```diff
@@ -57,6 +57,9 @@ class Parameters
   bool
   viscous_problem() const;
 
+  bool
+  viscous_term_is_nonlinear() const;
+
```
nfehn, line 62: As written somewhere else, this change should not be made.

include/exadg/incompressible_navier_stokes/user_interface/parameters.cpp
```diff
@@ -556,23 +556,13 @@ Parameters::check(dealii::ConditionalOStream const & pcout) const
...
-    {
-      AssertThrow(
-        implicit_convective_problem(),
-        dealii::ExcMessage(
-          "The current implementation only calls a Newton solver, if we have an implicit convective problem. Treat nonlinear convective and viscous terms alike."));
-    }
+    AssertThrow(
+      treatment_of_variable_viscosity == TreatmentOfVariableViscosity::Explicit,
+      dealii::ExcMessage(
+        "An implicit treatment of the variable viscosity field (rendering the viscous step of the dual splitting scheme nonlinear regarding the unknown velocity field) is currently not implemented for the dual splitting scheme."));
```
nfehn, line 565: As a minor comment: Maybe you could split up this message into smaller pieces (as we do in other places of the code) for the purpose of indentation.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_coupled_solver.cpp
richardschu: Got that, thanks!

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_pressure_correction.cpp
richardschu: also corrected the condition in the `driver_steady_problems.cpp`

include/exadg/incompressible_navier_stokes/user_interface/parameters.cpp
richardschu: now it is only used in the `parameters.h/.cpp`, I can remove it.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_coupled_solver.cpp
richardschu: I pulled out the update of the mesh velocity at previous timesteps and the computation of `sum_alpha_ui` as they are both identical for the two code paths.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_coupled_solver.cpp
```diff
@@ -225,22 +226,48 @@ TimeIntBDFCoupled<dim, Number>::do_timestep_solve()
     ((this->time_step_number - 1) % this->param.update_preconditioner_coupled_every_time_steps ==
      0);
 
-  if(this->param.nonlinear_problem_has_to_be_solved())
+  // calculate Sum_i (alpha_i/dt * u_i) and store
```
richardschu, line 229: pulled out the `sum_alpha_ui`

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_coupled_solver.cpp
```diff
@@ -225,22 +226,48 @@ TimeIntBDFCoupled<dim, Number>::do_timestep_solve()
...
+  for(unsigned int i = 1; i < solution.size(); ++i)
   {
-    VectorType sum_alphai_ui(solution[0].block(0));
+    sum_alphai_ui.add(this->bdf.get_alpha(i) / this->get_time_step_size(), solution[i].block(0));
+  }
 
-    // calculate Sum_i (alpha_i/dt * u_i)
-    sum_alphai_ui.equ(this->bdf.get_alpha(0) / this->get_time_step_size(), solution[0].block(0));
-    for(unsigned int i = 1; i < solution.size(); ++i)
+  // Update the convective term when using an ALE formulation
```
richardschu, line 237: pulled out the update of convective velocity in an ALE formulation

include/exadg/incompressible_navier_stokes/time_integration/driver_steady_problems.cpp
```diff
@@ -163,7 +164,7 @@ DriverSteadyProblems<dim, Number>::do_solve(double const time, bool unsteady_pro
     if(this->param.right_hand_side)
       pde_operator->evaluate_add_body_force_term(rhs, time);
 
-    // Newton solver
+    // Newton solver with viscosity update
```
nfehn, line 167: This comment "with viscosity update" seems to be irritating. It looks like the Newton solver will always do a viscosity update. However, it will only be done if the nonlinear viscosity term is treated implicitly, which is only one reason for the if-statement `this->param.nonlinear_problem_has_to_be_solved()` above to become `true`.

So I think I would actually skip this here.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_coupled_solver.cpp
```diff
@@ -225,22 +226,48 @@ TimeIntBDFCoupled<dim, Number>::do_timestep_solve()
...
+  if(this->param.nonlinear_problem_has_to_be_solved())
+  {
     VectorType rhs(sum_alphai_ui);
     pde_operator->apply_mass_operator(rhs, sum_alphai_ui);
     if(this->param.right_hand_side)
       pde_operator->evaluate_add_body_force_term(rhs, this->get_next_time());
 
+    // Add the convective term to the right-hand side of the equations
+    // if the convective term is treated explicitly (additive decomposition):
+    // evaluate convective term and add extrapolation of convective term to the rhs (-> minus sign!)
```
nfehn, line 263: "evaluate convective term and" needs to be removed here in order to avoid irritations.

include/exadg/incompressible_navier_stokes/user_interface/parameters.cpp
```diff
@@ -556,23 +556,14 @@ Parameters::check(dealii::ConditionalOStream const & pcout) const
   }
 
   // VARIABLE VISCOSITY MODELS
-  if(viscosity_is_variable())
+  if(viscosity_is_variable() &&
```
nfehn, line 559: `&&` -> `and`

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_coupled_solver.cpp
nfehn: Was this part also copied before or why does this change appear now?

nfehn:
looks good apart from some concluding comments!

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_coupled_solver.cpp
richardschu: adapted both :)

PR #481, opened 2023-06-07
implement Chebyshev smoother in a more encapsulated way

nfehn:
The problem with the previous implementation was that `dealii::PreconditionChebyshev` is templated regarding the preconditioner. However, not all such preconditioner have a common base class in ExaDG. This implies if-else branches on the level of data types / constructors, which was currently realized in `MultigridPreconditionerBase`. However, this somewhat "pollutes" the multigrid preconditioner class that should have a higher astraction level. Hence, it makes in my opinion more sense to implement such "wrapping functionality" within `ExaDG::ChebyshevSmoother`. Moreover, the circumstance that `ExaDG::MGCoarseChebyshev` currently supports only point-Jacobi preconditioning is now realized more naturally, with the `Assert` appearing in the respective class `ExaDG::MGCoarseChebyshev` (which one potentially wants to extend to block-Jacobi in the future).

With the present changes, the `ChebyshevSmoother` follows closely the design of other smoothers in `ExaDG`.

Note that `const`-correctness has been improved along the way for other smoothers in ExaDG.

### DIFF
--- a/include/exadg/solvers_and_preconditioners/multigrid/coarse_grid_solvers.h
+++ b/include/exadg/solvers_and_preconditioners/multigrid/coarse_grid_solvers.h
@@ -358,15 +358,21 @@ public:
 
   typedef dealii::LinearAlgebra::distributed::Vector<MultigridNumber> VectorType;
 
-  typedef ChebyshevSmoother<Operator, VectorType, dealii::DiagonalMatrix<VectorType>> Chebyshev;
+  typedef dealii::PreconditionChebyshev<Operator, VectorType, dealii::DiagonalMatrix<VectorType>>
+    DealiiChebyshev;
 
-  MGCoarseChebyshev(Operator const &   coarse_operator_in,
-                    SolverData const & solver_data_in,
-                    bool const         operator_is_singular_in)
+  MGCoarseChebyshev(Operator const &                          coarse_operator_in,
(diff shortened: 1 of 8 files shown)


kronbichler:
I agree with the general direction that puts everything inside a class for the smoother, such that the multigrid interface does not have to handle the switch cases. However, I am also not entirely happy about the `if` statements inside the smoother class, if we think towards additional "inner" preconditions in Chebyshev that are neither the diagonal nor the block-diagonal, say overlapping Schwarz. At the same time, we cannot create a virtual base class for the preconditioner inside `dealii::PreconditionChebyshev`, because the deal.II implementation relies on the ability to introspect the `Preconditioner` object to see if it can run vector operations in parallel. This means that we need a layer inside `ExaDG::ChebyshevSmoother` that can create different options for the inner precondition from an `enum` (as now), but has a uniform interface to be able to make all code a template. The current PR can be seen as a step in that direction, and it can later be extended or made more modular inside `ExaDG::ChebyshevSmoother`.

kronbichler (approved):

nfehn:
Maybe I should run tests manually using Chebyshev+BlockJacobi before merging ...

nfehn:
Chebyshev+block-Jacobi seems to be working well.

PR #491, opened 2023-06-12
constructor of Grid has not been called for precursor domain in solver_precursor

nfehn:


kronbichler (approved):

richardschu (approved):
does the job! thank for the quick fix!

PR #492, opened 2023-06-13
Refactor create_triangulation() functionality /enable global-coarsening MG for all apps

nfehn:
Currently, the utility functions `GridUtilities::create_triangulation()` do not construct the `Triangulation` object, which is currently done in `Grid::initialize()`. However, this implies that parameters such as e.g. `dealii:construct_multigrid_hierarchy` are spread over several places, and that e.g. an additional lambda function needs to be defined when creating explicitly the coarse triangulations for multigrid.

With the suggestions of the present PR, `create_triangulation()` functions also call the constructor, which is more intuitive in my opinion. No additional call of `Grid::initialize()` is needed in the `Application`. All grid generation functionality now concentrates in `Application::create_grid()`, which is in my opinion easier to overview for users of the library than is the case currently.

~~Due to the removal of `Grid::initialize()` calling constructors, several test cases will not run successfully in the current preliminary state of the PR, since not all `Application::create_grid()` functions make use of the "new" utility functions in `GridUtilities`. This, however, needs to be adjusted anyways since several months (note that even with the present code version, many test cases can and probably will fail for certain parameters, such as those invoking global coarsening multigrid). @peterrum @kronbichler could you comment on the idea of this PR already now, because all the remaining changes are technical, touch many applications, and should at best only be done once?~~

All applications should now principally run successfully for all the possible triangulation, element type, multigrid parameter settings in ExaDG (in particular global-coarsening MG and simplices (if implemented)).


include/exadg/grid/grid.h
```diff
@@ -45,19 +45,6 @@ class Grid
...
-   */
-  Grid()
-  {
-  }
-
-  /**
-   * Initialize function
-   */
-  void
-  initialize(GridData const & data, MPI_Comm const & mpi_comm);
```
nfehn, line 59: for full flexibility, `Grid` should not depend on `GridData`, which is achieved with the present PR.

include/exadg/grid/grid_utilities.h
```diff
@@ -300,31 +336,13 @@ create_coarse_triangulations(
...
-        fine_triangulation.n_global_levels() - 1);
+      std::vector<std::shared_ptr<dealii::Triangulation<dim>>> coarse_triangulations =
+        std::vector<std::shared_ptr<dealii::Triangulation<dim>>>(
+          fine_triangulation.n_global_levels() - 1);
 
       coarse_periodic_face_pairs =
         std::vector<PeriodicFacePairs<dim>>(coarse_triangulations.size());
 
-      // lambda function for creating the coarse triangulations
-      auto const lambda_create_level_triangulation =
```
nfehn, line 310: This lambda function has been quite un-intuitive in the present code version in my opinion.

include/exadg/incompressible_navier_stokes/driver.h
```diff
@@ -72,7 +72,7 @@ get_dofs_per_element(std::string const & input_file,
                      unsigned int const  degree)
 {
   PressureDegree pressure_degree = PressureDegree::MixedOrder;
-  OperatorType   operator_type;
+  OperatorType   operator_type   = OperatorType::InverseMassOperator;
```
nfehn, line 75: This is a bugfix somewhat independent of the present PR. A few lines below (lines 91-92), an error is thrown if this variable is not initialized. @jh66637 FYI

applications/incompressible_flow_with_transport/rising_bubble/application.h
```diff
@@ -226,7 +226,7 @@ class Application : public FTI::ApplicationBase<dim, Number>
 
     // MATHEMATICAL MODEL
     param.problem_type                = ProblemType::Unsteady;
-    param.equation_type               = EquationType::Convection;
+    param.equation_type               = EquationType::ConvectionDiffusion;
```
nfehn, line 229: due to PR #496, one should be able to use `Convection` only as in the previous code version. I will change this back to the original version.

applications/poisson/throughput/application.h
```diff
@@ -77,31 +77,48 @@ class Application : public ApplicationBase<dim, n_components, Number>
...
+          // do nothing
+        }
+        else if(mesh_type == MeshType::Curvilinear)
+        {
+          curvilinear_mesh = true;
+        }
+        else
+        {
+          AssertThrow(false, dealii::ExcMessage("Not implemented."));
+        }
```
nfehn, line 103: can be simplified as in other applications

Issue #493, opened 2023-06-14
PointwiseOutputGenerator crashes for compressible Navier-Stokes


nfehn:
For the turbulent channel example in CompNS, the simulation crashes in `ExaDG::PointwiseOutputGeneratorBase::write_quantity()`

The error message is 
```bash
    void ExaDG::extract_component_from_tensors(dealii::Vector<RangeNumberType>&, const std::vector<dealii::Tensor<1, dim, Number> >&, unsigned int) [with int n_components = 5; Number = double]
The violated condition was: 
    ::dealii::deal_II_exceptions::internals::compare_less_than(comp, n_components)
Additional information: 
    Index 5 is not in the half-open range [0,5).
```

@jh66637 can you confirm this?

jh66637:
I can confirm this. I place a PR to resolve it.

PR #494, opened 2023-06-14
fix writing Rho_E in pointwise output of CNSE

jh66637:
Closes https://github.com/exadg/exadg/issues/493

The index at which to access Rho_E was wrong.


nfehn (approved):
looks reasonable

PR #496, opened 2023-06-15
Enable OperatorBase::rhs()/evaluate() for is_dg=false or evaluate_face_integrals=false

nfehn:
`OperatorBase::rhs()` and `OperatorBase::evaluate()` currently assume `is_dg = true` and `evaluate_face_integrals() = true`. Since no asserts are placed in the code, this can be considered a bug in the present code version.

The problem occurred for example when solving scalar transport using the `CombinedOperator` but treating convection explicitly and omitting the diffusive term. Then, `CombinedOperator` should only evaluate cell integrals for the mass operator. However, the code currently runs into the face integrals due to shortcomings / incomplete functionality in `OperatorBase::rhs()` and `OperatorBase::evaluate()`.

For the implementation in `OperatorBase`, we have to distinguish 4 cases:
- DG with face integrals
- DG without face integrals
- CG with boundary face integrals
- CG without boundary face integrals

Some explanations for review:
- (Boundary) face integrals occur for CG disretizations only in the form of inhomogeneous boundary conditions. Hence, the function `OperatorBase::apply()` (homogeneous operator) does not distinguish between the two cases for CG (because `apply()` == homogeneous operator only).
- `OperatorBase::rhs()` (inhomogeneous operator) only considers inhomogeneous boundary face integrals. Therefore, there is no need to distinguish between DG and CG. The distinction between CG/DG happens somewhat deeper in `OperatorBase::boundary_face_loop_inhom_operator()`.
- `OperatorBase::evaluate()` (homogeneous+inhomogeneous operator) needs a 4th case as compared to `OperatorBase::apply()` because there is also the case of CG with boundary face integrals. In fact, for the case `evaluate_face_integrals()==false`, one could combine the CG and DG branches because only a cell loop remains, which is the same for CG/DG. However, for better readability and consistency with the other functions in `OperatorBase`, I think it is advantageous to show the 4 cases explicitly.

### DIFF
--- a/include/exadg/operators/operator_base.cpp
+++ b/include/exadg/operators/operator_base.cpp
@@ -328,15 +328,18 @@ OperatorBase<dim, Number, n_components>::rhs_add(VectorType & rhs) const
   VectorType tmp;
   tmp.reinit(rhs, false);
 
-  matrix_free->loop(&This::cell_loop_empty,
-                    &This::face_loop_empty,
-                    &This::boundary_face_loop_inhom_operator,
-                    this,
-                    tmp,
-                    tmp);
+  if(evaluate_face_integrals())
+  {
+    matrix_free->loop(&This::cell_loop_empty,
+                      &This::face_loop_empty,
+                      &This::boundary_face_loop_inhom_operator,
+                      this,
+                      tmp,
+                      tmp);
 
-  // multiply by -1.0 since the boundary face integrals have to be shifted to the right hand side
-  rhs.add(-1.0, tmp);
+    // multiply by -1.0 since the boundary face integrals have to be shifted to the right hand side
+    rhs.add(-1.0, tmp);
+  }
 
   if(not is_dg)
   {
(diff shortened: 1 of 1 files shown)


kronbichler (approved):

kronbichler:
This change seems reasonable. I think we might want to tackle the case of interleaving vector operations and cell loops (for the homogeneous operator) soon, where this preparation and further cleanups certainly help.

peterrum:
@kronbichler What about adding asserts to `FEEvaluation::evaluate()`, `::gather_evaluate`, `::integrate()` , `::integrate_scatter()` that `EvaluationFlags::nothing` is not set?

kronbichler:
Somewhere in the code we have early return statements in case we do not evaluate any quantities: https://github.com/dealii/dealii/blob/c03ccfab612106a51f8d052c2b69355eac756c21/include/deal.II/matrix_free/evaluation_kernels.h#L254-L255
I guess you are thinking of telling the user that nothing has happened? I guess it depends on whether one would have constructed codes doing no evaluation, and if that would break some codes. It does not make a lot of sense to not evaluate anything, and all `get_xxx` functions should return an error, but we might be able to change it.

peterrum (approved):

peterrum:
> I guess you are thinking of telling the user that nothing has happened?

Exactly. As far as I see, the fact that the buffers in `FEEvaluation` are reused is a very big pitfall. In the case of the current issue, it leads to a simple gather/scatter operation although one would like to scatter nothing. Another case is that someone did not submit gradients (maybe the gradients have been evaluated) but nevertheless tests the solutions on the quadrature points with the gradient of the test functions. In the last year, I had to help to debug such bugs in different user codes at least 3-4 times.

kronbichler:
> the fact that the buffers are reused is a very big pitfall.

It is a pitfall for sure, but it is necessary for performance. We might think if there are better ways to warn users, e.g. in the setting you describe regarding the `gather` and `scatter` option.

kronbichler:
That said, I would be happy to change the code to use different slots for the evaluated quantities and the quantities to be tested, but we should make sure that we do not regress in performance. As many test projects from 2018-2020 have already higher performance than `FEEvaluation`, I think we have plenty of things to do to improve performance, so regression even if single-digit percents is really a no-go for me.

PR #497, opened 2023-06-15
use new enum utility functions for throughput studies

nfehn:
follow up to #382

@jh66637 FYI

### DIFF
--- a/include/exadg/utilities/throughput_parameters.h
+++ b/include/exadg/utilities/throughput_parameters.h
@@ -86,6 +86,7 @@ measure_operator_evaluation_time(std::function<void(void)> const & evaluate_oper
   return wall_time;
 }
 
+template<typename OperatorType>
 struct ThroughputParameters
 {
   ThroughputParameters()
@@ -105,7 +106,7 @@ struct ThroughputParameters
     prm.enter_subsection("Throughput");
     {
       prm.add_parameter(
-        "OperatorType", operator_type, "Type of operator.", dealii::Patterns::Anything(), true);
+        "OperatorType", operator_type, "Operator type.", Patterns::Enum<OperatorType>(), true);
       prm.add_parameter("RepetitionsInner",
                         n_repetitions_inner,
                         "Number of operator evaluations.",
@@ -123,10 +124,11 @@ struct ThroughputParameters
   void
   print_results(MPI_Comm const & mpi_comm)
   {
-    print_throughput(wall_times, operator_type, mpi_comm);
+    print_throughput(wall_times, ExaDG::Utilities::enum_to_string(operator_type), mpi_comm);
   }
 
-  std::string operator_type = "Undefined";
+  // type of PDE operator
+  OperatorType operator_type = OperatorType();
 
   // number of repetitions used to determine the average/minimum wall time required
   // to compute the matrix-vector product
--- a/include/exadg/utilities/throughput_parameters.h
+++ b/include/exadg/utilities/throughput_parameters.h
@@ -128,7 +128,7 @@ struct ThroughputParameters
   }
 
   // type of PDE operator
-  OperatorType operator_type = OperatorType();
+  OperatorType operator_type = Utilities::default_constructor<OperatorType>();
 
   // number of repetitions used to determine the average/minimum wall time required
   // to compute the matrix-vector product
--- a/include/exadg/utilities/throughput_parameters.h
+++ b/include/exadg/utilities/throughput_parameters.h
@@ -138,7 +138,7 @@ struct ThroughputParameters
   unsigned int n_repetitions_inner = 100; // take the average of inner repetitions
   unsigned int n_repetitions_outer = 1;   // take the minimum of outer repetitions
 
-  // global variable used to store the wall times for different polynomial degrees and problem sizes
+  // variable used to store the wall times for different polynomial degrees and problem sizes
   mutable std::vector<std::tuple<unsigned int, dealii::types::global_dof_index, double>> wall_times;
 };
 } // namespace ExaDG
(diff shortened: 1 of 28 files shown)


include/exadg/utilities/throughput_parameters.h
```diff
@@ -123,10 +124,11 @@ struct ThroughputParameters
   void
   print_results(MPI_Comm const & mpi_comm)
   {
-    print_throughput(wall_times, operator_type, mpi_comm);
+    print_throughput(wall_times, ExaDG::Utilities::enum_to_string(operator_type), mpi_comm);
   }
 
-  std::string operator_type = "Undefined";
+  // type of PDE operator
+  OperatorType operator_type = OperatorType();
```
nfehn, line 131: The most essential change of this PR, which causes to template `ThroughputParameters`. The rest is technical.

jh66637: Does this work? I think we should use `OperatorType operator_type=OperatorType::Undefined` instead. This requires the `OperatorType` to have a field `Undefined` and therefore the other approach would be more generic, but I think it does not work. If I test the current initialization on a minimum example, it seems the initialization does not point to a valid enum:
```
enum class Test
{
  a = 1,
  b
};

int
main()
{
  Test t = Test();
  std::cout << "enum name " << magic_enum::enum_name(t) << std::endl;
  return 0;
}
Output: enum name
```

include/exadg/utilities/throughput_parameters.h
```diff
@@ -123,10 +124,11 @@ struct ThroughputParameters
   void
   print_results(MPI_Comm const & mpi_comm)
   {
-    print_throughput(wall_times, operator_type, mpi_comm);
+    print_throughput(wall_times, ExaDG::Utilities::enum_to_string(operator_type), mpi_comm);
```
jh66637, line 127: Could we instead have a `print_throughput()` overload which takes an enums as we have for printing parameters?

jh66637:
I really like the changes. In my opinion, much cleaner and safer than passing around strings :)

include/exadg/utilities/throughput_parameters.h
nfehn: What do you mean by overload? A templated `template<typename OperatorType> void print_throughput()` function?

The `print_throughput()` function does not need more than the string. This is why I prefered to do it like this.

include/exadg/utilities/throughput_parameters.h
nfehn: I have found this https://lifecs.likai.org/2010/07/c-enum-default-constructor.html, which is quite similar to your example.

But doesn't the new enum functionality check anyway if the string provided in the input file can be translated to one of the values of the enum type? I thought we would receive an error message then? Or do you mean if we do not set `true` as last parameter in `add_parameter(..., true)`?

jh66637: Can you also try with a non-existing value for `OperatorType`? Maybe an empty `Selection` pattern allows for all values, and we don't get the nice error messages from the `ParameterHandler` but instead error messages that stings could not be converted to the enum. 

And I think `OperatorType operator_type = OperatorType();` is equivalent to `OperatorType operator_type;`. Can you try this? Personally, I am confused by `EnumType = EnumType()` because I don't know what this implies.

nfehn: > And I think OperatorType operator_type = OperatorType(); is equivalent to OperatorType operator_type;. Can you try this? Personally, I am confused by EnumType = EnumType() because I don't know what this implies.

No, this is not the same. This is what I reported a few days ago and what I observed again today. Which is why I added `= OperatorType()`.

jh66637: In my minimal example, it is the same. Since also the issue you reported works on my machine, can we just go with `OperatorType operator_type=OperatorType::Undefined` since we know exactly what we are doing in this case? 

Alternatively, we could introduce another enum utility function to initialize with the first value
```
template <typename EnumType>
EnumType
first_value()
{
return magic_enum::enum_values<EnumType>()[0];
}
```

include/exadg/utilities/throughput_parameters.h
jh66637: I worked really hard to hide away the boilerplate `enum_to_string()` conversions in a lot of places to clean up the code. I think this is a fairly small change and keeps the whole `enum2string` stuff at one place in `print_functions.h`.

What you are describing can not happen due to the `constexpr` in https://github.com/exadg/exadg/blob/master/include/exadg/utilities/print_functions.h#L83.

I think the following could make both of us happy :)
```
if constexpr (Utilities::is_enum<ParameterType>())
   ...
else if consexpr (std::is_same_v<ParameterType, std::string>)
   ...
else
   static_assert(false,"print_throughput() can only be called with enum or string as second argument.")
```

nfehn: We should only introduce a template if we do not know the concrete type. For `print_throughput()` we know the type, namely `std::string`. `Utilities::enum_to_string()` is a utility function that one is free to use anywhere. The benefit of your hard work is still there, namely that this function is generic (and not written for every single enum type in ExaDG).

If you are strictly against using `Utilities::enum_to_string()` in the class `ThroughputParameters`, you would have to be also against `Patterns::Enum<OperatorType>()` a few lines above in my opinion (because in both cases we assume that the template parameter `OperatorType` is an enum. Note that there is concrete reasoning to use a template for `ThroughputParameters`, because we do not know the concrete enum definition (every PDE module in ExaDG defines its own enum for the operator type). This motivation is not given in my opinion for the function `print_throughput()`.

PR #499, opened 2023-06-16
enable running FDA benchmark without precursor

nfehn:
This PR should fix the problem that `./incompressible_navier_stokes/fda_benchmark/solver` (the solver without precursor) is currently not running for the FDA benchmark.

Overall, the application design for precursor simulations (relying on inheritance) should be changed, since it is very hard to keep the overview of what is going on if one runs such an `Application` (deriving from `ApplicationBasePrecursor`, itself deriving from `ApplicationBase`) without precursor.

### DIFF
--- a/applications/incompressible_navier_stokes/backward_facing_step/application.h
+++ b/applications/incompressible_navier_stokes/backward_facing_step/application.h
@@ -789,8 +789,6 @@ private:
 
 } // namespace ExaDG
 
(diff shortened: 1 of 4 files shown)


richardschu (approved):
Looks good to me. :+1:  I assume you tested that it runs and gives correct results, so it should be fine!

kronbichler (approved):

PR #506, opened 2023-06-26
Enable hanging nodes for continuous Galerkin discretizations

### LINKED ISSUE #386: Usage of read_dof_values_plain()
In comments https://github.com/exadg/exadg/issues/383#issuecomment-1466891274 and https://github.com/exadg/exadg/issues/383#issuecomment-1467580096, the question came up whether we use `read_dof_values_plain()` correctly in ExaDG, in particular the question whether the current implementation will support hanging nodes or will do wrong calculations in case of hanging nodes.

### LINKED ISSUE #446: Strange const_cast of const reference without reset
In `Poisson::Operator::solve`, one input parameter is `VectorType const & rhs`, but in the function body, the constness is then cast away with `const_cast` and the values in `rhs` are changed.

See:

https://github.com/exadg/exadg/blob/85e1ce3c9c678f97e6f048b47b36b105ea028202/include/exadg/poisson/spatial_discretization/operator.cpp#L513-L519

Firstly, this is dangerous and would result in undefined behavior if the passed in object `rhs` were actually `const`, which it luckily isn't in this case, hence it works.

Secondly, the values in `rhs` are not reset again, so getting a `const` reference is actually lying to the outside world. The vector `rhs` is not used elsewhere in the function, which could have somehow justified its being a `const` reference.

So I would suggest just modifying the input parameter to `VectorType & rhs` and ditch the `const_cast`.

Please correct me if I'm missing something.

nfehn:
This PR aims to realize local refinements / hanging-nodes for continuous Galerkin discretizations for nonlinear problems (e.g. nonlinear structural mechanisms) with special requirements in terms of constraints / inhomogeneous BC as discussed in issue #386.

closes #386 

closes #446 (@bugrahantemur please take a look at the changes and let me know if you do not agree)

TODOs:
- ~~introduce second `AffineConstraints` object for `Poisson` spatial discretization (which is the second case of continuous Galerkin discretizations in ExaDG apart from the `Structure` module; other modules use continuous Galerkin only inside MG, which is unproblematic since the operators used in MG preconditioning are always homogeneous)~~
- ~~`set_inhomogeneous_boundary_values()` will be removed from `OperatorBase::evaluate()` functions. Make sure that all modules currently calling these functions are adapted accordingly.~~
- ~~initialize `dof_index_inhomogeneous` with an invalid number and make sure that a correct value has been set whenever using `dof_index_inhomogeneous`~~

### DIFF
--- a/include/exadg/operators/operator_base.cpp
+++ b/include/exadg/operators/operator_base.cpp
@@ -889,6 +889,19 @@ OperatorBase<dim, Number, n_components>::reinit_face(IntegratorFace &   integrat
 {
   integrator_m.reinit(face);
   integrator_p.reinit(face);
+
+  reinit_face_additional(integrator_m, integrator_p, face);
+}
+
+template<int dim, typename Number, int n_components>
+void
+OperatorBase<dim, Number, n_components>::reinit_face_additional(IntegratorFace &   integrator_m,
+                                                                IntegratorFace &   integrator_p,
+                                                                unsigned int const face) const
+{
+  (void)integrator_m;
+  (void)integrator_p;
+  (void)face;
 }
 
 template<int dim, typename Number, int n_components>
@@ -897,6 +910,18 @@ OperatorBase<dim, Number, n_components>::reinit_boundary_face(IntegratorFace &  
                                                               unsigned int const face) const
 {
   integrator_m.reinit(face);
+
+  reinit_boundary_face_additional(integrator_m, face);
+}
+
+template<int dim, typename Number, int n_components>
+void
+OperatorBase<dim, Number, n_components>::reinit_boundary_face_additional(
+  IntegratorFace &   integrator_m,
+  unsigned int const face) const
+{
+  (void)integrator_m;
+  (void)face;
 }
 
 template<int dim, typename Number, int n_components>
@@ -1003,6 +1028,24 @@ OperatorBase<dim, Number, n_components>::reinit_face_cell_based(
   {
     integrator_p.reinit(cell, face);
   }
+
+  reinit_face_cell_based_additional(integrator_m, integrator_p, cell, face, boundary_id);
+}
+
+template<int dim, typename Number, int n_components>
+void
+OperatorBase<dim, Number, n_components>::reinit_face_cell_based_additional(
+  IntegratorFace &                 integrator_m,
+  IntegratorFace &                 integrator_p,
+  unsigned int const               cell,
+  unsigned int const               face,
+  dealii::types::boundary_id const boundary_id) const
+{
+  (void)integrator_m;
+  (void)integrator_p;
+  (void)cell;
+  (void)face;
+  (void)boundary_id;
 }
 
 template<int dim, typename Number, int n_components>
--- a/include/exadg/operators/operator_base.cpp
+++ b/include/exadg/operators/operator_base.cpp
@@ -344,8 +344,12 @@ OperatorBase<dim, Number, n_components>::rhs_add(VectorType & rhs) const
     src_tmp.reinit(rhs, false);
     dst_tmp.reinit(rhs, false);
 
-    // Since src_tmp = 0, the function evaluate_add() only computes the inhomogeneous part of the
-    // operator.
+    // Set constrained degrees of freedom according to inhomogeneous Dirichlet boundary conditions.
+    //  The rest of the vector remains unchanged.
+    set_inhomogeneous_boundary_values(src_tmp);
+
+    // Since src_tmp = 0 apart from inhomogeneous boundary data, the function evaluate_add() only
+    // computes the inhomogeneous part of the operator.
     this->evaluate_add(dst_tmp, src_tmp);
     // Minus sign since we compute the contribution to the right-hand side.
     rhs -= dst_tmp;
@@ -386,12 +390,6 @@ OperatorBase<dim, Number, n_components>::evaluate_add(VectorType &       dst,
     AssertThrow(data.dof_index_inhomogeneous != dealii::numbers::invalid_unsigned_int,
                 dealii::ExcMessage("dof_index_inhomogeneous is uninitialized."));
 
-    // Set constrained degrees of freedom according to inhomogeneous Dirichlet boundary conditions.
-    //  The rest of the vector remains unchanged.
-    // TODO: do this outside OperatorBase::evaluate_add()/rhs_add()
-    VectorType src_inhom = src;
-    set_inhomogeneous_boundary_values(src_inhom, time);
-
     // Start with a zero destination vector.
     VectorType tmp;
     tmp.reinit(dst, false);
@@ -405,11 +403,11 @@ OperatorBase<dim, Number, n_components>::evaluate_add(VectorType &       dst,
                         &This::boundary_face_loop_inhom_operator,
                         this,
                         tmp,
-                        src_inhom);
+                        src);
     }
     else
     {
-      matrix_free->cell_loop(&This::cell_loop_full_operator, this, tmp, src_inhom);
+      matrix_free->cell_loop(&This::cell_loop_full_operator, this, tmp, src);
     }
 
     // This is the function of type evaluate_add().
@@ -944,11 +942,10 @@ OperatorBase<dim, Number, n_components>::do_boundary_integral_continuous(
 
 template<int dim, typename Number, int n_components>
 void
-OperatorBase<dim, Number, n_components>::set_inhomogeneous_boundary_values(VectorType & solution,
-                                                                           double const time) const
+OperatorBase<dim, Number, n_components>::set_inhomogeneous_boundary_values(
+  VectorType & solution) const
 {
   (void)solution;
-  (void)time;
 
   AssertThrow(
     false,
--- a/include/exadg/operators/operator_base.cpp
+++ b/include/exadg/operators/operator_base.cpp
@@ -348,7 +348,7 @@ OperatorBase<dim, Number, n_components>::rhs_add(VectorType & rhs) const
     matrix_free->initialize_dof_vector(temp2, data.dof_index);
     // TODO: temp2 might now contain non-zero values in Dirichlet degrees of freedom, which were
     // previously part of AffineConstraints but no longer if we use dof_index_inhomogeneous here!?
-    matrix_free->cell_loop(&This::cell_loop_inhom_operator, this, temp2, temp1);
+    matrix_free->cell_loop(&This::cell_loop_full_operator, this, temp2, temp1);
     rhs -= temp2;
   }
 }
@@ -366,6 +366,9 @@ void
 OperatorBase<dim, Number, n_components>::evaluate_add(VectorType &       dst,
                                                       VectorType const & src) const
 {
+  // TODO: since we now use dof_index_inhomogeneous in some routines (where Dirichlet degrees of
+  // freedom are no longer part of AffineConstraints, this function might write to degrees of
+  // freedom in the dst vector that have not been touched before.
   if(is_dg)
   {
     if(evaluate_face_integrals())
@@ -386,7 +389,7 @@ OperatorBase<dim, Number, n_components>::evaluate_add(VectorType &       dst,
   {
     if(evaluate_face_integrals())
     {
-      matrix_free->loop(&This::cell_loop,
+      matrix_free->loop(&This::cell_loop_full_operator,
                         &This::face_loop_empty,
                         &This::boundary_face_loop_inhom_operator,
                         this,
@@ -395,7 +398,7 @@ OperatorBase<dim, Number, n_components>::evaluate_add(VectorType &       dst,
     }
     else
     {
-      matrix_free->cell_loop(&This::cell_loop, this, dst, src);
+      matrix_free->cell_loop(&This::cell_loop_full_operator, this, dst, src);
     }
   }
 }
@@ -1030,12 +1033,15 @@ OperatorBase<dim, Number, n_components>::create_standard_basis(unsigned int     
 
 template<int dim, typename Number, int n_components>
 void
-OperatorBase<dim, Number, n_components>::cell_loop_inhom_operator(
+OperatorBase<dim, Number, n_components>::cell_loop_full_operator(
   dealii::MatrixFree<dim, Number> const & matrix_free,
   VectorType &                            dst,
   VectorType const &                      src,
   Range const &                           range) const
 {
+  AssertThrow(not is_dg,
+              dealii::ExcMessage("This function should not be called for is_dg = true."));
+
   IntegratorCell integrator =
     IntegratorCell(matrix_free, this->data.dof_index_inhomogeneous, this->data.quad_index);
 
--- a/include/exadg/operators/operator_base.cpp
+++ b/include/exadg/operators/operator_base.cpp
@@ -331,11 +331,11 @@ OperatorBase<dim, Number, n_components>::rhs_add(VectorType & rhs) const
 
   if(not is_dg)
   {
-    // Set constrained degrees of freedom according to Dirichlet boundary conditions. The rest of
-    // the vector contains zeros.
+    // Set constrained degrees of freedom according to inhomogeneous Dirichlet boundary conditions.
+    //  The rest of the vector contains zeros.
     VectorType temp1;
     matrix_free->initialize_dof_vector(temp1, data.dof_index);
-    set_constrained_values(temp1, time);
+    set_inhomogeneous_boundary_values(temp1, time);
 
     // Perform matrix-vector product using read_dof_values_plain() with temp1 as src-vector to
     // obtain the inhomogeneous action of the operator. Subsequently, we shift the resulting vector
@@ -923,15 +923,16 @@ OperatorBase<dim, Number, n_components>::do_boundary_integral_continuous(
 
 template<int dim, typename Number, int n_components>
 void
-OperatorBase<dim, Number, n_components>::set_constrained_values(VectorType & solution,
-                                                                double const time) const
+OperatorBase<dim, Number, n_components>::set_inhomogeneous_boundary_values(VectorType & solution,
+                                                                           double const time) const
 {
   (void)solution;
   (void)time;
 
-  AssertThrow(false,
-              dealii::ExcMessage(
-                "OperatorBase::set_constrained_values() has to be overwritten by derived class!"));
+  AssertThrow(
+    false,
+    dealii::ExcMessage(
+      "OperatorBase::set_inhomogeneous_boundary_values() has to be overwritten by derived class!"));
 }
 
 template<int dim, typename Number, int n_components>
--- a/include/exadg/operators/operator_base.cpp
+++ b/include/exadg/operators/operator_base.cpp
@@ -213,7 +213,7 @@ OperatorBase<dim, Number, n_components>::initialize_dof_vector(VectorType & vect
 
 template<int dim, typename Number, int n_components>
 void
-OperatorBase<dim, Number, n_components>::set_constrained_values_to_zero(VectorType & vector) const
+OperatorBase<dim, Number, n_components>::set_constrained_dofs_to_zero(VectorType & vector) const
 {
   for(unsigned int const constrained_index :
       matrix_free->get_constrained_dofs(this->data.dof_index))
@@ -344,7 +344,7 @@ OperatorBase<dim, Number, n_components>::rhs_add(VectorType & rhs) const
     src_tmp.reinit(rhs, false);
     dst_tmp.reinit(rhs, false);
 
-    // Since src_tmp = 0, the function evaluate_add only computes the inhomogeneous part of the
+    // Since src_tmp = 0, the function evaluate_add() only computes the inhomogeneous part of the
     // operator.
     this->evaluate_add(dst_tmp, src_tmp);
     // Minus sign since we compute the contribution to the right-hand side.
@@ -385,6 +385,7 @@ OperatorBase<dim, Number, n_components>::evaluate_add(VectorType &       dst,
   {
     // Set constrained degrees of freedom according to inhomogeneous Dirichlet boundary conditions.
     //  The rest of the vector remains unchanged.
+    // TODO: do this outside OperatorBase::evaluate_add()/rhs_add()
     VectorType src_inhom = src;
     set_inhomogeneous_boundary_values(src_inhom, time);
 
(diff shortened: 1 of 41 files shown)


nfehn:
@kronbichler the lengthy comment in commit https://github.com/exadg/exadg/pull/506/commits/f02f2c1f8ff6270dba32d50f6c535b2f71f69534 is kind of a summary of our discussion from #386 (... or what I took home from this discussion). Could you have a look at this commit?

include/exadg/operators/operator_base.cpp
```diff
@@ -1031,7 +1032,7 @@ OperatorBase<dim, Number, n_components>::cell_loop_dbc(
   for(auto cell = range.first; cell < range.second; ++cell)
   {
     this->reinit_cell(cell);
-
+    // TODO
     integrator->read_dof_values_plain(src);
```
nfehn, line 1036: I think we need a separate `IntegratorCell` here, initialized with `dof_index_inhomogeneous`. The problem is that `reinit_cell()` in the line above is hard-coded for the member variable `integrator`, which is initialized with the standard `dof_index`. Hence, issue #507 needs to be addressed.

include/exadg/structure/spatial_discretization/operator.h
```diff
@@ -379,12 +385,30 @@ class Operator : public dealii::Subscriptor, public Interface::Operator<Number>
...
+  // dealii::MatrixFree/FEEvaluation. When using the standard AffineConstraints object including
+  // Dirichlet boundary conditions, inhomogeneous boundary data would be ignored by
+  // dealii::FEEvaluation::read_dof_values(). While dealii::FEEvaluation::read_dof_values_plain()
+  // would take into account inhomogeneous Dirichlet data using the standard AffineConstraints
+  // object, hanging-node constraints would not be resolved correctly. The solution/workaround is to
+  // use dealii::FEEvaluation::read_dof_values() for a correct handling of hanging nodes, but to
+  // exclude Dirichlet degrees of freedom from the AffineConstraints object so that it is possible
+  // to read inhomogeneous boundary data (to be set correctly in separate routines) when calling
+  // dealii::FEEvaluation::read_dof_values().
+  dealii::AffineConstraints<Number> affine_constraints_periodicity_and_hanging_nodes;
```
kronbichler, line 401: This sounds good.

include/exadg/structure/spatial_discretization/operator.cpp
```diff
@@ -134,15 +134,26 @@ Operator<dim, Number>::distribute_dofs()
 
   // affine constraints
   affine_constraints.clear();
+  if(param.large_deformation)
```
nfehn, line 137: I think we need this for linear problems calling `rhs()/rhs_add()` as well, right?

include/exadg/structure/spatial_discretization/operator.h
nfehn: The restriction to nonlinear problems in this comment is probably not corrrect, see comment/question https://github.com/exadg/exadg/pull/506/files#r1243604142.

include/exadg/poisson/spatial_discretization/operator.cpp
```diff
@@ -509,26 +526,22 @@ Operator<dim, n_components, Number>::solve(VectorType &       sol,
...
-  if(param.spatial_discretization == SpatialDiscretization::CG)
+    // Set Dirichlet degrees of freedom according to Dirichlet boundary condition.
+    laplace_operator.set_time(time);
+    laplace_operator.set_inhomogeneous_boundary_values(sol);
+  }
+  else
   {
-    laplace_operator.set_constrained_values(sol, time);
+    n_iterations = iterative_solver->solve(sol, rhs);
   }
```
bugrahantemur, line 543: This definitely closes the issue #446 :+1: . Setting the boundary values after the solution step (in contrast to before the solution step in the previous implementation) is intended right?

Apart from that I suggest writing:

```cpp
unsigned int const n_iterations = iterative_solver->solve(sol, rhs);

if (param.spatial_discretization == SpatialDiscretization::CG)
{
    // Set Dirichlet degrees of freedom according to Dirichlet boundary condition.
    laplace_operator.set_time(time);
    laplace_operator.set_inhomogeneous_boundary_values(sol);
}

return n_iterations;
```

nfehn:
The code seems to crash in `OperatorBase::add_diagonal()` with the error message

```bash
An error occurred in line <7847> of file </home/code/dealii/include/deal.II/matrix_free/fe_evaluation.h> in function
    void dealii::FEEvaluation<dim, fe_degree, n_q_points_1d, n_components_, Number, VectorizedArrayType>::evaluate(dealii::EvaluationFlags::EvaluationFlags) [with int dim = 3; int fe_degree = -1; int n_q_points_1d = 0; int n_components_ = 1; Number = float; VectorizedArrayType = dealii::VectorizedArray<float, 8>]
The violated condition was: 
    this->dof_values_initialized == true
Additional information: 
    (none)
```

What about the TODO here: https://github.com/exadg/exadg/blob/980cc102c71166964d15c58cebad5557e7bda0b3/include/exadg/operators/operator_base.cpp#L432-L448

include/exadg/poisson/spatial_discretization/operator.cpp
nfehn: Yes, this is intended. Before, we have actually set the dofs twice. But there should actually be no need to do it before solving (because the inhomogeneous parts have been shifted to the rhs and we only solve for the non-constrained dofs; while constrained dofs are set after solving). So I think we should avoid such non-unique logic that we had before.

Thanks for the suggestion! Several changes ended up in an ugly code ;)

kronbichler:
The code inside deal.II is complicated, because it tracks down the constraints and computes their influence on the local degrees of freedom, doing a long preparation https://github.com/dealii/dealii/blob/3bba74760bff6724e1766bf22e4ee06eaf98bfc1/include/deal.II/matrix_free/tools.h#L422-L854 and then https://github.com/dealii/dealii/blob/3bba74760bff6724e1766bf22e4ee06eaf98bfc1/include/deal.II/matrix_free/tools.h#L872-L908 to extract the entries. I recently rewrote around 300 of the 500 lines of that code in https://github.com/dealii/dealii/pull/15135 and https://github.com/dealii/dealii/pull/15147 because the initial implementation was performing poorly - the details are not so important, but it shows that re-writing this in ExaDG is probably not a good idea.

I see two options here:
- We equip deal.II with a lambda that triggers once per cell next to the `phi->reinit(cell)` inside the first link I posted. This leads to a second lambda function next to the operation to be provided or where we can provide a default. The problem here is that if we want to support both deal.II-9.5 and master, we need an `ifdef` and still option #2.
- We keep track of what `cell` has been used before,
```c++
  unsigned int last_cell = numbers::invalid_unsigned_int;

 dealii::MatrixFreeTools:: 
   compute_diagonal<dim, -1, 0, n_components, Number, dealii::VectorizedArray<Number>>( 
     *matrix_free, 
     diagonal, 
     [&](auto & integrator) -> void { 
       if (last_cell != integrator.get_current_cell_index())
         this->reinit_cell(integrator, integrator.get_current_cell_index()); 
       ...
```
and this way only call it once per cell, not once per column. We still use another evaluator here, but that one is hidden inside `compute_diagonal` and nothing we can do something about.

I usually do not like lambda functions, because they are also complicated for a user to understand. But here it might make sense to go down that way because it is a bit cleaner for ExaDG nonetheless in my opinion.

nfehn:
We should only do it once per cell and not once per column. So the lambda function I was thinking of would be called by the deal.II code only once per cell, just as `integrator.reinit(cell)`.

Overall this is in line with my general design ideas/wishes regarding generic marix-free operators: The user wants to specify (i) `reinit()` functions and (ii) what needs to be done on a quadrature point (the "kernel" operation). The evaluate/integrate steps as well as the quadrature point loop should be generic code not visible in user code ... At least this is what I would formulate as a design goal for a generic matrix-free code.

kronbichler:
> Overall this is in line with my general design ideas/wishes regarding generic marix-free operators: The user wants to specify (i) reinit() functions and (ii) what needs to be done on a quadrature point (the "kernel" operation).

Yes indeed, this is what one would like to have, fully agree.

nfehn:
I would like to avoid the above variant with `last_cell`, because the code is hard to understand from the user perspective if one does not know the implementation in deal.II. The problem is that `integrator` comes from the deal.II code, while `last_cell` is injected/catched from the ExaDG code and the body of the lambda does not only depend on both of them, it may also change both variables (i.e. they are not const). I think this is exactly what makes lambdas hard-to-understand to error-prone, so I agree with the general comment on lambdas. One should not mis- or over-use them.

nfehn:
@kronbichler I wonder why the tests have passed in PR #508 given that the line with `reinit_cell()` has been changed in PR #508.

nfehn:
@kronbichler I think I resolved the `reinit_cell()` problem for now in ExaDG. I also updated the multigrid preconditioner for `Structure` (adding another `AffineConstraints` object). Ready for review from my side ...

include/exadg/operators/operator_base.cpp
```diff
@@ -380,19 +385,31 @@ OperatorBase<dim, Number, n_components>::evaluate_add(VectorType &       dst,
...
     if(evaluate_face_integrals())
     {
-      matrix_free->loop(&This::cell_loop,
+      matrix_free->loop(&This::cell_loop_full_operator,
                         &This::face_loop_empty,
                         &This::boundary_face_loop_inhom_operator,
                         this,
-                        dst,
+                        tmp,
                         src);
```
kronbichler, line 404: Question: If you run `MatrixFree::loop` or `MatrixFree::cell_loop`, the default behavior is to add into the `dst` vector. This means that we should be able to avoid the temporary vector here and thus save a bit of memory capacity and memory transfer (the extra copy does 4 vector accesses). We do the same in the DG case above.

include/exadg/operators/operator_base.cpp
```diff
@@ -849,6 +866,19 @@ OperatorBase<dim, Number, n_components>::reinit_cell(IntegratorCell &   integrat
                                                      unsigned int const cell) const
 {
   integrator.reinit(cell);
+
+  reinit_cell_additional(integrator, cell);
```
kronbichler, line 870: Is `additional` a good name? I would intuitively choose `auxiliary`, but maybe there are even more precise words?

include/exadg/operators/operator_base.cpp
```diff
@@ -849,6 +866,19 @@ OperatorBase<dim, Number, n_components>::reinit_cell(IntegratorCell &   integrat
...
+
+template<int dim, typename Number, int n_components>
+void
+OperatorBase<dim, Number, n_components>::reinit_cell_additional(IntegratorCell &   integrator,
+                                                                unsigned int const cell) const
+{
+  (void)integrator;
+  (void)cell;
+
+  // overwrite this function in derived classes if additional initialization is necessary
```
kronbichler, line 881: Regarding the name, should we use the common C++ idiom,
```suggestion
  // override this function in derived classes if additional initialization is necessary
```

include/exadg/operators/operator_base.cpp
```diff
@@ -923,15 +978,15 @@ OperatorBase<dim, Number, n_components>::do_boundary_integral_continuous(
...
   (void)solution;
-  (void)time;
 
-  AssertThrow(false,
-              dealii::ExcMessage(
-                "OperatorBase::set_constrained_values() has to be overwritten by derived class!"));
+  AssertThrow(
+    false,
+    dealii::ExcMessage(
+      "OperatorBase::set_inhomogeneous_boundary_values() has to be overwritten by derived class!"));
```
kronbichler, line 989:
```suggestion
      "OperatorBase::set_inhomogeneous_boundary_values() has to be overridden by derived class!"));
```
Here and elsewhere?

include/exadg/operators/operator_base.h
```diff
@@ -229,10 +243,19 @@ class OperatorBase : public dealii::Subscriptor
    * rhs only make sense for linear operators (but they have e.g. no meaning for linearized
    * operators of nonlinear problems). For this reason, these functions are currently defined
    * 'virtual' to provide the opportunity to override and assert these functions in derived classes.
+   *
+   * For continuous Galerkin discretizations, this function calls internally the member function
```
kronbichler, line 247: Sorry for nit-picking, but isn't it more grammatically correct to use
```suggestion
   * For continuous Galerkin discretizations, this function internally calls the member function
```

include/exadg/operators/operator_base.h
```diff
@@ -243,10 +266,18 @@ class OperatorBase : public dealii::Subscriptor
    * (but they have e.g. no meaning for linearized operators of nonlinear problems). For this
    * reason, these functions are currently defined 'virtual' to provide the opportunity to override
    * and assert these functions in derived classes.
+   *
+   * Unlike the function rhs(), this function does internally not call the function
```
kronbichler, line 270:
```suggestion
   * Unlike the function rhs(), this function does not internally call the function
```

include/exadg/operators/operator_base.h
```diff
@@ -296,21 +327,29 @@ class OperatorBase : public dealii::Subscriptor
...
 
+  void
+  reinit_face_cell_based(IntegratorFace &                 integrator_m,
+                         IntegratorFace &                 integrator_p,
+                         unsigned int const               cell,
+                         unsigned int const               face,
+                         dealii::types::boundary_id const boundary_id) const;
+
+  /*
+   * These methods have to be overwritten by derived classes because these functions are
```
kronbichler, line 349:
```suggestion
   * These methods have to be overridden by derived classes because these functions are
```

include/exadg/operators/operator_base.h
```diff
@@ -420,13 +469,19 @@ class OperatorBase : public dealii::Subscriptor
                         IntegratorFace & integrator_2) const;
 
   /*
-   * This function applies Dirichlet BCs for continuous Galerkin discretizations.
+   * This function calculates cell integrals for the full (= homogeneous + inhomogeneous) part,
+   * where inhomogeneous part may occur for continuous Galerkin discretizations due to inhomogeneous
+   * Dirichlet BCs. A prerequisite to call this function is to set inhomogeneous Dirichlet degrees
+   * of freedom appropriate in the DoF-vector src. In case the DoF-vector src is zero apart from the
```
kronbichler, line 475:
```suggestion
   * of freedom appropriately in the DoF-vector src. In case the DoF-vector src is zero apart from the
```

include/exadg/poisson/spatial_discretization/operator.cpp
```diff
@@ -163,6 +163,9 @@ Operator<dim, n_components, Number>::distribute_dofs()
 
       dealii::DoFTools::make_periodicity_constraints<dim, dim, Number>(periodic_faces_dof,
                                                                        affine_constraints);
+
+      dealii::DoFTools::make_periodicity_constraints<dim, dim, Number>(
+        periodic_faces_dof, affine_constraints_periodicity_and_hanging_nodes);
     }
```
kronbichler, line 169: The above two functions are potentially expensive. I recommend to only call the above functions for `affine_constraints_periodicity_and_hanging_nodes` above, then call `close()` on that object, and finally call
```c++
  affine_constraints.copy_from(affine_constraints_periodicity_and_hanging_nodes);
```
before going on with the code below. I think that also documents more clearly that the regular constraints are an extended variant, and always have the same baseline constraints. (This would avoid the need for the commend you add in line 193 in my opinion.)

include/exadg/structure/preconditioners/multigrid_preconditioner.cpp
```diff
@@ -126,6 +126,35 @@ MultigridPreconditioner<dim, Number>::update()
...
+  {
+    Map_DBC               dirichlet_bc_empty;
+    Map_DBC_ComponentMask dirichlet_bc_empty_component_mask;
+    this->do_initialize_dof_handler_and_constraints(false,
+                                                    fe,
+                                                    dirichlet_bc_empty,
+                                                    dirichlet_bc_empty_component_mask,
+                                                    dof_handlers_inhomogeneous,
+                                                    constraints_inhomogeneous);
+  }
```
kronbichler, line 155: Just for clarity: We need this additional step because we might interpolate some fields to the multigrid levels, where using the Dirichlet boundary conditions are wrong. In case we had a different/better way to transfer the linearization point (or, if it ever gets relevant, a possible ALE mesh deformation we might impose in a structural problem), we would not need these additions, right? That is certainly a different problem for another pull request, but I think it is at least very clear where these additional data structures would need to be eliminated if we do not need them any more.

include/exadg/structure/spatial_discretization/operator.cpp
```diff
@@ -152,6 +158,9 @@ Operator<dim, Number>::distribute_dofs()
 
     dealii::DoFTools::make_periodicity_constraints<dim, dim, Number>(periodic_faces_dof,
                                                                      affine_constraints);
+
+    dealii::DoFTools::make_periodicity_constraints<dim, dim, Number>(
+      periodic_faces_dof, affine_constraints_periodicity_and_hanging_nodes);
```
kronbichler, line 163: Same as above.

include/exadg/structure/spatial_discretization/operator.h
```diff
@@ -213,24 +214,23 @@ class Operator : public dealii::Subscriptor, public Interface::Operator<Number>
   void
   prescribe_initial_velocity(VectorType & velocity, double const time) const final;
 
+  /*
+   * This computes computes the intial acceleration field by evaluating all PDE terms for the given
```
kronbichler, line 218:
```suggestion
   * This computes computes the initial acceleration field by evaluating all PDE terms for the given
```

include/exadg/structure/spatial_discretization/operator.h
```diff
@@ -379,12 +393,25 @@ class Operator : public dealii::Subscriptor, public Interface::Operator<Number>
...
-  std::string const dof_index                = "dof";
-  std::string const dof_index_mass           = "dof_mass";
+  // To treat inhomogeneous Dirichlet BCs correctly in the context of matrix-free operator
+  // evaluation using dealii::MatrixFree/FEEvaluation, we need a separate AffineConstraints
+  // object containing only periodicity and hanging node constraints.
+  // When using the standard AffineConstraints object including Dirichlet boundary conditions,
+  // inhomogeneous boundary data would be ignored by dealii::FEEvaluation::read_dof_values().
+  // While dealii::FEEvaluation::read_dof_values_plain() would take into account inhomogeneous
+  // Dirichlet data using the standard AffineConstraints object, hanging-node constraints would
+  // not be resolved correctly.
```
kronbichler, line 404: Let me lay out two alternative designs to this comment as a discussion before we finalize the design for now, not necessary to change anything here:
- `read_dof_values_plain` could be kept by calling `AffineConstraints::distribute` on the object containing only hanging nodes and periodicity constraints, but I think the way we choose now is clearer in terms of the evaluators.
- Of course it would be even nicer if we could encode two different sets of constraints into a single `dof_index` of `MatrixFree` and call `read_dof_values()` and `read_dof_values_unconstrained` (not reflecting the more strict category), but I do not think we are there yet. I am not even sure if it can be realized without ambiguity from the perspective of ExaDG in all cases we have.

As I said, I do not believe we want to go for a different route, but these are the two other reasonable options I see, so I wanted to state them once again for the completeness of discussion.

kronbichler (approved):
I mostly agree with this PR. Below, I list some cosmetic changes, plus some code design issues regarding the setup of constraints, plus one discussion as a point of documentation.

include/exadg/operators/operator_base.cpp
nfehn: The wording should mean "in addition to what is reinitialized in the base class". I am not sure if "auxiliary" can reflect this? An alternative I thought about was "derived", in the sense that this is reinitialization done by "derived" classes.

kronbichler: Yes, I like "derived".

nfehn: I consider both options equivalent, so I will change this. Overall, these are all private member functions, so I think it is not super-important right now that we fix the naming.

include/exadg/poisson/spatial_discretization/operator.cpp
nfehn: see https://github.com/exadg/exadg/pull/506/commits/db18a5ec73c5460a1bc5f524ef5f7c86a9570738

include/exadg/operators/operator_base.cpp
nfehn: I guess this code is some leftover, because we had - at least temporarily - also other functionality in this function.

nfehn: see https://github.com/exadg/exadg/pull/506/commits/c45b4bbd3acac43cbc8f0059d07505f60bd629fa

include/exadg/structure/preconditioners/multigrid_preconditioner.cpp
nfehn: I do not fully understand this comment. Maybe let's discuss this in person.

include/exadg/poisson/spatial_discretization/operator.cpp
nfehn: I think it makes sense to introduce a utility function and be more strict with "don't repeat yourself" even if the code is just a few calls to dealii utilities ... see https://github.com/exadg/exadg/pull/506/commits/d2d07b9f51f07a1fae918baddc1ed4de014097d9.

include/exadg/structure/preconditioners/multigrid_preconditioner.cpp
nfehn: Please also note commit https://github.com/exadg/exadg/pull/506/commits/c3dc80b3fed2e58b759358138869363fa0d55733

include/exadg/poisson/spatial_discretization/operator.cpp
```diff
@@ -135,57 +136,44 @@ Operator<dim, n_components, Number>::distribute_dofs()
...
+  // Affine constraints are only relevant for continuous Galerkin discretizations.
   if(param.spatial_discretization == SpatialDiscretization::CG)
   {
+    affine_constraints_periodicity_and_hanging_nodes.clear();
+
+    add_hanging_node_and_periodicity_constraints(affine_constraints_periodicity_and_hanging_nodes,
+                                                 *this->grid,
+                                                 dof_handler);
+
+    affine_constraints_periodicity_and_hanging_nodes.close();
```
nfehn, line 148: this line seems to have introduced a bug in debug mode, see PR #535, since we later call `copy_from` and then add additional constraints.

nfehn: The solution is to call `close()` in the end once all the constraints have been added (which somewhat hinders writing modular code ...)

PR #512, opened 2023-07-03
slightly refactor throughput studies for Poisson module

nfehn:


include/exadg/poisson/spatial_discretization/operator.h
```diff
@@ -115,34 +118,6 @@ class Operator : public dealii::Subscriptor
...
-                     MPI_Comm const &                           mpi_comm) const;
-
-  void
-  calculate_system_matrix(dealii::PETScWrappers::MPI::SparseMatrix & system_matrix) const;
-
-  void
-  vmult_matrix_based(VectorTypeDouble &                               dst,
-                     dealii::PETScWrappers::MPI::SparseMatrix const & system_matrix,
-                     VectorTypeDouble const &                         src) const;
-#endif
```
nfehn, line 144: This has not been used at all in the previous code version.

I remove this here since this functionality has not been used and since `DEAL_II_WITH_...` should not be visible in such high-level code in my opinion. We would also have to explicitly differentiate between trilinos/petsc in `Driver`, which has not been done so far (which means that the previous implementation in `Driver` is also incomplete).

PR #513, opened 2023-07-04
introduce utility function for creation of dealii::FiniteElement

nfehn:
The same code has been used at multiple places. This PR aims to eliminate code duplications regarding the creation of `dealii::FiniteElement`.

### DIFF
--- a/include/exadg/convection_diffusion/preconditioners/multigrid_preconditioner.cpp
+++ b/include/exadg/convection_diffusion/preconditioners/multigrid_preconditioner.cpp
@@ -34,6 +34,7 @@ namespace ConvDiff
 template<int dim, typename Number>
 MultigridPreconditioner<dim, Number>::MultigridPreconditioner(MPI_Comm const & mpi_comm)
   : Base(mpi_comm),
+    degree_velocity(1),
     pde_operator(nullptr),
     mg_operator_type(MultigridOperatorType::Undefined),
     mesh_is_moving(false)
@@ -53,6 +54,8 @@ MultigridPreconditioner<dim, Number>::initialize(
   Map_DBC const &                             dirichlet_bc,
   Map_DBC_ComponentMask const &               dirichlet_bc_component_mask)
 {
+  this->degree_velocity = fe.degree;
+
   this->pde_operator     = &pde_operator;
   this->mg_operator_type = mg_operator_type;
   this->mesh_is_moving   = mesh_is_moving;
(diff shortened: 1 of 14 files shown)


include/exadg/operators/finite_element.h
```diff
@@ -0,0 +1,128 @@
...
+/**
+ * Creates and returns a shared pointer to dealii::FiniteElement<dim> depending on the element type,
+ * the type of function space, the number of components of the solution field, and the polynomial
+ * degree of the shape functions. Regarding the function space, the code assumes an H^1 conforming
+ * space (Continuous Galerkin) if is_dg (= L^2-conforming space, Discontinuous Galerkin) is false.
+ */
+template<int dim>
+std::shared_ptr<dealii::FiniteElement<dim>>
+create_finite_element(ElementType const & element_type,
+                      bool const          is_dg,
```
kronbichler, line 46: We don't have the infrastructure right now, but I would prefer to pass
```suggestion
                      SpatialDiscretization(Type) const          discretization_type,
```
here, rather than `true/false`. When I say we are not there, I mean that we currently define the `SpatialDiscretization` enum only in some module, but not everywhere. Maybe we also do not need to re-introduce this parameter in ExaDG, as deal.II already has an enum with the various conformity settings https://github.com/dealii/dealii/blob/61a8f375f62f7ada3edd1f269e3dce1bd5973ccf/include/deal.II/fe/fe_data.h#L262-L296. We probably do not want to write `dealii::FiniteElementData<dim>::Conformity` but might introduce a shortcut. If we have consensus on this topic, I can introduce that in a follow-up.

include/exadg/operators/finite_element.h
```diff
@@ -0,0 +1,128 @@
...
+                      unsigned int const  n_components,
+                      unsigned int const  degree)
+{
+  std::shared_ptr<dealii::FiniteElement<dim>> fe;
+
+  if(n_components == 1)
+  {
```
kronbichler, line 53: I think technically we do not need to make the distinction between a scalar element, say `FE_Q<dim>(degree)`, and an `FESystem` with 1 component, `FESystem<dim>(FE_Q<dim>(degree), 1)`. The matrix-free evaluators will reduce this to the same code (at least I would assume so, otherwise it is a bug), so our central routines do not differ. The code working with `dealii::FEValues` will probably be somewhat more expensive because `dealii::FEValues` does some extra roundtrip for `FESystem` to copy things around, but we only use it in uncritical places (from a performance point of view) that I would be tempted to uniformly use one setting.

include/exadg/operators/finite_element.h
```diff
@@ -0,0 +1,128 @@
...
+    if(is_dg)
+    {
+      if(element_type == ElementType::Hypercube)
+      {
+        fe = std::make_shared<dealii::FE_DGQ<dim>>(degree);
```
kronbichler, line 58: I think that in the longer run, we would probably want to increase the flexibility here. For example, in the future we will probably want to use `FE_DGQHermite` in some circumstances, and `FE_DGQArbitraryNodes<dim>(QGauss<dim>(degree + 1))` in some others because they might be beneficial for certain setups. That is a longer-term objective that cannot be resolved in this PR, but I would like to plan for how we would support it. Maybe the comment above regarding `Conformity` is then not the right choice, but rather an enum encoding all types. The important part is that we should have a name that does not care about whether we have `Hypercube`, `Simplex` or (at a later stage), `Wedge` and `Pyramid` elements for supporting mixed geometries. The advantage would then be a more flexible choice in the space at the place where we call `create_finite_element` by this enum, and yet be flexible in our interfaces. What do you think?

kronbichler (approved):
I have some comments regarding the longer-term plans as explained above, but regarding the present changes I agree with this PR.

include/exadg/operators/finite_element.h
nfehn: Yes, let's wait with this and do it somewhere else. Note also that `OperatorBase` and `MultigridPreconditionerBase` use `bool is_dg`. This is why I intentionally followed the current procedure also in the present PR.

include/exadg/operators/finite_element.h
nfehn: I think one would need several enums. One for CG/DG/Hdiv/etc., one for GL/GLL/Hermite/etc., one for the element type. Not all combinations of these enums will be supported, which needs to be asserted in my opinion.

include/exadg/operators/finite_element.h
nfehn: Indeed, I was unsure here and did not want to introduce a performance penalty for the scalar case. If you want, I can adjust the code as you suggest.

include/exadg/convection_diffusion/preconditioners/multigrid_preconditioner.cpp
```diff
@@ -250,25 +253,25 @@ MultigridPreconditioner<dim, Number>::initialize_operator(unsigned int const lev
...
-                                               fe,
+                                               n_components,
                                                dirichlet_bc,
                                                dirichlet_bc_component_mask);
 
   if(data.convective_problem and
      data.convective_kernel_data.velocity_type == TypeVelocityField::DoFVector)
   {
-    dealii::FESystem<dim> fe_velocity(dealii::FE_DGQ<dim>(fe.degree), dim);
+    dealii::FESystem<dim> fe_velocity(dealii::FE_DGQ<dim>(degree_velocity), dim);
```
nfehn, line 269: When looking again at the code, I think this line should be adjusted as well, e.g. in order to be prepared for simplex.

include/exadg/operators/finite_element.h
kronbichler: Yes, let's do that now.

PR #515, opened 2023-07-07
Fix include file for dealii::Vector

kronbichler:
https://github.com/dealii/dealii/pull/15626 removed `LinearAlgebra::Vector` and the associated include file `deal.II/lac/la_vector.h` without deprecated. Since we are using `dealii::Vector`, we should have used the right include file from the start. We essentially only used the forward declaration of `la_vector.h` if I see things correctly, but we should be explicit in what this file uses anyway.

Edit: The second commit avoids creating a temporary for a face access, which is necessary with https://github.com/dealii/dealii/pull/15445

Both commits should be backwards compatible.


kronbichler:
Unfortunately, the CI won't let me create this PR without also resolving the other build problem due to recent deal.II updates (I had planned to do that separate to keep things separate). Let me push that one here in a minute.

PR #520, opened 2023-07-25
introduce utility function for creation of quadrature rule

### LINKED ISSUE #430: Use get_gauss_type_quadrature<dim>() to avoid if/else branches for hypercube/simplex?
In addition to using `get_gauss_type_quadrature<dim>()`, we should think about introducing utility functions (realized as free functions) to be called by different modules of **ExaDG**, instead of writing these lines of code creating a quadrature rule in every module.

This issue arises from comment https://github.com/exadg/exadg/pull/351/files#r1187186746

### LINKED ISSUE #517: Introduce utility function for creating quadrature objects similar to #513
In analogy to PR #513 for `FiniteElement`.

@kronbichler regarding quadrature rules for `MatrixFree`, we can always write `<dim>`, right? At some places I think we write `<1>` in the code.

nfehn:
Similar to PR #513, now addressing the quadrature rule. The goal is to use the same utility function for all modules (and for all element types).

closes #517
closes #430

### DIFF
--- a/applications/compressible_navier_stokes/couette/tests/couette.output
+++ b/applications/compressible_navier_stokes/couette/tests/couette.output
@@ -75,11 +75,14 @@ Discontinuous Galerkin finite element discretization:
   degree of 1D polynomials:                  3
   number of dofs per cell:                   64
   number of dofs (total):                    128
-  number of 1D q-points (std):               4
-  number of 1D q-points (over-conv):         4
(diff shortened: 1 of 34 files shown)


nfehn:
So far, only the `Poisson` module has been changed, but I want to apply the changes to all modules in the present PR.

peterrum (approved):

Issue #524, opened 2023-07-27
Remove function calculate_minimum_vertex_distance()?


nfehn:
This function does not take into account the mapping. Hence, it can not be expected to yield correct results for mesh deformation problems / ALE.

I suggest to remove this function from ExaDG.

It should be checked that  - whenever querying the minimum vertex distance in ExaDG - the current grid (with time-dependent deformation) is used.

@kronbichler @richardschu @peterrum

nfehn:
The same holds for the function `calculate_aspect_ratio_vertex_distance()`.

PR #528, opened 2023-07-30
Use mapping in calculate_minimum_vertex_distance

kronbichler:
In the discussion of #524, it is suggested to remove the function. This PR shows how to correctly calculate this information also on deformed geometries. Obviously, we should still assess whether the functionality is still needed, and this PR is merely there to demonstrate that the code changes are small (we just need to pass in the mapping and take the deformed vertices).

### DIFF
--- a/include/exadg/grid/calculate_characteristic_element_length.h
+++ b/include/exadg/grid/calculate_characteristic_element_length.h
@@ -36,24 +36,23 @@ namespace ExaDG
 template<int dim>
 inline double
 calculate_minimum_vertex_distance(dealii::Triangulation<dim> const & triangulation,
+                                  dealii::Mapping<dim> const &       mapping,
                                   MPI_Comm const &                   mpi_comm)
 {
-  double diameter = 0.0, min_cell_diameter = std::numeric_limits<double>::max();
+  double min_cell_diameter = std::numeric_limits<double>::max();
 
   for(auto const & cell : triangulation.active_cell_iterators())
   {
     if(cell->is_locally_owned())
     {
-      diameter = cell->minimum_vertex_distance();
-      if(diameter < min_cell_diameter)
-        min_cell_diameter = diameter;
+      auto const vertices = mapping.get_vertices(cell);
+      for(unsigned int i = 0; i < vertices.size(); ++i)
+        for(unsigned int j = i + 1; j < vertices.size(); ++j)
+          min_cell_diameter = std::min(min_cell_diameter, vertices[i].distance_square(vertices[j]));
     }
   }
 
-  double const global_min_cell_diameter =
-    -dealii::Utilities::MPI::max(-min_cell_diameter, mpi_comm);
-
-  return global_min_cell_diameter;
+  return dealii::Utilities::MPI::min(min_cell_diameter, mpi_comm);
 }
 
 inline double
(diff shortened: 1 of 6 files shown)


nfehn:
I like this PR. I think we should generalize the existing code and use the mapping ... it is certainly an improvement of the present code. We can later discuss to which extent we want to replace certain functionality or not.

nfehn:
One side effect of this PR could be that test cases are not longer running in stable manner, since the present changes affect the calculation of the CFL condition or CFL-related time step restriction. This could potentially mean that CFL numbers set in the application-files are now in the unstable regime. Note that this might affect also applications that are not part of our test suite.

nfehn:
An alternative for the future could be to run over the more general code for evaluating the CFL condition (with local "aspect ratio" / "grid size" computation), but using a DoF vector for the velocity that has been initialized with a constant value, the value that is provided by the user in the application. Of course, the user prescribes only a maximum velocity **magnitude**, so the question would be in which direction to prescribe the vectorial velocity (for a conservative estimate, one would probably set the prescribed maximum velocity magnitude in all coordinate direction, since we do not know the critical direction beforehand).

richardschu:
> An alternative for the future could be to run over the more general code for evaluating the CFL condition (with local "aspect ratio" / "grid size" computation), but using a DoF vector for the velocity that has been initialized with a constant value, the value that is provided by the user in the application. Of course, the user prescribes only a maximum velocity **magnitude**, so the question would be in which direction to prescribe the vectorial velocity (for a conservative estimate, one would probably set the prescribed maximum velocity magnitude in all coordinate direction, since we do not know the critical direction beforehand).

that conservative option is for testing purposes only, right?

include/exadg/grid/calculate_maximum_aspect_ratio.h
```diff
@@ -73,19 +50,27 @@ calculate_aspect_ratio_vertex_distance(dealii::Triangulation<dim> const & triang
...
     if(cell->is_locally_owned())
     {
-      double minimum_vertex_distance = cell->minimum_vertex_distance();
-      double maximum_vertex_distance = calculate_maximum_vertex_distance<dim>(cell);
-      // normalize so that a uniform Cartesian mesh has aspect ratio = 1
+      auto const vertices                = mapping.get_vertices(cell);
+      double     minimum_vertex_distance = std::numeric_limits<double>::max();
+      double     maximum_vertex_distance = 0;
+      for(unsigned int i = 0; i < vertices.size(); ++i)
+        for(unsigned int j = i + 1; j < vertices.size(); ++j)
```
richardschu, line 57: I like the not double-checking :+1:

richardschu (approved):
Everything here considers minimum element length - I assume the other PR opened by @nfehn considers the timestep selection given a velocity field and this is a demo, right?
edit: I checked the already merged PR just now.

include/exadg/grid/calculate_characteristic_element_length.h
```diff
@@ -36,24 +36,23 @@ namespace ExaDG
...
+        for(unsigned int j = i + 1; j < vertices.size(); ++j)
+          min_cell_diameter = std::min(min_cell_diameter, vertices[i].distance_square(vertices[j]));
     }
   }
 
-  double const global_min_cell_diameter =
-    -dealii::Utilities::MPI::max(-min_cell_diameter, mpi_comm);
-
-  return global_min_cell_diameter;
+  return dealii::Utilities::MPI::min(min_cell_diameter, mpi_comm);
```
nfehn, line 55: The square root seems to be missing in this function.

Since I am currently working on this, I will propose the changes of the present PR in a separate PR.

PR #535, opened 2023-08-03
fix order of constraints closing in poisson and structure operator

richardschu:
copying an AffineConstraints object seems to copy the `closed` state, such that adding constraints afterwards triggers an assert in debug mode. This affects Poisson and Structure modules.

### DIFF
--- a/include/exadg/structure/spatial_discretization/operator.cpp
+++ b/include/exadg/structure/spatial_discretization/operator.cpp
@@ -131,8 +131,6 @@ Operator<dim, Number>::distribute_dofs()
                                                *this->grid,
                                                dof_handler);
 
-  affine_constraints_periodicity_and_hanging_nodes.close();
-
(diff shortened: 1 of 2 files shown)


nfehn:
thanks for finding this out!

include/exadg/structure/spatial_discretization/operator.cpp
```diff
@@ -164,6 +162,8 @@ Operator<dim, Number>::distribute_dofs()
   // call deal.II utility function to add Dirichlet constraints
   add_homogeneous_dirichlet_constraints(affine_constraints, dof_handler, map_bid_to_mask);
 
+  // compress constraints *once after* complete setup
```
nfehn, line 165: Maybe we should add a comment in line 142 that `affine_constraints_periodicity_and_hanging_nodes` must be in "un-closed" state since we later want to add additional constraints.

I think such a comment would help me to better understand the code.

PR #536, opened 2023-08-03
simplify affine_constraints setup in Poisson and Structure module

richardschu:
alternative to the fix in #535, we can merge the constraints, such that the `clear()` and `close()` calls are closer to each other and further modularisation would be possible as well. With `copy_from` we would not need to construct `locally_relevant_dofs` (right?), which in this proposed version is needed. @kronbichler do you prefer one version over the other?

### DIFF
--- a/include/exadg/poisson/spatial_discretization/operator.cpp
+++ b/include/exadg/poisson/spatial_discretization/operator.cpp
@@ -91,18 +91,9 @@ Operator<dim, n_components, Number>::distribute_dofs()
                                                  *this->grid,
                                                  dof_handler);
 
-    affine_constraints_periodicity_and_hanging_nodes.close();
-
     // copy periodicity and hanging node constraints, and add further constraints stemming from
     // Dirichlet boundary conditions
-    affine_constraints.clear();
-
-    dealii::IndexSet locally_relevant_dofs;
-    dealii::DoFTools::extract_locally_relevant_dofs(dof_handler, locally_relevant_dofs);
-    affine_constraints.reinit(locally_relevant_dofs);
-
-    // affine constraints to copy from need to be non-closed to add further constraints
-    affine_constraints.merge(affine_constraints_periodicity_and_hanging_nodes);
+    affine_constraints.copy_from(affine_constraints_periodicity_and_hanging_nodes);
 
     // use all the component masks defined by the user
     std::map<dealii::types::boundary_id, dealii::ComponentMask> map_bid_to_mask =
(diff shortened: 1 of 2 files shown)


kronbichler:
Looking at the code, it seems that I would prefer `copy_from()`, which indeed copies the `local_lines` of the other object - however, I also note that the code in deal.II is a bit of a mess, failing to define a clear interface for the MPI case and then doing different things for `copy_from()` and `merge` in terms of the locally relevant dofs. We cannot fix that in deal.II with this PR, but overall I like to have the places that set up the constraints nearby. Note that `close()` needs then to be called after `affine_constraints.copy_from(onstraints_periodicity_and_hanging_nodes)` because otherwise we cannot add new constraints.

richardschu:
> Looking at the code, it seems that I would prefer `copy_from()`, which indeed copies the `local_lines` of the other object - however, I also note that the code in deal.II is a bit of a mess, failing to define a clear interface for the MPI case and then doing different things for `copy_from()` and `merge` in terms of the locally relevant dofs. We cannot fix that in deal.II with this PR, but overall I like to have the places that set up the constraints nearby. Note that `close()` needs then to be called after `affine_constraints.copy_from(onstraints_periodicity_and_hanging_nodes)` because otherwise we cannot add new constraints.

Thank you for taking a look! So we should actually keep it as is, do I understand that correctly? ... I do neither have a strong opinion nor preference here, I just thought this was "the way we do things here" :smile: 
So I can I just close this PR?

kronbichler:
I think we should leave the `copy_from` call there. But, as I observed from inspecting the deal.II source code, we could remove the code
```
    affine_constraints.clear();
    dealii::IndexSet locally_relevant_dofs;
    dealii::DoFTools::extract_locally_relevant_dofs(dof_handler, locally_relevant_dofs);
    affine_constraints.reinit(locally_relevant_dofs);
```
now, as the `copy_from` function clears the previous content and copies over the the `local_lines`, which is the data structure holding the `locally_relevant_dofs`. I had to look myself that this is indeed the case, which is why I complained about the confusing state in deal.II (of which I am partially guilty I believe).

richardschu:
> I think we should leave the `copy_from` call there. But, as I observed from inspecting the deal.II source code, we could remove the code
> 
> ```
>     affine_constraints.clear();
>     dealii::IndexSet locally_relevant_dofs;
>     dealii::DoFTools::extract_locally_relevant_dofs(dof_handler, locally_relevant_dofs);
>     affine_constraints.reinit(locally_relevant_dofs);
> ```
> 
> now, as the `copy_from` function clears the previous content and copies over the the `local_lines`, which is the data structure holding the `locally_relevant_dofs`. I had to look myself that this is indeed the case, which is why I complained about the confusing state in deal.II (of which I am partially guilty I believe).

ah thanks, I forgot my initial comment about the `locally_relevant_dofs` again and was a bit confused. So now, we can at least omit some setup :100:

kronbichler (approved):

nfehn:
I see this PR as an improvement which should be merged.

My opinion is that we should for the future try to realize more modular code. With "modular", I mean writing code like this

```C++
...
setup_constraints_a();
setup_constraints_b();
...
```

We currently have:
```C++
setup_constraints_a();
setup_constraints_b();
close_constraints_a():
```

PR #537, opened 2023-08-03
Comment and refactor structure material

richardschu:
In preparation for adding further hyperelastic structure models, I suggest the following changes to the structure material implementation. This PR introduces some comments explaining the St. Venant model to make it easier accesible, but most notably renames the `apply_C` function, which essentially returns the directional derivative of the second Piola Kirchhoff stress tensor in a general total Lagrangian implementation.  To make this function more general, I changed the arguments from the strain tensor alone to the linearization point and shape function gradient. With this, one can easily include invariant-based nonlinear material models as well.


include/exadg/structure/material/library/st_venant_kirchhoff.cpp
```diff
@@ -149,35 +155,40 @@ StVenantKirchhoff<dim, Number>::evaluate_stress(
...
-    S[0][0] = f0 * E[0][0] + f1 * E[1][1] + f1 * E[2][2];
-    S[1][1] = f1 * E[0][0] + f0 * E[1][1] + f1 * E[2][2];
-    S[2][2] = f1 * E[0][0] + f1 * E[1][1] + f0 * E[2][2];
-    S[0][1] = f2 * (E[0][1] + E[1][0]);
-    S[1][2] = f2 * (E[1][2] + E[2][1]);
-    S[0][2] = f2 * (E[0][2] + E[2][0]);
-    S[1][0] = f2 * (E[0][1] + E[1][0]);
-    S[2][1] = f2 * (E[1][2] + E[2][1]);
-    S[2][0] = f2 * (E[0][2] + E[2][0]);
+    S[0][0] = f0 * strain_measure[0][0] + f1 * (strain_measure[1][1] + strain_measure[2][2]);
```
richardschu, line 158: note that we could still pass gradient of the test function in case of linear elasticity, since we do not assume that strain measure is symmetric.

include/exadg/structure/material/library/st_venant_kirchhoff.cpp
```diff
@@ -149,35 +155,40 @@ StVenantKirchhoff<dim, Number>::evaluate_stress(
...
-StVenantKirchhoff<dim, Number>::apply_C(
-  dealii::Tensor<2, dim, dealii::VectorizedArray<Number>> const & E,
+StVenantKirchhoff<dim, Number>::second_piola_kirchhoff_stress_derivative(
+  dealii::Tensor<2, dim, dealii::VectorizedArray<Number>> const & Grad_delta,
+  dealii::Tensor<2, dim, dealii::VectorizedArray<Number>> const & F_lin,
   unsigned int const                                              cell,
   unsigned int const                                              q) const
 {
-  return evaluate_stress(E, cell, q);
+  // Exploit linear stress-strain relationship and minor symmetries in C
```
richardschu, line 187: please check this comment @nfehn

include/exadg/structure/material/material.h
```diff
@@ -41,15 +41,27 @@ class Material
...
+  /*
+   * Evaluate directional derivative of the 2nd Piola-Kirchhoff stress tensor given the displacement
+   * shape function gradient Grad_delta and deformation gradient at current linearization point
+   * F_lin
+   */
   virtual dealii::Tensor<2, dim, dealii::VectorizedArray<Number>>
-  apply_C(dealii::Tensor<2, dim, dealii::VectorizedArray<Number>> const & E,
-          unsigned int const                                              cell,
-          unsigned int const                                              q) const = 0;
+  second_piola_kirchhoff_stress_derivative(
```
richardschu, line 60: note the changed arguments here

include/exadg/structure/spatial_discretization/operators/linear_operator.cpp
```diff
@@ -35,11 +35,14 @@ LinearOperator<dim, Number>::do_cell_integral(IntegratorCell & integrator) const
 
   for(unsigned int q = 0; q < integrator.n_q_points; ++q)
   {
-    // engineering strains (material tensor is symmetric)
-    tensor const gradient = integrator.get_gradient(q);
+    // engineering strains
+    tensor const symmetric_gradient = integrator.get_symmetric_gradient(q);
```
richardschu, line 39: Should I revert to just `gradient`? This works due to the specific implementation, e.g., 
`S[0][1] = f2 * (strain_measure[0][1] + strain_measure[1][0]);`
for `dim = 2`, right?
Should I revert this adaptation for performance reasons @kronbichler ?

include/exadg/structure/spatial_discretization/operators/linear_operator.cpp
```diff
@@ -35,11 +35,14 @@ LinearOperator<dim, Number>::do_cell_integral(IntegratorCell & integrator) const
...
 
-    // Cauchy stresses
-    tensor const sigma = material->apply_C(gradient, integrator.get_current_cell_index(), q);
+    // Cauchy stresses, only valid for linear elasticity
+    tensor const sigma =
```
richardschu, line 42: also note here that I am not using the `second_piola_kirchhoff_stress_derivative` here to have a single implementation for linear elasticity and finite strain elasticity using the St Venant Kirchhoff model.

include/exadg/structure/material/library/st_venant_kirchhoff.cpp
```diff
@@ -149,35 +155,40 @@ StVenantKirchhoff<dim, Number>::evaluate_stress(
...
-    S[2][0] = f2 * (E[0][2] + E[2][0]);
+    S[0][0] = f0 * strain_measure[0][0] + f1 * (strain_measure[1][1] + strain_measure[2][2]);
+    S[1][1] = f0 * strain_measure[1][1] + f1 * (strain_measure[0][0] + strain_measure[2][2]);
+    S[2][2] = f0 * strain_measure[2][2] + f1 * (strain_measure[0][0] + strain_measure[1][1]);
+    S[0][1] = f2 * (strain_measure[0][1] + strain_measure[1][0]);
+    S[1][2] = f2 * (strain_measure[1][2] + strain_measure[2][1]);
+    S[0][2] = f2 * (strain_measure[0][2] + strain_measure[2][0]);
+    S[1][0] = S[0][1];
+    S[2][1] = S[1][2];
+    S[2][0] = S[0][2];
```
nfehn, line 166: this looks good

include/exadg/structure/material/library/st_venant_kirchhoff.cpp
```diff
@@ -149,35 +155,40 @@ StVenantKirchhoff<dim, Number>::evaluate_stress(
...
-    S[0][2] = f2 * (E[0][2] + E[2][0]);
-    S[1][0] = f2 * (E[0][1] + E[1][0]);
-    S[2][1] = f2 * (E[1][2] + E[2][1]);
-    S[2][0] = f2 * (E[0][2] + E[2][0]);
+    S[0][0] = f0 * strain_measure[0][0] + f1 * (strain_measure[1][1] + strain_measure[2][2]);
+    S[1][1] = f0 * strain_measure[1][1] + f1 * (strain_measure[0][0] + strain_measure[2][2]);
+    S[2][2] = f0 * strain_measure[2][2] + f1 * (strain_measure[0][0] + strain_measure[1][1]);
+    S[0][1] = f2 * (strain_measure[0][1] + strain_measure[1][0]);
+    S[1][2] = f2 * (strain_measure[1][2] + strain_measure[2][1]);
+    S[0][2] = f2 * (strain_measure[0][2] + strain_measure[2][0]);
```
nfehn, line 163: Here, I would no longer "symmetrize" inside the function, but pass a symmetric strain measure into the function.

include/exadg/structure/material/library/st_venant_kirchhoff.cpp
```diff
@@ -149,35 +155,40 @@ StVenantKirchhoff<dim, Number>::evaluate_stress(
...
   }
   else
   {
-    S[0][0] = f0 * E[0][0] + f1 * E[1][1];
-    S[1][1] = f1 * E[0][0] + f0 * E[1][1];
-    S[0][1] = f2 * (E[0][1] + E[1][0]);
-    S[1][0] = f2 * (E[0][1] + E[1][0]);
+    S[0][0] = f0 * strain_measure[0][0] + f1 * strain_measure[1][1];
+    S[1][1] = f1 * strain_measure[0][0] + f0 * strain_measure[1][1];
+    S[0][1] = f2 * (strain_measure[0][1] + strain_measure[1][0]);
```
nfehn, line 172: The above comment https://github.com/exadg/exadg/pull/537/files#r1284082874 also applies here in the 2d case

include/exadg/structure/material/library/st_venant_kirchhoff.h
```diff
@@ -69,16 +69,28 @@ class StVenantKirchhoff : public Material<dim, Number>
...
+    unsigned int const                                              cell,
+    unsigned int const                                              q) const final;
 
 private:
+  /*
+   * The second Piola-Kirchhoff stress tensor S is given as S = lambda * I * tr(E) + 2 mu E,
+   * with E being the Green-Lagrange strain tensor and Lamee parameters lambda and mu.
+   * Factor out coefficients for faster computation
+   * Sii = f0 * Eii + f1 * Eij, for i, j = 1, ..., dim and i != j, and
+   * Sij = f2 * (Eij + Eji),    for i, j = 1, ..., dim and i != j.
```
nfehn, line 90: this also needs to be adjusted according to https://github.com/exadg/exadg/pull/537/files#r1284082874

include/exadg/structure/material/library/st_venant_kirchhoff.cpp
```diff
@@ -149,35 +155,40 @@ StVenantKirchhoff<dim, Number>::evaluate_stress(
...
+  dealii::Tensor<2, dim, dealii::VectorizedArray<Number>> const & Grad_delta,
+  dealii::Tensor<2, dim, dealii::VectorizedArray<Number>> const & F_lin,
   unsigned int const                                              cell,
   unsigned int const                                              q) const
 {
-  return evaluate_stress(E, cell, q);
+  // Exploit linear stress-strain relationship and minor symmetries in C
+  // when forming directional derivative Du(PK2) = Du(C : E) = C : Du(E)
+  // = C : (0.5 * ( transpose(F_lin) * Grad_delta + transpose(Grad_delta) * F_lin ))
+  // = C : (transpose(F_lin) * Grad_delta )
```
nfehn, line 190: This explanation should appear in the declaration of the function in the base class in my opinion. Or is this not valid in general?

The "u" in "Du" is not clear to me. The function name "derivative" is somewhat misunderstandable, since the "derivative" is multiplied by an increment, giving a "Delta" along a certain "direction", right?

From the second to the third line, do we assume that C is symmetric? If so, this should also be reflected in the comment.

include/exadg/structure/material/library/st_venant_kirchhoff.cpp
```diff
@@ -149,35 +155,40 @@ StVenantKirchhoff<dim, Number>::evaluate_stress(
...
+  return (this->second_piola_kirchhoff_stress(transpose(F_lin) * Grad_delta, cell, q));
```
nfehn, line 191: If we change the implementation according to comment https://github.com/exadg/exadg/pull/537/files#r1284082874, this line will probably no longer be correct, but we would need to pass in a symmetric tensor, right?

include/exadg/structure/material/library/st_venant_kirchhoff.h
```diff
@@ -69,16 +69,28 @@ class StVenantKirchhoff : public Material<dim, Number>
...
+    unsigned int const                                              cell,
+    unsigned int const                                              q) const final;
 
   dealii::Tensor<2, dim, dealii::VectorizedArray<Number>>
-  apply_C(dealii::Tensor<2, dim, dealii::VectorizedArray<Number>> const & E,
-          unsigned int const                                              cell,
-          unsigned int const                                              q) const final;
+  second_piola_kirchhoff_stress_derivative(
+    dealii::Tensor<2, dim, dealii::VectorizedArray<Number>> const & Grad_delta,
+    dealii::Tensor<2, dim, dealii::VectorizedArray<Number>> const & F_lin,
```
nfehn, line 80: I am sure if this change should be made in the present PR.

The names are somewhat cryptic. Better would be "gradient_of_displacement_increment" and "deformation_gradient_in_linearization_point".

include/exadg/structure/spatial_discretization/operators/linear_operator.cpp
```diff
@@ -35,11 +35,14 @@ LinearOperator<dim, Number>::do_cell_integral(IntegratorCell & integrator) const
...
+      material->second_piola_kirchhoff_stress(symmetric_gradient,
```
nfehn, line 43: This PR replaces `apply_C()` by `second_piola_kirchhoff_stress_derivative()`, so I would expect that `second_piola_kirchhoff_stress_derivative()` is called here. The new version should also be valid, but I think not all of these changes should be mixed into a single PR.

include/exadg/structure/material/material.h
```diff
@@ -41,15 +41,27 @@ class Material
...
-                  unsigned int const                                              cell,
-                  unsigned int const                                              q) const = 0;
+  second_piola_kirchhoff_stress(
+    dealii::Tensor<2, dim, dealii::VectorizedArray<Number>> const & strain_measure,
+    unsigned int const                                              cell,
+    unsigned int const                                              q) const = 0;
 
+  /*
+   * Evaluate directional derivative of the 2nd Piola-Kirchhoff stress tensor given the displacement
+   * shape function gradient Grad_delta and deformation gradient at current linearization point
```
nfehn, line 56: The above function implements the function `S(E)`, while the present function refers to the directional derivative of a function `S(d)`, with `d` being the displacement vector. Do you agree?

This is currently not reflected in the function names, but I think this would be necessary. To be more precise, maybe the documentation should say `(D S(d) / D d)|_{d_lin} * delta_d`. Would you agree to this formula?

include/exadg/structure/material/library/st_venant_kirchhoff.cpp
richardschu: Now that we have the stress *and* strain measure implementation based on the deformation gradient, we *could* make use of this little trick again (the version I have currently in the code). If we want to go for the more obvious/straight-forward option (without symmetrizing), I can do that as well!

include/exadg/structure/spatial_discretization/operators/linear_operator.cpp
richardschu: I think actually in the `linear_operator.cpp` calling the derivative is more confusing, as we only want to evaluate the stress actually! ...once again one of these points were linear elasticity and St Venant solid were very similar, but other more general models are not so much as far as I am concerned.

include/exadg/structure/material/material.h
richardschu: This is true! First point we addressed by changing the interface to the displacement gradient, which is more general than the strain measure. (For some material models, we do not really want E in the nonlinear case but actually C to compute invariants, so a common ground is Gradient(displacement)).
The second point I forgot, will adapt the name accordingly and push ASAP.

richardschu:
@nfehn and I discussed today, that choosing the gradient of the displacement field makes most sense in the interface when computing second Piola-Kirchhoff stresses and the corrresponding directional derivative with respect to the dispalcement. Before this PR, the specific strain measures used in the (non-)linear operators made use of a symmetrization **inside** `apply_C` (which is now similar to `second_piola_kirchhoff_stress_symmetrize`) to reduce the number of needed operations. I did not give up on this nice "trick", since all the entire code that needs to fit together is now sitting in a single file. Therefore, I also left the comments regarding this symmetrization in place. The base operator now also has a parameter `large_deformation`, which is **not** defined by the user in the application but is automatically assigned during operator setup. 
For now this boolean is used to differentiate linear and nonlinear cases *during stress computation*, which should be addressed in a future PR. Then, instead of using the boolean parameter in `material_handler.initialize()` as an input parameter for the constructor of `StVenantKirchhoff`, we can create actually *different* objects depending on `large_deformation` and insert these into the `material_map`. This way, we could avoid the if statement.

include/exadg/structure/material/library/st_venant_kirchhoff.cpp
```diff
@@ -94,10 +98,10 @@ StVenantKirchhoff<dim, Number>::get_f2_factor() const
   Number const nu           = data.nu;
   Type2D const type_two_dim = data.type_two_dim;
 
-  return (dim == 3) ? (1. - 2. * nu) / 2. / (1. + nu) / (1. - 2. * nu) :
+  return (dim == 3) ? (1. - 2. * nu) * 0.5 / ((1. + nu) * (1. - 2. * nu)) :
                       (type_two_dim == Type2D::PlaneStress ?
-                         ((1. - nu) / 2. / (1. - nu * nu)) :
-                         ((1. - 2. * nu) / 2. / (1. + nu) / (1. - 2. * nu)));
+                         ((1. - nu) * 0.5 / (1. - nu * nu)) :
+                         ((1. - 2. * nu) * 0.5 / ((1. + nu) * (1. - 2. * nu))));
```
nfehn, line 104: only a minor comment, since you made these changes: one could write the factor of two inside the (new) brackets. Could be easier to read, but is probably also a matter of taste.

include/exadg/structure/material/library/st_venant_kirchhoff.h
```diff
@@ -88,6 +96,22 @@ class StVenantKirchhoff : public Material<dim, Number>
   Number
   get_f2_factor() const;
 
+  /*
+   * The second Piola-Kirchhoff stress tensor S is given as S = lambda * I * tr(E) + 2 mu E, with E
+   * being the Green-Lagrange strain tensor and Lamee parameters lambda and mu. This leads to
+   * Sii = f0 * Eii + f1 * Eij, for i, j = 1, ..., dim and i != j, and
```
nfehn, line 102: I think here one was to write `sum_{i != j} E_ij`, i.e. the sum is missing, because `S_ii` is not a sum but a single diagonal entry.

include/exadg/structure/material/material.h
```diff
@@ -41,15 +41,28 @@ class Material
...
-                  unsigned int const                                              cell,
-                  unsigned int const                                              q) const = 0;
+  second_piola_kirchhoff_stress(
+    dealii::Tensor<2, dim, dealii::VectorizedArray<Number>> const & gradient_displacement,
+    unsigned int const                                              cell,
+    unsigned int const                                              q) const = 0;
 
+  /*
+   * Evaluate the directional derivative with respect to the displacement of the 2nd Piola-Kirchhoff
+   * stress tensor given the displacement increment shape function gradient *gradient_increment*
```
nfehn, line 56: I think one should better remove "shape function" here, because the argument is already the interpolation of shape functions into a q-point. 

So I would write `gradient of the displacment increment, short "gradient_increment"`. 

A minor thing: I would add "." at the end of the comments.

include/exadg/structure/spatial_discretization/operators/linear_operator.cpp
```diff
@@ -35,11 +35,11 @@ LinearOperator<dim, Number>::do_cell_integral(IntegratorCell & integrator) const
...
   for(unsigned int q = 0; q < integrator.n_q_points; ++q)
   {
-    // engineering strains (material tensor is symmetric)
-    tensor const gradient = integrator.get_gradient(q);
-
-    // Cauchy stresses
-    tensor const sigma = material->apply_C(gradient, integrator.get_current_cell_index(), q);
+    // Cauchy stresses, only valid for linear elasticity
+    tensor const sigma =
+      material->second_piola_kirchhoff_stress(integrator.get_gradient(q),
```
nfehn, line 40: why not the variant `_symmetrize()` here?

include/exadg/structure/spatial_discretization/operators/nonlinear_operator.cpp
```diff
@@ -326,20 +324,21 @@ NonLinearOperator<dim, Number>::do_cell_integral(IntegratorCell & integrator) co
     // kinematics
     tensor const Grad_delta = integrator.get_gradient(q);
 
-    tensor const F_lin = get_F<dim, Number>(integrator_lin->get_gradient(q));
+    tensor const Grad_displacement = integrator_lin->get_gradient(q);
```
nfehn, line 327: minor comment: `Grad_lin` or `Grad_d_lin` would be shorter and in line with the other variable names `F_lin`, `S_lin`.

include/exadg/structure/material/library/st_venant_kirchhoff.cpp
```diff
@@ -149,35 +157,57 @@ StVenantKirchhoff<dim, Number>::evaluate_stress(
...
   }
 
   return S;
 }
 
 template<int dim, typename Number>
 dealii::Tensor<2, dim, dealii::VectorizedArray<Number>>
-StVenantKirchhoff<dim, Number>::apply_C(
-  dealii::Tensor<2, dim, dealii::VectorizedArray<Number>> const & E,
+StVenantKirchhoff<dim, Number>::second_piola_kirchhoff_stress(
```
nfehn, line 183: Ah, now I see, the name `symmetrize` is only inside this function.

nfehn (approved):

PR #542, opened 2023-08-06
fix include tokens

richardschu:
although we might not enforce include tokens super strict, I found these two not fitting the scheme/being a bit misleading. I think we are generally aiming for something like `FOLDER_FILENAME_H`.

### DIFF
--- a/include/exadg/operators/mass_kernel.h
+++ b/include/exadg/operators/mass_kernel.h
@@ -19,8 +19,8 @@
  *  ______________________________________________________________________
  */
 
-#ifndef INCLUDE_OPERATORS_MASS_MATRIX_OPERATOR_H_
-#define INCLUDE_OPERATORS_MASS_MATRIX_OPERATOR_H_
+#ifndef INCLUDE_OPERATORS_MASS_KERNEL_H_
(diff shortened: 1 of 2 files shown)


include/exadg/operators/mass_kernel.h
```diff
@@ -19,8 +19,8 @@
  *  ______________________________________________________________________
  */
 
-#ifndef INCLUDE_OPERATORS_MASS_MATRIX_OPERATOR_H_
-#define INCLUDE_OPERATORS_MASS_MATRIX_OPERATOR_H_
+#ifndef INCLUDE_OPERATORS_MASS_KERNEL_H_
+#define INCLUDE_OPERATORS_MASS_KERNEL_H_
```
peterrum, line 23: I think it is time to switch to `#pragma once`.

richardschu: feels like something we should discuss before someone is changing **all** the files :laughing:
Is the argument that it is not part of the standard still true actually? I read [this discussion](https://stackoverflow.com/questions/787533/is-pragma-once-a-safe-include-guard), which is from 2009 ...
Any other opinions on this @nfehn @kronbichler ?

PR #547, opened 2023-08-08
added assertion for weak damping

richardschu:
I forgot this assertion. The mass operator was not used in the `Stationary` or `Quasi-Static` cases, but this makes it more obvious and safer.

### DIFF
--- a/include/exadg/structure/user_interface/parameters.cpp
+++ b/include/exadg/structure/user_interface/parameters.cpp
@@ -96,21 +96,14 @@ Parameters::check() const
   {
     AssertThrow(restarted_simulation == false,
                 dealii::ExcMessage("Restart has not been implemented."));
-
-    if(weak_damping_active)
-    {
-      AssertThrow(weak_damping_coefficient > 0.0,
-                  dealii::ExcMessage("Weak linear damping requires positive coefficient."));
-    }
   }
-  else
+
+  if(weak_damping_active)
   {
-    AssertThrow(weak_damping_active == false,
+    AssertThrow(problem_type == ProblemType::Unsteady,
                 dealii::ExcMessage("Weak damping only well-defined for ProblemType::Unsteady."));
-    AssertThrow(
-      not(std::abs(weak_damping_coefficient) > 0.0),
-      dealii::ExcMessage(
-        "Do not set nonzero weak damping coefficient for ProblemType::Steady or ProblemType::QuasiStatic."));
+    AssertThrow(weak_damping_coefficient > 0.0,
+                dealii::ExcMessage("Weak damping coefficient defined positive."));
   }
 
   // SPATIAL DISCRETIZATION


include/exadg/structure/spatial_discretization/operator.cpp
```diff
@@ -241,6 +241,8 @@ Operator<dim, Number>::compute_scaling_factor_mass(double const scaling_factor_a
   double scaling_factor_mass = scaling_factor_acceleration;
   if(param.weak_damping_active)
   {
+    AssertThrow(param.problem_type == ProblemType::Unsteady,
+                dealii::ExcMessage("Weak damping only well-defined for instationary problems."));
```
nfehn, line 245: I think one should assert this when checking the parameters in the respective member function of `Parameters`. The spatial operator then applies the weak damping whenever `param.weak_damping_active == true`. Put differently, `Parameters::check()` aims to avoid (potentially redundant) parameter checks in spatial operators / time integration routines, ...

richardschu: that is absolutely true, sorry I forgot about the discussion we had some weeks ago! will update the PR now.

kronbichler (approved):

include/exadg/structure/user_interface/parameters.cpp
```diff
@@ -103,6 +103,15 @@ Parameters::check() const
...
     }
   }
+  else
+  {
+    AssertThrow(weak_damping_active == false,
+                dealii::ExcMessage("Weak damping only well-defined for ProblemType::Unsteady."));
+    AssertThrow(
+      not(std::abs(weak_damping_coefficient) > 0.0),
+      dealii::ExcMessage(
+        "Do not set nonzero weak damping coefficient for ProblemType::Steady or ProblemType::QuasiStatic."));
```
nfehn, line 113: Why is there a second assert? I thought the parameter `weak_damping_coefficient` will only be used if `weak_damping_active = true`, where we already have the assert a few lines above. The current solution looks a bit like "doppelt hält besser"? I think there would be simpler solutions.

My suggestion for aspects related to weak damping in `Parameters::check()` would be something like
```c++
if(problem_type == ProblemType::Steady or problem_type == ProblemType::QuasiStatic)
{
  AssertThrow(weak_damping_active == false, ...);
}

if (weak_damping_active)
{
  AssertThrow(weak_damping_coefficient > 0.0, ...);
}
```
or even simpler
```c++
if (weak_damping_active)
{
  AssertThrow(problem_type == ProblemType::Unsteady, ...);
  AssertThrow(weak_damping_coefficient > 0.0, ...);
}
```

richardschu: Yes, `weak_damping_active` controls whether or not the damping is actually used, irrespective of the `weak_damping_coefficient` set.
I wanted it to be more verbose, such that for every possible case there is an assertion, if the combination does not make sense.
I mean that:
1. I should get an error, if I request weak damping for `QuasiStatic` or `Steady` (both for `weak_damping_active` or `weak_damping_coefficient` independently).
2. I should get an error if I am in the `Unsteady`, requesting weak damping, but giving a coefficient <= 0.

That is at least a really verbose version.

Your second option is quite nice, we could even remove some code in the lines just above (100-104). So if we accept that not all cases are covered, e.g., `weak_damping_coefficient=1e-3` and `weak_damping_active=false` does not use damping and one might think "but I gave a coefficient :imp: ", then lets go for the second option you provided!

nfehn: > So if we accept that not all cases are covered

I think this is exactly what we want. Imagine that the functionality of a solver/module is extended more and more, but one wants to simulate something very basic. Then, the user should not need to worry (regarding parameter setttings) about functionality that is turned off (like `weak_damping_active = false`). We also want this logic for another reason: Imagine one wants to compare simulations with and without damping, by changing the parameters multiple times in the application/input file from one version back to the other (and so on). Then, one should only need to change the boolean parameter in the application/input file, not multiple parameters.

If the user sets a damping coefficient to simulate damping but deactivates the damping via the boolean parameter, this would be a wrong use of the parameters. For this purpose, we have the screen output, showing to the user that an actual simulation does not use damping.

richardschu: Yep, I see your points - and can go with that take! :smiley:
commit incoming

Issue #548, opened 2023-08-08
Structure::NonlinearOperator reinit_cell_derived fails to access ghost entries


richardschu:
Running the structure/bar application with multiple cores leads to assertion thrown in `NonlinearOperator::reinit_cell_derived()` (see stacktrace below). This happens only when using the NonlinearOperator together with `Preconditioner::Multigrid` (this is exactly the out-of-the-box setup).

I will try to fix this, should not be too hard to find as the `navier_stokes/poisuille` is indeed running with `mpirun -n 2`.

Here is the error + stacktrace:

`
An error occurred in line <1631> of file </homes/wiss/schussri/dealii-candi/dealii/include/deal.II/lac/la_parallel_vector.h> in function
    Number dealii::LinearAlgebra::distributed::Vector<Number, MemorySpace>::local_element(dealii::LinearAlgebra::distributed::Vector<Number, MemorySpace>::size_type) const [with Number = float; MemorySpace = dealii::MemorySpace::Host; dealii::LinearAlgebra::distributed::Vector<Number, MemorySpace>::size_type = unsigned int]
The violated condition was: 
    local_index < locally_owned_size() || vector_is_ghosted == true
Additional information: 
    You tried to read a ghost element of this vector, but it has not
    imported its ghost values.

Stacktrace:
-----------
#0  solver: dealii::LinearAlgebra::distributed::Vector<float, dealii::MemorySpace::Host>::local_element(unsigned int) const
#1  solver: void dealii::FEEvaluationBase<2, 2, float, false, dealii::VectorizedArray<float, 8ul> >::read_write_operation<dealii::LinearAlgebra::distributed::Vector<float, dealii::MemorySpace::Host> const, dealii::internal::VectorReader<float, dealii::VectorizedArray<float, 8ul> > >(dealii::internal::VectorReader<float, dealii::VectorizedArray<float, 8ul> > const&, std::array<dealii::LinearAlgebra::distributed::Vector<float, dealii::MemorySpace::Host> const*, 2ul> const&, std::array<std::vector<dealii::ArrayView<dealii::LinearAlgebra::distributed::Vector<float, dealii::MemorySpace::Host>::value_type const, dealii::MemorySpace::Host>, std::allocator<dealii::ArrayView<dealii::LinearAlgebra::distributed::Vector<float, dealii::MemorySpace::Host>::value_type const, dealii::MemorySpace::Host> > > const*, 2ul> const&, std::bitset<8ul> const&, bool) const
[...]

PR #549, opened 2023-08-08
Fix exadg issue 548

richardschu:
Fix issue #548 by calling introduced `NonLinearOperator::update_ghost_values_linearization_vector()` on each coarse level.
I tried to do this in the `MultigridPreconditioner::initialize` for `nonlinear==true` (i.e., only once initially), but that was not enough. `update_ghost_values` is only called `if(not displacement_lin.has_ghost_values())`, meaning only when necessary.

Problem is: in `NonLinearOperator::initialize` and `NonLinearOperator::set_solution_linearization`, we always call `displacement_lin.update_ghost_values()`, that is, initially and whenever it is adapted. Hence, we **should** actually always have `displacement_lin.has_ghost_values() == true`, which is clearly **not** the case.

So this is a fix, but maybe someone has a better idea on where things go wrong?


richardschu:
this is a temporary fix until we have a fix in deal.II available.

PR #550, opened 2023-08-09
remove naive version of aspect ratio computation

nfehn:
I decided to remove the "naive" version (not really being used currently), in order to make upcoming changes in ExaDG easier for generalizations in terms of mapping/simplex/etc.


richardschu (approved):
:+1:

PR #551, opened 2023-08-09
OperatorBase: Make methods virtual

bergbauer:
I would like to make these methods overwritable for CutDG operators.


richardschu:
:rocket:

PR #552, opened 2023-08-09
Allow choosing multigrid number type

bergbauer:


include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.h
```diff
@@ -57,11 +57,11 @@ class MGCoarseGridBase;
 
 namespace ExaDG
 {
-template<int dim, typename Number>
+template<int dim, typename Number, typename MultigridNumber_ = float>
 class MultigridPreconditionerBase : public PreconditionerBase<Number>
 {
 public:
-  typedef float MultigridNumber;
+  typedef MultigridNumber_ MultigridNumber;
```
nfehn, line 64: I think this line can/should be removed (using `MultigridNumber` a few lines above).

bergbauer: Unfortunately not, because we access this typedef in other classes. (We cannot infer the template argument directly.)

nfehn: but then only because of the fact that you use a default parameter for the template, right? I am not sure if one should then use a default value.

bergbauer: No, that's independent of the default parameter. We need a typedef for the other classes. I chose the default parameter to retain the interface. We could of course be more explicit here and set the template argument where needed.

PR #555, opened 2023-08-09
A Robin boundary condition for structure module

richardschu:
This PR introduces a general BC of the form
`traction = P*N = - k * displacement - c * velocity - exterior_pressure`.
For this we need several adaptations:

1. `boundary_mass_operator` to shift the discretized remainder of the velocity term to the rhs (similar to the weak damping or general time integration terms, but now with a mass term on the boundary only). This will also come in handy, if we want to project values on the boundary.
2. changes in `operator_base` such that CG discretizations involve some boundary integrals if `evaluate_boundary_integrals() == true` in the fundamental `apply`, `apply_add`, `evaluate`, `evaluate_add` ... operations **and** the remaining MF loops used in MG setup, preconditioner update etc. (building the system matrix or its inverse diagonal)
3. consider the new boundary integral in the linear and nonlinear operators in the structure module
4. scaling factors and updates of the new `BoundaryMassOperator` in `structure/operator.cpp` and the according interface.
5. tests for the new functionality

I have this now in a single PR to thoroughly check if the behavior is as desired (see the tests) and the changes in `OperatorBase` actually work as expected to get good Multigrid convergence. Also, with this you can see the intended use (and extensibility towards Robin coupling conditions in FSI). If need be, I can split into several parts:
1. changes in `OperatorBase`
2. add exterior pressure (Robin condition part 1)
3. add spring (Robin condition part 1)
4. add `BoundaryMassOperator`
5. add dashpot (use `BoundaryMassOperator` and temporal scaling)

Whoever is reviewing this (probably @nfehn :1st_place_medal:), let's have an in-person discussion first!


include/exadg/structure/user_interface/boundary_descriptor.h
```diff
@@ -70,6 +71,16 @@ struct BoundaryDescriptor
...
 
+  // Robin boundary condition of the form
+  // + ( v_h, k * d_h + c * d/dt(d_h) +  p * N )
+  // or
+  // + ( v_h, k * N * (d_h . N) + c * N . (d/dt(d_h) . N) + p * N )
+  // using normal projections of displacement/velocity terms controlled via the
+  // std::array<bool, 2> for the displacement (index 0) and velocity terms (index 1)
+  // The std::array<double, 3> contains the parameters k (index 0), c (index 1) and p (index 2).
+  std::map<dealii::types::boundary_id, std::pair<std::array<bool, 2>, std::array<double, 3>>>
+    robin_k_c_p_param;
```
nfehn, line 82: This design makes the code very hard to read in my opinion, because in the implementation (`SpatialOperator` classes), one only sees `.first`, `.second`, etc., without concrete names.

Here, the data structure should look like
```c++
std::map<dealii::types::boundary_id, ExaDG::RobinParameters> robin_bc;
```

and inside `RobinParameters` variables like `.stiffness`, etc. with readable names. Instead of `robin_k_c_p_param` (where it is hard to guess the meaning), we should follow the naming `robin_bc` for consistency with other boundary types.

include/exadg/structure/spatial_discretization/operators/nonlinear_operator.cpp
```diff
@@ -275,27 +282,83 @@ template<int dim, typename Number>
...
     {
-      tensor F = get_F<dim, Number>(integrator.get_gradient(q));
-      vector N = integrator.get_normal_vector(q);
-      // da/dA * n = det F F^{-T} * N := n_star
-      // -> da/dA = n_star.norm()
-      vector n_star = determinant(F) * transpose(invert(F)) * N;
-      // t_0 = da/dA * t
-      traction *= n_star.norm();
+      if(operator_type == OperatorType::inhomogeneous or operator_type == OperatorType::full)
+      {
```
nfehn, line 299: From the many if-statements introduced in this function, I am not sure whether we have a good design right now for the new Robin-functionality in the present PR. Note that the free function `calculate_neumann_value()` already contains if-else-statements. I am also not convinced right now that the way `OperatorBase` is extended is a good design.

include/exadg/structure/spatial_discretization/operators/nonlinear_operator.cpp
```diff
@@ -275,27 +282,83 @@ template<int dim, typename Number>
...
+        traction -= calculate_neumann_value<dim, Number>(
```
nfehn, line 300: I don't like that a minus sign is introduced here, and that the sign is also changed below in function `submit_value()`. 

(As a sidenote, the present change turns an assignment into an addition/subtraction, so why didn't you actually write `traction = - xyz;` to introduce a minus sign compared to the previous version? However, we should keep the previous version in my opinion. Robin is a new boundary type, that is optional.)

include/exadg/structure/spatial_discretization/operators/nonlinear_operator.cpp
```diff
@@ -275,27 +282,83 @@ template<int dim, typename Number>
...
   BoundaryType boundary_type = this->operator_data.bc->get_boundary_type(boundary_id);
 
   for(unsigned int q = 0; q < integrator.n_q_points; ++q)
   {
-    auto traction = calculate_neumann_value<dim, Number>(
-      q, integrator, boundary_type, boundary_id, this->operator_data.bc, this->time);
+    vector traction;
 
-    if(this->operator_data.pull_back_traction)
+    // integrate standard (stored) traction or exterior pressure on Robin boundaries
```
nfehn, line 294: This comment reads like Robin is "the only" boundary type. I think it holds for the present PR as a whole: We should aim at a modular code where Robin is one specific feature and where it would be very easy in the future to remove this feature again (with "easy" I mean minimal code changes). Even if we will never remove it again, this way of thinking might help for the design.

include/exadg/structure/spatial_discretization/operator.cpp
```diff
@@ -141,6 +141,14 @@ Operator<dim, Number>::fill_matrix_free_data(MatrixFreeData<dim, Number> & matri
   if(param.body_force)
     matrix_free_data.append_mapping_flags(BodyForceOperator<dim, Number>::get_mapping_flags());
 
+  if(param.problem_type == ProblemType::Unsteady)
+  {
+    // The boundary mass operator is only required, if we have damping terms from the Robin
+    // condition, i.e., in the unsteady case.
+    matrix_free_data.append_mapping_flags(
+      BoundaryMassOperator<dim, Number, dim /* n_components */>::get_mapping_flags());
+  }
```
nfehn, line 150: Similar to this comment https://github.com/exadg/exadg/pull/555/files#r1457203165: I think we do not always have `Robin` type BCs, so I would expect an if-statement here like `if(robin){...}`?

include/exadg/structure/spatial_discretization/operators/nonlinear_operator.cpp
```diff
@@ -275,27 +282,83 @@ template<int dim, typename Number>
...
+                               (N * integrator.get_value(q)));
+            }
+            else
+            {
+              traction +=
+                coefficient_velocity * this->scaling_factor_mass_boundary * integrator.get_value(q);
+            }
+          }
+        }
+      }
```
nfehn, line 358: This code looks like a duplication of the one above in `linear_operator.cpp`. Is there any difference between these two version?

In the previous code version, one could say we had a function call `calculate_neumann_value()` in both cases (linear operator and nonlinear operator), but here we have a sequence of instructions and calculations that are duplicated, which should definitely be avoided.

include/exadg/operators/operator_base.cpp
```diff
@@ -1354,56 +1412,63 @@ OperatorBase<dim, Number, n_components>::face_loop_diagonal(
...
-    IntegratorFace(matrix_free, true, this->data.dof_index, this->data.quad_index);
-  IntegratorFace integrator_p =
-    IntegratorFace(matrix_free, false, this->data.dof_index, this->data.quad_index);
-
-  // create temporal array for local diagonal
-  unsigned int const                                     dofs_per_cell = integrator_m.dofs_per_cell;
-  dealii::AlignedVector<dealii::VectorizedArray<Number>> local_diag(dofs_per_cell);
-
-  for(auto face = range.first; face < range.second; ++face)
+  if(is_dg)
```
nfehn, line 1415: I think this is a mixture of different styles. In other places, we use functions `face_loop_empty()` so that we see on the top level (`matrix_free->loop()/cell_loop()`) which type of loops we have. I think we need a consistent way, otherwise we can here and there introduce if-statements "for safety". So I think we should not write `if(is_dg)` here.

include/exadg/operators/operator_base.cpp
```diff
@@ -823,10 +861,10 @@ OperatorBase<dim, Number, n_components>::internal_calculate_system_matrix(
   SparseMatrix & system_matrix) const
 {
   // assemble matrix locally on each process
-  if(evaluate_face_integrals() and is_dg)
+  if(evaluate_face_integrals())
```
nfehn, line 864: I find this logic irritating or not consistent with other parts of `OperatorBase` where we pass `empty` functions to `matrix_free->loop()`. I think we should distinguish between `dg` and `cg` on the top level, and inside each variant introduce another if depending on whether we have face loops or not. If we enter a `face_loop`, we will then also do the face integrals. See also comment https://github.com/exadg/exadg/pull/555/files#r1457261288

nfehn:
I think it makes sense to split this PR into parts as you suggest above. 

The starting point are probably changes made to `OperatorBase`. If I understand the present PR correctly, the main change is essentially about allowing - for continuous Galerkin discretizations - homogeneous boundary face integrals (while so far only boundary face integrals of inhomogeneous type are possible for continuous Galerkin, which stem from usual Neumann BCs)?

From the code review  I did so far, I see mainly three PRs (1. OperatorBase, 2. introduction of BoundaryMassOperator, 3. all the rest; but we should do them one after the other and then the picture might also become clearer).

PR #556, opened 2023-08-10
change order of function calls in IncNS::SpatialOperatorBase initialization routines

nfehn:
The order seems to be a bit weird currently, probably because the functionality has grown over time.

This PR aims to clean up these things.

@richardschu please check if you have any concerns regarding the new order of function calls.


kronbichler (approved):

richardschu (approved):
I am happy with the changes, shoudl be fine :+1:

PR #560, opened 2023-08-11
Introduce StVenantKirchhoffBase

richardschu:
This PR introduces a base class for the St Venant solid, whichs' derived classes `StVenantKirchhoffLargeDeformation` and `...SmallDeformation` implement the  `second_piola_kirchhoff_stress` and `second_piola_kirchhoff_stress_displacement_derivative` differently (the derivative of S is now asserted for linear elasticity, if we ever try to solve linear elasticity with a Newton solver). I did not go for free functions for all of the shared functionality, even if other material models might have coefficients to store and initialize/update as well. The main reason lies in the number of coefficients and types thereof being different and we do not want to introduce them in the `Material` class. The second thing this PR achieves is that we get rid of the `if(large_deformation)` in the stress evaluation.

I only did this because I promised I would do it and I am not sure if this actually pays off, since we can assert `second_piola_kirchhoff_stress_displacement_derivative` for small displacements and I am not sure if all this is worth it for 1 `if` and 1 assert ...

### DIFF
--- a/include/exadg/structure/material/library/st_venant_kirchhoff.cpp
+++ b/include/exadg/structure/material/library/st_venant_kirchhoff.cpp
@@ -248,7 +248,7 @@ StVenantKirchhoffSmallDeformation<dim, Number>::
   (void)deformation_gradient;
 
   AssertThrow(false,
-              dealiiExcMessage(
+              dealii::ExcMessage(
                 "This code path is not implemented in NonLinearOperator::do_cell_integral ."));
 
   // Exploit linear stress-strain relationship and symmetrizing in
--- a/include/exadg/structure/material/library/st_venant_kirchhoff.cpp
+++ b/include/exadg/structure/material/library/st_venant_kirchhoff.cpp
@@ -170,7 +170,7 @@ StVenantKirchhoffBase<dim, Number>::second_piola_kirchhoff_stress_symmetrize(
   unsigned int const cell,
   unsigned int const q) const
 {
-  dealii::Tensor<2, dim, dealii::VectorizedArray<Number>> S;
+  tensor S;
 
   if(E_is_variable)
   {
--- a/include/exadg/structure/material/library/st_venant_kirchhoff.cpp
+++ b/include/exadg/structure/material/library/st_venant_kirchhoff.cpp
@@ -247,6 +247,10 @@ StVenantKirchhoffSmallDeformation<dim, Number>::
 {
   (void)deformation_gradient;
 
+  AssertThrow(false,
+              dealiiExcMessage(
+                "This code path is not implemented in NonLinearOperator::do_cell_integral ."));
+
   // Exploit linear stress-strain relationship and symmetrizing in
   // second_piola_kirchhoff_stress_symmetrize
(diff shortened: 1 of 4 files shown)


richardschu:
I added an assert now, since the NonLinearOperator hard-codes the P=F*S linearization and I will not bloat the code for a case we are currently not considering (Newton + linear elasticity). After all, this PR is for discussion mainly.

include/exadg/structure/material/library/st_venant_kirchhoff.cpp
```diff
@@ -180,41 +204,75 @@ StVenantKirchhoff<dim, Number>::second_piola_kirchhoff_stress_symmetrize(
...
-  {
-    return (this->second_piola_kirchhoff_stress_symmetrize(
-      get_E<dim, Number>(get_F<dim, Number>(gradient_displacement)), cell, q));
-  }
-  else
-  {
-    return (this->second_piola_kirchhoff_stress_symmetrize(gradient_displacement, cell, q));
-  }
+  return (this->second_piola_kirchhoff_stress_symmetrize(
+    get_E<dim, Number>(get_F<dim, Number>(gradient_displacement)), cell, q));
```
nfehn, line 213: The `_symmetrize()` function has its origin in the linear case. Hence, it is not intuitive that the nonlinear case, where `E` is already symmetric, now also calls the `_symmetrize()` function.

include/exadg/structure/material/library/st_venant_kirchhoff.cpp
```diff
@@ -180,41 +204,75 @@ StVenantKirchhoff<dim, Number>::second_piola_kirchhoff_stress_symmetrize(
...
+  (void)deformation_gradient;
+
+  AssertThrow(false,
+              dealii::ExcMessage(
+                "This code path is not implemented in NonLinearOperator::do_cell_integral ."));
+
+  // Exploit linear stress-strain relationship and symmetrizing in
+  // second_piola_kirchhoff_stress_symmetrize
+  return (this->second_piola_kirchhoff_stress_symmetrize(gradient_increment, cell, q));
+}
```
nfehn, line 257: I think this function actually shows that inheritance might not be the method of choice to combine linear/nonlinear behavior. Inheritance is appropriate if one has a true "is-a" relationship, but a `StVenantKirchhoffSmallDeformation` does not seem to "be a" `StVenantKirchhoffBase` as this function shows.

nfehn:
I think I would not decide for the present changes relying on inheritance. I do not want to say that two classed for linear/nonlinear St-Venant-Kirchhoff per-se do not make sense, but I think I would not decide for a common base class. I would rather decide to realize the common functionality (e.g. `f_i` factors) as free functions, or as a class that is going to be used in the two variants of the St-Venant-Kirchhoff material via a member variable (called "composition" from a design perspective). Overall, I think there is not much pressure currently to change the current version with the if-statement.

richardschu:
Yep, I agree, I do not like all the hassle to avoid a single `if` tbh. For now I agree to not apply these changes; I view that as a little experiment :smile: . But even with composition I am not sure if I would like it. Maybe I will try that out later for me to compare.

PR #563, opened 2023-08-14
introduce a second Operator::setup() function for easy use in applications

nfehn:
We currently have to write lengthy code in the `Driver` classes in order to set up the spatial operators. In many cases, the `MatrixFree(Data)` objects can be created inside the `Operator` class. I therefore suggest to introduce a second `setup()` function in order to make it easier to write solvers/drivers for new problems/applications.

The original `setup()` function with `MatrixFree(Data)` as arguments will still be necessary whenever we need to create the `MatrixFree` object outside the spatial operator class. The typical use case would be multiphysics coupling with one `MatrixFree` object to be used/handed over to many single-field solvers.

~~So far, I only addressed the Poisson module.~~ One can see from the changes that the same sequence of instructions is currently copied three times in ExaDG regarding the Poisson module. The present PR removes this redundancy.

The use of ALE methods requires a special treatment because of the function doing `matrix_free->update_mapping(mapping)`, which is not `const`. Hence, the usual shared pointer to `MatrixFree const` does not work in this case.

### DIFF
--- a/include/exadg/structure/spatial_discretization/operator.h
+++ b/include/exadg/structure/spatial_discretization/operator.h
@@ -184,9 +184,16 @@ public:
   void
   fill_matrix_free_data(MatrixFreeData<dim, Number> & matrix_free_data) const;
 
-  /*
-   * Setup function. Initializes basic operators. This function does not perform the setup
-   * related to the solution of linear systems of equations.
+  /**
+   * Call this setup() function if the dealii::MatrixFree object can be set up by the present class.
+   */
+  void
+  setup();
+
+  /**
+   * Call this setup() function if the dealii::MatrixFree object needs to be created outside this
+   * class. The typical use case would be multiphysics-coupling with one MatrixFree object handed
+   * over to several single-field solvers.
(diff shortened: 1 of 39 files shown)


include/exadg/structure/spatial_discretization/operator.h
```diff
@@ -184,17 +184,24 @@ class Operator : public dealii::Subscriptor, public Interface::Operator<Number>
...
+   * over to several single-field solvers.
    */
   void
   setup(std::shared_ptr<dealii::MatrixFree<dim, Number> const> matrix_free,
         std::shared_ptr<MatrixFreeData<dim, Number> const>     matrix_free_data);
 
   /*
    * This function initializes operators, preconditioners, and solvers related to the solution of
-   * linear systems of equation required for implicit formulations.
+   * (non-)linear systems of equation.
```
richardschu, line 204:
```suggestion
   * (non-)linear systems of equations.
```

richardschu (approved):

PR #564, opened 2023-08-15
Clean up order of initialization steps in flow-transport driver

nfehn:
A preparation for upcoming PRs.


nfehn:
could someone please review this PR?

richardschu (approved):
Looks good, but I would feel way more confident, if we had a test in the `incompressible_flow_with_trasnsport` module! Did you test this?
What do you think about introducing a minimal test in one of the applications?

nfehn:
I think I will merge this one now to avoid inconsistencies in #599

PR #565, opened 2023-08-15
Extend div- and conti-penalty parameters to ElementType::Simplex

nfehn:
When calculating the penalty parameters, we derive a characteristic element length h from the element volume V_e (which is obtained from numerical integration / quadrature) assuming an "equilateral" shape of the element.

For hypercube elements, the relation is V_e = h^dim. However, this does not hold for simplicial elements. The present PR aims to fix these calculations in order to allow simplicial elements. Note that this PR aims to make the implementation consistent with the original idea regarding the construction of these penalty parameters (so far only realized for hypercube elements). It might be a separate topic to investigate which formula is actually ideally suited from a theoretical/methodological perspective.

### DIFF
--- a/include/exadg/grid/calculate_characteristic_element_length.h
+++ b/include/exadg/grid/calculate_characteristic_element_length.h
@@ -61,8 +61,9 @@ calculate_minimum_vertex_distance(dealii::Triangulation<dim> const & triangulati
  * by the number of nodes per coordinate direction. Hence, the result depends on the function space
  * (H^1 vs. L^2).
  */
-inline double
-calculate_characteristic_element_length(double const       element_length,
+template<typename Number>
+inline Number
+calculate_characteristic_element_length(Number const       element_length,
                                         unsigned int const fe_degree,
                                         bool const         is_dg)
 {
--- a/include/exadg/grid/calculate_characteristic_element_length.h
+++ b/include/exadg/grid/calculate_characteristic_element_length.h
@@ -27,6 +27,9 @@
 #include <deal.II/fe/mapping.h>
 #include <deal.II/grid/tria.h>
 
+// ExaDG
+#include <exadg/grid/grid_data.h>
+
 namespace ExaDG
 {
 /**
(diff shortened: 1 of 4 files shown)


include/exadg/grid/calculate_characteristic_element_length.h
```diff
@@ -55,16 +59,55 @@ calculate_minimum_vertex_distance(dealii::Triangulation<dim> const & triangulati
...
+
+    if(dim == 2)
+      factor = 4.0 / std::sqrt(3.0);
+    else if(dim == 3)
+      factor = 6.0 * std::sqrt(2.0);
+    else
+      AssertThrow(false, dealii::ExcMessage("Only dim = 2, 3 implemented."));
+
+    return std::exp(std::log(element_volume * factor) / (double)dim);
+  }
```
richardschu, line 91: As far as I know, we have for
`dim==2`: `A=a^2 * sqrt(3)/4`, so, all good,
but for
`dim==3`: `V=a^3 * sqrt(2)/12`, see, e.g., [link](https://keisan.casio.com/exec/system/1223345456) or [link](https://www.mathematische-basteleien.de/tetrahedron.htm)
which would give a different factor for the three-dimensional case?

include/exadg/grid/calculate_characteristic_element_length.h
```diff
@@ -55,16 +59,55 @@ calculate_minimum_vertex_distance(dealii::Triangulation<dim> const & triangulati
...
-calculate_characteristic_element_length(double const       element_length,
-                                        unsigned int const fe_degree,
-                                        bool const         is_dg)
+template<typename Number>
+inline Number
+calculate_high_order_element_length(Number const       element_length,
+                                    unsigned int const fe_degree,
+                                    bool const         is_dg)
 {
   unsigned int n_nodes_1d = fe_degree;
```
richardschu, line 113: interesting dependency on `is_dg` ... the number of node-to-node segments is always fe_degree+1, or what is the point here??

include/exadg/grid/calculate_characteristic_element_length.h
kronbichler: Isn't `(sqrt(2)/12)^{-1}` the same as `6 * sqrt(2)`?

richardschu: seems I need a coffee ...

include/exadg/grid/calculate_characteristic_element_length.h
nfehn: What we calculate here is the distance between nodes, assuming an equidistant distribution of nodes.

richardschu: > interesting dependency on `is_dg` ... the number of node-to-node segments is always fe_degree+1, or what is the point here??

i meant fe_degree-1, but nevermind, we talked about this in person ;)

PR #566, opened 2023-08-15
Enable inverse mass operator for simplex elements in all modules

### LINKED ISSUE #575: Behavior of block-Jacobi preconditioner unclear when called with identical dst, src vectors
The `BlockJacobiPreconditioner` is defined for DG discretizations. Hence, calling 

```c++
block_jacobi.vmult(dst, dst);
```
should be possible for performance reasons.

However, it seems as if the behavior of `OperatorBase::apply_inverse_block_diagonal(dst, dst)`, see https://github.com/exadg/exadg/blob/e8462c43964630daf4340b41723673a59c194156/include/exadg/operators/operator_base.cpp#L538-L566 is undefined in this regard.

Looking at the code, it seems as if `cell_loop_apply_inverse_block_diagonal_matrix_based()` allows identical `dst, src` vectors, since we read from `begin_dof_values` -> apply the inverse block matrix -> write into `begin_dof_values`.

nfehn:
addresses the inverse mass operator.

closes #575

### DIFF
--- a/include/exadg/operators/inverse_mass_operator.h
+++ b/include/exadg/operators/inverse_mass_operator.h
@@ -31,6 +31,7 @@
 #include <exadg/operators/inverse_mass_parameters.h>
 #include <exadg/operators/mass_operator.h>
 #include <exadg/solvers_and_preconditioners/preconditioners/block_jacobi_preconditioner.h>
+#include <exadg/solvers_and_preconditioners/preconditioners/jacobi_preconditioner.h>
 
 namespace ExaDG
 {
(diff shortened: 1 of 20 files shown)


richardschu (approved):

nfehn:
Let me change this to a draft, since I still need to address some aspects.

include/exadg/operators/inverse_mass_operator.h
```diff
@@ -81,16 +83,20 @@ class InverseMassOperator
...
+                dealii::ExcMessage("InverseMassOperator only implemented for DG!"));
+
+    // Currently, the inverse mass realized as matrix-free operator evaluation is only available
+    // in deal.II for tensor-product elements
+    if(fe.base_element(0).dofs_per_cell == dealii::Utilities::pow(fe.degree + 1, dim))
     {
-      // this checks if we have a tensor-product element
-      if(fe.base_element(0).dofs_per_cell == dealii::Utilities::pow(fe.degree + 1, dim))
-        explicit_matrix_free_inverse_mass_available = true;
+      inverse_mass_available_as_matrix_free_operator = true;
```
kronbichler, line 94: Let me add a (currently only theoretical) note of what we had discussed at some point in the past: Even if we have a tensor product of homogeneous degree, we might still want to do an iterative solution, e.g. because we have a (strongly deformed) geometry such that the evaluation of the mass operator is done with more quadrature points than `p+1` (or `k+1`). Since the inverse is strongly linked to `n_q_points_1d == p + 1`, we would then need to have different quadrature formulas for the forward evaluation and the inverse, and might want to use the iterative solver to gain this accuracy of the more accurate integration.

kronbichler: This comment does not mean that there is something specific to be changed, but it might be useful to reflect this in a comment somewhere that it is not a strict separation between "inverse via tensor product available" and "iterative inversion".

kronbichler (approved):
Despite still in the draft state, I had a careful look at the changes and agree, apart from a possible comment somewhere.

include/exadg/operators/inverse_mass_operator.h
nfehn: Do you mean we should extend the above if-statement and check if the number of quadrature points is `n_q_points_1d == p + 1`? Otherwise, we would set `inverse_mass_available_as_matrix_free_operator = false` and use the iterative solution similar to the simplex case.

Another option would be to hand the responsibility over to the user and let the user specify which method to choose. This would allow to e.g. use the iterative solver also in cases where the explicit tensor-product inverse would principally be available (currently, to run this version (e.g. for performance comparisons), one would need to "hack" the code). When deciding for this option, the implementation of the inverse mass operator would have to assert that the user does not set `inverse_mass_available_as_matrix_free_operator = true` is cases where deal.II does actually not support this. Deciding for this option would imply having *more control* and using *less automation*, but I would also agree to such an option.

include/exadg/solvers_and_preconditioners/preconditioners/elementwise_preconditioners.h
```diff
@@ -148,6 +159,14 @@ class InverseMassPreconditioner
   {
     integrator = std::make_shared<Integrator>(matrix_free, dof_index, quad_index);
     inverse    = std::make_shared<CellwiseInverseMass>(*integrator);
+
+    dealii::FiniteElement<dim> const & fe = matrix_free.get_dof_handler(dof_index).get_fe();
+
+    AssertThrow(
+      fe.conforms(dealii::FiniteElementData<dim>::L2) and
+        (fe.base_element(0).dofs_per_cell == dealii::Utilities::pow(fe.degree + 1, dim)),
```
nfehn, line 167: @kronbichler If I understand your other comment correctly, we would have to assert here (where we consider an inverse mass preconditioner for the elementwise, local problem) that the number of quadrature points is equal to the number of nodes?

nfehn:
The parameter `implement_block_diagonal_preconditioner_matrix_free` does not make sense for the H(div)-case. Hence, we should be strict and introduce a separate data struct for the H(div) case, see commit https://github.com/exadg/exadg/pull/566/commits/58e5920b32bd0c6456262cb22763b28b1733be66.

include/exadg/operators/inverse_mass_operator.h
kronbichler: I would personally prefer to have control, especially as you mention the performance tests. On the other hand, I'm wondering if the number of parameters to set would become overwhelming at some point - we talk about a single one here, to be set in case the default to use the matrix-free operator is not available. I think we can then proceed as you suggest, and assert in case the matrix-free operator is selected.

include/exadg/solvers_and_preconditioners/preconditioners/elementwise_preconditioners.h
kronbichler: Yes, exactly, otherwise the operator will not work. Or maybe it works, but it does not compute an appropriate inverse but some kind of pseudo inverse.

kronbichler:
> The parameter `implement_block_diagonal_preconditioner_matrix_free` does not make sense for the H(div)-case. Hence, we should be strict and introduce a separate data struct for the H(div) case

I am not entirely sure about this. If I read the commit right, we now have two different data structures instead, either targeting one or the other case, so there is redundancy on another level. In the end, we will have some ambiguity either way; I would personally prefer to have a single `param.inverse_mass_operator` and ignore the `implement_block_diagonal_preconditioner_matrix_free` in such a case because it feels less redundant. Nonetheless, there is no clear decision, which means the author of the PR should decide.

nfehn:
When making the changes, I felt the comments are clearer now, which somewhat supported the decision to go for a second struct. Note also that we had the parameters twice before the present PR. It was yesterday that I thought about unifying them. Another aspect is the fact that the inverse mass operator can also be used as a preconditioner in the DG case, while this is not available in the H(div) case. The separate structs communicate more clearly to the user that the inverse mass is something special in the H(div) case compared to the DG case, in my opinion.

kronbichler (approved):

include/exadg/operators/inverse_mass_operator.h
nfehn: Thinking more about this, I think one would need an enum

```c++
enum class
InverseMassType
{
  MatrixfreeOperator,
  ElementwiseKrylovSolver,
  BlockMatrices
};
````
instead of the boolean parameter
```c++
bool implement_block_diagonal_solver_matrix_free;
````

The second and third option of the above enum are available for all DG elements. In case the first option is set by the user, we need to assert that the current restrictions of the cellwise inverse mass operator in deal.II are met, in order to have useful error messages for ExaDG users. As a default value, I would initialize this variable with `InverseMassType::MatrixfreeOperator`. It is not the most general version, but the one that is used most frequently. I think this is better than using one of the other options as default, with the consequence that many users will not be aware of the fact that a faster option would be available for `ElementType::Hypercube`.

@kronbichler Do you agree?

File: include/exadg/solvers_and_preconditioners/preconditioners/elementwise_preconditioners.h (hunk 2 above)
Comment (nfehn) [reply include/exadg/solvers_and_preconditioners/preconditioners/elementwise_preconditioners.h:167]: Ok, then I need to find out how to ask `MatrixFree` for the number of 1D quadrature points given a `quad_index`.

File: include/exadg/operators/inverse_mass_operator.h (hunk 1 above)
Comment (kronbichler) [reply include/exadg/operators/inverse_mass_operator.h:94]: Yes, very good choice for both the enum and the default choice!

File: include/exadg/solvers_and_preconditioners/preconditioners/elementwise_preconditioners.h (hunk 2 above)
Comment (kronbichler) [reply include/exadg/solvers_and_preconditioners/preconditioners/elementwise_preconditioners.h:167]: You should be able to ask
```c++
matrix_free.get_shape_info(0, quad_index).data[0].n_q_points_1d;
```

nfehn: thanks! already included in the newest version

nfehn:
I think I am done for now. I opened another PR regarding the use of identical dst, src vectors for elementwise Krylov solvers. For the block-Jacobi preconditioner (used for the inverse mass operator), using identical dst, src vectors should be possible / work correctly.

kronbichler (approved):
Good, let me merge this right away.

PR #571, opened 2023-08-16
Add cylinder flag fsi test

richardschu:
. add a 2D test for the FSI application `cylinder_with_flag`
. introduce a scaling factor for the `end_time` used in the application to shorten the test duration
. compute reference values of velocity, pressure and displacement to compare in the generated output ("absolute error" taking zero as "exact solution")
. suppress iteration count and wall time output by mesh motion `ale_mapping->update` called during ALE update in fluid solver


richardschu:
... this is to remind me that I should check if the test was actually picked up

include/exadg/fluid_structure_interaction/single_field_solvers/fluid.h
```diff
@@ -148,10 +148,9 @@ SolverFluid<dim, Number>::setup(std::shared_ptr<FluidFSI::ApplicationBase<dim, N
   // initialize time_integrator
   helpers_ale = std::make_shared<HelpersALE<Number>>();
 
+  bool print_solver_info = is_test ? false : time_integrator->print_solver_info();
```
nfehn, line 151: I am not sure if this works as intended because `time_integrator->print_solver_info()` might be `true/false` depending on the time step (say solver info is going to be printed every 10th time step). However, the variable `print_solver_info` will then take the same value throughout a whole simulation, doesn't it?

richardschu: I moved it inside now, this is work in progress :)

applications/fluid_structure_interaction/cylinder_with_flag/CMakeLists.txt
```diff
@@ -30,3 +30,5 @@ TARGETNAME(TARGET_NAME ${CMAKE_CURRENT_SOURCE_DIR})
 PROJECT(${TARGET_NAME})
 
 EXADG_PICKUP_EXE(solver.cpp ${TARGET_NAME} solver)
+
+ADD_SUBDIRECTORY(tests)
```
kronbichler, line 34: Can you please insert a line break at the end of the file.

applications/fluid_structure_interaction/cylinder_with_flag/application.h
```diff
@@ -178,19 +200,20 @@ class Application : public FluidFSI::ApplicationBase<dim, Number>
 
     // PHYSICAL QUANTITIES
     param.start_time = 0.0;
-    param.end_time   = END_TIME;
+    param.end_time   = END_TIME * end_time_factor;
```
kronbichler, line 203: Should we really introduce a factor for the time via the input file, but keep the `END_TIME` as a separate variable? I would prefer to specify the actual final time, because setting information in two places is usually confusing.

kronbichler (commented):
A few random comments.

applications/fluid_structure_interaction/cylinder_with_flag/CMakeLists.txt
richardschu: sure, is something like [this](https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file) the reason?

PR #573, opened 2023-08-17
Optional extrapolation for FSI

richardschu:
Add an additional parameter to let the user decide, whether or not a second-order extrapolation shall be used.

This can be beneficial for problems where the extrapolation would lead to divergence or slower convergence (e.g. when I couple a Windkessel model to the fluid, such that the outlet pressure is rapidly changing)

### DIFF
--- a/applications/fluid_structure_interaction/pressure_wave/input.json
+++ b/applications/fluid_structure_interaction/pressure_wave/input.json
@@ -17,6 +17,7 @@
         "AbsTol": "1.e-9",
         "RelTol": "1.e-3",
         "OmegaInit": "0.3",
+        "UseExtrapolation": "true",
         "PartitionedIterMax": "100",
         "ReusedTimeSteps": "10",
         "GeometricTolerance": "1.e-8"
--- a/applications/fluid_structure_interaction/pressure_wave/input.json
+++ b/applications/fluid_structure_interaction/pressure_wave/input.json
@@ -17,7 +17,7 @@
         "AbsTol": "1.e-9",
         "RelTol": "1.e-3",
         "OmegaInit": "0.3",
-        "UseExtrapolation": "true",
+        "InitialGuessCouplingScheme": "SolutionExtrapolatedToEndOfTimeStep",
         "PartitionedIterMax": "100",
         "ReusedTimeSteps": "10",
         "GeometricTolerance": "1.e-8"
--- a/include/exadg/fluid_structure_interaction/acceleration_schemes/parameters.h
+++ b/include/exadg/fluid_structure_interaction/acceleration_schemes/parameters.h
@@ -40,6 +40,7 @@ struct Parameters
       abs_tol(1.e-12),
       rel_tol(1.e-3),
       omega_init(0.1),
+      use_extrapolation(true),
       reused_time_steps(0),
       partitioned_iter_max(100),
       geometric_tolerance(1.e-10)
@@ -62,6 +63,12 @@ struct Parameters
                         "Initial relaxation parameter.",
                         dealii::Patterns::Double(0.0, 1.0),
                         true);
+      prm.add_parameter(
+        "UseExtrapolation",
+        use_extrapolation,
+        "Use second order extrapolation in time as initial guess for FSI coupling scheme.",
+        dealii::Patterns::Bool(),
+        false);
       prm.add_parameter("ReusedTimeSteps",
                         reused_time_steps,
                         "Number of time steps reused for acceleration.",
@@ -85,6 +92,7 @@ struct Parameters
   double             abs_tol;
   double             rel_tol;
   double             omega_init;
+  bool               use_extrapolation;
   unsigned int       reused_time_steps;
   unsigned int       partitioned_iter_max;
 
--- a/include/exadg/fluid_structure_interaction/acceleration_schemes/parameters.h
+++ b/include/exadg/fluid_structure_interaction/acceleration_schemes/parameters.h
@@ -63,12 +63,11 @@ struct Parameters
                         "Initial relaxation parameter.",
                         dealii::Patterns::Double(0.0, 1.0),
                         true);
-      prm.add_parameter(
-        "UseExtrapolation",
-        use_extrapolation,
-        "Use second order extrapolation in time as initial guess for FSI coupling scheme.",
-        dealii::Patterns::Bool(),
-        false);
+      prm.add_parameter("UseExtrapolation",
+                        use_extrapolation,
+                        "Extrapolate for initial guess in coupling scheme.",
+                        dealii::Patterns::Bool(),
+                        false);
       prm.add_parameter("ReusedTimeSteps",
                         reused_time_steps,
                         "Number of time steps reused for acceleration.",
--- a/include/exadg/fluid_structure_interaction/acceleration_schemes/parameters.h
+++ b/include/exadg/fluid_structure_interaction/acceleration_schemes/parameters.h
@@ -34,6 +34,15 @@ enum class AccelerationMethod
   IQN_IMVLS
 };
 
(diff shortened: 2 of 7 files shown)


include/exadg/fluid_structure_interaction/driver.cpp
```diff
@@ -231,9 +231,13 @@ Driver<dim, Number>::coupling_structure_to_fluid(bool const extrapolate) const
   VectorType velocity_structure;
   structure->pde_operator->initialize_dof_vector(velocity_structure);
   if(extrapolate)
+  {
```
richardschu, line 234: I will pass iteration and use `get_velocity_n()` instead of `..._np()` to not assume the velocity from t^n still be stored in the vector for t^{n+1}

richardschu:
I forgot to consider `coupling_structure_to_ale(d);` uses an extrapolation as well!

include/exadg/fluid_structure_interaction/driver.cpp
```diff
@@ -223,17 +223,28 @@ Driver<dim, Number>::coupling_structure_to_ale(VectorType const & displacement_s
...
+    }
+    else
+    {
+      velocity_structure = structure->time_integrator->get_velocity_n();
+    }
+  }
   else
+  {
     velocity_structure = structure->time_integrator->get_velocity_np();
+  }
```
nfehn, line 247: I think we should introduce a ember function here similar to `update_structure_displacement()` introduced in this PR. The logic seems to be the same apart from replacing `displacement` by `velocity`, right?

include/exadg/fluid_structure_interaction/acceleration_schemes/parameters.h
```diff
@@ -62,6 +63,11 @@ struct Parameters
                         "Initial relaxation parameter.",
                         dealii::Patterns::Double(0.0, 1.0),
                         true);
+      prm.add_parameter("UseExtrapolation",
+                        use_extrapolation,
+                        "Extrapolate for initial guess in coupling scheme.",
+                        dealii::Patterns::Bool(),
+                        false);
```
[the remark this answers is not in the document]
richardschu: I translate `use_extrapolation` in the sense that an extrapolation in the first FSI iteration of a given time step should be used. An `enum` might be more descriptive. I am not sure we mean the same thing. Please see the comment below.

include/exadg/fluid_structure_interaction/driver.cpp
richardschu: yes! the motivation for the other function was that we use it in multiple places, while a new `update_structure_velocity()` only here. But I guess they follow the same logic, so why not ;)

include/exadg/fluid_structure_interaction/acceleration_schemes/parameters.h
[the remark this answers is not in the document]
richardschu: this PR is not about semi-implicit methods, maybe thats what you might be expecting given our conversations?

include/exadg/fluid_structure_interaction/driver.cpp
```diff
@@ -281,16 +292,18 @@ Driver<dim, Number>::apply_dirichlet_neumann_scheme(VectorType &       d_tilde,
   fluid->solve_ale();
 
   // update velocity boundary condition for fluid
-  coupling_structure_to_fluid(iteration == 0);
+  coupling_structure_to_fluid(iteration);
 
   // solve fluid problem
-  fluid->time_integrator->advance_one_timestep_partitioned_solve(iteration == 0);
+  fluid->time_integrator->advance_one_timestep_partitioned_solve(iteration ==
+                                                                 0 /* use_extrapolation */);
```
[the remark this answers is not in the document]
richardschu: I am not trying to implement the idea that we want to use the old solution values at `t_n` as starting guess for the fluid and structure solvers. This we might need for future comparison, but the present methodology to use the last iterates in the individual solvers makes a lot of sense (one of the points we wanted to address in the planned FSI paper).

What I am aiming for in this PR is:

keep the logic for re-using iterates in the individual field solvers, 
and
just for the first FSI coupling loop, 
make the use of an extrapolated structure velocity at the interface optional, since it is the first thing used to couple (structure -> ALE and structure -> fluid)
for the ALE, this was missing (as this is never really the problem, but the incompressible fluid might "overreact", which is why this PR might be interesting for larger/varying time steps)

include/exadg/fluid_structure_interaction/driver.cpp
richardschu: i will come up with a new version this evening, have to teach now ;)

include/exadg/fluid_structure_interaction/driver.cpp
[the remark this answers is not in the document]
nfehn: I think I understood it correctly, at least the basic intention. The only thing I am not sure about is that you are mentioning that only the structure velocity is affected by this PR. According to my understanding, also the displacements and `coupling_structure_to_ale(d);` are affected by this PR, because

1. call `update_structure_displacement(d, k);` -> calculation of displacement affected by the present PR
2. call `apply_dirichlet_neumann_scheme(d_tilde, d, k);` given `d` from 1., which then calls immediately
3. `coupling_structure_to_ale(d);` -> hence also affected by the present PR ?!
4. having solved the ALE problem, call `coupling_structure_to_fluid(iteration);`, where the calculation of the structure velocity is affected by the present PR

Do you agree?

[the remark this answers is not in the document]
richardschu: yes this is all correct, it should be "make use of an extrapolated structure velocity _and displacement_ at the interface optional". I have corrected the comment (marked bold).
The reason why I did not wrap the ALE update like:
`if(iteration > 0 or this->parameters.use_extrapolation)`
`{`
`  coupling_structure_to_ale(displacement_structure);`
`  fluid->solve_ale();`
`  coupling_structure_to_fluid(iteration);`
 `}`
is that the previous iteration does not update the fluid mesh anymore after solving the structure in the Dirichlet-Neumann coupling loop and reaching convergence. Hence, also if no extrapolation is used, that is, using the structural displacement at the previous time `t_n`, which is the last recent iterate (which lead to convergence), the mesh has to be updated. So all cases require an ALE update in my opinion.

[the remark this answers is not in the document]
richardschu: the `bool use_extrapolation` was indeed misleading I think.
One could not know, if we mean the FSI coupling loop initial guess or the individual field solvers!
the enum name is now quite long, but I use autocomplete anyways, hence prefer a more descriptive name.
but what do you think about:

`enum class InitialGuessFSI`
`{`
`  ExtrapolatedInTime,`
`  LastTimestepValue`
`}`

and hence a variable `initial_guess_fsi` (I always derive the variable name from the enum name to find it easily)

applications/fluid_structure_interaction/pressure_wave/input.json
```diff
@@ -17,6 +17,7 @@
         "AbsTol": "1.e-9",
         "RelTol": "1.e-3",
         "OmegaInit": "0.3",
+        "UseExtrapolation": "true",
```
nfehn, line 20: probably not a bug, but we should remove this again, or use the enum instead

nfehn:
Looks good overall. Let's fix the .json file and then merge.

applications/fluid_structure_interaction/pressure_wave/input.json
richardschu: ah sry forgot that

richardschu: I ran it now just to be sure:
pressure wave example
extrapolation: 6.78 iterations on avg
I will update #574 now and let you knwo when its ready!

PR #574, opened 2023-08-17
Semi implicit fsi

richardschu:
... based on #573 -> partly same lines are affected, address #573 first.
Add an Enum `UpdateMethod` to control if we want **implicit coupling**, **geometric explicit**, or **pressure&structure** (added-mass stable) or **velocity&structure** (added-viscosity stable) implicit coupling.
Member variables `update_velocity` and `update_pressure` control execution of sub-steps in the splitting/projection fluid time integrators.

### DIFF
--- a/include/exadg/fluid_structure_interaction/acceleration_schemes/parameters.h
+++ b/include/exadg/fluid_structure_interaction/acceleration_schemes/parameters.h
@@ -34,10 +34,20 @@ enum class AccelerationMethod
   IQN_IMVLS
 };
 
+enum class UpdateMethod
+{
+  Undefined,
+  Implicit,
+  GeometricExplicit,
+  ImplicitPressureStructure,
+  ImplicitVelocityStructure
+};
+
 struct Parameters
 {
   Parameters()
     : acceleration_method(AccelerationMethod::Undefined),
+      update_method(UpdateMethod::Implicit),
       abs_tol(1.e-12),
       rel_tol(1.e-3),
       omega_init(0.1),
@@ -58,6 +68,8 @@ struct Parameters
                         "Acceleration method.",
                         Patterns::Enum<AccelerationMethod>(),
                         true);
+      prm.add_parameter(
+        "UpdateMethod", update_method, "Update method.", Patterns::Enum<UpdateMethod>(), false);
       prm.add_parameter(
         "AbsTol", abs_tol, "Absolute solver tolerance.", dealii::Patterns::Double(0.0, 1.0), true);
       prm.add_parameter(
@@ -92,6 +104,7 @@ struct Parameters
   }
 
   AccelerationMethod acceleration_method;
+  UpdateMethod       update_method;
   double             abs_tol;
   double             rel_tol;
   double             omega_init;
--- a/include/exadg/fluid_structure_interaction/acceleration_schemes/parameters.h
+++ b/include/exadg/fluid_structure_interaction/acceleration_schemes/parameters.h
@@ -41,6 +41,7 @@ struct Parameters
       abs_tol(1.e-12),
       rel_tol(1.e-3),
       omega_init(0.1),
+      use_extrapolation(true),
       reused_time_steps(0),
       partitioned_iter_max(100),
       geometric_tolerance(1.e-10)
@@ -66,6 +67,11 @@ struct Parameters
                         "Initial relaxation parameter.",
                         dealii::Patterns::Double(0.0, 1.0),
                         true);
+      prm.add_parameter("UseExtrapolation",
+                        use_extrapolation,
+                        "Extrapolate for initial guess in coupling scheme.",
+                        dealii::Patterns::Bool(),
+                        false);
       prm.add_parameter("ReusedTimeSteps",
                         reused_time_steps,
                         "Number of time steps reused for acceleration.",
@@ -89,6 +95,7 @@ struct Parameters
   double             abs_tol;
   double             rel_tol;
   double             omega_init;
+  bool               use_extrapolation;
   unsigned int       reused_time_steps;
   unsigned int       partitioned_iter_max;
 
--- a/include/exadg/fluid_structure_interaction/driver.cpp
+++ b/include/exadg/fluid_structure_interaction/driver.cpp
@@ -223,17 +223,28 @@ Driver<dim, Number>::coupling_structure_to_ale(VectorType const & displacement_s
 
 template<int dim, typename Number>
 void
-Driver<dim, Number>::coupling_structure_to_fluid(bool const extrapolate) const
+Driver<dim, Number>::coupling_structure_to_fluid(unsigned int iteration) const
 {
   dealii::Timer sub_timer;
   sub_timer.restart();
 
   VectorType velocity_structure;
   structure->pde_operator->initialize_dof_vector(velocity_structure);
-  if(extrapolate)
-    structure->time_integrator->extrapolate_velocity_to_np(velocity_structure);
+  if(iteration == 0)
+  {
+    if(parameters.use_extrapolation)
+    {
+      structure->time_integrator->extrapolate_velocity_to_np(velocity_structure);
+    }
+    else
+    {
+      velocity_structure = structure->time_integrator->get_velocity_n();
+    }
+  }
   else
+  {
     velocity_structure = structure->time_integrator->get_velocity_np();
+  }
 
   structure_to_fluid->update_data(velocity_structure);
 
@@ -281,16 +292,18 @@ Driver<dim, Number>::apply_dirichlet_neumann_scheme(VectorType &       d_tilde,
   fluid->solve_ale();
 
   // update velocity boundary condition for fluid
-  coupling_structure_to_fluid(iteration == 0);
+  coupling_structure_to_fluid(iteration);
 
   // solve fluid problem
-  fluid->time_integrator->advance_one_timestep_partitioned_solve(iteration == 0);
+  fluid->time_integrator->advance_one_timestep_partitioned_solve(iteration ==
+                                                                 0 /* use_extrapolation */);
 
   // update stress boundary condition for solid
-  coupling_fluid_to_structure(/* end_of_time_step = */ true);
+  coupling_fluid_to_structure(true /* end_of_time_step */);
 
   // solve structural problem
-  structure->time_integrator->advance_one_timestep_partitioned_solve(iteration == 0);
+  structure->time_integrator->advance_one_timestep_partitioned_solve(iteration ==
+                                                                     0 /* use_extrapolation */);
 
   d_tilde = structure->time_integrator->get_displacement_np();
 }
--- a/include/exadg/fluid_structure_interaction/driver.cpp
+++ b/include/exadg/fluid_structure_interaction/driver.cpp
@@ -282,30 +282,83 @@ Driver<dim, Number>::coupling_fluid_to_structure(bool const end_of_time_step) co
 
 template<int dim, typename Number>
 void
-Driver<dim, Number>::apply_dirichlet_neumann_scheme(VectorType &       d_tilde,
-                                                    VectorType const & d,
-                                                    unsigned int       iteration) const
+Driver<dim, Number>::solve_subproblem_mesh(VectorType const & displacement_structure) const
 {
-  coupling_structure_to_ale(d);
+  // update displacement condition at interface
+  coupling_structure_to_ale(displacement_structure);
 
   // move the fluid mesh and update dependent data structures
   fluid->solve_ale();
+}
 
+template<int dim, typename Number>
+void
+Driver<dim, Number>::solve_subproblem_fluid(unsigned int const iteration,
+                                            bool const         update_velocity,
+                                            bool const         update_pressure) const
+{
   // update velocity boundary condition for fluid
   coupling_structure_to_fluid(iteration);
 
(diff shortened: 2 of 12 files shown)


richardschu:
@nfehn this together with #573 might be a good starting point for the FSI paper.
Should we work on merging these first and then proceed from there, or should I work from this branch straight away?
Also, should we test the fully (potentially useless) correct linearization of the boundary traction pull-back term in the structure module?

nfehn:
I think we should in fact work on getting things merged. I can look at this PR today or during the week.

nfehn:
I started with #573 according to what you have written in the description of this PR.

richardschu:
> I started with #573 according to what you have written in the description of this PR.
thanks! I am on it :100:

richardschu:
its a draft now since i want to check the pressure correction again, I think there is an inconsistency for the semi-implicit variant rn. I will switch to non-draft once ready for review!

include/exadg/fluid_structure_interaction/driver.cpp
```diff
@@ -258,30 +258,82 @@ Driver<dim, Number>::coupling_fluid_to_structure(bool const end_of_time_step) co
...
   structure->time_integrator->advance_one_timestep_partitioned_solve(iteration ==
                                                                      0 /* use_extrapolation */);
+}
+
+template<int dim, typename Number>
+void
+Driver<dim, Number>::apply_dirichlet_neumann_scheme(VectorType & displacement_structure_updated,
+                                                    VectorType const & displacement_structure,
+                                                    unsigned int       iteration) const
+{
```
richardschu, line 301: the content in the `if` statement here is quite repetitive and could be shortened.
But I think this way its easier to read ... I changed back and forth multiple times joining some cases and splitting only the semi-impilcit ones off ... only to realize that it is easier to read this way.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf.cpp
```diff
@@ -171,10 +173,24 @@ TimeIntBDF<dim, Number>::ale_update()
...
-TimeIntBDF<dim, Number>::advance_one_timestep_partitioned_solve(bool const use_extrapolation)
+TimeIntBDF<dim, Number>::advance_one_timestep_partitioned_solve(bool const use_extrapolation,
+                                                                bool const update_velocity,
+                                                                bool const update_pressure)
 {
   this->use_extrapolation = use_extrapolation;
   this->store_solution    = true;
+  this->update_velocity   = update_velocity;
+  this->update_pressure   = update_pressure;
+
```
richardschu, line 184: checking if one uses `BDFCoupledSolution` and a semi-implicit method could be done at setup, but only with the combined `Parameters` from fluid and FSI. This is did not do _yet_, but put the assert here.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_pressure_correction.cpp
```diff
@@ -318,19 +318,26 @@ void
...
+  {
+    VectorType pressure_increment;
+    pressure_increment.reinit(pressure_np, false /* init with zero */);
 
-  pressure_step(pressure_increment);
+    pressure_step(pressure_increment);
 
-  projection_step(pressure_increment);
+    projection_step(pressure_increment);
+  }
```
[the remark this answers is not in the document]
richardschu: I have solved this issue now reusing the anyways stored `pressure_increment_last_iter`.
One could also recompute the pressure increment (using a `+div(u)` term for the rotational schemes, so its not a simple vector addition unfortunately).
But since recovering velocity and pressure is only relevant for multiphysics problems (via `advance_one_timestep_partitioned_solve()`), where we always store the solution, we might as well use it :smile:

include/exadg/operators/operator_base.h
```diff
@@ -270,13 +270,16 @@ class OperatorBase : public dealii::Subscriptor
 
     if(dof_handler.has_level_dofs())
     {
```
richardschu, line 272: this is actually covered in
#713 but was needed to run.

richardschu:
@nfehn I refactored this and its ready for you to look at! thanks in advance!

nfehn:
Thanks for notifying. I can now start looking into this PR in more detail. Sorry, I was quite busy with teaching duties.

include/exadg/fluid_structure_interaction/driver.cpp
```diff
@@ -258,30 +258,82 @@ Driver<dim, Number>::coupling_fluid_to_structure(bool const end_of_time_step) co
...
+    solve_subproblem_ale(displacement_structure);
+    solve_subproblem_fluid(iteration, true /* update_velocity */, true /* update_pressure */);
+    solve_subproblem_structure(iteration);
+  }
+  else if(parameters.update_method == UpdateMethod::GeometricExplicit)
+  {
+    if(iteration == 0)
+    {
+      solve_subproblem_ale(displacement_structure);
+    }
```
nfehn, line 313: We should make sure that, - if we skip this function - `MatrixFree` does not perform any updates of the geometry terms and/or that no other part of the code triggers any "geometry updates" in the single field solvers (which could cause computational costs that can be avoided in this case). I assume you already did this. So this comment is just there to not forget this.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_dual_splitting.cpp
```diff
@@ -316,20 +316,30 @@ TimeIntBDFDualSplitting<dim, Number>::do_timestep_solve()
...
-  convective_step();
+  if(this->update_velocity)
+  {
+    convective_step();
+  }
 
-  pressure_step();
+  if(this->update_pressure)
+  {
+    pressure_step();
```
nfehn, line 326: Assuming we have `update_velocity == false` and `update_pressure == true` and that we do not solve for the ALE problem: What does the fluid time stepper actually solve for in this case? It solves the pressure but why does this actually change the solution? I understand that in this case we have a coupling "fluid -> structure" via the stress boundary condition (which is affected by the pressure solution), but what exactly is the coupling "structure -> fluid" in this case? Is it only the coupling via the pressure Neumann boundary condition that remains?

Before looking at this code, I would have expected that this UpdateMethod solves the convective step (which is explicit and which affects the pressure rhs-vector) as well, but then only skip the steps after the `pressure_step()` (which are implicit and which might be expensive). If we implement this variant as described here, it is important that we make sure that the `convective_step()` does not manipulate any velocity dof vectors apart from temporary vectors (or vectors of local scope). It could then potentially happen that we falsify the "true" velocity solution (which is "unconverged" in the FSI sense, but taking into account the viscous term and the projection onto divergence free vectors).

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_pressure_correction.cpp
```diff
@@ -318,19 +318,35 @@ void
...
 
-  // evaluate convective term once the final solution at time
-  // t_{n+1} is known
-  evaluate_convective_term();
+  // evaluate convective term once the final solution at time t_{n+1} is known
+  if(this->update_velocity)
+  {
+    evaluate_convective_term();
+  }
```
nfehn, line 349: This change could be dangerous in general. I am not completely sure without looking into the code in detail, but we might (or, right now, I actually think that we) need this step so that the next time step is computed correctly. This would imply that we need it in the last FSI iteration even if `this->update_velocity == false`. We do not know at this point whether this will be the last FSI iteration. So if this reasoning is true, we would have to call this function always.

include/exadg/operators/operator_base.h
nfehn: has been merged, so you could rebase?

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_pressure_correction.cpp
```diff
@@ -318,19 +318,35 @@ void
...
 
   VectorType pressure_increment;
-  pressure_increment.reinit(pressure_np, false /* init with zero */);
-
-  pressure_step(pressure_increment);
+  if(this->update_pressure)
+  {
+    pressure_increment.reinit(pressure_np, false /* omit_zeroing_entries */);
+    pressure_step(pressure_increment);
+  }
```
nfehn, line 331: A similar question as here: https://github.com/exadg/exadg/pull/574/files#r1860844732

If `update_velocity == false` and `update_pressure == true` the coupling "structure -> fluid" would only be via the velocity DBC entering the pressure rhs vector, correct?

include/exadg/fluid_structure_interaction/acceleration_schemes/parameters.h
```diff
@@ -34,6 +34,15 @@ enum class AccelerationMethod
...
 };
 
+enum class UpdateMethod
+{
+  Undefined,
+  Implicit,
+  GeometricExplicit,
+  ImplicitPressureStructure,
+  ImplicitVelocityStructure
+};
```
nfehn, line 44: I think I would not implement it like this, but instead use three boolean parameters:

```C++
bool update_ale;
bool update_fluid_velocity;
bool update_fluid_pressure;
```
Default values are `true` such that the standard Dirichlet-Neumann-scheme with all couplings is the default.

Advantages:
- intuitive to understand for a user in my opinion
- the notion of implicit and explicit in the above `enum` is not clear a priori in my opinion without documentation. Moreover, the terms `ImplicitPressureStructure` and `ImplicitVelocityStructure` do not intuitively communicate that they inherently involve `GeometricExplicit` in my opinion.
- the above `enum` does not include the variant `GeometricImplicit_PressureStructure` or `GeometricImplicit_VelocityStructure`. This example also shows why I do not like the `Implicit`-wording in the above enum. One might in fact expect `GeometricImplicit` for the last two variants because of `Implicit` in the name, but this is not the case!?
- The above enum requires an if-else if-implementation with currently three branches. Realizing also the two additional variants mentioned here would require one/two more branche(s) following the current implementation style.
- using 3 booleans instead would be very close to the actual implementation of the FSI coupling scheme and the fluid single field solver with exactly these parameters in the interface. 
- The implementation in the FSI code for all 6 variants would collaps to a single "branch"
```C++
if(iter == 0 or update_ale)
  solve_subproblem_ale(displacement_structure);
solve_subproblem_fluid(iteration,  iter==0 or update_velocity, iter== 0 or update_pressure);
solve_subproblem_structure(iteration);
```

for all 6 variants with the 3 boolean parameters (principally 8 variants, but `update_velocity = update_pressure = false` is excluded since this is equivalent to no FSI-coupling at all).

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_dual_splitting.cpp
richardschu: yes, it just updates the pressure given new boundary data.
in the dual-splitting scheme, it enters via $h_p$:
pressure step:
image
with $h_p$ defined as:
image
(and where the read part should be considered when extending towards gen Newtonian fluids, coming soon!)

The semi-implicit versions are not unique in general,  but I wanted to get to the cheapest scheme possible.
The relevant BCs do capture the change of interface data (a change in the displacement "reaches" the PPE).
This is the case since `do_timestep_solve()` in any case involves
` pde_operator->interpolate_velocity_dirichlet_bc(velocity_dbc_np, this->get_next_time());`

I think we follow the credo 
"update_velocity == false means not touching the velocity fields"
we should not come into a situation, where we would have to execute the entire scheme to get a valid `velocity_np`. Vice versa, same holds true for the pressure!

So in general I would say i) we have to be careful not to leave solution vectors in an "invalid" state (meaning not considering for some terms) and ii) that we do not perform updates that are not necessary (for efficiency).
Or do you mean something different here? Am I overlooking something?

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_pressure_correction.cpp
richardschu: The semi-implicit version in this PR focus on the FSI case only. 
(the fluid solvers are also semi-implicit for some combinations, see `TreatmentOfConvectiveTerm`)
within a fluid solver *alone*, however we always call `do_timestep_solve()` with 
`update_pressure == true` and `update_velocity == true`
as they are initialized with `true`.

So whatever happens in `evaluate_convective_term()` can now only lead to changed behavior in the FSI case.
FSI always needs the ALE formulation. Some people might want to do this, but this combination is asserted for now.
Anyways, this function does nothing for the ALE case:
image

so I think we should be good.
The question wether or not FSI always requires ALE is I think a different one.

include/exadg/fluid_structure_interaction/acceleration_schemes/parameters.h
richardschu: Good point, this can easily be way more intuitive and more general, i.e., covering more combinations.
 I would want to name them a bit differently though, since with the current name, one might think they are otherwise not updated at all. But I can't find a different name; let's keep the ones you suggested! (or any ideas?)

richardschu:
@nfehn this would be ready for you to look at if you have time :christmas_tree: :santa: :gift:

richardschu:
@nfehn if you have time to look at this again ... would be greatly appreciated!

PR #579, opened 2023-09-06
split multistep and BDF

jh66637:
This PR splits the Multistep mechanisms from the BDF mechanisms and is a direct follow-up of https://github.com/exadg/exadg/pull/495. Splitting the mechanisms, needed for any multistep scheme, from those that are only needed for BDF makes it simple to add other Multistep methods, such as Adams-Bashforth.

`update_time_integrator_constants()` is made pure virtual, since every multistep class manages its own coefficients and, thus, needs to implement this function itself. All other functions which are only used in BDF schemes are moved to the BDF class.

@nfehn

### DIFF
--- a/include/exadg/time_integration/time_int_bdf_base.h
+++ b/include/exadg/time_integration/time_int_bdf_base.h
@@ -1,0 +1,64 @@
+/*  ______________________________________________________________________
+ *
+ *  ExaDG - High-Order Discontinuous Galerkin for the Exa-Scale
+ *
+ *  Copyright (C) 2021 by the ExaDG authors
+ *
+ *  This program is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *  ______________________________________________________________________
+ */
+
+#ifndef INCLUDE_EXADG_TIME_INTEGRATION_TIME_INT_BDF_BASE_H_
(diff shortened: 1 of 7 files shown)


nfehn:
If I remember correctly, there was a preceding PR on BDF/multistep, that has already been merged? If so, could you refer to this one in the description of this PR just for completeness of information/documentation?

include/exadg/time_integration/time_int_bdf_base.h
```diff
@@ -0,0 +1,64 @@
...
+protected:
+  void
+  update_time_integrator_constants() override;
+
+  /*
+   * Time integration constants. The extrapolation scheme is not necessarily used for a BDF time
+   * integration scheme with fully implicit time stepping, implying a violation of the Liskov
+   * substitution principle (OO software design principle). However, it does not appear to be
+   * reasonable to complicate the inheritance due to this fact.
+   */
```
nfehn, line 58: good comment!

include/exadg/time_integration/time_int_multistep_base.h
```diff
@@ -65,12 +63,6 @@ class TimeIntBDFBase : public TimeIntBase
   void
   timeloop_steady_problem();
 
-  /*
-   * Setters and getters.
-   */
-  double
-  get_scaling_factor_time_derivative_term() const;
-
```
nfehn, line 73: For clarity: Is there a reason why you removed this function from the interface, while you made the function `update_time_integrator_constants()` below pure virtual in the present PR?

This seems to be the main change of this PR (and could also be briefly mentioned in the PR description).

nfehn (approved):
looks good overall.

include/exadg/time_integration/time_int_multistep_base.h
jh66637: ```update_time_integrator_constants()``` is needed for all multistep methods. The scaling factor for the time derivative term is only needed for BDF (at least, I didn't use it in the Adams methods). I would suggest proceeding like this and if you feel it would be helpful also for Adams methods move it back to the base class. Would this be OK?

nfehn: A question is also whether `update_time_integrator_constants()` should still be in the base class, given that we no longer have the member variables for the time integrator constants in the base class? As you suggest for the other question above, we can also keep this question in mind for the upcoming changes and then decide for a "final" design.

PR #582, opened 2023-09-26
Cell-centric loops: assert that no locally refinded meshes are used

peterrum:


peterrum:
@richardschu Hope this gives better error messages.

include/exadg/operators/operator_base.cpp
```diff
@@ -425,6 +425,11 @@ OperatorBase<dim, Number, n_components>::add_diagonal(VectorType & diagonal) con
   {
     if(data.use_cell_based_loops)
     {
+      AssertThrow(
+        not evaluate_face_integrals() ||
```
nfehn, line 429: `evaluate_face_integrals()` will be `true` due to the if-statement a few lines above. So I believe this code is hard to understand/read.

nfehn:
I would suggest that we always provide some information in the description of a PR, so that people not involved get some context.

include/exadg/operators/operator_base.cpp
nfehn: Principally, we decided to write `or` in ExaDG. This comment will probably be superfluous due to the other comment above suggesting to remove one of the booleans.

peterrum: I have removed the additional check.

kronbichler (approved):

nfehn:
I think we should discuss why these checks are not done here: https://github.com/exadg/exadg/blob/master/include/exadg/matrix_free/categorization.h, which would be a better separation of concerns in my opinion than inserting this check many times in `OperatorBase` (and potentially elsewhere).

I might also miss something ... As written above, this PR does not provide information in the description.

richardschu:
> I think we should discuss why these checks are not done here: https://github.com/exadg/exadg/blob/master/include/exadg/matrix_free/categorization.h, which would be a better separation of concerns in my opinion than inserting this check many times in `OperatorBase` (and potentially elsewhere).

So you mean to move [assert](https://github.com/exadg/exadg/pull/582/files#diff-57bc5fe81cd78ec816b05e45d32c1a9e842014a5b71484160259366b9c6a4fd1R428-R431), [assert](https://github.com/exadg/exadg/pull/582/files#diff-57bc5fe81cd78ec816b05e45d32c1a9e842014a5b71484160259366b9c6a4fd1R503-R506) and [assert](https://github.com/exadg/exadg/pull/582/files#diff-57bc5fe81cd78ec816b05e45d32c1a9e842014a5b71484160259366b9c6a4fd1R558-R560) to [catergorization.h](https://github.com/exadg/exadg/blob/master/include/exadg/matrix_free/categorization.h), right?
The `if` statements in lines 1512 and 1812 of òperator_base.cpp` are required I think.

I need this in some form to proceed with #586 ...

nfehn:
I think we should at least have a discussion about this. Personally, I do not see `OperatorBase` in the responsibility to check whether certain loops are possible. The fact that the same assert is inserted three times in  `OperatorBase` might support this perspective. I have two questions:

- Why does not dealii/matrixfree do the assert, because it is in my opinion actually dealii that implements the loop and hence should check that functionality is not mis-used?
- Why not doing this check at a central place in ExaDG? I suggested categorization.h because this is about cell-based face loops.

richardschu:
>     * Why does not dealii/matrixfree do the assert, because it is in my opinion actually dealii that implements the loop and hence should check that functionality is not mis-used?

Does the assert message: "Cell based face loops do not support locally refined meshes." refer to deal.II or could this be done differently in ExaDG, such that deal.II is not responsible and/or can not check if we are using it correctly? 

>     * Why not doing this check at a central place in ExaDG? I suggested categorization.h because this is about cell-based face loops.

We can, I have no problem with that. Having the assert right before one would run into the problem makes it a bit easier to understand on the other side.

PR #585, opened 2023-10-26
Add Adams-Bashforth-Moulton time integrator

jh66637:
@nfehn This PR introduces the Adams-Bashforth-Moulton time integrator. The class takes an operator, similar to the Runge--Kutta time-integrators. ALE could be easily incorporated into the class, but this is for another PR.

*  `initialize_former_solutions` is renamed to `initialize_former_multistep_dof_vectors`. This is because for the Adams methods, no former solutions are stored. Instead, former vectors of evaluated operators are stored. 
* While the PR does not introduce mesh movement, the push-back of the vectors happens with ALE in mind. In the ALE case one has to compute the prediction on an old mesh instance, move the mesh and compute the correction on the new mesh instance. To avoid splitting up the function that performs one time step, or to provide information on how to move the mesh to the class, the function starts with the correction of the predicted solution. As a consequence, the pushback of the vectors has to be done inside  the function and not afterward. This very specific for ABM. For standard multistep methods (AB, AM, and BDF) the push-back has to be done after the time step, and therefore I think the multistep time integration should still call the push-back function as it is currently implemented.


include/exadg/time_integration/time_int_abm_base.h
```diff
@@ -0,0 +1,266 @@
...
+  ABTimeIntegratorConstants ab;
+  AMTimeIntegratorConstants am;
+
+  // solution vector
+  VectorType solution;
+  // temporary vector to store prediction
+  VectorType prediction;
+
+  // store evaluated operators from previous time steps
+  VectorType              evaluated_operators_np;
```
nfehn, line 260: Regarding the naming, it should be `evaluated_operator_np`, since it is only one time vector/operator in my opinion.

include/exadg/time_integration/time_int_abm_base.h
```diff
@@ -0,0 +1,266 @@
...
+    pde_operator->prescribe_initial_conditions(solution, get_time());
+  }
+
+  void
+  initialize_former_multistep_dof_vectors() final
+  {
+    if(start_with_low_order)
+    {
+      if(vec_evaluated_operators.size() > 0)
+        evaluate_operator(vec_evaluated_operators[0], solution, get_time());
```
nfehn, line 121: What is the content of `solution` at this point? Only one line below the initial condition is prescribed. I am irritated.

include/exadg/time_integration/time_int_abm_base.h
```diff
@@ -0,0 +1,266 @@
...
+
+#include <exadg/time_integration/ab_constants.h>
+#include <exadg/time_integration/am_constants.h>
+#include <exadg/time_integration/push_back_vectors.h>
+#include <exadg/time_integration/time_int_multistep_base.h>
+
+namespace ExaDG
+{
+template<typename Operator, typename VectorType>
+class TimeIntABMBase : public TimeIntMultistepBase
```
nfehn, line 34: I tend to write `ABM` explicitly as `AdamsBashforthMoulton` in the class name.

In my opinion, a comment / some documentation for the class should explain that this is an explicit time integrator of predictor-corrector type. One possible interpretation of ABM could be that this class unites AB time integrators and AM time integrators, and potentially also the predictor-corrector version ABM.

include/exadg/time_integration/time_int_abm_base.h
```diff
@@ -0,0 +1,266 @@
...
+    // In case the mesh is moving, the mesh has to be updated between prediction and correction.
+    // Pushing back the vectors here makes it possible to move the mesh before do_timestep_solve(),
+    // thus this class does not have to know how to move the mesh.
+    if(vec_evaluated_operators.size() > 0)
+    {
+      push_back(vec_evaluated_operators);
+      std::swap(vec_evaluated_operators[0], evaluated_operators_np);
+    }
+
+    // predict next timestep
```
nfehn, line 178: Are you predicting the solution at time `t_{n+2}` here? If so, this should be specified explicitly in the comment. Note that you also write "next time step" some lines above, but it seems that you mean different time instants here and in the comments above. It might also be that I do not understand it.

You already did the push back of vectors above. Hence, I wonder whether the time integrator constants used in the line below are actually correct?

In my opinion, the logic in ExaDG's time integrators that time integrator constants are updated at the beginning of a time step and the push back of vectors is done at the end of a time step should not be broken.

include/exadg/time_integration/time_int_abm_base.h
```diff
@@ -0,0 +1,266 @@
...
+  virtual double
+  recalculate_time_step_size() const = 0;
+
+  void
+  do_timestep_solve() final
+  {
+    dealii::Timer timer;
+    timer.restart();
+
+    // predict operator for next time step
```
nfehn, line 162: This comment seems somewhat imprecise. What I would like more is e.g. "evaluate operator given the predicted solution"?

nfehn:
The aspect with ALE seems to be somewhat complicated. In one comment, I raised the concern whether the update of the time integrator constants and the push-back of vectors get out-of-sync, which would lead to wrong behavior / consistency errors in case of adaptive time step sizes.

Why not going for the following solution:

```C++
class
TimeIntABM
{
public:
void
do_timestep()
{
  AssertThrow(ale == false, ...);
  do_timestep_predict();
  do_timestep_correct();
}

void
do_timestep_predict();

void
do_timestep_correct();
};
```

Driver for ALE case:
```C++
while (/*time loop not finished*/)
{
  do_timestep_predict();
  ale_update();
  do_timestep_correct();
}
```
I think this could be an intuitive implementation of what one would otherwise have to explain in more detail in the code.

Passing a lambda to the time integrator for the `ale_update()` could be an alternative. However, for coupled problems (with multiple single field solvers operating on the same mesh) one wants to do the ALE update only once. This would complicate again the solution with a lambda passed to the time integrator. So the solution suggested above could be quite flexible, not only for realizing ALE but also for ALE+multiphysics.

What do you think @jh66637 ?

jh66637:
@nfehn This works fine for me.

include/exadg/time_integration/time_int_abm_base.h
jh66637: [ebc3e2a](https://github.com/exadg/exadg/pull/585/commits/ebc3e2a9db5d954fc665222adf52a981ac34070c)

include/exadg/time_integration/time_int_abm_base.h
jh66637: `solution` is correctly initialized at this point. The initial condition does not have to be prescribed again. [d295d6a](https://github.com/exadg/exadg/pull/585/commits/d295d6a60072842c980ccebd0eb64ee62e41811b)

include/exadg/time_integration/time_int_abm_base.h
jh66637: [d16cfd3](https://github.com/exadg/exadg/pull/585/commits/d16cfd35392b2acbf7f018d2db2649329285102b)

include/exadg/time_integration/time_int_abm_base.h
jh66637: This is superfluous since we agreed to split the time step computation

jh66637:
@nfehn I restructured as suggested in [7a46047](https://github.com/exadg/exadg/pull/585/commits/7a460471668fbef3a38a2f3721bd820f0b14be2a). I will ping you once I tested if I broke something.

include/exadg/time_integration/time_int_abm_base.h
jh66637: Done in [7a46047](https://github.com/exadg/exadg/pull/585/commits/7a460471668fbef3a38a2f3721bd820f0b14be2a)

jh66637:
@nfehn I checked the results with the restructured and everything works as expected. Note, that I will add tests once a minimal version of the acoustic module is merged.

include/exadg/time_integration/time_int_abm_base.h
```diff
@@ -0,0 +1,270 @@
...
+  {
+    AssertThrow(order_ >= 1,
+                dealii::ExcMessage("Oder of ABM time integrator has to be at least 1."));
+  }
+
+  void
+  print_iterations() const
+  {
+    // explicit time integration -> one iteration
+    print_list_of_iterations(pcout, {"Adams-Bashforth-Moulton"}, {1});
```
nfehn, line 75: If I remember correctly, we do it somewhat differently in other parts of ExaDG (Explicit == 0 iterations), see e.g. the dual-splitting scheme: https://github.com/exadg/exadg/blob/56272c15cceb9ecc1300a176c7c5f11cb78604ca/include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_dual_splitting.cpp#L954

nfehn:
Apart from the `print_iterations()` function, this would be ready-to-merge from my side. 

@jh66637 please let me know once this is ready from your side.

include/exadg/time_integration/time_int_abm_base.h
jh66637: I changed it to zero iterations :)

nfehn (approved):

PR #586, opened 2023-10-26
Adaptive mesh refinement for ConvDiff applications

richardschu:
This PR is based on #582 and developed in collaboration with @peterrum and @nfehn.
It introduces AMR to the convection diffusion module, including a test making up for 50% (750 lines) of the changes added here.

We aim to extend the current design to the other modules as well. It goes as follows:

----

The driver gets 3 new functions:
1) `mark_cells_for_coarsening_and_refinement(tria, solution)` : Flags cells for refinement and coarsening depending on some `solution` vector or other criteria depending on the application. That is, the driver can decide depending on the application how to refine/coarsen. For now, this just uses a free function adopting the Kelly error estimator.
2) `setup_after_coarsening_and_refinement()` : repeats the necessary parts after AMR was executed. Similar to `Driver::setup()`, but omits some unnecessary/undesired parts.
3) `do_adaptive_refinement(time_step_number)`: actually executes functions to refine/coarsen, to be called in `Driver::solve()` similar to ALE update routines. Ensure right sequence of function calls.

----

We have free functions (in parts taken from MeltPoolDG):
1) `trigger_coarsening_and_refinement_now(amr_data, time_step_number)` to be used in the driver to check, if we want to execute AMR in the current time step.
2) `limit_coarsening_and_refinement(tria, amr_data)`, which based only on the triangulation sets/clears refine flags on the triangulation
3) `any_cells_flagged_for_coarsening_or_refinement` : check, if any refinement/coarsening flags are set (possibly sparing the interpolation and amr setup).
4) `mark_cells_kelly_error_estimator` : standard implementation of the Kelly-error-estimator-based refinement/coarsening (setting flags in the triangulation).

----

In the `TimeIntegrator`, the relevant functions are:

`time_integrator->prepare_coarsening_and_refinement();` and
`time_integrator->interpolate_after_coarsening_and_refinement();`,
where the solution vector pointers owned by the time integrator are passed to the spatial operator

*public* `get_solution_np()` for the refinement criterion evaluated in the driver
and a *private* `get_vectors()` to pass the vector pointers to the spatial operator.

Former two are general for all time integrators, hence introduced in `TimeIntBase`, but trigger assertions when not overwritten by derived classes. Overwriting the functions in the derived time integrators such as `TimeIntBDF` as in the present case, allows packing the relevant vectors (which differ from time integrator to time integrator).  'get_solution_np()' could also be considered in the base class, but this is optional, since refinement does not *need* to be based on some solution vector.

----

The spatial operator also has some `prepare_coarsening_and_refinement(vectors)` and `interpolate_after_coarsening_and_refinement(vectors)` and is responsible for the interpolation of the vectors provided by the time integrator.

----

The `ExaDG::SolutionTransfer` is the last ingredient and was inspired by MeltPoolDG as well (there, these were free functions using lambdas). It needs the `DoFHandler` provided during initialization.
Therefore it can get the (p::d::)`Triangulation` and can initialize the correct (p::d::)`SolutionTransfer`. The vector storage for the `dealii::Triangulation` is a private class member and is set up/deleted automatically when calling `interpolate_after_coarsening_and_refinement(vectors)`. This class will be extended towards multiple `DoFHandlers` and vectors of vector pointers to enforce correct calling sequence in the future.


include/exadg/operators/solution_transfer.h
```diff
@@ -0,0 +1,147 @@
...
+      {
+        vectors[i]->copy_locally_owned_data_from(vectors_new_grid[i]);
+      }
+    }
+
+    // Clear the containers holding actual copies or pointers depending
+    // on the SolutionTransfer type. In both cases, the containers need
+    // to be in scope when execute_coarsening_and_refinement is called.
+    vectors_old_grid.clear();
+  }
```
richardschu, line 127: vector container is cleared after using it.

include/exadg/operators/solution_transfer.h
```diff
@@ -0,0 +1,147 @@
...
+template<int dim, typename VectorType>
+class SolutionTransfer
+{
+public:
+  /*
+   * Constructor.
+   */
+  SolutionTransfer(dealii::DoFHandler<dim> const & dof_handler_in)
+  {
+    dof_handler = &dof_handler_in;
```
richardschu, line 42: note that we will change this to std::vectors of `DoFHandler`s and `std::vector<VectorType *>` in the future to enforce a correct sequence of interpolation. This will be needed for the combination with ALE already (grid velocities and coords etc))

include/exadg/convection_diffusion/user_interface/application_base.h
```diff
@@ -131,6 +131,12 @@ class ApplicationBase
     return grid;
   }
 
+  std::shared_ptr<Grid<dim>>
+  get_grid_non_const()
+  {
+    return grid;
+  }
```
nfehn, line 138: This is somewhat identical to making `grid` a public member of `ApplicationBase`. I am against such subtle changes, which  actually change the design a lot (since responsibilities become unclear).

We need to think about two options: (i) should the application execute the relevant functions changing the triangulation during AMR, or (ii) should the driver be responsible for the `Grid` object, meaning that `grid` becomes a member variable of `Driver` and is handed over to `ApplicationBase` e.g. to create the grid.

include/exadg/convection_diffusion/user_interface/parameters.cpp
```diff
@@ -256,6 +257,29 @@ Parameters::check() const
...
+  {
+    AssertThrow(not ale_formulation,
+                dealii::ExcMessage("Combination of adaptive mesh refinement "
+                                   "and ALE formulation not implemented."));
+
+    AssertThrow(temporal_discretization == TemporalDiscretization::BDF,
+                dealii::ExcMessage("Combination of adaptive mesh refinement "
+                                   "and explicit time integration not implemented."));
+
+    AssertThrow(preconditioner != Preconditioner::BlockJacobi,
```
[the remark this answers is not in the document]
nfehn: This assert and the next assert is superfluous in my opinion. The implementation of the `BlockJacobi` preconditioner/smoother (currently the responsibility of `OperatorBase`) needs to make sure that a valid type of face loops is chosen with corresponding asserts in such routines. In this sense, AMR is related to `use_cell_based_face_loops` and `BlockJacobi` to `use_cell_based_face_loops`, but exception messages relating AMR to `BlockJacobi` are in my opinion hard to understand for a user.

include/exadg/convection_diffusion/user_interface/parameters.cpp
```diff
@@ -508,6 +532,8 @@ Parameters::print_parameters_spatial_discretization(dealii::ConditionalOStream c
 
   print_parameter(pcout, "Polynomial degree", degree);
 
+  print_parameter(pcout, "Enable adaptivity", enable_adaptivity);
+
```
nfehn, line 536: I think our standard procedure woul be to print the AMR parameters `if(enable_adaptivity)`.

include/exadg/operators/adaptive_mesh_refinement.h
```diff
@@ -0,0 +1,163 @@
...
+#include <deal.II/distributed/solution_transfer.h>
+#include <deal.II/dofs/dof_tools.h>
+#include <deal.II/grid/grid_refinement.h>
+#include <deal.II/numerics/error_estimator.h>
+#include <deal.II/numerics/solution_transfer.h>
+
+
+namespace ExaDG
+{
+struct AdaptiveMeshRefinementData
```
nfehn, line 40: would need a print function, see other parameter structs in ExaDG.

include/exadg/operators/adaptive_mesh_refinement.h
```diff
@@ -0,0 +1,163 @@
...
+      {
+        cell->clear_refine_flag();
+        cell->clear_coarsen_flag();
+      }
+    }
+  }
+}
+
+/**
+ * Loop over the locally owned cells and check for refinement/coarsening flags set.
```
nfehn, line 99: Involves an MPI call, so I am not sure if the documentation of this function is correct. The comment could maybe appear inside the function.

include/exadg/operators/operator_base.cpp
```diff
@@ -1499,42 +1499,45 @@ OperatorBase<dim, Number, n_components>::cell_based_loop_diagonal(
     }
 
     // loop over all faces and gather results into local diagonal local_diag
-    unsigned int const n_faces = dealii::ReferenceCells::template get_hypercube<dim>().n_faces();
-    for(unsigned int face = 0; face < n_faces; ++face)
+    if(evaluate_face_integrals())
```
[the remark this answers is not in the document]
nfehn: I see. I think we should do this separately so that the present PR shows the relevant changes.

include/exadg/operators/solution_transfer.h
```diff
@@ -0,0 +1,147 @@
...
+    std::vector<VectorType const *> vectors_old_grid_ptr(vectors.size());
+    for(unsigned int i = 0; i < vectors.size(); ++i)
+    {
+      vectors[i]->update_ghost_values();
+
+      if(is_parallel_distributed_triangulation())
+      {
+        vectors_old_grid_ptr[i] = vectors[i];
+      }
+      else
```
nfehn, line 61: Is this branch also relevant for serial triangulations? I am not that familiar with serial, but `locally_owned_elements()`/`get_mpi_communicator()` looks somewhat strange in this context?

include/exadg/operators/solution_transfer.h
```diff
@@ -0,0 +1,147 @@
...
+  SolutionTransfer(dealii::DoFHandler<dim> const & dof_handler_in)
+  {
+    dof_handler = &dof_handler_in;
+  }
+
+  void
+  prepare_coarsening_and_refinement(std::vector<VectorType *> & vectors)
+  {
+    // Container vectors_old_grid hold vectors for interpolation *REQUIRED* for
+    // interpolate_after_coarsening_and_refinement. Thus makes an actual copy.
```
nfehn, line 49: Just as a minor comment, I would use bracket () after function names, to make intuitively clear that we are talking about a function.

include/exadg/operators/adaptive_mesh_refinement.h
```diff
@@ -0,0 +1,163 @@
...
+};
+
+/**
+ * Check if adaptive mesh refinement should be triggered depending on the time_step_number.
+ */
+inline bool
+trigger_coarsening_and_refinement_now(const AdaptiveMeshRefinementData & amr_data,
+                                      const unsigned int                 time_step_number)
+{
+  return ((time_step_number == 0) or not(time_step_number % amr_data.trigger_every_n_time_steps));
```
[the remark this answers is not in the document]
nfehn: For now: We should not pass amr_data to this function in my opinion, but only the relevant parameter. We might have different AMRData structs in the future, while this function could then no longer be used even though the functionality is actually the same.

For the future: Moreover, we should actually realize such functionality generically in ExaDG (i.e. independently of AMR), since this appears for many different problems (e.g. postprocessing, output of solver information, update of preconditioners). As you have written, let's consider this a future task.

include/exadg/convection_diffusion/driver.cpp
```diff
@@ -188,28 +187,152 @@ Driver<dim, Number>::ale_update() const
...
       {
         time_integrator->advance_one_timestep_pre_solve(true);
 
-        ale_update();
-
         time_integrator->advance_one_timestep_solve();
 
+        // Adapt the mesh before post_solve(), in order to recalculate the
+        // time step size based on the new mesh.
+        if(time_integrator->get_number_of_time_steps() > 0)
```
nfehn, line 309: From a design perspective, I do not see why this if-statement appears here and not further down in the sequence of function calls.

include/exadg/convection_diffusion/driver.cpp
```diff
@@ -188,28 +187,152 @@ Driver<dim, Number>::ale_update() const
...
+  }
+  else
+  {
+    std::shared_ptr<TimeIntBDF<dim, Number>> bdf_time_integrator =
+      std::dynamic_pointer_cast<TimeIntBDF<dim, Number>>(time_integrator);
+
+    if(trigger_coarsening_and_refinement_now(application->get_parameters().amr_data,
+                                             time_step_number))
+    {
+      mark_cells_coarsening_and_refinement(tria, bdf_time_integrator->get_solution_np());
```
nfehn, line 273: In my opinion, just before this function call we should assert that the above dynamic cast was successful. Then, the above if-else-branch for BDF time integration with corresponding assert can be removed and the check appears more locally, i.e. only for the functionality where time_integrator vs. bdf_time_integrator is relevant.

include/exadg/convection_diffusion/driver.cpp
```diff
@@ -188,28 +187,152 @@ Driver<dim, Number>::ale_update() const
...
+        Assert(false, dealii::ExcMessage("ALE and AMR combination not implemented."));
+        pde_operator->initialize_dof_vector_velocity(velocity);
+        pde_operator->interpolate_velocity(velocity, time_integrator->get_time());
+        velocity_ptr = &velocity;
+      }
+
+      auto & grid = *application->get_grid_non_const();
+      grid.coarse_triangulations =
+        dealii::MGTransferGlobalCoarseningTools::create_geometric_coarsening_sequence(
+          *application->get_grid_non_const()->triangulation);
```
nfehn, line 235: I do not like this design. The update of the coarse triangulations in case something has changed is the responsibility of the `Grid` class. So I believe we need to think more fundamentally about the design. As a rule, we should not allow to return non-const pointers to `Grid`.

Moreover, `MGTransferGlobalCoarseningTools::create_geometric_coarsening_sequence()` seems to be just one variant of our create-grid-functionality in ExaDG. Hence, this function call should in my opinion not appear on the driver level, but inside some other class/free function responsible for these grid-related functionalities.

Another design aspect is that this code should in my opinion not appear in `setup_after_coarsening_and_refinement()` but at the place where the fine triangulation is manipulated during AMR.

include/exadg/convection_diffusion/driver.cpp
```diff
@@ -188,28 +187,152 @@ Driver<dim, Number>::ale_update() const
...
+
+  VectorType const * velocity_ptr = nullptr;
+  VectorType         velocity;
+
+  if(application->get_parameters().problem_type == ProblemType::Unsteady)
+  {
+    if(application->get_parameters().temporal_discretization == TemporalDiscretization::BDF)
+    {
+      std::shared_ptr<TimeIntBDF<dim, Number>> time_integrator_bdf =
+        std::dynamic_pointer_cast<TimeIntBDF<dim, Number>>(time_integrator);
```
nfehn, line 222: This should be shifted to the place where it is needed/used.

include/exadg/convection_diffusion/driver.cpp
```diff
@@ -188,28 +187,152 @@ Driver<dim, Number>::ale_update() const
...
+{
+  pde_operator->initialize_dof_handler_and_constraints();
+  pde_operator->setup();
+
+  postprocessor->setup(*pde_operator);
+
+  VectorType const * velocity_ptr = nullptr;
+  VectorType         velocity;
+
+  if(application->get_parameters().problem_type == ProblemType::Unsteady)
```
nfehn, line 217: Is this necessary at all? We only run into such functionality during time stepping, right?

include/exadg/operators/solution_transfer.h
kronbichler: Yes, it also applies to serial triangulations. The crucial difference is type of solution transfer object used in lines 78 or 85 below, where one needs vectors to stay alive during the transfer (the non-`is_parallel_distributed_triangulation` case) or if the data is stored inside the transfer class somehow.

include/exadg/convection_diffusion/driver.cpp
kronbichler: I guess this depends on the use case: I could imagine a setup where one wants to implement a loop to compute the solution, evaluate some error estimator, and then run again, until a predefined number of refinement steps have been done or the error is below a certain threshold. It would be similar to the loop over refinements we have now, but acting on a different level (where not everything is set up again). It could be also possible to do it on another level of the code, but in general it can make sense to change the mesh also in stationary problems and keep as much of the data alive, in a similar vein as for time-dependent problems.

nfehn:
I think we already converged to quite a good basis. I see one major problem regarding the responsibility / write-access of the grid class. My suggestion is the following:

I think it is most reasonable to shift responsibility of the `Grid` object from `Application` to `Driver`, since AMR changes the grid/triangulation, something we did not think about before. This implies
- changing some of the interfaces of the application class, (in particular construction and creation of grid/mapping like `Application::setup()` becoming `Application::setup(Grid<dim> & grid)`, removing all the getter functions, etc.), with the hope that this does not make things too complicated inside the application base and derived classes,
- equipping the `Grid` class with certain functionality for AMR, adding `Grid` as a member variable to `Driver`, and calling in the driver something like `grid->amr_related_functionality()`, which internally cares about coarse triangulations for example. This would be natural since `Driver` owns the `Grid` object.

I suggest to try out this idea as a separate PR independently of the present one. Then we see more clearly the implications of such a design. We changed grid-related code more than once in the past and I might currently not be aware of all the implications of such a change. It would be enough to do this only for the `ConvDiff` module for now. Aftewards. the present PR could be adapted to such a new design.

include/exadg/convection_diffusion/driver.cpp
nfehn: Yes, we are probably more flexible when removing this if-statement here.

include/exadg/convection_diffusion/user_interface/parameters.cpp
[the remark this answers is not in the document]
richardschu: Yep; I removed the asserts.

include/exadg/convection_diffusion/driver.cpp
richardschu: I removed it here, since it is not necessary. I did not want to do AMR in the first time step for no reason whatsoever, which was purely arbitrary ... comparing the test results, we see slightly lower errors now doing one more refinement.

include/exadg/convection_diffusion/driver.cpp
richardschu: My logic here was to simply assert everything I did not test such that if we extend in the future we at least know, where certain cases were not really considered yet.
If you check out [the if-else branch for ProblemType::Steady ind Driver::setup()](https://github.com/richardschu/exadg/blob/adaptivity_restructured/include/exadg/convection_diffusion/driver.cpp#L156-L166), we actually need a different setup. But the steady case does not consider AMR yet as of this PR. Three options: i) remove the if and wait till we have to adapt again once we decide to have AMR for the steady case; ii) implement needed setup for the steady case already despite the rest of the PR does not consider `ProblemType::Steady`; iii) leave as is such that we know in the future where we have to adapt.
I vote for iii) here - am I missing something?

include/exadg/convection_diffusion/user_interface/application_base.h
richardschu: this is still open! separate PR moving `grid` needed.

include/exadg/convection_diffusion/driver.cpp
richardschu: this is still open! separate PR moving `grid` needed.

include/exadg/operators/operator_base.cpp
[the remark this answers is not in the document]
richardschu: Ok, so I will push on #582 to do the requested changes and merge later.

richardschu:
Thanks for your remarks!
Missing stuff:
.) #582 and remove changes here for smaller diff.
.) I will open a separate PR to test the approach @nfehn suggested to remove the `get_grid_non_const` and then we see how it would impact this PR.

include/exadg/operators/operator_base.cpp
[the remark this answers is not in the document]
richardschu: changes to `operator_base` were moved to #595

include/exadg/convection_diffusion/driver.cpp
richardschu: Thank you for pointing this out. We are currently missing out on the mappings that are potentially present even if we do not employ ALE.
For now, I will first in a separate PR move the `grid` (and static mapping) to the `driver`, where the related ALE functionality already resides, and then come back here, apply changes to fit that approach and then introduce a member function in Grid, that updates the coarse triangulations and the respective mappings to be called from the driver like so:

`tria.execute_coarsening_and_refinement();`

`// TODO: update the mapping OR ale_mapping, both owned by the Driver`

`grid.update_coarse_triangulations_and_mappings(mapping);`

What we still miss then is the `TODO` mentioned here, which is easy to achieve for a mapping of non-`MappingQCache` type, otherwise I would get the displacement vector described by `MappingQCache` (new member function get_mapping_dof_vector or maybe even exists) and then interpolate this vector as well.

how does that sound?

richardschu:
I will merge #596 (i.e., master after merge) and then we see where this goes :)

include/exadg/convection_diffusion/user_interface/application_base.h
richardschu: `grid` was moved to the driver in #596.
this member function was removed

applications/convection_diffusion/rotating_hill/application.h
```diff
@@ -180,6 +206,11 @@ class Application : public ApplicationBase<dim, Number>
 
         // hypercube volume is [left,right]^dim
         dealii::GridGenerator::hyper_cube(tria, left, right);
+
+        unsigned int const frequency   = 1;
+        double const       deformation = (right - left) * 0.1;
+        apply_deformed_cube_manifold(tria, left, right, deformation, frequency);
+
```
richardschu, line 213: Note tha we use `this->param.mapping_degree = this->param.degree` now.
Hence, we test the case, where the mapping is higher-order, but stationary.
AMR does not yet support `MappingDoFVector` (`MappingQCache`), which is the next step.
For this, we need a strategy to get the DoF vector from the `MappingDoFVector`, attach it to the `SolutionTransfer`, interpolate and update the `MappingQCache` after executing AMR.
This is for a future PR.

include/exadg/convection_diffusion/driver.cpp
```diff
@@ -188,28 +187,154 @@ Driver<dim, Number>::ale_update() const
...
+                                   application->get_parameters().degree +
+                                     1 /* n_face_quadrature_points */,
+                                   application->get_parameters().amr_data);
+}
+
+template<int dim, typename Number>
+void
+Driver<dim, Number>::setup_after_coarsening_and_refinement()
+{
+  // Update mapping
```
richardschu, line 209: I put some assertions here since we do not yet cover (i) ALE mappings (the hierarchy goes `DeformedMappingBase`  is a `MappingDoFVector` is a `MappingQCache`), (ii) Mappings based on `MappingQCache`. So I added a `DeformedCubeManifold` and `mapping_degree = degree` to test the remaining case of a static, but non-affine `mapping` in `rotating_hill/application.h`. Later, we will consider the interpolation of a DoFVector describing the mapping and can hence support the rest here as well!

richardschu (commented):
This next commit consider the `grid` and `mapping` in the Driver, which was moved in #596. Additionally, we add some manifold in the `rotating_hill/application.h` to test that a stationary mapping is in fact working.

richardschu:
With `0d9cf03` we introduce `GridUtilities::setup_after_coarsening_and_refinement()`, which's job is to update `coarse_triangulations` and `coarse_periodic_face_pairs` if required, that is, in the case of h-multigrid.
Also, this way I found a bug, which caused:
(i) create/update the `coarse_triangulations` even if they are not needed, and 
(ii) storing the fine triangulation twice(!) since we did not `coarse_triangulations.pop_back()` after calling `MGTransferGlobalCoarseningTools::create_geometric_coarsening_sequence`.
`FullyDistributed::Triangulation` is asserted for now, since we neither want to call `lambda_create_triangulation` again, nor have it available in the driver. As discussed with @nfehn, this is a topic to be covered in the future.

include/exadg/convection_diffusion/driver.cpp
```diff
@@ -188,28 +188,158 @@ Driver<dim, Number>::ale_update() const
...
+                "Combination of adaptive mesh refinement and ALE not implemented."));
+
+  std::shared_ptr<dealii::MappingQCache<dim>> mapping_q_cache =
+    std::dynamic_pointer_cast<dealii::MappingQCache<dim>>(mapping);
+  AssertThrow(
+    mapping_q_cache.get() == 0,
+    dealii::ExcMessage(
+      "Combination of adaptive mesh refinement and dealii::MappingQCache not implemented."));
+
+  pde_operator->initialize_dof_handler_and_constraints();
```
richardschu, line 222: this is not optimal here.

include/exadg/convection_diffusion/spatial_discretization/operator.h
```diff
@@ -89,6 +90,12 @@ class Operator : public dealii::Subscriptor, public Interface::Operator<Number>
   void
   setup_solver(double const scaling_factor_mass = -1.0, VectorType const * velocity = nullptr);
 
+  /**
+   * Initializes dealii::DoFHandlers and dealii::AffineConstraints.
+   */
+  void
+  initialize_dof_handler_and_constraints();
```
nfehn, line 97: We have to be careful with making this public. This function resets shared pointers that are used in other public functions e.g. in `fill_matrix_free_data()`. Hence, other classes might rely on what has been set in the constructor of `ConvDiff::Operator`.

nfehn: see PR #607

nfehn: I think the function should remain private and we should introduce a new public member function `setup_after_adaptive_mesh_refinement()`, which internally uses as much as possible existing member functions.

include/exadg/convection_diffusion/driver.cpp
```diff
@@ -143,28 +144,124 @@ Driver<dim, Number>::ale_update() const
...
+                "Combination of adaptive mesh refinement and ALE not implemented."));
+
+  std::shared_ptr<dealii::MappingQCache<dim>> mapping_q_cache =
+    std::dynamic_pointer_cast<dealii::MappingQCache<dim>>(mapping);
+  AssertThrow(
+    mapping_q_cache.get() == 0,
+    dealii::ExcMessage(
+      "Combination of adaptive mesh refinement and dealii::MappingQCache not implemented."));
+
+  pde_operator->setup_after_coarsening_and_refinement();
```
richardschu, line 178: this is the only new thing since the last commit :)

include/exadg/convection_diffusion/driver.cpp
```diff
@@ -143,28 +144,124 @@ Driver<dim, Number>::ale_update() const
...
+  // AMR is only implemented for implicit timestepping.
+  if(trigger_coarsening_and_refinement_now(
+       application->get_parameters().amr_data.trigger_every_n_time_steps, time_step_number))
+  {
+    std::shared_ptr<TimeIntBDF<dim, Number>> bdf_time_integrator =
+      std::dynamic_pointer_cast<TimeIntBDF<dim, Number>>(time_integrator);
+
+    AssertThrow(bdf_time_integrator.get(),
+                dealii::ExcMessage("Combination of adaptive mesh refinement "
+                                   "and explicit time integration not implemented."));
```
nfehn, line 197: Is this check needed at all? Below, we have the call `time_integrator->prepare_coarsening_and_refinement();`. Isn't this the place where one would check this?

include/exadg/convection_diffusion/driver.cpp
```diff
@@ -143,28 +144,124 @@ Driver<dim, Number>::ale_update() const
...
+
+  pde_operator->setup_after_coarsening_and_refinement();
+
+  postprocessor->setup(*pde_operator);
+}
+
+
+template<int dim, typename Number>
+void
+Driver<dim, Number>::do_adaptive_refinement(unsigned int const time_step_number)
```
nfehn, line 186: I did not fully understand this comment: https://github.com/exadg/exadg/pull/586#discussion_r1386303815

Is the idea that this function can treat both the steady and unsteady case? If so, one would move the next if statement `if(trigger_coarsening_and_refinement_now(time_step_size))` outside this function (which has the advantage that we do not need to pass the `time_step_number` to this function), and introduce another if-else somewhat below (I will make another comment ...)

include/exadg/convection_diffusion/driver.cpp
```diff
@@ -143,28 +144,124 @@ Driver<dim, Number>::ale_update() const
...
+    mark_cells_coarsening_and_refinement(*grid->triangulation,
+                                         bdf_time_integrator->get_solution_np());
+
+    limit_coarsening_and_refinement(*grid->triangulation, application->get_parameters().amr_data);
+
+    if(any_cells_flagged_for_coarsening_or_refinement(*grid->triangulation))
+    {
+      grid->triangulation->prepare_coarsening_and_refinement();
+
+      time_integrator->prepare_coarsening_and_refinement();
```
nfehn, line 208: related to this comment: https://github.com/exadg/exadg/pull/586/files#r1404332107
```C++
if(unsteady)
  time_integrator->prepare_coarsening_and_refinement();
else
  driver_steady->prepare_coarsening_and_refinement();
```

should allow us to use the same function `do_adaptive_refinement()` for both the steady and unsteady case. Of course, we can simply ignore the steady case for now, but comment https://github.com/exadg/exadg/pull/586/files#r1404332107 could already be realized.

include/exadg/convection_diffusion/driver.cpp
```diff
@@ -143,28 +144,124 @@ Driver<dim, Number>::ale_update() const
...
+  std::shared_ptr<dealii::MappingQCache<dim>> mapping_q_cache =
+    std::dynamic_pointer_cast<dealii::MappingQCache<dim>>(mapping);
+  AssertThrow(
+    mapping_q_cache.get() == 0,
+    dealii::ExcMessage(
+      "Combination of adaptive mesh refinement and dealii::MappingQCache not implemented."));
+
+  pde_operator->setup_after_coarsening_and_refinement();
+
+  postprocessor->setup(*pde_operator);
```
nfehn, line 180: I think this can be problematic. Often, postprocessors have some history (e.g. collect statistics over time).

The standard postprocessor for `ConvDiff` contains `ErrorCalculator` and `OutputGenerator`. They work with the `DoFHandler`. Do they need a re-setup() after adaptive mesh refinement? Of course, applications might implement any postprocessor with additional features as a derived class. Then, however, such a call of `postprocessor->setup()` can be dangerous. So I think we need to discuss this aspect of postprocessing with AMR more fundamentally. 

My current understanding is the following: postprocessing can be anything and we don't know this here in the `Driver`. So all we can do is to prepare a member function `Postprocessor::setup_after_coarsening_and_refinement()` in which the postprocessor either re-initializes certain postprocessing functionality, or where one could assert that a certain postprocessing functionality must be in-active because it does not support AMR in its current form. If this function is virtual like `setup()`, any special application can care about this as needed. In contrast, we probably do not want to make every postprocessing functionality capable of AMR.

include/exadg/convection_diffusion/driver.cpp
richardschu: Now, `Driver::do_adaptive_refinement` might be used for `ProblemType::Steady` and `ProblemType::Unsteady` problems. We lack the `driver_steady->prepare_coarsening_and_refinement()` and `driver_steady->interpolate_after_coarsening_after_refinement()`. I asserted the relevant lines in `Driver::do:adaptive_refinement`.

include/exadg/convection_diffusion/driver.cpp
richardschu: The point is that neither the explicit time integrator nor the corresponding base class have the `get_solution_np`, which gives a solution vector at $t^{n+1}$ to base the adaptive refinement on. That is the reason for the cast.
I did not introduce `get_solution_np()` in the base class, but I could? Once we have `time_integrator_rk->get_solution_np()`, I can remove the cast here completely. WIth this, the check for a successful cast would also be removed! ... I would rather not add explicit timestepping in this PR, ok?

include/exadg/convection_diffusion/driver.cpp
richardschu: Good catch! Shoudl we go for pure virtual to force implementation in some derived class, such that one is forced to think about what happens after AMR? Otherwise, this might generate bugs that are somewhat tricky to find ...
The ConvDiff applications use `OutputGenerator` and `ErrorCalculator`, yes (see exadg/include/exadg/convection_diffusion/postprocessor/postprocessor.h).

.) The `ErrorCalculator` (its `time_control`, so when it is triggered etc.) should not be reset by AMR. So I intriduced `ErrorCalculator<dim, Number>::setup_after_coarsening_and_refinement`, which just updates the DoFHandler and the Mapping.

.) Similar thing apply to `OutputGenerator`, where we can also skip output of boundary IDs etc., which is normally done at setup. Optionally, we can also include that output functionality in the setup.

include/exadg/convection_diffusion/driver.cpp
```diff
@@ -238,7 +237,23 @@ Driver<dim, Number>::solve()
...
+        {
+          // AMR is only implemented for implicit timestepping.
+          std::shared_ptr<TimeIntBDF<dim, Number>> bdf_time_integrator =
+            std::dynamic_pointer_cast<TimeIntBDF<dim, Number>>(time_integrator);
+
+          AssertThrow(bdf_time_integrator.get(),
+                      dealii::ExcMessage("Combination of adaptive mesh refinement "
+                                         "and explicit time integration not implemented."));
+
+          mark_cells_coarsening_and_refinement(*grid->triangulation,
```
richardschu, line 252: if we were to introduce `get_solution_np` in the time integrator base class, we could get rid of the dynamic casts. For explicit timestepping, I would have to introduce another one in the future.

include/exadg/convection_diffusion/postprocessor/postprocessor_base.h
```diff
@@ -68,6 +68,12 @@ class PostProcessorBase : public PostProcessorInterface<Number>
 
   virtual void
   setup(Operator<dim, Number> const & pde_operator) = 0;
+
+  /*
+   * In the derived classes, one might need to take some actions after coarsening and refinement.
+   */
+  virtual void
+  setup_after_coarsening_and_refinement(Operator<dim, Number> const & pde_operator) = 0;
```
richardschu, line 76: pure virtual.

include/exadg/postprocessor/error_calculation.cpp
```diff
@@ -134,6 +134,16 @@ ErrorCalculator<dim, Number>::setup(dealii::DoFHandler<dim> const &   dof_handle
     create_directories(error_data.directory, mpi_comm);
 }
 
+template<int dim, typename Number>
+void
+ErrorCalculator<dim, Number>::setup_after_coarsening_and_refinement(
```
richardschu, line 139: compare this to the `setup` just above these lines here.

include/exadg/postprocessor/output_generator_scalar.cpp
```diff
@@ -113,6 +113,16 @@ OutputGenerator<dim, Number>::setup(dealii::DoFHandler<dim> const & dof_handler_
   }
 }
 
+template<int dim, typename Number>
+void
+OutputGenerator<dim, Number>::setup_after_coarsening_and_refinement(
+  dealii::DoFHandler<dim> const & dof_handler_in,
+  dealii::Mapping<dim> const &    mapping_in)
+{
+  dof_handler = &dof_handler_in;
```
richardschu, line 122: compared to the `setup` just above these lines, I did not include the output of the boundary IDs processor ID etc, since this can generate quite some output.

richardschu (commented):
Postprocessor has also some `setup_after_coarsening_and_refinement`.

include/exadg/postprocessor/output_generator_scalar.cpp
nfehn: Why did you introduce these functions? I thought the DoFHandler and Mapping objects stay alive during adaptive mesh refinement, so why is there a need to reset the pointers?

include/exadg/postprocessor/error_calculation.cpp
nfehn: probably similar as for `OutputGenerator`

include/exadg/convection_diffusion/postprocessor/postprocessor.cpp
```diff
@@ -49,6 +49,18 @@ PostProcessor<dim, Number>::setup(Operator<dim, Number> const & pde_operator)
                          pp_data.output_data);
 }
 
+template<int dim, typename Number>
+void
+PostProcessor<dim, Number>::setup_after_coarsening_and_refinement(
+  Operator<dim, Number> const & pde_operator)
```
nfehn, line 55: Here, I would like to skip the argument `pde_operator`. This is also an object that stays valid during AMR. Currently, the postprocessor in the `ConvDiff` module does not store a pointer to this operator, but other modules like `IncNS` already do this. So I think we should do something similar here.

include/exadg/postprocessor/output_generator_scalar.cpp
richardschu: I think we can simply skip the postprocessor->setup_after_coarsening_and_refinement(); since in the present case it does not need to do anything as `dof_handler` and `mapping` are changed, but the pointers still valid. I got confused it seems ...

include/exadg/convection_diffusion/postprocessor/postprocessor_base.h
richardschu: should we even keep this then?

include/exadg/convection_diffusion/postprocessor/postprocessor.cpp
richardschu: The argument is not needed since the whole funciton is not needed?

richardschu: I just checked `IncNS::PostProcessor` and there we do in `setup(Operator const & pde_operator)`:
`navier_stokes_operator = &pde_operator;`
So I understand your comment in the sense that we should skip the `pde_operator` argument for the ConvDiff module.

include/exadg/postprocessor/output_generator_scalar.cpp
richardschu: that is true, it was completely useless.

richardschu:
@nfehn could you please check it once again? I removed the superfluous `setup_after_coarsening_and_refinement()` functions of `OutputGeneratorScalar` and `ErrorCalculator`.

applications/convection_diffusion/rotating_hill/tests/convective_implicit_time_int.output
```diff
@@ -111,175 +111,175 @@ Setup Multistep time integrator ...
...
 
-  Time step size:                            1.0040e-03
+  Time step size:                            6.8306e-04
 
 ... done!
 
 Starting time loop ...
 
 Calculate error for all fields at time t = 0.0000e+00:
-  Relative error (L2-norm): 5.63691e-03
```
nfehn, line 125: Why do we have these changes in errors? Almost an order of magnitude larger?

applications/convection_diffusion/rotating_hill/tests/convective_implicit_time_int_amr.output
```diff
@@ -0,0 +1,782 @@
...
+
+Discontinuous Galerkin finite element discretization:
+
+  degree of 1D polynomials:                  4
+  number of dofs per cell:                   25
+  number of dofs (total):                    1450
+
+Setup convection-diffusion operator ...
+
+... done!
```
nfehn, line 195: I think we can improve this output, i.e. it looks like one can currently not extract from the output that AMR is performed.

nfehn: In order to not make the output too lengthy, a one-liner per AMR step should be enough in my opinion.

include/exadg/convection_diffusion/driver.cpp
nfehn: I think this is ok for now. However, one could change the exc-message and write that "only BDF time integration is implemented for AMR", or similar. (I think the present message is only valid as long as no other than the two time stepping routines currently available for `ConvDiff` is implemented.)

include/exadg/convection_diffusion/user_interface/parameters.cpp
```diff
@@ -256,6 +257,28 @@ Parameters::check() const
...
+    AssertThrow(use_cell_based_face_loops == false,
+                dealii::ExcMessage("Cell-based face loops not supported for "
+                                   "adaptively refined meshes."));
+
+    if(involves_h_multigrid())
+    {
+      AssertThrow(grid.create_coarse_triangulations,
+                  dealii::ExcMessage("Coarse triangulations required for locally refined meshes"
+                                     "and h-multigrid."));
+    }
```
nfehn, line 279: Can we write this at a more central place? The idea behind is that we do not want to write such asserts for every module for which AMR is introduced. How about doing this check inside the `MultigridPreconditionerBase` class?

include/exadg/convection_diffusion/user_interface/parameters.h
```diff
@@ -233,6 +234,10 @@ class Parameters
   // polynomial degree of shape functions
   unsigned int degree;
 
+  // enable adaptive mesh refinement
+  bool                       enable_adaptivity;
+  AdaptiveMeshRefinementData amr_data;
```
nfehn, line 239: I don't remember exactly: Did we discuss whether these parameters should belong to `GridData` or whether they should not? One idea would be that inside `GridData` one could easily make certain checks/asserts that will then hold for all the modules, e.g. that `enable_adaptivity` is currently not possible with `Simplex`. I am not sure if e.g. this particular condition is implemented somewhere in the code currently?

nfehn: To be more precise: I think it still makes sense to keep all the parameters currently in `AdaptiveMeshRefinementData` in a separate struct related to AMR, in order to not mix these parameters with other grid parameters. One might want to define different types of AMR-data, so it could make sense to keep `amr_data` outside `grid_data`. Having `enable_adaptivity` inside `grid_data`, however, would easily allow the assert mentioned above. what do you think?

include/exadg/grid/grid_utilities.h
```diff
@@ -467,6 +467,104 @@ create_triangulation_with_multigrid(
...
+    AssertThrow(false, dealii::ExcMessage("Invalid parameter triangulation_type."));
+  }
+}
+
+/**
+ * Function to update the grid members after adaptive mesh refinement with Grid interface.
+ */
+template<int dim>
+inline void
+setup_after_coarsening_and_refinement(Grid<dim> &      grid,
```
nfehn, line 550: This function only seems to exist because of the if-statement below. I think the code would simplify overall when writing on the calling side

```C++
if(param.involves_h_multigrid()`
  create_coarse_triangulations_after_coarsening_and_refinement(...);
```
and then removing the present function.

For comparison, the function https://github.com/exadg/exadg/blob/b651712b088348f4e6684244a7040e961459f14e/include/exadg/grid/grid_utilities.h#L419 writes the if-statement inside the function, but there we can not pull it outside as we can here.

include/exadg/convection_diffusion/time_integration/time_int_bdf.cpp
```diff
@@ -591,6 +654,13 @@ TimeIntBDF<dim, Number>::set_velocities_and_times(
   times      = times_in;
 }
 
+template<int dim, typename Number>
+dealii::LinearAlgebra::distributed::Vector<Number> const &
+TimeIntBDF<dim, Number>::get_solution_np() const
+{
+  return (this->solution[0]);
```
nfehn, line 661: This class also has a member variable `solution_np`, right? Then, the implementation of this function seems to be ambiguous/undefined. Strictly speaking, `get_solution_np()` has to return `solution_np` in my opinion. Is the reason behind the place in the time stepping / AMR algorithm at which this function is called?

Generally speaking, the consistency of `time_step_number`, `time`, and of the solution vectors should be private content of the time integrator in my opinion.

My understanding of a function like `get_solution_np()` is to receive the solution at time `t_n+1` as long as the time step `n -> n + 1` has not been completed. Once a time step is completed (but a new one has not been initiated so far), the function `get_solution_np()` does not have a meaning, because we do not have a guarantee that `solution_np` exists at all. A time integrator could realize `solution_np` as a temporary vector which is deleted once a time step has been finished. This could be the reason why there is currently no `get_solution_np()` functionality in general.

applications/convection_diffusion/rotating_hill/tests/convective_implicit_time_int.output
richardschu: Because now we are using a `DeformedCubeManifold`:
https://github.com/exadg/exadg/pull/586/files/8e95faacb4023d638fca4e13f7c53971bb2573b9#diff-2b1dc07e6c57bc454d21f974e7dcb5d08b9d1e66f0c0b04207e7ad4d38c9f7edR211-R215
by default in that application. It distorts the grid quite a bit, but I didi this to show that the mapping is not lost during AMR.

applications/convection_diffusion/rotating_hill/tests/convective_implicit_time_int_amr.output
richardschu: Ok, I will do something like "Performing adaptive mesh refinement: Number of dofs (total): ...". So **that** one line basically.

richardschu: Hmmmm ... I introduced now `print_discretization_info` that differentiates these cases. In ExaDG, we frequently have some "Doing something" "... done!" output, but that was not the case here, so I pulled this out of the `initialize_dof_handler_and_constraints`.
The setup now has some `bool const verbosity = true`, which regulates the "output depth". Honestly I do not like this too much, should we just skip the "Setup conv-diff op." "...done!" for the case when AMR is enabled?
Please check 
https://github.com/exadg/exadg/pull/586/commits/18bbf4d09c111951be71f4c1d9254edf5356b466

include/exadg/convection_diffusion/user_interface/parameters.cpp
richardschu: In `MultigridPreconditionerBase`, we do not have the parameters readily available to check these conditions. But I think `GridUtilities::setup_after_coarsening_and_refinement()` might be well suited! I asserted simplex elements and the condition we checked before in parameters. This achieves the point regarding application independency, as well. Please check
https://github.com/exadg/exadg/pull/586/commits/bbfad6f3df4cdaab742ff80c8e9e584a1e08c272

include/exadg/grid/grid_utilities.h
richardschu: I totally see the point. Motivation was, that I wanted to have a call like `GridUtilities::setup_after_coarsening_and_refinement(grid, ...)` to make clear, that one needs to call this.
Now, we will have the if in all drivers, but I can see your point!
So in the current state, where this is the only thing we need to do with the grid after AMR, I totally agree.
Maybe in the future, we will shift things around again. I will address this in a common commit with the above suggested change to `GridUtilities`, since the Asserts I also mentioned before are then moved. Stay tuned :smile:

include/exadg/grid/grid_utilities.h
```diff
@@ -467,6 +467,104 @@ create_triangulation_with_multigrid(
   }
 }
 
+/**
+ * Function to update the coarse_triangulations and coarse_periodic_face_pairs given the
+ * fine_triangulation and fine_periodic_face_pairs.
```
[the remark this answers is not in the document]
richardschu: Aaaaah much better! Thanks; I addressed these comments in
https://github.com/exadg/exadg/pull/586/commits/dd277bd552bdd52f8c0d93480834ce4e88539310
The Asserts are also now in `create_coarse_triangulations_after_coarsening_and_refinement()`

include/exadg/convection_diffusion/time_integration/time_int_bdf.cpp
richardschu: this one is still open, will look at it tomorrow morning! :+1:

include/exadg/convection_diffusion/driver.cpp
richardschu: coming with the next push tomorrow morning :+1:

include/exadg/convection_diffusion/time_integration/time_int_bdf.cpp
richardschu: I agree on the interpretation of function `time_integrator->get_solution_np()`.
I think the naming I introduced in this PR is very misleading, since there is a member variable `solution_np`, but we actually return the most recent solution vector.

The `solution_np` vector contains the extrapolation at some point in the time stepper, but I wanted to have a solution at the current or even previous timestep. The sole purpose was to have some (close to?) consistent and recent enough vector for AMR.

In AMR, we call what is now named `time_integrator->get_solution_np()` after `time_integrator->advance_one_timestep_solve()`, but before updating the vectors in `time_integrator->advance_one_timestep_post_solve()`. So, the vectors are actually not the same, with `solution[0]` containing the old time step's solution still, but `solution_np` having the most recent solution (which has just been computed at this point).

Long story short: I will just return `solution_np` as the name suggests, ok?
 
Interestingly, the error increases now a tiny bit for the tests, but the mesh is also different.
So I think its just in general better to use the most recent vector `solution_np`.

richardschu: see
https://github.com/exadg/exadg/pull/586/commits/62f80c3b6cda67b715b8b3099a3bc18532f6b2b4

applications/convection_diffusion/rotating_hill/tests/convective_implicit_time_int_amr.output
nfehn: I agree that we do not want to have "Setup conv-diff op." "...done!" during timestepping with AMR.

applications/convection_diffusion/rotating_hill/tests/convective_implicit_time_int.output
nfehn: I think we should not change the current test case. Another test case specifically for AMR is introduced here anyway, right?

include/exadg/convection_diffusion/user_interface/parameters.h
nfehn: I think I mean something different here. For every new module, we would have to add this boolean parameter, while we would not have to add it in case it would be part of `GridData`. 

The check between `enable_adaptivity` and `simplex` inside `GridData` is more *encapsulated* than in some free function (which one does not have to use; users/developers might write different functions potentially and would then have to think of such asserts again).

Maybe let's discuss this in person.

include/exadg/convection_diffusion/user_interface/parameters.cpp
```diff
@@ -256,6 +257,29 @@ Parameters::check() const
   // SPATIAL DISCRETIZATION
   grid.check();
 
+  if(enable_adaptivity)
+  {
+    AssertThrow(not ale_formulation,
+                dealii::ExcMessage("Combination of adaptive mesh refinement "
+                                   "and ALE formulation not implemented."));
+
+    AssertThrow(temporal_discretization == TemporalDiscretization::BDF,
```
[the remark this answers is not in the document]
nfehn: somewhere else we discussed the exception message I think.

include/exadg/convection_diffusion/user_interface/parameters.cpp
```diff
@@ -256,6 +257,21 @@ Parameters::check() const
...
+                dealii::ExcMessage("Combination of adaptive mesh refinement "
+                                   "and explicit time integration not implemented."));
+
+    AssertThrow(use_cell_based_face_loops == false,
+                dealii::ExcMessage("Cell-based face loops not supported for "
+                                   "adaptively refined meshes."));
```
nfehn, line 272: I could imagine we can realize this assert more generically (using `triangulation.has_hanging_nodes()` rather than `enable_adaptivity`), so that we do not have to add such an assert for every module. If I remember correctly, we introduced this recently. See https://github.com/exadg/exadg/blob/6fd3bee0a4db8455e42ebc0e067ea6a2ad8cc66c/include/exadg/matrix_free/categorization.h#L41

include/exadg/convection_diffusion/spatial_discretization/operator.cpp
```diff
@@ -323,6 +345,8 @@ Operator<dim, Number>::setup()
 
   if(param.use_cell_based_face_loops)
     Categorization::do_cell_based_loops(*grid->triangulation, mf_data->data);
+
+  mf->clear();
```
nfehn, line 349: probably not needed because default constructor is called a few lines above?

include/exadg/convection_diffusion/spatial_discretization/operator.cpp
```diff
@@ -88,14 +90,29 @@ Operator<dim, Number>::initialize_dof_handler_and_constraints()
...
-  print_parameter(pcout, "number of dofs (total)", dof_handler.n_dofs());
+  if(verbosity)
+  {
+    print_parameter(pcout, "degree of 1D polynomials", param.degree);
+    print_parameter(pcout, "number of dofs per cell", fe->n_dofs_per_cell());
+    print_parameter(pcout, "number of dofs (total)", dof_handler.n_dofs());
+  }
+  else
+  {
+    print_parameter(pcout, "number of dofs (total)", dof_handler.n_dofs());
```
nfehn, line 114: can't we just write this line in `setup_after_coarsening_and_refinement()` and then skip all the parameters (enable_adaptivity, verbosity) here? I think this would be more simple overall.

include/exadg/convection_diffusion/spatial_discretization/operator.cpp
```diff
@@ -311,8 +328,13 @@ Operator<dim, Number>::setup_operators()
 
 template<int dim, typename Number>
 void
-Operator<dim, Number>::setup()
+Operator<dim, Number>::setup(bool const verbosity)
 {
+  if(verbosity)
+  {
+    pcout << std::endl << "Setup convection-diffusion operator ..." << std::endl;
+  }
```
nfehn, line 336: I would like to avoid introducing boolean parameters for these public member functions.

Let me make another suggestion that could be more elegant in terms of the interfaces of this class:

```C++
public:
void setup()
{
  // write output

  // call private member function
  do_setup();

 // write output
}

private:
void do_setup(){...}
```

And then call `do_setup()` in `setup_after_coarsening_and_refinement()`. This should allow to avoid the parameters and if-statements.

include/exadg/convection_diffusion/spatial_discretization/operator.cpp
richardschu: yes, removed it!

include/exadg/convection_diffusion/user_interface/parameters.cpp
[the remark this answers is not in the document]
richardschu: Reformulated it :)

include/exadg/convection_diffusion/spatial_discretization/operator.cpp
richardschu: restructured it!

include/exadg/convection_diffusion/user_interface/parameters.cpp
nfehn: Let's discuss this also in person and see which options we have.

include/exadg/convection_diffusion/user_interface/parameters.cpp
nfehn: with PR #617, the respective assert in `do_cell_based_loops()` will be triggered as intended, so that we can then remove this assert on the level of the ConvDiff module.

PR #587, opened 2023-10-28
Add acoustic operator

jh66637:
@nfehn This PR adds a minimal version of an operator for the acoustic conservation equations. Its discretization in the strong from reads:
discre
The operator uses Lax--Friedrichs fluxes
flux
Since the fluxes use both primal variables, the operator is merged (not split into separate gradient and divergence operators) to keep the amount of flops low. 

In addition to the operator, the `BoundaryDescriptor` is needed, as well as relevant `enum` types and a function that prescribes the boundary conditions. 

The way how to prescribe the boundary conditions is different compared to the standard way in `ExaDG`. This design choice was made for non-matching interfaces (that are not present in this PR, but will be introduced in another PR). In my opinion this choice simplifies the operator also for BC application. This approach could be generalized for the rest of `ExaDG` in the future, in any way I would like to keep it here to be able to introduce the non-matching interfaces easily.

The operator is comparably template rich. If compile times become a problem in the future, it would be possible to shift everything into the `.cpp` file and add a class in the `.h` which only provides the interface. 

---------------------------
Once this PR is merged, a minimal acoustic solver can be implemented in a single PR since the minimal solver has no peculiarities.

### DIFF
--- a/include/exadg/acoustic_conservation_equations/spatial_discretization/operators/operator.h
+++ b/include/exadg/acoustic_conservation_equations/spatial_discretization/operators/operator.h
@@ -1,0 +1,493 @@
+/*  ______________________________________________________________________
+ *
+ *  ExaDG - High-Order Discontinuous Galerkin for the Exa-Scale
+ *
+ *  Copyright (C) 2023 by the ExaDG authors
+ *
+ *  This program is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *  ______________________________________________________________________
+ */
+
+#ifndef EXADG_ACOUSTIC_CONSERVATION_EQUATIONS_SPATIAL_DISCRETIZATION_OPERATORS_OPERATOR_H_
+#define EXADG_ACOUSTIC_CONSERVATION_EQUATIONS_SPATIAL_DISCRETIZATION_OPERATORS_OPERATOR_H_
+
+#include <deal.II/lac/la_parallel_block_vector.h>
+
+#include <exadg/acoustic_conservation_equations/spatial_discretization/operators/weak_boundary_conditions.h>
+#include <exadg/acoustic_conservation_equations/user_interface/boundary_descriptor.h>
+#include <exadg/acoustic_conservation_equations/user_interface/enum_types.h>
+#include <exadg/matrix_free/integrators.h>
+#include <exadg/operators/integrator_flags.h>
+
+namespace ExaDG
+{
+namespace Acoustics
+{
+namespace Operators
+{
+template<int dim, typename Number>
+class Kernel
+{
+  using scalar = dealii::VectorizedArray<Number>;
+  using vector = dealii::Tensor<1, dim, dealii::VectorizedArray<Number>>;
+
+  using CellIntegratorP = CellIntegrator<dim, 1, Number>;
+  using CellIntegratorU = CellIntegrator<dim, dim, Number>;
+
+public:
+  /*
+   * Volume flux for the momentum equation, i.e., the term occurring in the volume integral for
+   * weak formulation (performing integration-by-parts)
+   */
+  inline DEAL_II_ALWAYS_INLINE //
+    vector
+    get_volume_flux_weak_momentum(CellIntegratorU const & velocity, unsigned int const q) const
+  {
+    // minus sign due to integration by parts
+    return -velocity.get_value(q);
+  }
+
+  /*
+   * Volume flux for the momentum equation, i.e., the term occurring in the volume integral for
+   * strong formulation (integration-by-parts performed twice)
+   */
+  inline DEAL_II_ALWAYS_INLINE //
+    scalar
+    get_volume_flux_strong_momentum(CellIntegratorU const & velocity, unsigned int const q) const
+  {
+    return velocity.get_divergence(q);
+  }
+
+
+  /*
+   * Volume flux for the mass equation, i.e., the term occurring in the volume integral for
+   * weak formulation (performing integration-by-parts)
+   */
+  inline DEAL_II_ALWAYS_INLINE //
+    scalar
+    get_volume_flux_weak_mass(CellIntegratorP const & pressure, unsigned int const q) const
+  {
+    // minus sign due to integration by parts
+    return -pressure.get_value(q);
+  }
+
+  /*
+   * Volume flux for the mass equation, i.e., the term occurring in the volume integral for
+   * strong formulation (integration-by-parts performed twice)
+   */
+  inline DEAL_II_ALWAYS_INLINE //
+    vector
+    get_volume_flux_strong_mass(CellIntegratorP const & pressure, unsigned int const q) const
+  {
+    return pressure.get_gradient(q);
+  }
+
+  /*
+   * Lax Friedrichs flux for the momentum equation.
+   */
+  inline DEAL_II_ALWAYS_INLINE //
+    vector
+    calculate_lax_friedrichs_flux_momentum(vector const & um,
+                                           vector const & up,
+                                           Number const & gamma,
+                                           scalar const & pm,
+                                           scalar const & pp,
+                                           vector const & n) const
+  {
+    return Number{0.5} * (um + up) + gamma * (pm - pp) * n;
+  }
+
+  /*
+   * Lax Friedrichs flux for the mass equation.
+   */
+  inline DEAL_II_ALWAYS_INLINE //
+    scalar
+    calculate_lax_friedrichs_flux_mass(scalar const & pm,
+                                       scalar const & pp,
+                                       Number const & tau,
+                                       vector const & um,
+                                       vector const & up,
+                                       vector const & n) const
+  {
+    return Number{0.5} * (pm + pp) + tau * (um - up) * n;
+  }
+};
+
+} // namespace Operators
+
+template<int dim>
+struct OperatorData
+{
+  OperatorData()
+    : dof_index_pressure(0),
+      dof_index_velocity(1),
+      quad_index(0),
+      formulation(Formulation::SkewSymmetric),
+      bc(nullptr),
+      density(-1.0),
+      speed_of_sound(-1.0)
+  {
+  }
+
+  unsigned int dof_index_pressure;
+  unsigned int dof_index_velocity;
+
+  unsigned int quad_index;
+
+  Formulation formulation;
+
+  std::shared_ptr<BoundaryDescriptor<dim> const> bc;
+
+  double density;
+  double speed_of_sound;
+};
+
+template<int dim, typename Number>
+class Operator
+{
+  using This = Operator<dim, Number>;
+
+  using BlockVectorType = dealii::LinearAlgebra::distributed::BlockVector<Number>;
+
+  using scalar = dealii::VectorizedArray<Number>;
+  using vector = dealii::Tensor<1, dim, dealii::VectorizedArray<Number>>;
+
+  using Range = std::pair<unsigned int, unsigned int>;
+
+  using CellIntegratorU = CellIntegrator<dim, dim, Number>;
+  using CellIntegratorP = CellIntegrator<dim, 1, Number>;
+
+  using FaceIntegratorU = FaceIntegrator<dim, dim, Number>;
+  using FaceIntegratorP = FaceIntegrator<dim, 1, Number>;
+
+public:
+  Operator()
+    : evaluation_time(Number{0.0}),
+      rhocc(Number{0.0}),
+      rho_inv(Number{0.0}),
+      tau(Number{0.0}),
+      gamma(Number{0.0})
+  {
+  }
+
+  void
+  initialize(dealii::MatrixFree<dim, Number> const & matrix_free_in,
+             OperatorData<dim> const &               data_in)
+  {
+    // Set Matrixfree and OperatorData.
+    matrix_free = &matrix_free_in;
+    data        = data_in;
+
+    // Precompute numbers that are needed in kernels.
+    rhocc   = (Number)(data_in.density * data_in.speed_of_sound * data_in.speed_of_sound);
+    rho_inv = (Number)(1.0 / data_in.density);
+    tau     = (Number)(0.5 * data_in.speed_of_sound * data_in.density);
+    gamma   = (Number)(0.5 / (data_in.speed_of_sound * data_in.density));
+
+    // Set integration flags.
+    if(data_in.formulation == Formulation::Weak)
+    {
+      integrator_flags_p.cell_evaluate  = dealii::EvaluationFlags::values;
+      integrator_flags_p.cell_integrate = dealii::EvaluationFlags::gradients;
+      integrator_flags_p.face_evaluate  = dealii::EvaluationFlags::values;
+      integrator_flags_p.face_integrate = dealii::EvaluationFlags::values;
+
+      integrator_flags_u.cell_evaluate  = dealii::EvaluationFlags::values;
+      integrator_flags_u.cell_integrate = dealii::EvaluationFlags::gradients;
+      integrator_flags_u.face_evaluate  = dealii::EvaluationFlags::values;
+      integrator_flags_u.face_integrate = dealii::EvaluationFlags::values;
+    }
+    else if(data_in.formulation == Formulation::Strong)
+    {
+      integrator_flags_p.cell_evaluate  = dealii::EvaluationFlags::gradients;
+      integrator_flags_p.cell_integrate = dealii::EvaluationFlags::values;
+      integrator_flags_p.face_evaluate  = dealii::EvaluationFlags::values;
+      integrator_flags_p.face_integrate = dealii::EvaluationFlags::values;
+
+      integrator_flags_u.cell_evaluate  = dealii::EvaluationFlags::gradients;
+      integrator_flags_u.cell_integrate = dealii::EvaluationFlags::values;
+      integrator_flags_u.face_evaluate  = dealii::EvaluationFlags::values;
+      integrator_flags_u.face_integrate = dealii::EvaluationFlags::values;
+    }
+    else if(data_in.formulation == Formulation::SkewSymmetric)
+    {
+      integrator_flags_p.cell_evaluate  = dealii::EvaluationFlags::gradients;
+      integrator_flags_p.cell_integrate = dealii::EvaluationFlags::gradients;
+      integrator_flags_p.face_evaluate  = dealii::EvaluationFlags::values;
+      integrator_flags_p.face_integrate = dealii::EvaluationFlags::values;
+
+      integrator_flags_u.cell_evaluate  = dealii::EvaluationFlags::values;
+      integrator_flags_u.cell_integrate = dealii::EvaluationFlags::values;
+      integrator_flags_u.face_evaluate  = dealii::EvaluationFlags::values;
+      integrator_flags_u.face_integrate = dealii::EvaluationFlags::values;
+    }
+    else
+    {
+      AssertThrow(false, dealii::ExcMessage("Not implemented."));
+    }
+  }
+
+  void
+  evaluate(BlockVectorType & dst, BlockVectorType const & src, double const time) const
+  {
+    do_evaluate(dst, src, true, time);
+  }
+
+  void
+  evaluate_add(BlockVectorType & dst, BlockVectorType const & src, double const time) const
+  {
+    do_evaluate(dst, src, false, time);
+  }
+
+private:
+  void
+  do_evaluate(BlockVectorType &       dst,
+              BlockVectorType const & src,
+              double const            time,
+              bool const              zero_dst_vector) const
+  {
+    evaluation_time = (Number)time;
+
+    matrix_free->loop(&This::cell_loop,
+                      &This::face_loop,
+                      &This::boundary_face_loop,
+                      this,
+                      dst,
+                      src,
+                      zero_dst_vector,
+                      dealii::MatrixFree<dim, Number>::DataAccessOnFaces::values,
+                      dealii::MatrixFree<dim, Number>::DataAccessOnFaces::values);
+  }
+
+  void
+  do_cell_integral(CellIntegratorP & pressure, CellIntegratorU & velocity) const
+  {
+    if(data.formulation == Formulation::Weak)
+    {
+      for(unsigned int q : pressure.quadrature_point_indices())
+      {
+        vector const flux_momentum = kernel.get_volume_flux_weak_momentum(velocity, q);
+        scalar const flux_mass     = kernel.get_volume_flux_weak_mass(pressure, q);
+
+        pressure.submit_gradient(rhocc * flux_momentum, q);
+        velocity.submit_divergence(rho_inv * flux_mass, q);
+      }
+    }
+    else if(data.formulation == Formulation::Strong)
+    {
+      for(unsigned int q : pressure.quadrature_point_indices())
+      {
+        scalar const flux_momentum = kernel.get_volume_flux_strong_momentum(velocity, q);
+        vector const flux_mass     = kernel.get_volume_flux_strong_mass(pressure, q);
+
+        pressure.submit_value(rhocc * flux_momentum, q);
+        velocity.submit_value(rho_inv * flux_mass, q);
+      }
+    }
+    else if(data.formulation == Formulation::SkewSymmetric)
+    {
+      for(unsigned int q : pressure.quadrature_point_indices())
+      {
+        vector const flux_momentum = kernel.get_volume_flux_weak_momentum(velocity, q);
+        vector const flux_mass     = kernel.get_volume_flux_strong_mass(pressure, q);
+
+        pressure.submit_gradient(rhocc * flux_momentum, q);
+        velocity.submit_value(rho_inv * flux_mass, q);
+      }
+    }
+    else
+    {
+      AssertThrow(false, dealii::ExcMessage("Not implemented."));
+    }
+  }
+
+
+  template<bool weight_neighbor, // = false for cell centric loops and boundary loops
+           typename ExteriorFaceIntegratorP,
+           typename ExteriorFaceIntegratorU>
+  void
+  do_face_integral(FaceIntegratorP &         pressure_m,
+                   ExteriorFaceIntegratorP & pressure_p,
+                   FaceIntegratorU &         velocity_m,
+                   ExteriorFaceIntegratorU & velocity_p) const
+  {
+    for(unsigned int q : pressure_m.quadrature_point_indices())
+    {
+      scalar const pm = pressure_m.get_value(q);
+      scalar const pp = pressure_p.get_value(q);
+      vector const um = velocity_m.get_value(q);
+      vector const up = velocity_p.get_value(q);
+      vector const n  = pressure_m.normal_vector(q);
+
+      vector const flux_momentum =
+        kernel.calculate_lax_friedrichs_flux_momentum(um, up, gamma, pm, pp, n);
+
+      scalar const flux_mass = kernel.calculate_lax_friedrichs_flux_mass(pm, pp, tau, um, up, n);
+
+      if(data.formulation == Formulation::Weak)
+      {
+        scalar const flux_momentum_weak = rhocc * flux_momentum * n;
+        vector const flux_mass_weak     = rho_inv * flux_mass * n;
+
+        pressure_m.submit_value(flux_momentum_weak, q);
+        velocity_m.submit_value(flux_mass_weak, q);
+
+        if constexpr(weight_neighbor)
+        {
+          // minus signs since n⁺ = - n⁻
+          pressure_p.submit_value(-flux_momentum_weak, q);
+          velocity_p.submit_value(-flux_mass_weak, q);
+        }
+      }
+      else if(data.formulation == Formulation::Strong)
+      {
+        pressure_m.submit_value(rhocc * (flux_momentum - um) * n, q);
+        velocity_m.submit_value(rho_inv * (flux_mass - pm) * n, q);
+
+        if constexpr(weight_neighbor)
+        {
+          // minus signs since n⁺ = - n⁻
+          pressure_p.submit_value(-rhocc * (flux_momentum - up) * n, q);
+          velocity_p.submit_value(-rho_inv * (flux_mass - pp) * n, q);
+        }
+      }
+      else if(data.formulation == Formulation::SkewSymmetric)
+      {
+        scalar const flux_momentum_weak = rhocc * flux_momentum * n;
+
+        pressure_m.submit_value(flux_momentum_weak, q);
+        velocity_m.submit_value(rho_inv * (flux_mass - pm) * n, q);
+
+        if constexpr(weight_neighbor)
+        {
+          // minus signs since n⁺ = - n⁻
+          pressure_p.submit_value(-flux_momentum_weak, q);
+          velocity_p.submit_value(-rho_inv * (flux_mass - pp) * n, q);
+        }
+      }
+      else
+      {
+        AssertThrow(false, dealii::ExcMessage("Not implemented."));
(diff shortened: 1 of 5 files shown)


include/exadg/acoustic_conservation_equations/spatial_discretization/operators/grad_div_operator.h
```diff
@@ -0,0 +1,479 @@
...
+      if constexpr(formulation == FormulationGradientDivergenceTerms::Weak)
+      {
+        pressure.submit_gradient(rhocc * kernel.get_volume_flux_weak_momentum(velocity, q), q);
+        velocity.submit_divergence(rho_inv * kernel.get_volume_flux_weak_mass(pressure, q), q);
+      }
+      else if constexpr(formulation == FormulationGradientDivergenceTerms::Strong)
+      {
+        pressure.submit_value(rhocc * kernel.get_volume_flux_strong_momentum(velocity, q), q);
+        velocity.submit_value(rho_inv * kernel.get_volume_flux_strong_mass(pressure, q), q);
+      }
```
kronbichler, line 285: If you care about computational efficiency, I would suggest a third alternative, namely to use the weak form `-(grad q, rho u)` and `(w, grad p)` with the respective face terms, see also https://github.com/kronbichler/exwave/blob/97000e8234f41c6cdd7ccd46ffb37979dd922644/source/wave_equation_operations.cc#L447-L451 where we did this based on a large number of efficiency experiments. The important ingredient is to use the gradient only on the scalar variable. This can be explained by the underlying sum factorization algorithms: If we are in 3D and compute the gradient of a scalar variable, we need 6 sum factorization sweeps (meaning 1D interpolation kernels going through all points), 3 for transforming each spatial direction to quadrature points and 3 for the components of the gradient in collocation space. The divergence of a vector field needs 18, 6 for each component. One could suspect that the evaluation routine could know whether we only need the divergence and the mesh could possibly be Cartesian, where in fact 9 sweeps would suffice, but we don't know that. The value of a vector-valued variable needs 9 sum factorization sweeps. The total cost is then `(6 + 9) * 2 = 30` sweeps for both the pressure and velocity terms. In your formulation you need `(3 + 18) + (6 + 9) = 36` sweeps, where the two terms are the first and second line of the equation in the PR description, respectively.

As a second comment, I note that your code is dangerous in the sense that `pressure.submit_gradient()` gets called before the `pressure` field is handed into the volume mass flux. Since the `get_xxx()` and `submit_xxx` refer to the same underlying storage, it is very easy to make the mistake we the `get_volume_flux_weak_mass` would also request the `pressure.get_gradient()` and then produce wrong results. Your code is fine because you request the value of the pressure (after you submit the gradient), which are in fact different places in memory. I think brittle behavior of `CellIntegrator` is something we need to think about in a future design update. (It is very important that those refer to the same data field when seen for all quadrature points because otherwise the caches would be overwhelmed, but while sitting on an individual point another interface would be better.) If you would switch to the code suggested by me, you would have to take this into account, because then we would always step on another function at some place.

As a final remark, the form suggested by me should also be the most robust one, at least according to my understanding of the wave equation. It leads to a skew-symmetric formulation, so even if your numerical quadrature is not exact (non-affine cells), the energy will be non-increasing. In either the `Weak` or `Strong` form derived by you, this cannot be mathematically guaranteed, even though it will often be OK.

jh66637: Thank you for the detailed review and the helpful suggestions. 

To avoid any potential future problems with the same underlying memory that is used by `evaluate()` and `submit()` I cache the fluxes using `get_value()` before submitting anything to the integrators. [7ed38c1](https://github.com/exadg/exadg/pull/587/commits/7ed38c127cf514012d8638793b320a4fd8403453)

I added the proposed formulation in [cff4e25](https://github.com/exadg/exadg/pull/587/commits/cff4e25dfca35317d155859f278d1658ad79167b) and made it the default one due to all the arguments you made.

I agree that the skew-symmetric formulation is the best choice. Quite some time ago I tested all formulations (including the skew-symmetric one) and could not see any differences in the results in practice. I didn't think about the performance implications back then, and obviously it is better to have the skew-symmetric formulation either way, since all the mentioned advantages come without drawbacks.

nfehn: Yes, it is good to have all options available. Since part of our work in ExaDG is developing discretization methods, it is important to be able to switch between different variants at any time by just changing some input parameters.

kronbichler:
I like this.

Let me add a short comment on the fluxes, nothing to be addressed right now but maybe of interest for later investigations: The local Lax-Friedrichs versions suggested here are somewhat dissipative. The experience we had in the past with Exwave was that using a Roe-type flux, i.e., a flux that only adds dissipation on the component normal to the interface in accordance with the eigenvector of the two non-zero eigenvalues of the wave equation, would give more accurate results. Instead of adding `normal * jump(u_h)`, it would use `jump(normal * u_h)` in the flux `normal p_h^*`.

nfehn:
At this point, I think we should discuss the problem "template rich". If I imagine further variants (e.g. Roe-type flux suggested above by @kronbichler), I have the concern that adding more and more template parameters (that take the role of just enums/booleans) increase compilation times more and more. 

I assume you did this for performance reasons? If so, however, the question would be whether this has any visible effect on performance at all? The acoustic operators seem to be quite simple in terms of calculations, aren't they? so intuitively I could imagine the cost of computations or an if-statement could be hidden behind the memory transfer?

include/exadg/acoustic_conservation_equations/spatial_discretization/operators/grad_div_operator.h
```diff
@@ -0,0 +1,537 @@
...
+                     Range const &                           face_range) const
+  {
+    // precompute numbers that are needed in kernels.
+    Number const rhocc   = (Number)(data.density * data.speed_of_sound * data.speed_of_sound);
+    Number const rho_inv = (Number)(1.0 / data.density);
+    Number const tau     = (Number)(0.5 * data.speed_of_sound * data.density);
+    Number const gamma   = (Number)(0.5 / (data.speed_of_sound * data.density));
+
+    FaceIntegratorP pressure_m(matrix_free_in, true, data.dof_index_pressure, data.quad_index);
+    BoundaryFaceIntegratorP<dim, Number> pressure_p(pressure_m, *data.bc->pressure);
```
nfehn, line 498: This seems to be the only place where the new type is used? From a general software development perspective, it is then however not clear to me why the new class `BoundaryFaceIntegratorP` is introduced at all. The interface of `do_face_integral()` could also be designed as passing the result of `ExpertiorFaceIntegratorP::get_value()` directly to `do_face_integral()`. Such a design would be clearer since it does not introduce implicit dependencies between template parameters.

include/exadg/acoustic_conservation_equations/spatial_discretization/operators/grad_div_operator.h
```diff
@@ -0,0 +1,537 @@
...
+        scalar const flux_momentum_weak = rhocc * flux_momentum * n;
+        vector const flux_mass_weak     = rho_inv * flux_mass * n;
+
+        pressure_m.submit_value(flux_momentum_weak, q);
+        velocity_m.submit_value(flux_mass_weak, q);
+
+        if constexpr(consider_exterior_cell)
+        {
+          // minus sign since n⁺ = - n⁻
+          pressure_p.submit_value(-flux_momentum_weak, q);
```
nfehn, line 374: From this line, I expected to find a function definition of `BoundaryFaceIntegratorP::submit_value()`, but I did not find it. 

For this reason, I actually do not like the present design very much. It seems to be too implicit in my opinion.

include/exadg/acoustic_conservation_equations/spatial_discretization/operators/grad_div_operator.h
```diff
@@ -0,0 +1,537 @@
...
+        vector const flux_mass     = kernel.get_volume_flux_strong_mass(pressure, q);
+
+        pressure.submit_gradient(rhocc * flux_momentum, q);
+        velocity.submit_value(rho_inv * flux_mass, q);
+      }
+      else
+      {
+        AssertThrow(false, dealii::ExcMessage("Not implemented."));
+      }
+    }
```
nfehn, line 333: Here, it would be very easy (i.e. with almost no code duplication) to write the for-loop inside the if-else-branches. Then, one would probably make the current motivation for `constexpr` and thte use of a template argument (for performance reasons) superfluous.

include/exadg/acoustic_conservation_equations/spatial_discretization/operators/grad_div_operator.h
```diff
@@ -0,0 +1,537 @@
...
+    {
+      do_evaluate_formulation<FormulationGradientDivergenceTerms::SkewSymmetric>(dst,
+                                                                                 src,
+                                                                                 time,
+                                                                                 zero_dst_vector);
+    }
+    else
+    {
+      AssertThrow(false, dealii::ExcMessage("Not implemented."));
+    }
```
nfehn, line 273: There might be more parameters than the weak vs. strong vs. skew-symmetric formulation that need to be evaluated in the innermost (q-point) loop. One example would be the type of flux (Lax-Friedrichs, Roe, central, ...). Since the number of branches in functions like this one would grow larger and larger, I am actually against the use of templates for such functionality.

include/exadg/acoustic_conservation_equations/spatial_discretization/operators/grad_div_operator.h
nfehn: As discussed in person, we keep this for now and see how other parts of ExaDG could be realized this way in the future.

include/exadg/acoustic_conservation_equations/spatial_discretization/operators/grad_div_operator.h
nfehn: see https://github.com/exadg/exadg/pull/587/files#r1377402916

include/exadg/acoustic_conservation_equations/user_interface/enum_types.h
```diff
@@ -0,0 +1,48 @@
...
+enum class FormulationGradientDivergenceTerms
+{
+  Undefined,
+  /** Weak form of pressure gradient and velocity divergence terms. */
+  Weak,
+  /** Strong form of pressure gradient and velocity divergence terms. */
+  Strong,
+  /** Strong form of the pressure gradient term and weak form of the velocity divergence term.
+   *  This way, the gradient is only used on scalar variables which reduces the number of sum
+   *  factorization sweeps in the cell loop.
```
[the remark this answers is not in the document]
jh66637: https://github.com/exadg/exadg/pull/587/commits/c6ce5387369f1b57c1cfd19082f34cea4505b88f

include/exadg/acoustic_conservation_equations/spatial_discretization/operators/grad_div_operator.h
```diff
@@ -0,0 +1,537 @@
...
+      {
+        AssertThrow(false, dealii::ExcMessage("Not implemented."));
+      }
+    }
+  }
+
+  template<FormulationGradientDivergenceTerms formulation,
+           bool                               consider_exterior_cell,
+           typename ExteriorFaceIntegratorP,
+           typename ExteriorFaceIntegratorU>
```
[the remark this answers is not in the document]
jh66637: The reason is to be able to write a single kernel for the boundary loop as well as for standard face loops and cell centric face loops. So the template parameters do not necessarily depend on each other if one thinks about cell centric loops. I adjusted the name of `consider_exterior_cell -> weight_neighbor` to make this more clear.

include/exadg/acoustic_conservation_equations/spatial_discretization/operators/grad_div_operator.h
jh66637: I agree, but it doesn't work for the face loop like this :/

include/exadg/acoustic_conservation_equations/spatial_discretization/operators/grad_div_operator.h
jh66637: I think the length of the function should not be the reason to decide for or against the template parameter (this could even be generalized with `magic_enum::for_each` from the latest magic enum version as far as I understand).

I cannot reliably measure (I have only access to my laptop atm) if it has significant impact on the performance to remove the template parameter and use a standard `if` instead. I would at least avoid `if` in the innermost loop. In the case of the face loop, moving the `if` out of the loop over quadrature points leads to code duplication. On the other hand removing the template from the function would be better in terms of consistency, i.e. each existing operator could be more easily adjusted to the current design (also the ones derived from `OperatorBase`). 
For me this is a trade-off and since I am not able to do the measurements I will adjust to your wishes.

@nfehn You would prefer to remove the template from the function and have some code duplication, correct?

nfehn: What do you mean with "code duplication"? Currently, we have the same function call three times, apart from the template, which one could see as some form of code duplication. I think this could be removed by not using templates for parameters like `Formulation`.

nfehn: In essence, and since we do not have indications from performance measurements that these kind of if-statements are performance relevant, I would use templates only if necessary in the sense of generic programming and where inheritence / dynamic polymorphism is not an option (e.g. because `dealii::FEEValuation` would then have to derive from a base / interface class like `ExADG::ExteriorFaceIntegratorBase`, which is not possible). This holds for the `ExteriorFaceIntegrator` data types of this PR. This reasoning does not hold for the `Formulation` parameters, so I would not use templates for these parameters.

This is not meant as an absolute statement. It is meant as a pragmatic decision (as a compromise between performance  and code complexity) in the absence of performance numbers giving indications of potential bottlenecks.

jh66637: @nfehn What I meant with code duplication if we shift the if statement out of the innermost loop is the following:
```      
if (formulation == Formulation::Weak)
    for(unsigned int q : pressure_m.quadrature_point_indices())
    {
      scalar const pm = pressure_m.get_value(q);//duplicated
      scalar const pp = pressure_p.get_value(q);//duplicated
      vector const um = velocity_m.get_value(q);//duplicated
      vector const up = velocity_p.get_value(q);//duplicated
      vector const n  = pressure_m.normal_vector(q);//duplicated

      vector const flux_momentum =
        kernel.calculate_lax_friedrichs_flux_momentum(um, up, gamma, pm, pp, n);//duplicated

      scalar const flux_mass = kernel.calculate_lax_friedrichs_flux_mass(pm, pp, tau, um, up, n);//duplicated
      
      // formuation specific stuff
     }
else if (formulation == ...)
    for(unsigned int q : pressure_m.quadrature_point_indices())
    {
      scalar const pm = pressure_m.get_value(q);//duplicated
      scalar const pp = pressure_p.get_value(q);//duplicated
      vector const um = velocity_m.get_value(q);//duplicated
      vector const up = velocity_p.get_value(q);//duplicated
      vector const n  = pressure_m.normal_vector(q);//duplicated

      vector const flux_momentum =
        kernel.calculate_lax_friedrichs_flux_momentum(um, up, gamma, pm, pp, n);//duplicated

      scalar const flux_mass = kernel.calculate_lax_friedrichs_flux_mass(pm, pp, tau, um, up, n);//duplicated
      
      // formuation specific stuff
     }
```
From your response, I think you are suggesting the following, correct?
```
    for(unsigned int q : pressure_m.quadrature_point_indices())
    {
      scalar const pm = pressure_m.get_value(q);
      scalar const pp = pressure_p.get_value(q);
      vector const um = velocity_m.get_value(q);
      vector const up = velocity_p.get_value(q);
      vector const n  = pressure_m.normal_vector(q);

      vector const flux_momentum =
        kernel.calculate_lax_friedrichs_flux_momentum(um, up, gamma, pm, pp, n);

      scalar const flux_mass = kernel.calculate_lax_friedrichs_flux_mass(pm, pp, tau, um, up, n);

      if (formulation == Formulation::Weak)
          // formuation specific stuff
      else if (formulation == ...)
          // formuation specific stuff
```
I would like to propose the following: 
* We go with the second version from above
* I continue with a PR to be able to run a minimal example
* We test the performance for all three versions. If I have access to a cluster at that time, I am happy to do it, otherwise I might have to ask for help.
* If we can measure a noticeable performance difference, we reintroduce the template parameter

I made the changes as proposed. If you agree to above procedure, the PR is ready to merge from my side.

kronbichler: Yes, this sounds like a good first step. We probably need to look into the variants a bit. I would personally not see templates as a form of code duplication (because it duplicates the machine code, not the maintained code), depending on how the code is setup up. The main part is that templates need a dispatch mechanism to select between the different variants at run-time, and how much work it is for the compiler to figure out different `if` statements compared to more compact statements.

As a side remark, I would point out that the compiler will have different heuristics that go in different directions, which can lead to "surprising" behavior where a template must not even uniformly lead to better performance, if the compiler e.g. decides to put common code into a function that comes with too much call overhead. Or that for dispatch is near the innermost part, like in the first variant you show here (but done via a templated inner function), the compile time or library size might not even be affected.

nfehn: We probably refer to different types of code duplication in this conversation. What I meant is the if-else statement with three times the same function call, the only difference being the template parameter.

@jh66637 I think we are of the same opinion regarding the face integrals (and that this is different than for the cell integrals). I was probably irritated because of the reference to conversation https://github.com/exadg/exadg/pull/587#discussion_r1377387760 (which discusses another aspect of "code duplication", which I did not mean in the present conversation).

jh66637: I think we are all on the same page :) 

I will add an issue to check the different versions and I think this PR can be merged, and I can continue adding a working acoustic solver :)

include/exadg/acoustic_conservation_equations/spatial_discretization/operators/weak_boundary_conditions.h
```diff
@@ -0,0 +1,165 @@
...
+#include <exadg/acoustic_conservation_equations/user_interface/boundary_descriptor.h>
+#include <exadg/functions_and_boundary_conditions/evaluate_functions.h>
+#include <exadg/matrix_free/integrators.h>
+
+namespace ExaDG
+{
+namespace Acoustics
+{
+/**
+ * Class to access boundary values with a similar interface to FaceIntegrators.
```
nfehn, line 34: I agree only partly on the statement that the interface is similar to `FaceIntegrator`. I think we should be more explicit here in the sense that we need generic programming regarding the two functions `reinit()`/`get_value()`, not more than this.

I am not fully happy with the current design. Most of the code of this class is duplicated between `BoundaryFaceIntegratorP` and `BoundaryFaceIntegratorU`, which should be avoided. The only difference is the implementation of `get_value()`, which is problem/operator-specific and where we have free functions in ExaDG. The use of free functions is a modular design in the sense that it keeps dependencies to a minimum. ExaDG implementations for other operators are more general in the sense of an additional parameter `operator_type` (https://github.com/exadg/exadg/blob/ee073c7979b3c695aca00ade2ab9ecfebc74a410/include/exadg/incompressible_navier_stokes/spatial_discretization/operators/weak_boundary_conditions.h#L72) needed for implicit time stepping, which does not appear here for the explicit acoustic operator, which always evaluates the `full` operator.


There are two open questions in my opinion:
- Why not introducing a base class `BoundaryFaceIntegratorBase`?
- How to realize `get_value()` in a generic way to be able to deal with all types of ExaDG operators? I see the three options of (i) templating `BoundaryFaceIntegratorBase` with `BoundaryDescriptor`, (ii) using problem specific derived classes `BoundaryFaceIntegratorDerived` (where an concrete boundary descriptor is used but which does not allow the short-cut in the `reinit()` function used currently), or (iii) the use of lambda functions in order to hide `BoundaryDescriptor` (and other problem-specific parameters).

include/exadg/acoustic_conservation_equations/spatial_discretization/operators/weak_boundary_conditions.h
```diff
@@ -0,0 +1,165 @@
...
+        *boundary_descriptor.dirichlet_bc.find(boundary_id)->second,
+        integrator_m.quadrature_point(q),
+        evaluation_time);
+      return -integrator_m.get_value(q) + Number{2.0} * g;
+    }
+    else
+    {
+      AssertThrow(false,
+                  dealii::ExcMessage("Boundary type of face is invalid or not implemented."));
+    }
```
nfehn, line 88: Here, one could call a free function `get_exterior_value()` like for other operators in ExaDG.

include/exadg/acoustic_conservation_equations/spatial_discretization/operators/weak_boundary_conditions.h
```diff
@@ -0,0 +1,165 @@
...
+  reinit(unsigned int const face)
+  {
+    auto const boundary_id_new = integrator_m.get_matrix_free().get_boundary_id(face);
+
+    // only update boundary_type if needed to avoid an unnecessary search in boundary_descriptor
+    if(boundary_id_new != boundary_id)
+    {
+      boundary_id   = boundary_id_new;
+      boundary_type = boundary_descriptor.get_boundary_type(boundary_id);
+    }
```
nfehn, line 63: This optimization could be shifted to `get_value()`. The (probably) expensive part `get_boundary_type()` (operating on `std::map`s) would not be done more often than currently, only the if-statement would be evaluated more often. Here, the question would again be whether such an if-statement has any relevance in terms of performance as compared to the other instructions/operations of `get_value()`.

The benefit would be that `BoundaryFaceIntegrator` does not depend on `BoundaryType` and `BoundaryDescriptor`, or that a base class `BoundaryFaceIntegratorBase` would not need `BoundaryType` and `BoundaryDescriptor` as template parameters.

include/exadg/acoustic_conservation_equations/spatial_discretization/operators/operator.h
```diff
@@ -0,0 +1,495 @@
...
+    BoundaryFaceIntegratorP<dim, Number> pressure_p(pressure_m, *data.bc->pressure);
+
+    FaceIntegratorU velocity_m(matrix_free_in, true, data.dof_index_velocity, data.quad_index);
+    BoundaryFaceIntegratorU<dim, Number> velocity_p(velocity_m, *data.bc->velocity);
+
+    for(unsigned int face = face_range.first; face < face_range.second; face++)
+    {
+      pressure_m.reinit(face);
+      pressure_m.gather_evaluate(src.block(0), integrator_flags_p.face_evaluate);
+      pressure_p.reinit(face);
```
nfehn, line 460: The `reinit()` function does not seem to be part of the "interface" `BoundaryFaceIntegratorBase`, because we are here operating on a concrete type, right?

Then, `reinit()` and `set_time()` could actually be combined.

include/exadg/acoustic_conservation_equations/spatial_discretization/operators/weak_boundary_conditions.h
nfehn: As written here https://github.com/exadg/exadg/pull/587/files#r1383107143, the "interface" / "similarity" with `FEFaceEvaluation` seems to be `get_value()` only.

nfehn: A question would then indeed be whether it makes sense that dealii  provides a base/interface class `FEFaceEvaluationInterface` with `get_value()`/`get_gradient()` only. `FEFaceIntegrationInterface` as a derived class would add the `submit()` functionality ("Integration" = testing, weighting) to `FEFaceEvaluationInterface` according to this idea.

jh66637: I didn't introduce a base class, since the amount of duplicated code is small. If this requested base class will be used in other parts of `ExaDG` in the future, I agree that we should have one. I will restructure as suggested, just a few answers to your comments:

* I think the description is correct: The interface **to access the value** is similar to `FEEval`.
* Obviously additional parameters such as `OperatorType` etc. can be easily added, but I think this is not relevant for this PR and should be done when all the other operators are adjusted to this design.
* I am in favor of templating the boundary descriptor. It is natural to fetch the boundary ID in the `reinit()` function and not in `get_value()`, also the things you didn't like about the `formulation` template do not apply here.
* I don't understand this comment https://github.com/exadg/exadg/pull/587#discussion_r1383122750

nfehn: The comment you did not understand raises the question whether a base class introducing the `get_value/gradient()` interface should be implemented in deal.II. It is just an idea that does not have to be resolved in the present PR.

> I think the description is correct: The interface to access the value is similar to FEEval.

I agree. However, it is not only similar, it needs to be identical to `FEEval`. At the same time, I think `reinit()` and `set_time()` are independent of `FEEval`. (Generally speaking, this is a disadvantage of using templates vs. base classes and dynamic polymorphism. The definition of interfaces is implicit and somewhat hard to extract from the code, since there is no base class to jump into and see what the required interface is. It can be extracted from the code, but this needs time.) Hence, I consider it important to document/explain this. So I want to make the documentation more precise than it is currently.

Overall, you do not have to change the code as suggested. I want to discuss aspects with pros and cons, and maybe also get other developers involved in the discussion. In the end, I am willing to merge a version that you consider final from your side. When then still have the discussions as a reference for upcoming PRs (by you or by other developers).

include/exadg/acoustic_conservation_equations/spatial_discretization/operators/weak_boundary_conditions.h
```diff
@@ -0,0 +1,152 @@
...
+// compute exterior velocity values for different boundary types
+template<int dim, typename Number>
+inline DEAL_II_ALWAYS_INLINE //
+  dealii::Tensor<1, dim, dealii::VectorizedArray<Number>>
+  calculate_exterior_value_velocity(unsigned int const                       q,
+                                    FaceIntegrator<dim, dim, Number> const & integrator_m,
+                                    BoundaryTypeU const &                    boundary_type,
+                                    dealii::types::boundary_id const         boundary_id,
+                                    BoundaryDescriptorU<dim> const &         boundary_descriptor,
+                                    Number const                             time)
```
nfehn, line 68: don't you get warnings regarding unused variables?

include/exadg/functions_and_boundary_conditions/boundary_face_integrator_base.h
```diff
@@ -0,0 +1,84 @@
...
+#include <deal.II/matrix_free/matrix_free.h>
+
+#include <exadg/utilities/enum_utilities.h>
+
+namespace ExaDG
+{
+/**
+ * Base class to access boundary values with a similar interface to FaceIntegrators.
+ * In the derived class a @c get_value() has to be implemented. Since the return type
+ * depends on the derived class, a pure virtual function for this purpose can not be added.
```
nfehn, line 34: I think we can still improve this description. According to my understanding, the main purpose of this class is to allow a simple call `face_integrator.get_value(q)` with just the quadrature point index `q` as argument(s). To enable this, we need a class that receives the relevant parameters ({MatrixFree, BoundaryDescriptor} and {face_id, time}) during construction and `reinit()`, respectively.

It is not mandatory for this PR and I can also change it later.

Other operators might also need the (normal) gradient on boundary faces. This is another reason why such a base class is useful in my opinion (it can be used equally for first-derivative and second-derivative operators).

nfehn (approved):
Looks good from my side. We should make sure that we do not get warning (because of unused variables). Apart from that, I could merge if you want, @jh66637.

include/exadg/acoustic_conservation_equations/spatial_discretization/operators/weak_boundary_conditions.h
kronbichler: > A question would then indeed be whether it makes sense that dealii provides a base/interface class `FEFaceEvaluationInterface` with `get_value()/get_gradient()` only. 

This is to some extent related to other discussions we had previously (regarding the "inefficiency" of the `FEEvaluation` interface and the associated `for` loops), where in fact being able to separate the operations at quadrature points from the data containers and the loop index would be helpful. If we consider adding another interface, what you desire as an interface would be possible.

One caveat is that the `get_value()/get_gradient()` cannot be virtual functions because this is one of those places where inlining does have a tangible effect and preventing inlining via function calls would be clearly visible. The other effect preventing a simple change into a `FEEvaluationInterface` as a base class is the complicated inheritance logic present in `FEEvaluation` already now, where the templates on the number of components are used to define different types for the `get_value()`, `get_gradient()` etc. interfaces, as a means to make as intuitive user interfaces as possible. Changing the order by which the classes are set up without introducing virtual functions is probably taking a considerable effort. I think that concentrating the effort on a **new class for the quadrature point operation**, trying to address the various limitations we learned in the past and focusing on usability on various architectures for the future, would be the best option forward. It could then naturally pick up the interface necessary here. We would still need to find a more suitable mechanism than dynamic polymorphism, but I think we are approaching the right problem.

kronbichler: Regarding the "complicated inheritance logic present", the class to look at is `FEEvaluationAccess` and its various specializations, https://www.dealii.org/developer/doxygen/deal.II/classFEEvaluationAccess.html

nfehn: I agree that we should evolve the ExaDG code for now and accept that this will be somewhat a temporary solution. I think we will still get ideas for the design of dealii/matrixfree from such developments in ExaDG .

include/exadg/acoustic_conservation_equations/spatial_discretization/operators/weak_boundary_conditions.h
jh66637: Good catch. Surprisingly, I didn't see warnings when linking to `ExaDG` to compile the operators and check if the results are correct. I fixed this.

include/exadg/acoustic_conservation_equations/spatial_discretization/operators/operator.h
```diff
@@ -0,0 +1,493 @@
...
+           typename ExteriorFaceIntegratorU>
+  void
+  do_face_integral(FaceIntegratorP &         pressure_m,
+                   ExteriorFaceIntegratorP & pressure_p,
+                   FaceIntegratorU &         velocity_m,
+                   ExteriorFaceIntegratorU & velocity_p) const
+  {
+    for(unsigned int q : pressure_m.quadrature_point_indices())
+    {
+      scalar const pm = pressure_m.get_value(q);
```
nfehn, line 327: In case of implicit problems with the solution of a linear system of equations, we need to be able to compute only the inhomogeneous action of the operator on boundary faces. This means that `pressure_m.get_value(q)` would not be sufficient, but we would need something like `calculate_interior_value(...)`, see ConvDiff module: https://github.com/exadg/exadg/blob/51a6a7e65e81c4a3816367da24f36a4d8fc813b9/include/exadg/convection_diffusion/spatial_discretization/operators/weak_boundary_conditions.h#L49-L69

@jh66637 how would you realize this? Following the design of the present PR, we could also template on the interior face integrator, increasing the number of template parameters by two. However, for boundary faces, this means we would have to define not only a class `ExteriorBoundaryFaceIntegrator`, but also a class `InteriorBoundaryFaceIntegrator`.

include/exadg/acoustic_conservation_equations/spatial_discretization/operators/operator.h
```diff
@@ -0,0 +1,493 @@
...
+        {
+          // minus signs since n⁺ = - n⁻
+          pressure_p.submit_value(-flux_momentum_weak, q);
+          velocity_p.submit_value(-rho_inv * (flux_mass - pp) * n, q);
+        }
+      }
+      else
+      {
+        AssertThrow(false, dealii::ExcMessage("Not implemented."));
+      }
```
nfehn, line 382: Given the difficulties in evaluating the interior/exterior values in a generic way and for all types of loops (see discussion here https://github.com/exadg/exadg/pull/587/files#r1389376858, isn't it actually this part of the code that can be written generically? @jh66637 In the non-matching case, you do not integrate over the neighboring face, right? Then, writing these lines of code as a kernel function ("kernel" in the meaning of a quadrature point function) would not require templating with `ExteriorFaceIntegratorP` and `ExteriorFaceIntegratorU`, because we are always operating on the "normal" `dealii::FaceIntegrator/FEFaceEvaluation`?

@kronbichler would be great if you could also share your opinion.

PR #589, opened 2023-10-31
Implement additive Schwarz Preconditioner

bergbauer:
This PR implements a matrix-based additive Schwarz preconditioner where a sparse matrix is assembled first, then cell blocks are cut out and factorized using LU factorization. It can then be used as a preconditioner or as a smoother in multigrid.

### DIFF
--- a/include/exadg/operators/multigrid_operator.h
+++ b/include/exadg/operators/multigrid_operator.h
@@ -135,15 +135,15 @@ public:
   }
 
   virtual void
-  apply_inverse_as_matrices(VectorType & dst, VectorType const & src) const final
+  apply_inverse_additive_schwarz_matrices(VectorType & dst, VectorType const & src) const final
   {
-    pde_operator->apply_inverse_as_matrices(dst, src);
+    pde_operator->apply_inverse_additive_schwarz_matrices(dst, src);
   }
 
   virtual void
-  compute_factorized_as_matrices() const final
+  compute_factorized_additive_schwarz_matrices() const final
   {
-    pde_operator->compute_factorized_as_matrices();
+    pde_operator->compute_factorized_additive_schwarz_matrices();
   }
 
 #ifdef DEAL_II_WITH_TRILINOS
--- a/include/exadg/operators/multigrid_operator.h
+++ b/include/exadg/operators/multigrid_operator.h
@@ -134,6 +134,18 @@ public:
     pde_operator->apply_inverse_block_diagonal(dst, src);
   }
 
+  virtual void
+  apply_inverse_as_matrices(VectorType & dst, VectorType const & src) const final
+  {
+    pde_operator->apply_inverse_as_matrices(dst, src);
+  }
+
+  virtual void
+  compute_factorized_as_matrices() const final
+  {
+    pde_operator->compute_factorized_as_matrices();
+  }
+
 #ifdef DEAL_II_WITH_TRILINOS
   void
   init_system_matrix(dealii::TrilinosWrappers::SparseMatrix & system_matrix,
--- a/include/exadg/operators/operator_base.cpp
+++ b/include/exadg/operators/operator_base.cpp
@@ -473,7 +473,7 @@ OperatorBase<dim, Number, n_components>::calculate_block_diagonal_matrices() con
     auto dofs =
       matrix_free->get_shape_info(this->data.dof_index).dofs_per_component_on_cell * n_components;
 
-    matrices.resize(matrix_free->n_cell_batches() * vectorization_length, CellMatrix(dofs, dofs));
+    matrices.resize(matrix_free->n_cell_batches() * vectorization_length, LAPACKMatrix(dofs, dofs));
 
     block_diagonal_preconditioner_is_initialized = true;
   }
@@ -488,7 +488,8 @@ OperatorBase<dim, Number, n_components>::calculate_block_diagonal_matrices() con
 
 template<int dim, typename Number, int n_components>
 void
-OperatorBase<dim, Number, n_components>::add_block_diagonal_matrices(BlockMatrix & matrices) const
+OperatorBase<dim, Number, n_components>::add_block_diagonal_matrices(
+  std::vector<LAPACKMatrix> & matrices) const
 {
   AssertThrow(is_dg, dealii::ExcMessage("Block Jacobi only implemented for DG!"));
 
@@ -714,7 +715,8 @@ OperatorBase<dim, Number, n_components>::update_block_diagonal_preconditioner() 
       // allocate memory only the first time
       auto dofs =
         matrix_free->get_shape_info(this->data.dof_index).dofs_per_component_on_cell * n_components;
-      matrices.resize(matrix_free->n_cell_batches() * vectorization_length, CellMatrix(dofs, dofs));
+      matrices.resize(matrix_free->n_cell_batches() * vectorization_length,
+                      LAPACKMatrix(dofs, dofs));
     }
 
     block_diagonal_preconditioner_is_initialized = true;
@@ -1624,8 +1626,8 @@ template<int dim, typename Number, int n_components>
 void
 OperatorBase<dim, Number, n_components>::cell_loop_block_diagonal(
   dealii::MatrixFree<dim, Number> const & matrix_free,
-  BlockMatrix &                           matrices,
-  BlockMatrix const &,
+  std::vector<LAPACKMatrix> &             matrices,
+  std::vector<LAPACKMatrix> const &,
   Range const & range) const
 {
   IntegratorCell integrator =
@@ -1660,8 +1662,8 @@ template<int dim, typename Number, int n_components>
 void
 OperatorBase<dim, Number, n_components>::face_loop_block_diagonal(
   dealii::MatrixFree<dim, Number> const & matrix_free,
-  BlockMatrix &                           matrices,
-  BlockMatrix const &,
+  std::vector<LAPACKMatrix> &             matrices,
+  std::vector<LAPACKMatrix> const &,
   Range const & range) const
 {
   IntegratorFace integrator_m =
@@ -1721,8 +1723,8 @@ template<int dim, typename Number, int n_components>
 void
 OperatorBase<dim, Number, n_components>::boundary_face_loop_block_diagonal(
   dealii::MatrixFree<dim, Number> const & matrix_free,
-  BlockMatrix &                           matrices,
-  BlockMatrix const &,
+  std::vector<LAPACKMatrix> &             matrices,
+  std::vector<LAPACKMatrix> const &,
   Range const & range) const
 {
   IntegratorFace integrator_m =
@@ -1763,8 +1765,8 @@ template<int dim, typename Number, int n_components>
 void
 OperatorBase<dim, Number, n_components>::cell_based_loop_block_diagonal(
   dealii::MatrixFree<dim, Number> const & matrix_free,
-  BlockMatrix &                           matrices,
-  BlockMatrix const &,
+  std::vector<LAPACKMatrix> &             matrices,
+  std::vector<LAPACKMatrix> const &,
   Range const & range) const
 {
   IntegratorCell integrator =
@@ -2239,7 +2241,7 @@ OperatorBase<dim, Number, n_components>::internal_compute_factorized_additive_sc
                                                                overlapped_block_matrix);
 
   // factorize and store cell matrices
-  matrices.resize(n_cells, CellMatrix(dofs_per_cell));
+  matrices.resize(n_cells, LAPACKMatrix(dofs_per_cell));
   for(unsigned int cell_batch_index = 0; cell_batch_index < matrix_free->n_cell_batches();
       ++cell_batch_index)
   {
--- a/include/exadg/operators/operator_base.cpp
+++ b/include/exadg/operators/operator_base.cpp
@@ -2199,6 +2199,7 @@ OperatorBase<dim, Number, n_components>::internal_compute_factorized_additive_sc
   std::vector<std::vector<dealii::types::global_dof_index>> all_dof_indices(
     n_cells, std::vector<dealii::types::global_dof_index>(dofs_per_cell));
 
+  // compute weights by counting the contributions to a DoF
   initialize_dof_vector(weights);
   for(unsigned int cell_batch_index = 0; cell_batch_index < matrix_free->n_cell_batches();
       ++cell_batch_index)
@@ -2222,6 +2223,7 @@ OperatorBase<dim, Number, n_components>::internal_compute_factorized_additive_sc
   }
   weights.compress(dealii::VectorOperation::add);
 
+  // prepare the weights vector for symmetric weighting
   for(unsigned int i = 0; i < weights.size(); ++i)
   {
     if(weights.in_local_range(i))
--- a/include/exadg/operators/operator_base.cpp
+++ b/include/exadg/operators/operator_base.cpp
@@ -21,6 +21,7 @@
 
 // deal.II
 #include <deal.II/dofs/dof_tools.h>
+#include <deal.II/lac/sparse_matrix_tools.h>
 #include <deal.II/lac/sparsity_tools.h>
 #include <deal.II/matrix_free/tools.h>
 
@@ -472,8 +473,7 @@ OperatorBase<dim, Number, n_components>::calculate_block_diagonal_matrices() con
     auto dofs =
       matrix_free->get_shape_info(this->data.dof_index).dofs_per_component_on_cell * n_components;
 
-    matrices.resize(matrix_free->n_cell_batches() * vectorization_length,
-                    dealii::LAPACKFullMatrix<Number>(dofs, dofs));
+    matrices.resize(matrix_free->n_cell_batches() * vectorization_length, CellMatrix(dofs, dofs));
 
     block_diagonal_preconditioner_is_initialized = true;
   }
@@ -714,8 +714,7 @@ OperatorBase<dim, Number, n_components>::update_block_diagonal_preconditioner() 
       // allocate memory only the first time
       auto dofs =
         matrix_free->get_shape_info(this->data.dof_index).dofs_per_component_on_cell * n_components;
-      matrices.resize(matrix_free->n_cell_batches() * vectorization_length,
-                      dealii::LAPACKFullMatrix<Number>(dofs, dofs));
+      matrices.resize(matrix_free->n_cell_batches() * vectorization_length, CellMatrix(dofs, dofs));
     }
 
     block_diagonal_preconditioner_is_initialized = true;
@@ -744,7 +743,8 @@ OperatorBase<dim, Number, n_components>::init_system_matrix(
   dealii::TrilinosWrappers::SparseMatrix & system_matrix,
   MPI_Comm const &                         mpi_comm) const
 {
-  internal_init_system_matrix(system_matrix, mpi_comm);
+  dealii::DynamicSparsityPattern dsp;
+  internal_init_system_matrix(system_matrix, dsp, mpi_comm);
 }
 
 template<int dim, typename Number, int n_components>
@@ -763,7 +763,8 @@ OperatorBase<dim, Number, n_components>::init_system_matrix(
   dealii::PETScWrappers::MPI::SparseMatrix & system_matrix,
   MPI_Comm const &                           mpi_comm) const
 {
-  internal_init_system_matrix(system_matrix, mpi_comm);
+  dealii::DynamicSparsityPattern dsp;
+  internal_init_system_matrix(system_matrix, dsp, mpi_comm);
 }
 
 template<int dim, typename Number, int n_components>
@@ -779,8 +780,9 @@ template<int dim, typename Number, int n_components>
 template<typename SparseMatrix>
 void
 OperatorBase<dim, Number, n_components>::internal_init_system_matrix(
-  SparseMatrix &   system_matrix,
-  MPI_Comm const & mpi_comm) const
+  SparseMatrix &                   system_matrix,
+  dealii::DynamicSparsityPattern & dsp,
+  MPI_Comm const &                 mpi_comm) const
 {
   dealii::DoFHandler<dim> const & dof_handler =
     this->matrix_free->get_dof_handler(this->data.dof_index);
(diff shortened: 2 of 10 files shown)


include/exadg/operators/operator_base.cpp
```diff
@@ -2153,6 +2155,173 @@ OperatorBase<dim, Number, n_components>::evaluate_face_integrals() const
...
+OperatorBase<dim, Number, n_components>::compute_factorized_as_matrices() const
+{
+#ifdef DEAL_II_WITH_TRILINOS
+  internal_compute_factorized_as_matrices<dealii::TrilinosWrappers::SparseMatrix>();
+#elif DEAL_II_WITH_PETSC
+  internal_compute_factorized_as_matrices<dealii::PETScWrappers::MPI::SparseMatrix>();
+#else
+  AssertThrow(n_mpi_processes == 1,
+              ExcMessage(
+                "Use Trilinos or Petsc for distributed sparse matrices needed for this function!"));
```
nfehn, line 2169: I think "Use" or "needed" is too much here

include/exadg/operators/operator_base.cpp
```diff
@@ -744,7 +743,8 @@ OperatorBase<dim, Number, n_components>::init_system_matrix(
   dealii::TrilinosWrappers::SparseMatrix & system_matrix,
   MPI_Comm const &                         mpi_comm) const
 {
-  internal_init_system_matrix(system_matrix, mpi_comm);
+  dealii::DynamicSparsityPattern dsp;
```
nfehn, line 746: Why is the variable `dsp` needed outside the function call?

include/exadg/poisson/user_interface/enum_types.h
```diff
@@ -68,6 +68,7 @@ enum class Preconditioner
   PointJacobi,
   BlockJacobi,
   AMG,
+  AdditiveSchwarz,
```
nfehn, line 71: just a minor comment: I think I would change the order of AMG and AdditiveSchwarz.

include/exadg/operators/operator_base.cpp
```diff
@@ -472,8 +473,7 @@ OperatorBase<dim, Number, n_components>::calculate_block_diagonal_matrices() con
     auto dofs =
       matrix_free->get_shape_info(this->data.dof_index).dofs_per_component_on_cell * n_components;
 
-    matrices.resize(matrix_free->n_cell_batches() * vectorization_length,
-                    dealii::LAPACKFullMatrix<Number>(dofs, dofs));
+    matrices.resize(matrix_free->n_cell_batches() * vectorization_length, CellMatrix(dofs, dofs));
```
nfehn, line 476: `CellMatrix` is defined in deal.II?

nfehn: found it

include/exadg/operators/operator_base.cpp
```diff
@@ -2153,6 +2155,173 @@ OperatorBase<dim, Number, n_components>::evaluate_face_integrals() const
...
+template<int dim, typename Number, int n_components>
+void
+OperatorBase<dim, Number, n_components>::apply_inverse_as_matrices(VectorType &       dst,
+                                                                   VectorType const & src) const
+{
+  if(is_dg)
+    matrix_free->cell_loop(&This::cell_loop_apply_inverse_block_diagonal_matrix_based,
+                           this,
+                           dst,
+                           src);
```
nfehn, line 2279: Is it correct that "Additive Schwarz"  always refers to single elements in the literature and not a batch of elements? In case of a batch of elements, additive Schwarz would be different from "block-Jacobi" also for DG. So I am not sure if we should implement it like this or actually assert that `is_dg==false` in case of additive Schwarz. @bergbauer How is it done for block-Jacobi, do we assert `is_dg==true` in that case? If so, it would probably make sense to do it similarly for additive Schwarz, i.e. only implement the continuous Galerkin case.

include/exadg/operators/operator_base.cpp
```diff
@@ -2153,6 +2155,173 @@ OperatorBase<dim, Number, n_components>::evaluate_face_integrals() const
...
+    }
+  }
+  weights.compress(dealii::VectorOperation::add);
+
+  for(unsigned int i = 0; i < weights.size(); ++i)
+  {
+    if(weights.in_local_range(i))
+      weights[i] = 1. / std::sqrt(weights[i]);
+  }
+  weights.update_ghost_values();
```
nfehn, line 2227: Could you explain via comments the idea behind the "weights" calculations?

include/exadg/operators/operator_base.cpp
```diff
@@ -2153,6 +2155,173 @@ OperatorBase<dim, Number, n_components>::evaluate_face_integrals() const
...
+      // get overlapped cell matrix
+      auto const & overlapped_cell_matrix =
+        overlapped_block_matrix[cell_batch_index * vectorization_length + v];
+
+      // store the cell matrix and renumber lexicographic if necessary
+      auto & lapack_matrix = matrices[cell_batch_index * vectorization_length + v];
+      if(is_dg)
+      {
+        lapack_matrix = overlapped_cell_matrix;
+      }
```
nfehn, line 2255: This is similar as the other comment on `is_dg`. It looks like the whole system matrix is assembled in the DG case? Does weights = 1 hold in the DG case?

include/exadg/operators/operator_base.cpp
```diff
@@ -2153,6 +2155,173 @@ OperatorBase<dim, Number, n_components>::evaluate_face_integrals() const
...
+  unsigned int const n_cells = matrix_free->n_cell_batches() * vectorization_length;
+
+  // assemble a temporary sparse matrix to cut out the blocks
+  SparseMatrix                   tmp_matrix;
+  dealii::DynamicSparsityPattern dsp;
+  internal_init_system_matrix(tmp_matrix, dsp, dof_handler.get_communicator());
+  internal_calculate_system_matrix(tmp_matrix);
+
+  // cut out overlapped block matrices
+  std::vector<dealii::FullMatrix<Number>> overlapped_block_matrix(
```
nfehn, line 2193: I would use "matrices" here since it is a vector of matrices.

include/exadg/operators/operator_base.cpp
```diff
@@ -2153,6 +2155,173 @@ OperatorBase<dim, Number, n_components>::evaluate_face_integrals() const
...
+    n_cells, dealii::FullMatrix<Number>(dofs_per_cell));
+
+  std::vector<std::vector<dealii::types::global_dof_index>> all_dof_indices(
```
nfehn, line 2196: I think somewhat more precise would be `dof_indices_all_cells` or `dof_indices_all_blocks` as a name. As the comment above, I think it helps when the variable name reflects that the variable is a `std::vector`.

include/exadg/operators/operator_base.cpp
nfehn: In particular, where does `1./std::sqrt()` come from?

include/exadg/operators/operator_base.h
```diff
@@ -106,7 +106,8 @@ class OperatorBase : public dealii::Subscriptor
 
   static unsigned int const vectorization_length = dealii::VectorizedArray<Number>::size();
 
-  typedef std::vector<dealii::LAPACKFullMatrix<Number>> BlockMatrix;
+  typedef dealii::LAPACKFullMatrix<Number> CellMatrix;
+  typedef std::vector<CellMatrix>          BlockMatrix;
```
nfehn, line 110: should be `BlockMatrices` in my opinion, or maybe we should always write explicitly `std::vector<CellMatrix>` where needed for clarity.

include/exadg/operators/operator_base.cpp
```diff
@@ -2153,6 +2155,173 @@ OperatorBase<dim, Number, n_components>::evaluate_face_integrals() const
...
+      weights[i] = 1. / std::sqrt(weights[i]);
+  }
+  weights.update_ghost_values();
+
+  dealii::SparseMatrixTools::restrict_to_full_matrices<SparseMatrix,
+                                                       dealii::DynamicSparsityPattern,
+                                                       Number>(tmp_matrix,
+                                                               dsp,
+                                                               all_dof_indices,
+                                                               overlapped_block_matrix);
```
bergbauer, line 2234: We need the sparsity pattern for this function call.

include/exadg/operators/operator_base.cpp
bergbauer: We need it for `dealii::SparseMatrixTools::restrict_to_full_matrices`. I have marked the usage below.

include/exadg/operators/operator_base.cpp
bergbauer: The idea is to weight contributions which have overlap with `1/n_{contributions to dof}`. To do this in a symmetric way we weight the local src vector and the local dst vector with `1./std::sqrt()`. I will add a comment!

include/exadg/operators/operator_base.cpp
bergbauer: For this special case of elementwise additive Schwarz and DG yes, we have only one contribution per dof, meaning weight = 1. For this special case additive Schwarz and Block Jacobi are identical. One can interpret Block Jacobi as a special additive Schwarz case.

nfehn: From a software side, it looks like we should make additive Schwarz and block-Jacobi "disjunct" implementations?

include/exadg/operators/operator_base.cpp
bergbauer: Additive Schwarz can be element wise but can also batch several elements meaning it is a more general term. We could call the current class `ElementwiseAdditiveSchwarz` to make this more clear. I am looking into strategies for batching but this is not ready for master yet and will be an extension for the future.

>So I am not sure if we should implement it like this or actually assert that is_dg==false in case of additive Schwarz.

I am strongly against asserting additive Schwarz for DG. One main advantage of additive Schwarz is that you can use it as a Smoother on any multigrid level (CG or DG). With BlockJacobi as a smoother no c-transfer is currently possible. Sure, currently it is equivalent to Block Jacobi for DG, which is why I chose to use the optimized path here. We can also run though the general path for DG if you think this is clearer from a code design perspective.
Applying the factorization should be much more expensive than zeroing the vector and applying the weighting anyways.

include/exadg/operators/operator_base.cpp
bergbauer: Yes, I think so. The only option I see is if we detect block-Jacobi behaviour of additive Schwarz we use specialised code paths for performance optimization (like we do here in the PR for the application of the element wise inverse).

include/exadg/operators/operator_base.cpp
nfehn: I think for a clearer design, the function `internal_compute_factorized_as_matrices()` should distinguish between DG and CG on the top level, just as the function `apply_inverse_as_matrices()`. It looks like weights are computed always, but are not used in the DG case, right? 

Whether we then remove one of the classes, merge them, rename `blockJacobi` or `AdditiveSchwarz`, would be a subsequent question in my opinion.

include/exadg/operators/multigrid_operator.h
```diff
@@ -134,6 +134,18 @@ class MultigridOperator : public MultigridOperatorBase<dim, Number>
     pde_operator->apply_inverse_block_diagonal(dst, src);
   }
 
+  virtual void
+  apply_inverse_as_matrices(VectorType & dst, VectorType const & src) const final
```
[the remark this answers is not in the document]
bergbauer: Fixed in 2004d4f71516629f5673edec689fcb647c5f557e

include/exadg/operators/operator_base.cpp
bergbauer: Fixed in c35081b80bce12dadc50eee1ae1788de959be394

include/exadg/operators/operator_base.cpp
bergbauer: I added comments in 51c103e259e52b893757735d59929de446234d56

include/exadg/operators/operator_base.h
bergbauer: Done in 04e1da7b41495fabec19ac2807737fb101f1726f

I decided to keep a typedef for `LAPACKMatrix` to shorten the name without changing the meaning. The std::vector reflects everywhere that we simply have a vector of LAPACK `matrices`.

include/exadg/operators/operator_base.cpp
bergbauer: Done in 0946e0ce697c568d48f462918088c27ec907664c

include/exadg/operators/operator_base.cpp
bergbauer: And this is also changed in 0946e0ce697c568d48f462918088c27ec907664c

include/exadg/operators/operator_base.cpp
bergbauer: I think we have two options here:

- we can currently distinguish between CG and DG on the top level, using the block-Jacobi path for DG and the general additive Schwarz path for CG because at the moment we only support cellwise additive Schwarz with weighting by contributions per DoF
- we run everything through the general additive Schwarz path for both CG and DG and for the `compute` and `apply` function

The current solution is a mix of the two options above because in the `compute` function the call `LAPACKMatrix::compute_lu_factorization` dominates the computational effort and it is only called once per solve. In the `apply` function, the dominant function call is `LAPACKMatrix::solve` which has lower complexity than the factorize call. Also `apply` is called in every iteration of a solver which makes optimization more important.

@peterrum Do you have an opinion on this topic as you thought about additive Schwarz extensively?

bergbauer: ec537287bdb9a94abf48b3083a5443abca3ec87e uses the `BlockJacobi` path for DG

include/exadg/operators/operator_base.cpp
```diff
@@ -2153,6 +2157,175 @@ OperatorBase<dim, Number, n_components>::evaluate_face_integrals() const
...
+        {
+          this->reinit_cell(integrator, cell);
+
+          integrator.read_dof_values(weights);
+          for(unsigned int i = 0; i < dofs_per_cell; ++i)
+            local_weights_vector[i] = integrator.begin_dof_values()[i];
+
+          integrator.read_dof_values(src);
+
+          for(unsigned int v = 0; v < vectorization_length; ++v)
```
nfehn, line 2307: The function above uses `n_filled_lanes` and not `vectorization_length`. Could this become a problem?

include/exadg/operators/operator_base.cpp
```diff
@@ -2153,6 +2157,175 @@ OperatorBase<dim, Number, n_components>::evaluate_face_integrals() const
...
+          {
+            for(unsigned int i = 0; i < dofs_per_cell; ++i)
+              local_vector[i] = integrator.begin_dof_values()[i][v] * local_weights_vector[i][v];
```
nfehn, line 2310: I think here I would again refer to the symmetric weighting (if I understand the implementation correctly ...).

nfehn:
The code looks overall good from my perspective. I raised some questions, mainly to improve my own understanding of the code (or `MatrixFree`).

include/exadg/operators/operator_base.cpp
nfehn: Put differently, what is the difference between `cell` used here and `cell_batch_index` used above?

include/exadg/operators/operator_base.cpp
```diff
@@ -2153,6 +2157,175 @@ OperatorBase<dim, Number, n_components>::evaluate_face_integrals() const
...
+        auto const & overlapped_cell_matrix =
+          overlapped_cell_matrices[cell_batch_index * vectorization_length + v];
+
+        // store the cell matrix and renumber lexicographic
+        auto & lapack_matrix = matrices[cell_batch_index * vectorization_length + v];
+
+        auto const & lex = matrix_free->get_shape_info().lexicographic_numbering;
+        for(unsigned int i = 0; i < dofs_per_cell; i++)
+          for(unsigned int j = 0; j < dofs_per_cell; j++)
+            lapack_matrix.set(i, j, overlapped_cell_matrix[lex[i]][lex[j]]);
```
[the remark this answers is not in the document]
bergbauer: `FE_Q` is numbered hierarchical but `FEEvaluation` expects lexicographic numbering. Should we add the same reference like here https://github.com/exadg/exadg/blob/ee073c7979b3c695aca00ade2ab9ecfebc74a410/include/exadg/operators/operator_base.cpp#L1894-L1902 ?

`FEEvaluation::get_dof_indices()` would indeed sometimes make sense, but we usually don't access the dof indices in performance-critical places.

We could also write this as a cell loop but I tend to write out the for loop over cell batches when we don't work on dof vectors.

include/exadg/operators/operator_base.cpp
bergbauer: > The function above uses n_filled_lanes and not vectorization_length. Could this become a problem?

It should not, `FEEvaluation::read_dof_values` and `FEEvaluation::distribute_local_to_global` ignore the lanes which are not filled.

bergbauer: > Put differently, what is the difference between cell used here and cell_batch_index used above?

That's a question of philosophy, we usually call the cell batch number `cell` in matrix-free loops. I like being explict here and would at least call it `cell_batch` or `cell_batch_index` like above.

Standard deal.II often writes `cell` for a `DofCellAccessor/CellAccessor` like in the function above. We should aim for a consistent naming, so for now `cell` for the cell batch index and `cell_v` for the accessor (to be consistent with current ExaDG)?

nfehn: But we still execute `matrix.solve()` for such non-filled lanes? If I understand it correctly, we did not compute / factorize such matrices before solving?

include/exadg/operators/operator_base.cpp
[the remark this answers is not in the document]
nfehn: I am somewhat confused. according to your explanations and https://dealii.org/current/doxygen/deal.II/structinternal_1_1MatrixFreeFunctions_1_1ShapeInfo.html#abad0471d12c4f07f3d83271377e24231, I would expect the code to read
```c++
lapack_matrix.set(lex[i], lex[j], overlapped_cell_matrix[i][j]);
```
because `lapack_matrix` uses lexicographic numerbing  and `overlapped_cell_matrix` hierarchical numbering, and because shape_info::lexicographic_numbering is a  "Renumbering from deal.II's numbering of cell degrees of freedom to lexicographic numbering used inside FEEvaluation"?

 Generally, I think one should be able to get dof indices in the meaning of / consistent with `integrator.read/begin_dof_values()`.

include/exadg/operators/operator_base.cpp
```diff
@@ -2153,6 +2157,175 @@ OperatorBase<dim, Number, n_components>::evaluate_face_integrals() const
...
+      for(unsigned int v = 0; v < n_filled_lanes; ++v)
+      {
+        // get overlapped cell matrix
+        auto const & overlapped_cell_matrix =
+          overlapped_cell_matrices[cell_batch_index * vectorization_length + v];
+
+        // store the cell matrix and renumber lexicographic
+        auto & lapack_matrix = matrices[cell_batch_index * vectorization_length + v];
+
+        auto const & lex = matrix_free->get_shape_info().lexicographic_numbering;
```
nfehn, line 2264: MatrixFree can have multiple DoFHandlers, so shouldn't there somewhere appear the dof-index?

nfehn: see https://dealii.org/current/doxygen/deal.II/classMatrixFree.html#a0f344f8afc12a87625bf247c75730db1

include/exadg/operators/operator_base.cpp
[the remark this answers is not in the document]
nfehn: I discussed this briefly with @peterrum. We think both the name `shape_info.lexicograhpic_numbering` and the deal.II docu is confusing. It should be `lexicographic_to_hierarchical_numbering` in my opinion.

include/exadg/operators/operator_base.cpp
nfehn: Regarding the naming: or something with "cell_iterator" / "cell_pointer" for the accessor to be clear.

include/exadg/operators/operator_base.cpp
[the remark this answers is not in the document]
bergbauer: The documentation is really confusing, I agree. The renumbering in correct as it currently is.

include/exadg/operators/operator_base.cpp
bergbauer: New comments 77374e24c2bc712fec7e8cd17f214cd7a49330a2

include/exadg/operators/operator_base.cpp
bergbauer: This is consistent to current `OperatorBase` 2ecad0db2861d7461986128eec087cbce80133bc

include/exadg/operators/operator_base.cpp
bergbauer: Fixed this and in another place 709ced2d727724f9d3795cf8dbf1540bf92ba552

include/exadg/operators/operator_base.cpp
bergbauer: And we have to use `n_filled_lanes` here, you are right 02d181ae3d97050a6e2c8ce38a75a2885f4cfaf6

Not easy to trigger with global refinements.

include/exadg/operators/operator_base.cpp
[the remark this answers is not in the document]
bergbauer: I renamed the variable but the docu has to be fixed in deal.II. 13e5d649da7b97d9c72ab7001e949634bc32c0eb

nfehn (approved):
looks good!

include/exadg/operators/operator_base.cpp
```diff
@@ -2153,6 +2158,176 @@ OperatorBase<dim, Number, n_components>::evaluate_face_integrals() const
...
 
+template<int dim, typename Number, int n_components>
+void
+OperatorBase<dim, Number, n_components>::compute_factorized_additive_schwarz_matrices() const
+{
+#ifdef DEAL_II_WITH_TRILINOS
+  internal_compute_factorized_additive_schwarz_matrices<dealii::TrilinosWrappers::SparseMatrix>();
+#elif DEAL_II_WITH_PETSC
+  internal_compute_factorized_additive_schwarz_matrices<dealii::PETScWrappers::MPI::SparseMatrix>();
+#else
```
nfehn, line 2169: When working on PR #681, I realized that in other places of ExaDG we want to be able to compile for both Trilinos and PETSc and then decide which code to use depending on a parameter (`SparseMatrixType` introduced in #681, `AMGType` already used in AMG preconditioners). So I think we should adapt this code to be consistent with other parts of ExaDG.

PR #592, opened 2023-11-03
Add minimal acoustic solver

jh66637:
I added the respective structure for a minimal acoustic solver (including one simple application).

There is not a lot going on. Yet, a lot of files had to be added. Everything is oriented on the rest of ExaDG and there are no differences in the structure. I know this PR is quite lengthy, but I couldn't imagine a way to further reduce the diff. 

I think the minimal solver should at least be able to calculate errors to ensure everything works as expected. The rest of the postprocessors will be added in additional PRs.

### DIFF
--- a/include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
+++ b/include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
@@ -1,0 +1,384 @@
+/*  ______________________________________________________________________
+ *
+ *  ExaDG - High-Order Discontinuous Galerkin for the Exa-Scale
+ *
+ *  Copyright (C) 2023 by the ExaDG authors
+ *
+ *  This program is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *  ______________________________________________________________________
+ */
+
+// deal.II
+#include <deal.II/numerics/vector_tools.h>
+
+// ExaDG
+#include <exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.h>
+#include <exadg/grid/mapping_dof_vector.h>
+#include <exadg/operators/finite_element.h>
+#include <exadg/operators/quadrature.h>
+#include <exadg/utilities/exceptions.h>
+
+namespace ExaDG
+{
+namespace Acoustics
+{
+template<int dim, typename Number>
+SpatialOperator<dim, Number>::SpatialOperator(
+  std::shared_ptr<Grid<dim> const>               grid_in,
+  std::shared_ptr<dealii::Mapping<dim> const>    mapping_in,
+  std::shared_ptr<BoundaryDescriptor<dim> const> boundary_descriptor_in,
+  std::shared_ptr<FieldFunctions<dim> const>     field_functions_in,
+  Parameters const &                             parameters_in,
+  std::string const &                            field_in,
+  MPI_Comm const &                               mpi_comm_in)
+  : Interface::SpatialOperator<Number>(),
+    grid(grid_in),
+    mapping(mapping_in),
+    boundary_descriptor(boundary_descriptor_in),
+    field_functions(field_functions_in),
+    param(parameters_in),
+    field(field_in),
+    dof_handler_p(*grid_in->triangulation),
+    dof_handler_u(*grid_in->triangulation),
+    mpi_comm(mpi_comm_in),
+    pcout(std::cout, dealii::Utilities::MPI::this_mpi_process(mpi_comm_in) == 0)
+{
+  pcout << std::endl
+        << "Construct acoustic conservation equations operator ..." << std::endl
+        << std::flush;
+
+  initialize_dof_handler_and_constraints();
+
+  pcout << std::endl << "... done!" << std::endl << std::flush;
+}
+
+template<int dim, typename Number>
+void
+SpatialOperator<dim, Number>::fill_matrix_free_data(
+  MatrixFreeData<dim, Number> & matrix_free_data) const
+{
+  // append mapping flags
+  matrix_free_data.append_mapping_flags(Operators::Kernel<dim, Number>::get_mapping_flags());
+
+  // dof handler
+  matrix_free_data.insert_dof_handler(&dof_handler_p, field + dof_index_p);
+  matrix_free_data.insert_dof_handler(&dof_handler_u, field + dof_index_u);
+
+  // constraint
+  matrix_free_data.insert_constraint(&constraint_p, field + dof_index_p);
+  matrix_free_data.insert_constraint(&constraint_u, field + dof_index_u);
+
+  // quadrature for pressure
+  std::shared_ptr<dealii::Quadrature<dim>> quadrature_p =
+    create_quadrature<dim>(param.grid.element_type, param.degree_p + 1);
+  matrix_free_data.insert_quadrature(*quadrature_p, field + quad_index_p);
+
+  // quadrature for velocity
+  std::shared_ptr<dealii::Quadrature<dim>> quadrature_u =
+    create_quadrature<dim>(param.grid.element_type, param.degree_u + 1);
+  matrix_free_data.insert_quadrature(*quadrature_u, field + quad_index_u);
+
+  // quadrature that works for pressure and velocity
+  std::shared_ptr<dealii::Quadrature<dim>> quadrature_p_u =
+    create_quadrature<dim>(param.grid.element_type, std::max(param.degree_p, param.degree_u) + 1);
+  matrix_free_data.insert_quadrature(*quadrature_p_u, field + quad_index_p_u);
+}
+
+template<int dim, typename Number>
+void
+SpatialOperator<dim, Number>::setup()
+{
+  // initialize MatrixFree and MatrixFreeData
+  std::shared_ptr<dealii::MatrixFree<dim, Number>> mf =
+    std::make_shared<dealii::MatrixFree<dim, Number>>();
+  std::shared_ptr<MatrixFreeData<dim, Number>> mf_data =
+    std::make_shared<MatrixFreeData<dim, Number>>();
+
+  fill_matrix_free_data(*mf_data);
+
+  mf->reinit(*get_mapping(),
+             mf_data->get_dof_handler_vector(),
+             mf_data->get_constraint_vector(),
+             mf_data->get_quadrature_vector(),
+             mf_data->data);
+
+  // Subsequently, call the other setup function with MatrixFree/MatrixFreeData objects as
+  // arguments.
+  this->setup(mf, mf_data);
+}
+
+template<int dim, typename Number>
+void
+SpatialOperator<dim, Number>::setup(
+  std::shared_ptr<dealii::MatrixFree<dim, Number> const> matrix_free_in,
+  std::shared_ptr<MatrixFreeData<dim, Number> const>     matrix_free_data_in)
+{
+  pcout << std::endl
+        << "Setup acoustic conservation equations operator ..." << std::endl
+        << std::flush;
+
+  // MatrixFree
+  matrix_free      = matrix_free_in;
+  matrix_free_data = matrix_free_data_in;
+
+  initialize_operators();
+
+  pcout << std::endl << "... done!" << std::endl << std::flush;
+}
+
+template<int dim, typename Number>
+dealii::MatrixFree<dim, Number> const &
+SpatialOperator<dim, Number>::get_matrix_free() const
+{
+  return *matrix_free;
+}
+
+template<int dim, typename Number>
+std::string
+SpatialOperator<dim, Number>::get_dof_name_pressure() const
+{
+  return field + dof_index_p;
+}
+
+template<int dim, typename Number>
+unsigned int
+SpatialOperator<dim, Number>::get_dof_index_pressure() const
+{
+  return matrix_free_data->get_dof_index(get_dof_name_pressure());
+}
+
+template<int dim, typename Number>
+std::string
+SpatialOperator<dim, Number>::get_dof_name_velocity() const
+{
+  return field + dof_index_u;
+}
+
+template<int dim, typename Number>
+unsigned int
+SpatialOperator<dim, Number>::get_dof_index_velocity() const
+{
+  return matrix_free_data->get_dof_index(get_dof_name_velocity());
+}
+
+template<int dim, typename Number>
+unsigned int
+SpatialOperator<dim, Number>::get_quad_index_pressure_velocity() const
+{
+  return matrix_free_data->get_quad_index(field + quad_index_p_u);
+}
+
+template<int dim, typename Number>
+unsigned int
+SpatialOperator<dim, Number>::get_quad_index_pressure() const
+{
+  return matrix_free_data->get_quad_index(field + quad_index_p);
+}
+
+template<int dim, typename Number>
+unsigned int
+SpatialOperator<dim, Number>::get_quad_index_velocity() const
+{
+  return matrix_free_data->get_quad_index(field + quad_index_u);
+}
+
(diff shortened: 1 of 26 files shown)


include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
```diff
@@ -0,0 +1,383 @@
...
+
+  // inverse mass operator velocity
+  {
+    InverseMassOperatorData data;
+    data.dof_index  = get_dof_index_velocity();
+    data.quad_index = get_quad_index_velocity();
+    inverse_mass_velocity.initialize(*matrix_free, data);
+  }
+
+  // acoustic operator pressure
```
nfehn, line 362: why pressure?

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
```diff
@@ -0,0 +1,383 @@
...
+ */
+template<int dim, typename Number>
+void
+SpatialOperator<dim, Number>::initialize_dof_vector(BlockVectorType & dst) const
+{
+  // pressure (1st block) + velocity (2nd block)
+  dst.reinit(2);
+
+  matrix_free->initialize_dof_vector(dst.block(0), get_dof_index_pressure());
+  matrix_free->initialize_dof_vector(dst.block(1), get_dof_index_velocity());
```
nfehn, line 263: I have seen these block-indices {0, 1} quite often in the code, also in the postprocessor. This could be somewhat error prone. We might have done it like this also in the past in other operators. How about introducing public static member variables `SpatialOperator::block_index_pressure` and `SpatialOperator::block_index_velocity`?

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.h
```diff
@@ -0,0 +1,242 @@
...
+   * Operators.
+   */
+
+  // acoustic operator
+  void
+  evaluate_acoustic_operator(BlockVectorType &       dst,
+                             BlockVectorType const & src,
+                             double const            time) const;
+
+  // inverse velocity mass operator
```
nfehn, line 163: why velocity?

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.h
```diff
@@ -0,0 +1,242 @@
...
+  dealii::DoFHandler<dim> dof_handler_u;
+
+  dealii::AffineConstraints<Number> constraint_p, constraint_u;
+
+  std::string const dof_index_p = "pressure";
+  std::string const dof_index_u = "velocity";
+
+  std::string const quad_index_p = "pressure";
+  std::string const quad_index_u = "velocity";
+  // quadrature that works for both, pressure and velocity
```
nfehn, line 217: could you explain what this means or for which purpose this quadrature is needed/used?

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
```diff
@@ -0,0 +1,383 @@
...
+  acoustic_operator.evaluate(dst, src, time);
+}
+
+template<int dim, typename Number>
+void
+SpatialOperator<dim, Number>::apply_inverse_mass_operator(BlockVectorType &       dst,
+                                                          BlockVectorType const & src) const
+{
+  inverse_mass_pressure.apply(dst.block(0), src.block(0));
+  inverse_mass_velocity.apply(dst.block(1), src.block(1));
```
nfehn, line 327: If I remember correctly, we have an `inverse_mass_all` operator in the compressible Navier-Stokes module. The main advantage is probably that in case of deformed elements, geometric mapping data is loaded only once during operator evaluation, and not multiple times for every physical quantity of the block vector.

@kronbichler what do you think? should be introduce such an `inverse_mass_all` also here?

include/exadg/acoustic_conservation_equations/throughput.h
```diff
@@ -0,0 +1,201 @@
...
+
+      return 0;
+    }
+  }
+
+  ExaDG::GeneralParameters                                    general(input_file);
+  ExaDG::HypercubeResolutionParameters                        resolution(input_file, general.dim);
+  ExaDG::ThroughputParameters<ExaDG::Acoustics::OperatorType> throughput(input_file);
+
+  prm.parse_input(input_file, "", true, true);
```
nfehn, line 147: For some reasons, I do not see the declaration of this parameter `prm`?

include/exadg/acoustic_conservation_equations/throughput.h
```diff
@@ -0,0 +1,201 @@
...
+  GeneralParameters general;
+  general.add_parameters(prm);
+
+  HypercubeResolutionParameters resolution;
+  resolution.add_parameters(prm);
+
+  ThroughputParameters<Acoustics::OperatorType> throughput;
+  throughput.add_parameters(prm);
+
+  try
```
nfehn, line 59: Do you remember the reason for this try-catch block or why we do not have the same in the solver file?

I have seen that we also have this block in other modules like `ConvDiff`.

include/exadg/acoustic_conservation_equations/user_interface/parameters.cpp
```diff
@@ -0,0 +1,167 @@
...
+  // TEMPORAL DISCRETIZATION
+  AssertThrow(calculation_of_time_step_size != TimeStepCalculation::Undefined,
+              dealii::ExcMessage("parameter must be defined"));
+
+  if(calculation_of_time_step_size == TimeStepCalculation::UserSpecified)
+    AssertThrow(time_step_size > 0., dealii::ExcMessage("parameter must be defined"));
+
+  if(adaptive_time_stepping)
+    AssertThrow(not(calculation_of_time_step_size == TimeStepCalculation::UserSpecified),
+                dealii::ExcMessage("parameter must be defined"));
```
nfehn, line 83: more precise - for now - could be `AsserThrow(not(adaptive_time_stepping)), ...)`.

include/exadg/acoustic_conservation_equations/user_interface/parameters.cpp
```diff
@@ -0,0 +1,167 @@
...
+#include <exadg/acoustic_conservation_equations/user_interface/parameters.h>
+#include <exadg/utilities/print_functions.h>
+
+namespace ExaDG
+{
+namespace Acoustics
+{
+// standard constructor that initializes parameters
+Parameters::Parameters()
+  : formulation(Formulation::Undefined),
```
nfehn, line 35: maybe add section name `// MATHEMATICAL MODEL` here

nfehn:
Thanks for this first acoustics solver! Looks very good overall!

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
jh66637: I am open for other ways. I just want to mention, that in the compressible INSE module the quantities are not stored in block vectors but in a single vector. The reason why this is not done here, is because we have to be able to access pressure and velocities in aero-acoustics (this can not be seen from this PR). Additionally, it is possible to use mixed orders for pressure and velocity for the acoustic module, which can only be achieved with block vectors (this can also not be seen immediately from this PR).

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.h
jh66637: this is a typo

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.h
jh66637: This might be confusing with the present PR. It is possible to use different orders for pressure and velocity, since we have a merged operator we need a quadrature rule that works for both orders, i.e. `n_q=(k_max+1)^dim`.

include/exadg/acoustic_conservation_equations/throughput.h
jh66637: I didn't understand why this try catch is used. I simply adapted to the rest of the modules. I am happy to remove it if you also do not see a need for it.

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.h
jh66637: [b81e2b0](https://github.com/exadg/exadg/pull/592/commits/b81e2b087d1b4b947127493ddf050e15bb9f59a4)

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
jh66637: [b81e2b0](https://github.com/exadg/exadg/pull/592/commits/b81e2b087d1b4b947127493ddf050e15bb9f59a4)

include/exadg/acoustic_conservation_equations/user_interface/parameters.cpp
jh66637: [2b35e7c](https://github.com/exadg/exadg/pull/592/commits/2b35e7c426c180de938d153403788ba3091e7767)

include/exadg/acoustic_conservation_equations/user_interface/parameters.cpp
jh66637: [2b35e7c](https://github.com/exadg/exadg/pull/592/commits/2b35e7c426c180de938d153403788ba3091e7767)

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
jh66637: Should this be defined in `SpatialOperator`? Dependent, where it is used, this requires "unnecessary" includes of the operator. 
What about introducing something similar to `numbers::invalid_timestep`, i.e. `Acoustics::numbers::block_index_p`?

include/exadg/acoustic_conservation_equations/throughput.h
jh66637: I think I forgot to build a "throughput app" when I liked against this version of `ExaDG`. I think there are two possible ways how to proceed here:

1) add a throughput app (no need to review `throughput.h` again).
2) remove `throughput.h` from this PR and introduce it with an app in a follow-up PR.

Probably it's best to go with 2). What do you think?

jh66637:
@nfehn Thank you for going through all these lines!

include/exadg/acoustic_conservation_equations/user_interface/numbers.h
```diff
@@ -0,0 +1,37 @@
...
+
+namespace ExaDG
+{
+namespace Acoustics
+{
+namespace block
+{
+unsigned int const pressure = 0;
+unsigned int const velocity = 1;
+} // namespace block
```
nfehn, line 33: Why did you make this information public? Isn't it the responsibility of the spatial operator, which also is responsible for the dof-handlers, initialization of dof-vectors, etc.

Also, it is not clear to me why this is part of the user-interface. In terms of the user interface (e.g. creating new acoustics applications), to which extent is it relevant to know these indices? My goal would be to make this a detail of the internal implementation.

The postprocessor might need this information. But there we already have `using AcousticsOperator = SpatialOperator<dim, Number>;`, so one could write `AcousticOperator::block_index_pressure` in the postprocessor.

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
nfehn: I don't think so, see also comment https://github.com/exadg/exadg/pull/592#discussion_r1390835643.

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.h
nfehn: So `quad_index_p/u` is only used for the mass matrix problem?

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
nfehn: I see, so we would have to figure out whether a combined inverse mass matrix is possible with block-vectors? The polynomial degrees might be different, but in case they are the same it could still be a reasonable optimization to have a combined inverse mass matrix. Or do you think the case of different polynomial degrees will be the standard case?

include/exadg/acoustic_conservation_equations/throughput.h
nfehn: Makes sense.

include/exadg/acoustic_conservation_equations/throughput.h
nfehn: Let's consider this a sub-ordinate problem for now, given discussion https://github.com/exadg/exadg/pull/592#discussion_r1389154815.

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.h
jh66637: Currently, yes. However, I plan to add convective terms where `quad_index_p/u` will also be needed.

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
nfehn: `Operators::Operator` also needs this information. But from there, one could pass this information to `SpatialOperator`, where `Operators::Operator` is a member variable.

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
jh66637: Using mixed degrees might yield a large potential for performance optimization while keeping the same convergence rates, but I didn't investigate this in detail.

If it is possible to have an `inverse_mass_all` operator for block vectors without touching `deal.II`, I think we should make use of it. I would suggest filing an issue for this and proceed with the current implementation in the meantime.

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.h
jh66637: I added a comment in [3cdb937](https://github.com/exadg/exadg/pull/592/commits/3cdb93776eac0fb46fe68be7691829cbc5e32e75)

include/exadg/acoustic_conservation_equations/throughput.h
jh66637: [4625de4](https://github.com/exadg/exadg/pull/592/commits/4625de41a8a9300cb76b1a7ec0161711eb723b7b)

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
jh66637: It will also be needed in future convective operators, so I think it should be defined in the `SpatialOperator`. I restructured accordingly and used a forward declaration to avoid cyclic dependencies.

include/exadg/acoustic_conservation_equations/user_interface/numbers.h
jh66637: I restructured accordingly [8b5e62c](https://github.com/exadg/exadg/pull/592/commits/8b5e62cf1ed3e467bd025f27a38c05fc45dfb543)

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
nfehn: Ok. Then it would be more natural in my opinion to inject these parameters into the operator via `OperatorData`. We follow a similar procedure in other modules where one `SpatialOperator` has different `Operators`, which have partial overlap in parameters. Note that all the `Operators` in ExaDG receive dof-index / quad-index in a very similar fashion from the `SpatialOperator` class. The current design with forward declaration is sub-optimal in  my opinion. It is also not really modular: Prinicipally, we want to be able write modified / extended / new `SpatialOperators` using existing `Operators`. With the present solution, `Operators` depend explicitly on one particular `SpatialOperator`, which is against the modular design of arranging existing `Operators` in another `SpatialOperator`.

nfehn:
Looks very good from my side apart from the forward-declaration issue (https://github.com/exadg/exadg/pull/592#discussion_r1391080691).

nfehn (approved):

kronbichler (approved):

jh66637:
Can we proceed here?

nfehn:
I will look at this later today. Personally, I would have preferred to have different commits to see what has been changed compared to the last review.

PR #596, opened 2023-11-08
Move grid and mapping to driver for ConvDiff module

richardschu:
... for the convection diffusion module.
This was discussed in #586 : ownership of grid and mapping should be in driver, such that it can execute the adaptive refinement. In the application's `create_grid` I added now a `Grid &` and `Mapping &`. Getters were removed.

### DIFF
--- a/include/exadg/convection_diffusion/driver.cpp
+++ b/include/exadg/convection_diffusion/driver.cpp
@@ -58,8 +58,7 @@ Driver<dim, Number>::setup()
 
   pcout << std::endl << "Setting up scalar convection-diffusion solver:" << std::endl;
 
-  grid = std::make_shared<Grid<dim>>();
-  application->setup(*grid, mapping);
+  application->setup(grid, mapping);
 
   if(application->get_parameters().ale_formulation) // moving mesh
   {
--- a/include/exadg/convection_diffusion/driver.cpp
+++ b/include/exadg/convection_diffusion/driver.cpp
@@ -58,16 +58,17 @@ Driver<dim, Number>::setup()
 
   pcout << std::endl << "Setting up scalar convection-diffusion solver:" << std::endl;
 
-  application->setup();
+  grid = std::make_shared<Grid<dim>>();
+  application->setup(*grid, mapping);
 
   if(application->get_parameters().ale_formulation) // moving mesh
   {
     std::shared_ptr<dealii::Function<dim>> mesh_motion =
       application->create_mesh_movement_function();
     ale_mapping = std::make_shared<DeformedMappingFunction<dim, Number>>(
-      application->get_mapping(),
+      mapping,
       application->get_parameters().degree,
-      *application->get_grid()->triangulation,
+      *grid->triangulation,
       mesh_motion,
       application->get_parameters().start_time);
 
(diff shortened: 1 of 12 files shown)


include/exadg/convection_diffusion/driver.cpp
```diff
@@ -58,16 +58,17 @@ Driver<dim, Number>::setup()
 
   pcout << std::endl << "Setting up scalar convection-diffusion solver:" << std::endl;
 
-  application->setup();
+  grid = std::make_shared<Grid<dim>>();
```
nfehn, line 61: I think this is an "inconsistency" in the sense that `mapping` is created inside `Application::setup()` (or https://github.com/exadg/exadg/blob/67e0beaa957e174a8b4fb526fdb703ba4aa4331f/include/exadg/grid/grid_utilities.h#L52), while `grid` is created outside. I would prefer to do it uniformly.


I think it would be in line with other code in ExaDG that `create()`-functions also create the object (i.e. call the constructor). Hence, this line should probably be shifted inside the `setup()` function.

include/exadg/convection_diffusion/user_interface/application_base.h
```diff
@@ -92,15 +91,15 @@ class ApplicationBase
     param.check();
     param.print(pcout, "List of parameters:");
 
-    // grid
+    // create grid and mapping owned by driver
     GridUtilities::create_mapping(mapping, param.grid.element_type, param.mapping_degree);
-    create_grid();
-    print_grid_info(pcout, *grid);
+    create_grid(grid, mapping);
```
nfehn, line 96: before this line, `grid = std::make_shared<Grid<dim>>();`

include/exadg/convection_diffusion/user_interface/application_base.h
```diff
@@ -92,15 +91,15 @@ class ApplicationBase
     param.check();
     param.print(pcout, "List of parameters:");
 
-    // grid
+    // create grid and mapping owned by driver
```
nfehn, line 94: just as a minor comment: I think I would not introduce here in the comment the notion `driver` and where this variable comes from. Considering ExaDG as a library, one may use the `ConvDiff::Application` and `Grid` classes in ExaDG or derived projects independently of `ConvDiff::Driver`.

nfehn:
The present PR gives the impression that we should place the mapping object inside `Grid`. I think for problems with static meshes (no ALE) it would be clear to do this. However, the problem is the case with moving meshes (ALE), where we did not find a good solution in the past for a clear "single-responsibility" design (i..e., is `grid->mapping` or `ale_moving_mapping` the relevant one for operator evaluation (MatrixFree)? How to avoid/exclude ambiguities in this context?). I would suggest that we keep this in mind for the future but proceed with the present PR in its current state regarding ownership of mapping.

include/exadg/convection_diffusion/user_interface/application_base.h
nfehn: One could argue to shift this line even inside `create_grid()` to be consistent in terms of the `create()`-nomenclature, but then every concrete application has to write this line. For this reason, I would probably suggest to do it just once in `Application::setup()`.

What do you think?

nfehn: One example where we do it like this would be `Application::create_postprocessor()`, where the object is created inside the function. In this sense, it would be consequent to shift `grid = std::make_shared<Grid<dim>>();` inside `create_grid()`.

richardschu: I also favor keeping `grid = std::make_shared<Grid<dim>>();` outside of `create_grid` just to not repeat it all of the `application.h` even though this is a bit "inconsistent" given the function's name.

nfehn (approved):

PR #599, opened 2023-11-15
Initialization of preconditioners

nfehn:
In general, initialization of preconditioners depends on parameters that change during a simulation (e.g. time step size, linearization vectors). Currently, we completely initialize preconditioners in the setup-phase of a solver. The usual sequence of function calls in the `Driver` during setup is

```C++
spatial_operator->setup();
time_integrator->setup();
spatial_operator->setup_solvers(time_integrator.get_...(), ...);
```
During many discussions, most recently with @richardschu, it turned out that this procedure is quite impractical and I wonder whether we could find a better solution.

So far, the preconditioner already receives automatically the relevant information for preconditioner updates during time-dependent simulations (time steppping) or nonlinear solvers (linearization vector). Hence, we principally already have the functionality to not trigger the preconditioner init/update in the driver, but let it be triggered somewhat "automatically" (typically realized by preconditioners having a reference/pointer to the `underlying_operator`, which is the object that has to be up-to-date and from which parameters can be queried). However, what we would need in my opinion is being able to ask the preconditioner whether it as been initialized so far, such that a Krylov solver can then trigger the initialization of the preconditioner.

I think such a procedure would simplify or clarify the code overall. One downside would be that in terms of compute timings, setup costs in the sense of preconditioner initialization would appear as "solver costs" or "time integration costs", even though the initialization could be done prior to starting the time loop (exactly the procedure we do currently). However, the idea would be that for preconditioners that do not depend on dynamic parameters (say the inverse mass matrix preconditioner), we can do the initialization of the preconditioner already in the constructor. Hence, for these simple cases, `preconditioner.is_initialized()` called in the Krylov solver during the first time step would evaluate to true and nothing would have to be done.

### DIFF
--- a/include/exadg/incompressible_navier_stokes/spatial_discretization/operator_coupled.cpp
+++ b/include/exadg/incompressible_navier_stokes/spatial_discretization/operator_coupled.cpp
@@ -68,26 +68,24 @@ OperatorCoupled<dim, Number>::setup_derived()
 
 template<int dim, typename Number>
 void
-OperatorCoupled<dim, Number>::setup_solvers(double const &     scaling_factor_time_derivative_term,
-                                            VectorType const & velocity)
+OperatorCoupled<dim, Number>::setup_preconditioners_and_solvers()
 {
   this->pcout << std::endl << "Setup incompressible Navier-Stokes solver ..." << std::endl;
 
-  Base::setup_solvers(scaling_factor_time_derivative_term, velocity);
-
-  initialize_block_preconditioner();
-
-  initialize_solver_coupled();
+  Base::setup_preconditioners_and_solvers();
 
   if(this->param.apply_penalty_terms_in_postprocessing_step)
-    this->setup_projection_solver();
+    Base::setup_projection_solver();
+
+  setup_block_preconditioner();
+  setup_solver_coupled();
 
   this->pcout << std::endl << "... done!" << std::endl;
 }
 
 template<int dim, typename Number>
 void
-OperatorCoupled<dim, Number>::initialize_solver_coupled()
+OperatorCoupled<dim, Number>::setup_solver_coupled()
 {
   linear_operator.initialize(*this);
 
@@ -391,7 +389,7 @@ OperatorCoupled<dim, Number>::evaluate_nonlinear_residual_steady(BlockVectorType
 
 template<int dim, typename Number>
 void
-OperatorCoupled<dim, Number>::initialize_block_preconditioner()
+OperatorCoupled<dim, Number>::setup_block_preconditioner()
 {
   block_preconditioner.initialize(this);
 
--- a/include/exadg/incompressible_navier_stokes/spatial_discretization/operator_coupled.cpp
+++ b/include/exadg/incompressible_navier_stokes/spatial_discretization/operator_coupled.cpp
@@ -171,11 +171,13 @@ unsigned int
 OperatorCoupled<dim, Number>::solve_linear_stokes_problem(BlockVectorType &       dst,
                                                           BlockVectorType const & src,
                                                           bool const &   update_preconditioner,
-                                                          double const & time,
                                                           double const & scaling_factor_mass)
 {
-  // Update linear operator
-  linear_operator.update(time, scaling_factor_mass);
+  // Update momentum operator
+  // We do not need to set the time here, because time affects the operator only in the form of
+  // boundary conditions. The result of such boundary condition evaluations is handed over to this
+  // function via the vector src.
+  this->momentum_operator.set_scaling_factor_mass_operator(scaling_factor_mass);
 
   linear_solver->update_preconditioner(update_preconditioner);
 
@@ -213,13 +215,9 @@ OperatorCoupled<dim, Number>::rhs_stokes_problem(BlockVectorType & dst, double c
 template<int dim, typename Number>
 void
 OperatorCoupled<dim, Number>::apply_linearized_problem(BlockVectorType &       dst,
-                                                       BlockVectorType const & src,
-                                                       double const &          time,
-                                                       double const & scaling_factor_mass) const
+                                                       BlockVectorType const & src) const
 {
   // (1,1) block of saddle point matrix
-  this->momentum_operator.set_time(time);
-  this->momentum_operator.set_scaling_factor_mass_operator(scaling_factor_mass);
   this->momentum_operator.vmult(dst.block(0), src.block(0));
 
   // Divergence and continuity penalty operators
@@ -256,8 +254,9 @@ OperatorCoupled<dim, Number>::solve_nonlinear_problem(BlockVectorType &  dst,
   // Update nonlinear operator
   nonlinear_operator.update(rhs_vector, time, scaling_factor_mass);
 
-  // Update linear operator
-  linear_operator.update(time, scaling_factor_mass);
+  // Update linearized momentum operator
+  this->momentum_operator.set_time(time);
+  this->momentum_operator.set_scaling_factor_mass_operator(scaling_factor_mass);
 
   // Solve nonlinear problem
   Newton::UpdateData update;
--- a/include/exadg/incompressible_navier_stokes/spatial_discretization/operator_coupled.cpp
+++ b/include/exadg/incompressible_navier_stokes/spatial_discretization/operator_coupled.cpp
@@ -70,8 +70,6 @@ template<int dim, typename Number>
 void
 OperatorCoupled<dim, Number>::setup_preconditioners_and_solvers()
 {
-  this->pcout << std::endl << "Setup incompressible Navier-Stokes solver ..." << std::endl;
-
   Base::setup_preconditioners_and_solvers();
 
   if(this->param.apply_penalty_terms_in_postprocessing_step)
@@ -79,8 +77,6 @@ OperatorCoupled<dim, Number>::setup_preconditioners_and_solvers()
 
   setup_block_preconditioner();
   setup_solver_coupled();
-
-  this->pcout << std::endl << "... done!" << std::endl;
 }
 
 template<int dim, typename Number>
--- a/include/exadg/incompressible_navier_stokes/spatial_discretization/operator_coupled.cpp
+++ b/include/exadg/incompressible_navier_stokes/spatial_discretization/operator_coupled.cpp
@@ -428,14 +428,15 @@ OperatorCoupled<dim, Number>::initialize_preconditioner_velocity_block()
 
   if(type == MomentumPreconditioner::PointJacobi)
   {
-    preconditioner_momentum = std::make_shared<JacobiPreconditioner<MomentumOperator<dim, Number>>>(
-      this->momentum_operator);
+    preconditioner_momentum =
+      std::make_shared<JacobiPreconditioner<MomentumOperator<dim, Number>>>(this->momentum_operator,
+                                                                            false);
   }
   else if(type == MomentumPreconditioner::BlockJacobi)
   {
     preconditioner_momentum =
       std::make_shared<BlockJacobiPreconditioner<MomentumOperator<dim, Number>>>(
-        this->momentum_operator);
+        this->momentum_operator, false);
   }
   else if(type == MomentumPreconditioner::InverseMassMatrix)
   {
--- a/include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
+++ b/include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
@@ -1633,7 +1633,7 @@ SpatialOperatorBase<dim, Number>::setup_projection_solver()
       // time step size has not been set. Hence, 'update_preconditioner = true' should be used for
       // the Jacobi preconditioner in order to use to correct diagonal for preconditioning.
       preconditioner_projection =
-        std::make_shared<JacobiPreconditioner<ProjOperator>>(*projection_operator);
+        std::make_shared<JacobiPreconditioner<ProjOperator>>(*projection_operator, false);
     }
     else if(param.preconditioner_projection == PreconditionerProjection::BlockJacobi)
     {
@@ -1642,7 +1642,7 @@ SpatialOperatorBase<dim, Number>::setup_projection_solver()
       // size has not been set. Hence, 'update_preconditioner = true' should be used for the Jacobi
       // preconditioner in order to use to correct diagonal blocks for preconditioning.
       preconditioner_projection =
-        std::make_shared<BlockJacobiPreconditioner<ProjOperator>>(*projection_operator);
+        std::make_shared<BlockJacobiPreconditioner<ProjOperator>>(*projection_operator, false);
     }
     else if(param.preconditioner_projection == PreconditionerProjection::Multigrid)
     {
--- a/include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
+++ b/include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
@@ -723,18 +723,9 @@ SpatialOperatorBase<dim, Number>::setup(
   // Finally, do set up of derived classes
   setup_derived();
 
+  setup_preconditioners_and_solvers();
+
   pcout << std::endl << "... done!" << std::endl << std::flush;
-}
-
-template<int dim, typename Number>
-void
-SpatialOperatorBase<dim, Number>::setup_solvers(double const &     scaling_factor_mass,
-                                                VectorType const & velocity)
-{
-  momentum_operator.set_scaling_factor_mass_operator(scaling_factor_mass);
-  momentum_operator.set_velocity_ptr(velocity);
-
-  // remaining setup of preconditioners and solvers is done in derived classes
 }
 
 template<int dim, typename Number>
--- a/include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
+++ b/include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
@@ -1568,6 +1568,17 @@ SpatialOperatorBase<dim, Number>::setup_projection_solver()
                                        projection_operator->get_dof_index(),
                                        projection_operator->get_quad_index());
     }
+    else if(param.preconditioner_projection == PreconditionerProjection::PointJacobi)
+    {
+      typedef Elementwise::JacobiPreconditioner<dim, dim, Number, ProjOperator> JACOBI;
+
+      elementwise_preconditioner_projection =
+        std::make_shared<JACOBI>(projection_operator->get_matrix_free(),
+                                 projection_operator->get_dof_index(),
+                                 projection_operator->get_quad_index(),
+                                 *projection_operator,
+                                 false);
+    }
     else
     {
       AssertThrow(false, dealii::ExcMessage("The specified preconditioner is not implemented."));
--- a/include/exadg/incompressible_navier_stokes/user_interface/enum_types.h
+++ b/include/exadg/incompressible_navier_stokes/user_interface/enum_types.h
@@ -370,6 +370,7 @@ enum class PreconditionerPressurePoisson
 {
   None,
   PointJacobi,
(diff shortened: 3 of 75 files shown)


richardschu (commented):
as discussed, the `update_preconditioner(bool const update_preconditioner) const override` will be removed. Otherwise good to go!

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_base.h
```diff
@@ -51,6 +69,12 @@ class PreconditionerBase
   {
     return std::make_shared<TimerTree>();
   }
+
+private:
+  virtual void
+  do_initialize() = 0;
```
kronbichler, line 75: I guess this function will eventually be `protected` rather than `private`, because we expect the derived classes to override the function?

kronbichler (commented):
I think this can make sense, albeit I would not overengineer things and flexibly see for the needs. I see no problem in solver costs being reported with preconditioner initialization/update, because those can always be annotated more clearly in case of need by changing the source code for the timing purposes (or a suitable longer-term solution).

The question not clear to me yet is what the mechanism to indicate that the preconditioner has become out of date, e.g. if the mesh (or time step) has changed, is supposed to be. We have the `update` functionality that we used before to update the parameters, but the question is whether we would expect the preconditioner to make a full setup (potentially changing vector sizes, parallel partitioning, etc.) during `update` then? Or is the preconditioner then destroyed when a full setup including mesh + dof_handler update is done?

include/exadg/solvers_and_preconditioners/solvers/iterative_solvers_dealii_wrapper.h
```diff
@@ -272,6 +277,11 @@ class SolverGMRES : public SolverBase<VectorType>
     }
     else
     {
+      if(not preconditioner.is_initialized())
+        preconditioner.initialize();
+      else if(update_preconditioner)
+        preconditioner.update();
```
nfehn, line 283: aims to make sure that preconditioner is not initialized and - immediately after this - updated again.

nfehn:
With overengineering, do you refer to the old or to the intended new implementation? The intention of this PR would be to simplify or clarify the code overall. Overengineering is in my understanding needless complexity and should be avoided of course.

You mentioned adaptive mesh refinement: When setting `refine_mesh_every_time_steps = 10` and `update_preconditioner_every_time_steps = 5`, the code would currently do a complete initialization of preconditioners every 10th time step, and do additional updates every 5th time step. This would mean there could be time steps where the diagonal (or Chebyshev smoother setup) are computed twice per time step. I think what we all want is to be able to detect such cases and avoid unnecessary initalization/update work? The aim of this PR is to find a good solution for such problems with a design where the user can understand the code and is able to use it correctly. The if-elseif-solution used here https://github.com/exadg/exadg/pull/599/files#r1394190961 is one suggestion for this problem.

Originally, I did not intend to address in this PR the invalidation of preconditioners (in case something fundamental happens like the partitioning changes). I believe we would indeed need such a mechanism and that the design could look similarly as in the present PR (e.g. an additional member function `set_uninitialized()` setting the private member function `has_been_initialized = false`). I could already implement this here. 

I think we need mechanisms to distinguish between mandatory and optional preconditioner "updates". My current understanding is that

- `initialize()` does a complete initialization, which is necessary in the very beginning and maybe after the grid/partitioning has been changed. The preconditioner is then ready-to-use after this function.
- `update()`: updates the preconditioner after certain parameters (like the time step size) have changed, e.g. recalculate the diagonal.

nfehn:
> as discussed, the update_preconditioner(bool const update_preconditioner) const override will be removed. Otherwise good to go!

It turned out that this is not possible because the current Newton solver implementation needs to be able to trigger a separate preconditioner update.

nfehn:
The motivation for this PR is probably best seen from commits https://github.com/exadg/exadg/pull/599/commits/2d6a0cdfa2240ddd0fefe4c53d6a62eebf518226 and https://github.com/exadg/exadg/pull/599/commits/bd777f78f18f0ec57815424a65331c74bac972e8.

nfehn:
Commit https://github.com/exadg/exadg/pull/599/commits/6e8911bde92963468eecdc7aba0d21b7ceaa02ee fixes a (potential) bug, principally already available in the previous implementation in my opinion.

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_coupled.cpp
```diff
@@ -213,13 +215,9 @@ OperatorCoupled<dim, Number>::rhs_stokes_problem(BlockVectorType & dst, double c
...
 void
 OperatorCoupled<dim, Number>::apply_linearized_problem(BlockVectorType &       dst,
-                                                       BlockVectorType const & src,
-                                                       double const &          time,
-                                                       double const & scaling_factor_mass) const
+                                                       BlockVectorType const & src) const
 {
   // (1,1) block of saddle point matrix
-  this->momentum_operator.set_time(time);
-  this->momentum_operator.set_scaling_factor_mass_operator(scaling_factor_mass);
```
nfehn, line 224: In the previous implementation, `scaling_factor_mass` is only set in `apply_linearized_problem(), which is the `vmult()` function of the Krylov solver. Hence, the preconditioner does not know about this factor.

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_coupled.cpp
```diff
@@ -171,11 +171,13 @@ unsigned int
 OperatorCoupled<dim, Number>::solve_linear_stokes_problem(BlockVectorType &       dst,
                                                           BlockVectorType const & src,
                                                           bool const &   update_preconditioner,
-                                                          double const & time,
                                                           double const & scaling_factor_mass)
 {
-  // Update linear operator
-  linear_operator.update(time, scaling_factor_mass);
```
nfehn, line 180: This line did actually not call `momentum_operator->set_scaling_factor_mass_operator()` in the previous implementation, such that the following line `linear_solver->update_preconditioner()` used a potentially wrong `scaling_factor_mass`. This can have a significant influence on iteration counts.

nfehn:
I think I addressed everything and the code would be ready for review.

kronbichler (approved):
Looks good to me, this is a good simplification.

nfehn:
I realized that for some reasons, certain preconditioning options have so far not been supported for certain modules. I added these options in commit https://github.com/exadg/exadg/pull/599/commits/ddf0704e93a844cd385596224cd2a4adcd5f55e6.

@bergbauer For the Structure module, I added `AdditiveSchwarz`.

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_coupled.cpp
```diff
@@ -258,8 +250,9 @@ OperatorCoupled<dim, Number>::solve_nonlinear_problem(BlockVectorType &  dst,
   // Update nonlinear operator
   nonlinear_operator.update(rhs_vector, time, scaling_factor_mass);
 
-  // Update linear operator
```
richardschu, line 261: I do not get line 262 in the old code - is `linear_operator` the *linearized* operator? Why was the `momentum_operator` not updatedpreviously? Is that the bug you were mentioning before?

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
```diff
@@ -1577,6 +1568,17 @@ SpatialOperatorBase<dim, Number>::setup_projection_solver()
                                        projection_operator->get_dof_index(),
                                        projection_operator->get_quad_index());
     }
+    else if(param.preconditioner_projection == PreconditionerProjection::PointJacobi)
```
richardschu, line 1571: Why is this showing up here? You just added a new elementwise preconditioner since you were at it?

include/exadg/operators/inverse_mass_operator.h
```diff
@@ -277,7 +277,8 @@ class InverseMassOperatorHdiv
     {
       preconditioner =
         std::make_shared<JacobiPreconditioner<MassOperator<dim, n_components, Number>>>(
-          mass_operator);
+          mass_operator, true);
```
richardschu, line 280: '/* initialize_preconditioners */' ?

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -82,13 +83,13 @@ MultigridPreconditionerBase<dim, Number, MultigridNumber>::initialize(
...
+  this->initialize_transfer_operators();
 
-  this->initialize_smoothers();
+  this->initialize_operators();
 
-  this->initialize_coarse_solver(operator_is_singular);
+  this->initialize_smoothers(initialize_preconditioners);
 
-  this->initialize_transfer_operators();
+  this->initialize_coarse_solver(operator_is_singular, initialize_preconditioners);
```
richardschu, line 93: Why is the sequence here changed as well? Does that have a reason?

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -662,6 +665,10 @@ void
 MultigridPreconditionerBase<dim, Number, MultigridNumber>::vmult(VectorType &       dst,
                                                                  VectorType const & src) const
 {
+  AssertThrow(not this->update_needed,
+              dealii::ExcMessage(
+                "Multigrid preconditioner can not be applied because it needs to be updated."));
+
```
richardschu, line 671: good assert!

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -670,6 +677,10 @@ unsigned int
 MultigridPreconditionerBase<dim, Number, MultigridNumber>::solve(VectorType &       dst,
                                                                  VectorType const & src) const
 {
+  AssertThrow(not this->update_needed,
+              dealii::ExcMessage(
+                "Multigrid preconditioner can not be applied because it needs to be updated."));
+
```
richardschu, line 683: good assert!

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -806,21 +820,20 @@ MultigridPreconditionerBase<dim, Number, MultigridNumber>::initialize_coarse_sol
...
     case MultigridCoarseGridSolver::AMG:
     {
-      if(data.coarse_problem.amg_data.amg_type == AMGType::ML)
-      {
-        coarse_grid_solver =
-          std::make_shared<MGCoarseAMG<Operator>>(coarse_operator, data.coarse_problem.amg_data);
-      }
-      else if(data.coarse_problem.amg_data.amg_type == AMGType::BoomerAMG)
+      if(data.coarse_problem.amg_data.amg_type == AMGType::ML or
+         data.coarse_problem.amg_data.amg_type == AMGType::BoomerAMG)
```
richardschu, line 832: :+1:

include/exadg/solvers_and_preconditioners/preconditioners/additive_schwarz_preconditioner.h
```diff
@@ -34,32 +34,41 @@ class AdditiveSchwarzPreconditioner : public PreconditionerBase<typename Operato
...
   void
-  update() final
+  vmult(VectorType & dst, VectorType const & src) const final
   {
-    underlying_operator.compute_factorized_additive_schwarz_matrices();
+    AssertThrow(
+      not this->update_needed,
+      dealii::ExcMessage(
+        "Additive Schwarz preconditioner can not be applied because it needs to be updated."));
+
```
richardschu, line 58: :+1:  better naming and good assert!

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
```diff
@@ -114,23 +113,25 @@ class PreconditionerML : public PreconditionerBase<Number>
...
+  {
+    amg.vmult(dst, src);
+  }
+
   void
   update() override
   {
-    // clear content of matrix since the next calculate_system_matrix-commands add their result
+    // clear content of matrix since calculate_system_matrix() adds the result
     system_matrix *= 0.0;
```
richardschu, line 126: is this cheaper than `=0` or equivalent?

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
```diff
@@ -220,6 +213,21 @@ class PreconditionerBoomerAMG : public PreconditionerBase<Number>
...
 
+  void
+  update() override
+  {
+    // clear content of matrix since the next calculate_system_matrix calls
+    // add their result; since we might run this on a sub-communicator, we
+    // skip the processes that do not participate in the matrix and have size
+    // zero
+    if(system_matrix.m() > 0)
+      system_matrix = 0.0;
```
richardschu, line 224: why not `*= 0` as above?

include/exadg/structure/spatial_discretization/operator.cpp
```diff
@@ -256,13 +256,15 @@ Operator<dim, Number>::setup_operators()
 
     mass_operator.set_scaling_factor(param.density);
 
+    // TODO: this comment can be removed once we merge setup() and setup_solvers()
```
richardschu, line 259: `setup_solver()`

include/exadg/structure/spatial_discretization/operator.cpp
```diff
@@ -373,12 +358,26 @@ Operator<dim, Number>::initialize_preconditioner()
...
     else
     {
       preconditioner = std::make_shared<JacobiPreconditioner<LinearOperator<dim, Number>>>(
-        elasticity_operator_linear);
+        elasticity_operator_linear, false);
+    }
+  }
+  else if(param.preconditioner == Preconditioner::AdditiveSchwarz)
+  {
+    if(param.large_deformation)
```
richardschu, line 371: :+1:  cool to see AdditiveSchwarz added as well!

include/exadg/incompressible_navier_stokes/user_interface/enum_types.h
```diff
@@ -370,6 +370,7 @@ enum class PreconditionerPressurePoisson
 {
   None,
   PointJacobi,
+  BlockJacobi,
```
richardschu, line 373: was this just forgottten?

include/exadg/solvers_and_preconditioners/solvers/iterative_solvers_dealii_wrapper.h
```diff
@@ -122,9 +122,12 @@ class SolverCG : public SolverBase<VectorType>
   void
   update_preconditioner(bool const update_preconditioner) const override
   {
-    if(solver_data.use_preconditioner and update_preconditioner)
+    if(solver_data.use_preconditioner)
     {
-      preconditioner.update();
+      if(preconditioner.needs_update() or update_preconditioner)
```
richardschu, line 127: ok so actually, `needs_update` is only `false` initially and causes a preconditioner update in the very first solver call. Afterwards, we force the update from the linear solver call, right?

richardschu (approved):
Please clearify some questions I added, otherwise good! I am happy that we could remove the complications in the drivers. Very cool!

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_coupled.cpp
nfehn: no, it is exactly the `momentum_operator` that needs it. However, when setting these parameters in the `vmult()` function, the preconditioner can not query this data from the `momentum_operator` during `preconditioner->update()`.

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_coupled.cpp
nfehn: The linear operator is just needed as the class handed over to the Krylov solver as the matrix or operator, providing the function `vmult()`. For our matrix-free implementation, `momentum_operator` is the relevant spatial operator.

The problem or potential bug was that `linear_operator->update()` did not immediately update the `momentum_operator`, which however would have been necessary for a correct update of the preconditioner.

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
nfehn: yes. I also commented in this commit here https://github.com/exadg/exadg/pull/599#issuecomment-1821040787

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
nfehn: Not strictly necessary, but I wanted to clarify the code by first doing the initialization steps not depending on `initialize_preconditioners`, and then do the rest that depends on `initialize_preconditioners`.

include/exadg/solvers_and_preconditioners/solvers/iterative_solvers_dealii_wrapper.h
nfehn: Yes, exactly. Let's put it this way: "The user is the king.", i.e. if the user says the preconditioner should be updated, we will do it. "With great power comes great responsibility", though ;)

include/exadg/incompressible_navier_stokes/user_interface/enum_types.h
nfehn: no, I added this preconditioner variant here. You are right in the sense that this type of preconditioner would have been available before the present PR. The reason is probably historically, i.e. the block-Jacobi preconditioner was added later to ExaDG, while for the pressure Poisson equation, the geometric multigrid preconditioner is our working horse.

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
nfehn: I would suggest to look at this somewhere else, or did I change the code here in this PR?

include/exadg/structure/spatial_discretization/operator.cpp
nfehn: good hint!

nfehn: please check again commit https://github.com/exadg/exadg/pull/599/commits/0b89496c22a68c1db3c494c66299aa90465d7a86

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_coupled.cpp
nfehn: ... because `preconditioner->update()` is called before the code runs into `operator->vmult()` in the Krylov solver.

include/exadg/operators/inverse_mass_operator.h
nfehn: I am somewhat undecided. I  can change it if you want. If so, we should do it consistently (i.e. there are probably also other places).

richardschu: we can do it or not do it, I do not think this is too strict ... but in line 139 above you had it and here in line 280 its not. So decide for one variant and make it consistent in the file :smile:

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
richardschu: no, code was like this before - I was just wondering.
We could open up an issue and see what the others think?

richardschu (approved):
thanks for addressing my questions, check out the comments: one regarding consistency of `false /* initialize_precodnitioner */` and one regarding the potential issue on `matrix = 0;` or `matrix *= 0;`
This PR is ready to merge :100:

PR #600, opened 2023-11-15
optimize Jacobi preconditioner implementation

### LINKED ISSUE #598: Performance of JacobiPreconditioner
I wonder if the current implementation of the Jacobi preconditioner is optimal
https://github.com/exadg/exadg/blob/51a6a7e65e81c4a3816367da24f36a4d8fc813b9/include/exadg/solvers_and_preconditioners/preconditioners/jacobi_preconditioner.h#L48-L54

If the vectors `dst`, `src` are not the same, we load the `dst` vector twice from main memory, right? Due to read-for-ownership, the data transfer would be 6 vectors (src, 2x dst, diagonal, 2x dst) in total in instead of 4 vectors in the optimal case (src, diagonal, 2x dst), right? So I am thinking of
```C++
  vmult(VectorType & dst, VectorType const & src) const final
  {
    if(dealii::PointerComparison::equal(&dst, &src))
        dst.scale(inverse_diagonal);
    else
        dst.elementwise_product(inverse_diagonal, src);
  }
```
I am not sure if deal.II provides something like `elementwise_product()`.

@kronbichler It would be great if you could comment on this.

nfehn:
closes #598 

(I did not insert a pragma because we have this type of loop multiple times in ExaDG and I think we (nowhere) write such a pragma. So I think we should do it everywhere consistently. Overall, such functionality should be free functions in deal.II (where the implementation also cares about pragmas).)

### DIFF
--- a/include/exadg/solvers_and_preconditioners/preconditioners/jacobi_preconditioner.h
+++ b/include/exadg/solvers_and_preconditioners/preconditioners/jacobi_preconditioner.h
@@ -47,9 +47,15 @@ public:
   void
   vmult(VectorType & dst, VectorType const & src) const final
   {
-    if(not dealii::PointerComparison::equal(&dst, &src))
-      dst = src;
-    dst.scale(inverse_diagonal);
+    if(dealii::PointerComparison::equal(&dst, &src))
+    {
+      dst.scale(inverse_diagonal);
+    }
+    else
+    {
+      for(unsigned int i = 0; i < dst.local_size(); ++i)
+        dst.local_element(i) = inverse_diagonal.local_element(i) * src.local_element(i);
+    }
   }
 
   void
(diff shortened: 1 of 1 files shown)


kronbichler:
> Overall, such functionality should be free functions in deal.II (where the implementation also cares about pragmas).

I don't think the separation between doing things in deal.II or ExaDG is along the definition of the pragma or not (which faciliates the generation of good machine code for the compiler by avoiding the check for some overlapping and similar assumptions, and in turn enable vectorization in more circumstances), but how generic an interface would be and how often it is used. If we do certain things only a few times inside ExaDG, I see ExaDG as the right place to do it, irrespective the optimization level. There is one caveat to this statement: If we would go towards GPUs, we would need to write this loop differently, but that is something to be addressed specifically when making the transition for a concrete problem.

peterrum:
https://github.com/dealii/dealii/blob/c379a048f1fc7e67f3344ac7e763026df3bea0c3/include/deal.II/lac/diagonal_matrix.h#L442C29-L442C45

PR #602, opened 2023-11-17
allow update of inverse mass operator and refactor block-Jacobi implementation

### LINKED ISSUE #601: Check consistency of InverseMassOperator for ALE problems
In some cases, we use a block-Jacobi preconditioner to represent the action of the inverse mass operator for DG methods (e.g. simplex elements, but possibly also for hexahedral elements), see https://github.com/exadg/exadg/blob/ac70ca39cc205cfdf1e5d1ecb7a8211ad4e119e8/include/exadg/operators/inverse_mass_operator.h#L106-L140

However, for problems with moving meshes (ALE), the block matrices are no longer correct since they are not updated over time in the current implementation (and there are also no mechanisms in the code that check this), so that it might happen that the inverse mass operator (which has to be the correct or exact inverse of the mass operator) is wrong.

@kronbichler @richardschu do you agree?

nfehn:
closes #601

In the past, we probably did not choose parameter combinations (ALE problems with inverse mass realized as matrix-based block-Jacobi preconditioner) that would have led to problems. The present PR aims to fix this potential bug.

@kronbichler @richardschu Someone interested in reviewing this PR? I would like to merge the present PR before continuing with #599.

### DIFF
--- a/include/exadg/operators/operator_base.cpp
+++ b/include/exadg/operators/operator_base.cpp
@@ -42,7 +42,6 @@ OperatorBase<dim, Number, n_components>::OperatorBase()
     is_dg(true),
     data(OperatorBaseData()),
     level(dealii::numbers::invalid_unsigned_int),
-    block_diagonal_preconditioner_is_initialized(false),
     n_mpi_processes(0)
 {
 }
@@ -462,32 +461,6 @@ OperatorBase<dim, Number, n_components>::add_diagonal(VectorType & diagonal) con
 
 template<int dim, typename Number, int n_components>
 void
-OperatorBase<dim, Number, n_components>::calculate_block_diagonal_matrices() const
-{
-  AssertThrow(is_dg, dealii::ExcMessage("Block Jacobi only implemented for DG!"));
-
-  // allocate memory only the first time
-  if(not(block_diagonal_preconditioner_is_initialized) or
-     matrix_free->n_cell_batches() * vectorization_length != matrices.size())
-  {
-    auto dofs =
-      matrix_free->get_shape_info(this->data.dof_index).dofs_per_component_on_cell * n_components;
-
-    matrices.resize(matrix_free->n_cell_batches() * vectorization_length, LAPACKMatrix(dofs, dofs));
-
-    block_diagonal_preconditioner_is_initialized = true;
-  }
-  // else: reuse old memory
-
-  // clear matrices
-  initialize_block_jacobi_matrices_with_zero(matrices);
-
-  // compute block matrices
-  add_block_diagonal_matrices(matrices);
-}
-
-template<int dim, typename Number, int n_components>
-void
 OperatorBase<dim, Number, n_components>::add_block_diagonal_matrices(
   std::vector<LAPACKMatrix> & matrices) const
 {
@@ -519,19 +492,6 @@ OperatorBase<dim, Number, n_components>::add_block_diagonal_matrices(
   {
     matrix_free->cell_loop(&This::cell_loop_block_diagonal, this, matrices, matrices);
   }
-}
-
-template<int dim, typename Number, int n_components>
-void
-OperatorBase<dim, Number, n_components>::apply_block_diagonal_matrix_based(
-  VectorType &       dst,
-  VectorType const & src) const
-{
-  AssertThrow(is_dg, dealii::ExcMessage("Block Jacobi only implemented for DG!"));
-  AssertThrow(block_diagonal_preconditioner_is_initialized,
-              dealii::ExcMessage("Block Jacobi matrices have not been initialized!"));
-
-  matrix_free->cell_loop(&This::cell_loop_apply_block_diagonal_matrix_based, this, dst, src);
 }
 
 template<int dim, typename Number, int n_components>
@@ -573,8 +533,6 @@ OperatorBase<dim, Number, n_components>::apply_inverse_block_diagonal_matrix_bas
   VectorType const & src) const
 {
   AssertThrow(is_dg, dealii::ExcMessage("Block Jacobi only implemented for DG!"));
-  AssertThrow(block_diagonal_preconditioner_is_initialized,
-              dealii::ExcMessage("Block Jacobi matrices have not been initialized!"));
 
   matrix_free->cell_loop(&This::cell_loop_apply_inverse_block_diagonal_matrix_based,
                          this,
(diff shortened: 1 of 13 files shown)


include/exadg/solvers_and_preconditioners/preconditioners/elementwise_preconditioners.h
```diff
@@ -192,6 +207,13 @@ class InverseMassPreconditioner
     integrator->reinit(cell);
   }
 
+  void
+  update() final
+  {
+    // no updates needed as long as the MatrixFree/Integrator object is up-to-date (which is not the
+    // responsibility of the present class).
+  }
```
nfehn, line 215: @kronbichler A few lines above in the constructor, there is the code

 ```C++
integrator = std::make_shared<Integrator>(matrix_free, dof_index, quad_index);
inverse = std::make_shared<CellwiseInverseMass>(*integrator);
```

I wonder whether these data structures are still correct if the mesh has been moved (with corresponding updates of matrix_free) in the background, or whether the constructor of `dealii::CellwiseInverseMass` would have to be called again? So it might be that the comment introduced for now is actually not correct.

Note that this is the `Elementwise::InverseMassPreconditioner`. For the "normal" or global `InverseMassOperator`, we create the `dealii::CellwiseInverseMass` inside the `cell_loop()`. We probably need to do this in a different way here in the `Elementwise` case, because we would otherwise create the object in every iteration of the elementwise Krylov solver (which is called for every cell). However, we could do some update inside the "global" `update()` function introduced in this PR, which is called once for all the cells (while the function `setup(cell)` then performs the setup related to a cell before the elementwise Krylov solver is called for every cell).

File: include/exadg/operators/operator_base.h [hunk 2]
```diff
@@ -320,6 +318,19 @@ class OperatorBase : public dealii::Subscriptor
   void
   initialize_block_diagonal_preconditioner_matrix_free() const;
 
+  // updates block diagonal preconditioner when using the matrix-free variant with elementwise
+  // iterative solvers and matrix-free operator evaluation.
+  void
+  update_block_diagonal_preconditioner_matrix_free() const;
+
+  // initializes block diagonal preconditioner for matrix-based variant
```
richardschu, line 326: lines 326 & 328: Can we add something like "using cell-wise solves." to make it a bite more verbose?

include/exadg/operators/operator_base.h
```diff
@@ -715,13 +716,6 @@ class OperatorBase : public dealii::Subscriptor
    */
   mutable VectorType weights;
 
-  /*
-   * We want to initialize the block diagonal preconditioner (block diagonal matrices or elementwise
-   * iterative solvers in case of matrix-free implementation) only once, so we store the status of
-   * initialization in a variable.
-   */
-  mutable bool block_diagonal_preconditioner_is_initialized;
```
richardschu, line 723: Before, we had this logic where we potentially initialize when calling `update`. So, the preconditioner was initialized when being updated. Now, we initialize in the constructor and later only update, hence do not need to keep track of the state using `preconditioner_is_initialized`, checking this variable and potentially initialize.

Two questions here:
i) Is the `underlying_operator` in all cases ready at the time we setup the preconditioner following the current strategy (remember, we were discussing the scaling factors and so on)? Before this PR, there *could* -- I am not saying there were, but there might be -- some changes to the `underlying_operator` inbetween preconditioner setup and calling the former `update` including the initialization step.
ii) In the parallel PR #599, we actually introduce the strategy of initializing when calling update if necessary (what we had here before this PR). I think we should keep this aspect of the previous implementation, but **do split** into `update` and `initialize` steps as done here. In your opinion, would this only needlessly complicate matters here, since it is always clear that the preconditioner has been initialized after calling the constructor, right?

These questions relate to the differences between the preconditioner and the solver and why we go different routes within ExaDG for each of these. It would be more flexible, if we would not need the `underlying_operator` to be initialized when constructing the preconditioner. The question is, if this is overengineered was introduced previously for a good reason?

include/exadg/operators/operator_base.h
nfehn: this is actually the matrix-based variant, where "solves" might be irritating as well, or at least not precise enough to differentiate from "cell-wise iterative solves"?

include/exadg/operators/operator_base.h
nfehn: I did not fully understand this. Where do you think did something change in the logic?

So far, we need the underlying operator to be initialized before creating/initializing the preconditioner. I think the present PR does not change this, and also should not change this because this is to be addressed separately.

include/exadg/solvers_and_preconditioners/preconditioners/block_jacobi_preconditioner.h
```diff
@@ -36,7 +36,7 @@ class BlockJacobiPreconditioner : public PreconditionerBase<typename Operator::v
     : underlying_operator(underlying_operator_in)
   {
     // initialize block Jacobi
-    underlying_operator.update_block_diagonal_preconditioner();
+    underlying_operator.initialize_block_diagonal_preconditioner();
```
nfehn, line 39: I think this is the important line of code related to this comment https://github.com/exadg/exadg/pull/602#discussion_r1398171895. I thought it is an advantage to get rid of the variable `OperatorBase::block_jacobi_preconditioner_is_initialized`.

In the long run, the responsibility should be shifted even more strongly to the `BlockJacobiPreconditioner` (the relevant data structures currently live in `OperatorBase`). I discussed this a lot with @bergbauer, but the solution is not that clear/easy and we still have to work on this.

include/exadg/operators/operator_base.h
richardschu: Ah, so the matrix-based variant always does `calculate_lu_factorization_block_jacobi(matrices);`, while the matrix-free application of the inverse mass matrix *and* the element Krylov solvers both follow the matrix-free path?
... then you are right ... maybe add "using cell-wise LU decomposition"?

include/exadg/operators/operator_base.h
richardschu: The logic isnt changed in this PR, but we had that variable `preconditioner_is_initialized`, which would allow similar logic as is the goal in #602, isn't it? Wouldn't that mean we remove the variable here and introduce it again later?

include/exadg/solvers_and_preconditioners/preconditioners/block_jacobi_preconditioner.h
richardschu: Yes I have seen this. I think the current behavior is fine, I was just thinking about #602, where we aim to trigger updates in the solver. There, one motivation was to not be dependent on construction sequences of solver/preconditioner/operator etc to have all the scaling factors for mass and stiffness matrices avaible when needed. This is just something we should have in the back of our heads here ... probably it just belongs to  that separate PR #602 .

nfehn: Agree, but - for clarifty - let me mention it should be PR #599 where this is done.

include/exadg/solvers_and_preconditioners/preconditioners/elementwise_preconditioners.h
nfehn: @kronbichler what do you think on this topic? Should I wait merging this PR?

kronbichler: Regarding correctness, the present code should be fine because `Integrator` (aka `FEEvaluation`) and `CellwiseInverseMatrix` do not hold a state regarding the geometry stored inside `MatrixFree`/`MappingInfo`, apart from pointers into these data structures. As long as only `MappingInfo::update_mapping()` (rather than `initialize()`) is called, these pointers should remain valid.

Regarding the general code design, it would be desirable to set up the temporary data such as `FEEvaluation` inside the loop over cells, where the call-back is made into the user code, and ideally have no `FEEvaluation` alive at all outside a matrix-free loop. This is because allowing objects to release their memory after use makes better use of data caches and, probably more importantly, avoids traversing long lists of data for use in `FEEvaluation` where the data is acquired (look at https://github.com/dealii/dealii/blob/4f8abe489f7ec609b88cfdf0d865cf1c9ee43b9c/include/deal.II/matrix_free/fe_evaluation.h#L2861 that calls https://github.com/dealii/dealii/blob/4f8abe489f7ec609b88cfdf0d865cf1c9ee43b9c/include/deal.II/matrix_free/matrix_free.h#L3039-L3052). Ideally, that list's length should be equal to the maximum of integrators present in a single cell or face work context. In that sense, every use case like the one via `shared_ptr` that persists across many calls is a use case we should strive to avoid. However, this is a long-standing design issue not related to the present change, and it would only be marginally be resolved by finding a new design here. However, we need to figure out a solution in a general re-design in the matrix-free loops that also includes how the deal.II functions are designed and how the call-backs look like. (I keep repeating this argument, so it should be well-known.)

include/exadg/operators/operator_base.cpp
```diff
@@ -627,6 +585,40 @@ OperatorBase<dim, Number, n_components>::initialize_block_diagonal_preconditione
...
+void
+OperatorBase<dim, Number, n_components>::update_block_diagonal_preconditioner_matrix_based() const
+{
+  // clear matrices
+  initialize_block_jacobi_matrices_with_zero(matrices);
+
+  // compute block matrices and add
+  add_block_diagonal_matrices(matrices);
+
+  calculate_lu_factorization_block_jacobi(matrices);
```
kronbichler, line 619: It has been like this before the PR, but in general we should aim to combine the initialization, computation/add, and factorization of a single matrix at a time, rather than doing it globally. In accordance with my other comment, this hints at what kind of data structures we might want to prefer in a future design.

kronbichler (approved):

PR #605, opened 2023-11-22
Add acoustic output generator

jh66637:
This PR adds an output generator to be able to visualize acoustic pressure and acoustic particle velocity.


nfehn (approved):

nfehn:
We can think about generalizing the `OutputGenerator` implementation for the future, because the code looks similar or redundant for the different modules. However, I think this is not necessary for the present PR.

PR #608, opened 2023-11-23
shift mapping/coarse_mappings to Grid class?

nfehn:
Discussion with @richardschu on adaptive mesh refinement (#586) let us rethink whether we should shift `mapping` and `coarse_mappings` to the `Grid` class. Note that the `Grid` and `Mapping` objects have been shifted recently to the `Driver` class for the `ConvDiff` module in PR #596, the motivation behind again being adaptive mesh refinement. Further, note that `coarse_triangulations` is already a member variable of `Grid`. `coarse_mappings` currently lives in `MultigridPreconditioner`. However, several multigrid preconditioners might require the same `coarse_mappings`. So the idea would be to have this variable once at a central place. In case of grid manipulations (like ALE and adaptive mesh refinement), we further believe to have better control over these data structures in the `Driver` class (which owns these objects) rather than somewhat deeper in the code e.g. in the `MultigridPreconditioner`.

The first commit shows the basic idea for the `ConvDiff` module (which supports e.g. also ALE functionality), where I did not touch `coarse_mappings` so far. Of course, function calls like `setup(grid, grid->mapping)` can now be simplified to `setup(grid)`.

### DIFF
--- a/applications/convection_diffusion/boundary_layer/application.h
+++ b/applications/convection_diffusion/boundary_layer/application.h
@@ -126,10 +126,8 @@ private:
   }
 
   void
-  create_grid(Grid<dim> & grid, std::shared_ptr<dealii::Mapping<dim>> & mapping) final
+  create_grid(Grid<dim> & grid) final
   {
-    (void)mapping;
-
     auto const lambda_create_triangulation =
       [&](dealii::Triangulation<dim, dim> &                        tria,
           std::vector<dealii::GridTools::PeriodicFacePair<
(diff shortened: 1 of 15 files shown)


nfehn:
The problem with shifting mapping/coarse_mappings into the `Grid` class is that in case of ALE-problems based on PDE-based grid deformation (Poisson-type or elasticity-type grid deformation), we also need to remember the undeformed mappings for the ALE grid deformation solver. The spatial operators of these PDE solvers rely on the `Grid` class. When using two grid objects (for the undeformed state and the deformed state) we have, however, a redundancy in the triangulation objects (which are the same). This implies in my opinion additional potential difficulties / misunderstandings for adaptive mesh refinement, where we manipulate the triangulations. The code might then become hard to understand.

PR #612, opened 2023-11-30
add asserts related to fullydistributed::Triangulation

nfehn:
I expect we would get wrong results when using `fullydistributed::Triangulation` in combination with manifolds or periodic faces. This problem has been documented in deal.II' issues https://github.com/dealii/dealii/issues/16312 and https://github.com/dealii/dealii/issues/16313.

I decided to add asserts for now, given that it is unclear how to fix these problems. In particular, Unless for simple problems (like hypercube grids), I do not see a general solution to apply the manifolds afterwards, i.e. after the fullydistributed triangulation has been created via `deal.II` functionality like `create_description_from_triangulation_in_groups()` functionality.

The first commit add asserts for the poisson/sine test case only. Many more problematic applications exist in ExaDG. Please check if you agree with the conditions and ExcMessage's of the added asserts.

I will then go through ExaDG and apply similar changes.

### DIFF
--- a/applications/acoustic_conservation_equations/throughput/application.h
+++ b/applications/acoustic_conservation_equations/throughput/application.h
@@ -86,24 +86,41 @@ private:
   {
     (void)mapping;
 
-    auto const lambda_create_triangulation =
-      [&](dealii::Triangulation<dim, dim> &                        tria,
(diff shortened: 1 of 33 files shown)


nfehn:
I think all applications have been addressed now.

PR #614, opened 2023-12-01
Invalid coarse mappings in multigrid

nfehn:
For the `poisson/sine` test case, I added a new "boundary-layer" manifold (typical of CFD applications to resolve turbulent boundary layers around bodies), which is described analytically by a `tanh()` function in wall-normal direction and a `dealii::ChartManifold`, and approximated in the numerical simulation by a quadratic mapping.

Then, using our default multigrid implementation with the same mapping degree on all h-levels (equivalent to the fine level mapping degree), will lead to invalid Jacobians on coarser multigrid levels, which is shown by the present test case. For the chosen parameters and the chosen grid stretch factor, refine level 2 (set in the input.json file) leads to a valid mapping on the fine grid (level 2), while coarser grids (level 1) have an invalid mapping.

The present PR proposes a test case that fails on purpose. Potential solutions to this problem are to be addressed by upcoming PRs.


applications/poisson/sine/application.h
```diff
@@ -269,10 +263,30 @@ class Application : public ApplicationBase<dim, n_components, Number>
...
         else
         {
           AssertThrow(false, dealii::ExcMessage("not implemented."));
         }
+
+        if(vector_local_refinements.size() > 0)
+          refine_local(tria, vector_local_refinements);
+
+        if(global_refinements > 0)
+          tria.refine_global(global_refinements);
```
nfehn, line 289: I shifted these lines of code.

@kronbichler I think it should be possible to first apply the manifolds (where we call push_forward on all the triangulation vertices) and then do the global or local refinements. Nevertheless, could you briefly look at this change?

nfehn:
With deal.II PR https://github.com/dealii/dealii/pull/16317, an assert is now triggered in matrix_free/mapping_info.templates.h.

applications/poisson/sine/application.h
```diff
@@ -269,10 +263,30 @@ class Application : public ApplicationBase<dim, n_components, Number>
...
+        {
+          dealii::Tensor<1, dim> dimensions;
+          for(unsigned int d = 0; d < dim; ++d)
+          {
+            dimensions[d] = right - left;
+          }
+
+          double const grid_stretch_factor = 2.8;
+
+          static BoundaryLayerManifold<dim> manifold(dimensions, grid_stretch_factor);
```
kronbichler, line 276: I think we do not need to make the manifolds static any more (that used to be needed before the manifold redesign in 2019/2020 I believe). So to limit the number of static variables, I would suggest to write
```suggestion
          BoundaryLayerManifold<dim> manifold(dimensions, grid_stretch_factor);
```
(I realize there are other places, too, but I guess we can start working in that direction.)

applications/poisson/sine/application.h
kronbichler: Do you mean in the context of what we discussed last week would handle, i.e., if we copy a triangulation for fully distributed triangulations, the manifolds should be kept alive and make sure to reposition points? On general levels and ignoring the creation of additional meshes for now, the answer would be yes, the manifold is globally defined in terms of the global mesh, so it should be fine to let it handle the deformation when creating new points.

nfehn: I don't think this will solve the problem that manifolds "get lost". I suggested this as an alternative implementation for the distributed::Tria case. The fullydistributed::Tria case would be asserted for now, see PR https://github.com/exadg/exadg/pull/612, which I consider still active.

PR #619, opened 2023-12-12
Add RHS for acoustic module

jh66637:
To be able to solve realistic acoustic problems, we need to be able to consider source terms in the mass conservation equation. This PR introduces this functionality and adds a test case with RHS for which the analytical solution is known. 
Additionally, a test is added that checks CFL time-stepping, the application of the RHS, and velocity Dirichlet BCs.

The PR is dependent on https://github.com/exadg/exadg/pull/619.

The acoustic conservation equations read:
```math
\frac{\partial p}{\partial t} + \rho c^2 \nabla\cdot\mathbf{u} = c^2 f 
```
```math
\frac{\partial \mathbf{u}}{\partial t}+\frac{1}{\rho}\nabla p =\mathbf{0}
```

Alternative
```math
 \frac{1}{c^2}\frac{\partial p}{\partial t} + \rho \nabla\cdot\mathbf{u} = f 
```
```math
\rho\frac{\partial \mathbf{u}}{\partial t}+\nabla p =\mathbf{0}
```

### DIFF
--- a/applications/acoustic_conservation_equations/circular_moving_gauss_pulse/tests/cfl_rhs_velocity_dirichlet.json
+++ b/applications/acoustic_conservation_equations/circular_moving_gauss_pulse/tests/cfl_rhs_velocity_dirichlet.json
@@ -1,0 +1,25 @@
+{
+    "General": {
+        "Precision": "double",
+        "Dim": "2",
+        "IsTest": "true"
+    },
+    "SpatialResolution": {
+        "DegreeMin": "5",
+        "DegreeMax": "5",
+        "RefineSpaceMin": "3",
+        "RefineSpaceMax": "3"
+    },
+    "TemporalResolution": {
+        "RefineTimeMin": "0",
+        "RefineTimeMax": "0"
+    },
+    "Application": {
+        "RuntimeInNumberOfPulseRotations": "0.1"
+    },
+    "Output": {
+        "OutputDirectory": "output/circular_moving_gauss_pulse/",
+        "OutputName": "circular_moving_gauss_pulse",
+        "WriteOutput": "false"
+    }
+}
--- a/include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
+++ b/include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
@@ -72,6 +72,10 @@ SpatialOperator<dim, Number>::fill_matrix_free_data(
   // append mapping flags
   matrix_free_data.append_mapping_flags(Operators::Kernel<dim, Number>::get_mapping_flags());
 
+  if(param.right_hand_side)
+    matrix_free_data.append_mapping_flags(
+      ExaDG::Operators::RHSKernel<dim, Number>::get_mapping_flags());
+
   // mapping flags required for CFL condition
   if(param.calculation_of_time_step_size == TimeStepCalculation::CFL)
   {
@@ -314,6 +318,9 @@ SpatialOperator<dim, Number>::evaluate(BlockVectorType &       dst,
   // shift to the right-hand side of the equation
   dst *= -1.0;
 
+  if(param.right_hand_side)
+    rhs_operator.evaluate_add(dst.block(block_index_pressure), time);
+
   apply_scaled_inverse_mass_operator(dst, dst);
 }
 
@@ -427,6 +434,16 @@ SpatialOperator<dim, Number>::initialize_operators()
     data.formulation          = param.formulation;
     data.bc                   = boundary_descriptor;
     acoustic_operator.initialize(*matrix_free, data);
+  }
+
+  // rhs operator
+  if(param.right_hand_side)
+  {
+    RHSOperatorData<dim> data;
+    data.dof_index     = get_dof_index_pressure();
+    data.quad_index    = get_quad_index_pressure();
+    data.kernel_data.f = field_functions->right_hand_side;
+    rhs_operator.initialize(*matrix_free, data);
   }
 }
 
--- a/include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.h
+++ b/include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.h
@@ -36,6 +36,7 @@
 #include <exadg/grid/grid.h>
 #include <exadg/matrix_free/matrix_free_data.h>
 #include <exadg/operators/inverse_mass_operator.h>
+#include <exadg/operators/rhs_operator.h>
 
 namespace ExaDG
 {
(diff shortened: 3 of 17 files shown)


applications/acoustic_conservation_equations/circular_moving_gauss_pulse/tests/cfl_rhs_velocity_dirichlet.json
```diff
@@ -0,0 +1,25 @@
+{
```
jh66637, line 1: The test takes approximately 4 seconds on my laptop.

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
```diff
@@ -306,6 +310,11 @@ SpatialOperator<dim, Number>::evaluate(BlockVectorType &       dst,
   // shift to the right-hand side of the equation
   dst *= -1.0;
 
+  if(param.right_hand_side)
+    rhs_operator.evaluate_scale_add(dst.block(block_index_pressure),
+                                    param.speed_of_sound * param.speed_of_sound,
+                                    time);
+
```
nfehn, line 317: From a design perspective, I am not sure whether we should do it like this. I first saw the definition of `evaluate_scale_add()` and expected the scaling factor to be a dynamic parameter, dynamic in the sense that it depends on the time step size for example (as we have for the mass operator). However, for the rhs-operator, which by definition operates on a `dealii::Function`, one might argue this scaling factor should appear inside `dealii::Function`.

Another reason is that the spatial derivative operators for acoustics are not written in `evaluate_scale()` logic, despite some terms are multiplied during operator evaluation by `c*c`. Hence, trying to understand the strong PDE model problem behind the code by just looking at the code (as I do right now), raises some questions/inconsistencies in my opinion.

include/exadg/acoustic_conservation_equations/user_interface/field_functions.h
```diff
@@ -40,6 +40,12 @@ struct FieldFunctions
    * beginning of the simulation.
    */
   std::shared_ptr<dealii::Function<dim>> initial_solution_velocity;
+
+  /*
+   * The function right_hand_side is used to evaluate the acoustic source term on the right-hand
+   * side of the mass conservation equation of the acoustic conservation equations.
+   */
+  std::shared_ptr<dealii::Function<dim>> right_hand_side;
```
nfehn, line 48: Somewhat related to this comment https://github.com/exadg/exadg/pull/619/files#r1426708752, this is not fully clear to me. It should be stated explicitly which unknowns/dofs (velocity vs. pressure) this term operates on.

The documentation should also be clear about which formulation we consider the strong form with right-hand side "f" and whether this right-hand side "f" applies `c*c` internally or needs to be scaled externally.

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.h
```diff
@@ -241,6 +242,11 @@ class SpatialOperator : public Interface::SpatialOperator<Number>
   InverseMassOperator<dim, 1, Number>   inverse_mass_pressure;
   InverseMassOperator<dim, dim, Number> inverse_mass_velocity;
 
+  /*
+   * RHS operator
+   */
```
nfehn, line 247: More information here, e.g. explaining `1`, similar to comment https://github.com/exadg/exadg/pull/619/files#r1426712232 would be helpful here.

nfehn:
In the description of this PR, could you provide the strong form of the PDE model problem including the right hand side? (Is there a scaling factor in front of the time derivative terms? is the right-hand side of the equations just "f" or some scaled version of "f"?) I think all my comments are related to this aspect.

jh66637:
I added the equations to the PR description :)

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
jh66637: Would you prefer to add a scaling factor to `RHSKernelData`?

include/exadg/acoustic_conservation_equations/user_interface/field_functions.h
jh66637: I will extend the documentation once we decided how to proceed.

nfehn:
Thanks for the extended description!

I assume this is the common notation in literature?

I thought about the different variants to implement this and one can probably justify each of them. So I think in the end we need a good documentation.

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
nfehn: This would have the advantage that we do not have to extend the interface of `RHSOperator` by a new function, and modify the existing interface of `RHSKernel`. In the current implementation, the kernel always multiplies by `scaling_factor` (even if we do not want to scale), so a `RHSKernelData::scaling_factor` initialized with `1.0` would probably be more straight forward.

nfehn:
Let me add a few more infos so that the performance aspects we talk about are more explicit: From a performance perspective, this change implies we do more flops for the rhs-operator (whenever it is used), while we avoid streaming  a typically large DoF-vector once more from memory (in those cases where we want to scale it). I think one can argue that the latter is more likely the bottleneck (i.e. the memory bandwidth), and that one additional multiplication done generally does not hurt.

jh66637:
The RHS should remain the same (i.e., $`f`$ should be given without the scaling factor) to be consistent with the dimensions if the RHS is an aeroacoustic source term. However, I think we could also shift the scaling to the temporal derivative. Avoiding to stream the vector again from memory, and only perform the additional flop when needed can then be achieved by a matrix-free cell loop that takes `operation_after_loop` as an argument in an `apply_scale()` routine. This implies again that we write the scaling factor at the same point as it is currently the case for the RHS. 

Streaming the vector again from memory to scale it is even worse in this particular case. We have to generate a new Vector, evaluate the RHS operator, scale the vector and finally add the vector to the Vector obtained after the acoustic operator.

jh66637:
The more I think about it, the more I am in favor of shifting the scaling factors to the temporal derivative. https://github.com/exadg/exadg/pull/630 implements the necessary `apply_scale` method for the inverse mass operator.

jh66637:
@nfehn In https://github.com/exadg/exadg/pull/631 we can see that it is better to solve for the alternative version (see PR description). This way, we can reduce the total amount of flops, we don't have to touch the rhs operator and the equations become more readable from the source code.

jh66637:
With https://github.com/exadg/exadg/pull/631 the PR is now trivial. I extended the documentation and resolved all conversations which are obsolete.

jh66637:
I rebased on top of master and updated everything. This should be ready now.

nfehn (approved):

PR #620, opened 2023-12-12
Add CFL based time step size computation for acoustics

jh66637:
It is tedious to specify the time step size manually. With this PR the user can choose CFL-based time-stepping. Note that I didn't use the implementation in an application on purpose: I.e., I will create a test in a separate PR that additionally checks different BCs and the RHS application.

### DIFF
--- a/include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
+++ b/include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
@@ -336,7 +336,7 @@ SpatialOperator<dim, Number>::calculate_time_step_cfl() const
 
   // The time-step size is not adapted every time-step. Thus, we are using
   // a constant function to pass in the speed of sound, even though it is
-  // be possible to optimize calculate_time_step_cfl_local() for this case.
+  // possible to optimize calculate_time_step_cfl_local() for this case.
 
   return calculate_time_step_cfl_local<dim, Number>(
     get_matrix_free(),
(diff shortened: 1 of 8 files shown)


include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
```diff
@@ -326,6 +327,29 @@ SpatialOperator<dim, Number>::apply_inverse_mass_operator(BlockVectorType &
...
+template<int dim, typename Number>
+double
+SpatialOperator<dim, Number>::calculate_time_step_cfl() const
+{
+  // In case of mixed-orders use the maximum fe_degree and the corresponding
+  // quadrature rule.
+
+  // The time-step size is not adapted every time-step. Thus, we are using
+  // a constant function to pass in the speed of sound, even though it is
+  // be possible to optimize calculate_time_step_cfl_local() for this case.
```
nfehn, line 339: "it is be possible"

include/exadg/time_integration/time_int_abm_base.h
```diff
@@ -88,6 +88,12 @@ class TimeIntAdamsBashforthMoultonBase : public TimeIntMultistepBase
     return solution;
   }
 
+  Operator const &
+  get_underlying_operator() const
+  {
+    return *pde_operator;
+  }
```
nfehn, line 95: I think this should be protected

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
jh66637: [3eeb568](https://github.com/exadg/exadg/pull/620/commits/3eeb568630fefc005322180097ce3ee411e0ace0)

nfehn (approved):

PR #621, opened 2023-12-12
Add Velocity Dirichlet and Pressure Neumann BCs for Acoustic Module

jh66637:
This PR adds velocity Dirichlet and pressure Neumann BCs for the acoustic module

### DIFF
--- a/applications/acoustic_conservation_equations/vibrating_membrane/application.h
+++ b/applications/acoustic_conservation_equations/vibrating_membrane/application.h
@@ -209,9 +209,8 @@ private:
   void
   set_boundary_descriptor() final
   {
-    this->boundary_descriptor->pressure->dirichlet_bc.insert(std::make_pair(
+    this->boundary_descriptor->pressure_dbc.insert(std::make_pair(
       1, std::make_shared<AnalyticalSolutionPressure<dim>>(modes, this->param.speed_of_sound)));
-    this->boundary_descriptor->velocity->neumann_bc.insert(1);
   }
 
   void
--- a/include/exadg/acoustic_conservation_equations/spatial_discretization/operators/operator.h
+++ b/include/exadg/acoustic_conservation_equations/spatial_discretization/operators/operator.h
@@ -479,10 +479,10 @@ private:
                      Range const &                           face_range) const
   {
     FaceIntegratorP pressure_m(matrix_free_in, true, data.dof_index_pressure, data.quad_index);
-    BoundaryFaceIntegratorP<dim, Number> pressure_p(pressure_m, *data.bc->pressure);
+    BoundaryFaceIntegratorP<dim, Number> pressure_p(pressure_m, *data.bc);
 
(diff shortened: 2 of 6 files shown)


nfehn (approved):

nfehn:
The implementation looks good.

I wonder whether we need to explain the meaning of "Neumann" boundary conditions and the relation between velocity and pressure. Or should we refer to the incompressible flow module? (The wording "Neumann BC" seems to be less clear than for the incompressible Navier-Stokes equations, because here in the acoustic modules it seems that for none of the quantities a Neumann BC in the sense of a normal gradient is prescribed?)

For incompressible flows, I think we started with a single `BoundaryType` and only later distinguished between velocity and pressure once there have been introduced more boundary types than just "Dirichlet"/"Neumann". Another aspect is the presence of a pressure Poisson equation with corresponding "pressure-only BCs" for the incompressible Navier-Stokes equations, which somewhat motivates to have a separate `BoundaryTypeP` with Dirichlet/Neumann in the meaning of a "standard Poisson solver". 

So as long as we have Dirichlet/Neumann only (with the opposite type between velocity and pressure), it could make sense - from a software design perspective - to introduce only a single `BoundaryType`?

jh66637:
> I wonder whether we need to explain the meaning of "Neumann" boundary conditions and the relation between velocity and pressure.

I oriented myself on the naming of the incompressible Navier--Stokes module. I am happy to add some documentation. We could also rename the fields but I think the naming is consistent to the rest of `ExaDG`, correct?

> [...] as we have Dirichlet/Neumann only [...], it could make sense [...] to introduce only a single BoundaryType

I have the code for a velocity admittance boundary condition (locally) in place. So we can not use the same boundary type later. I would like to keep both types to make it easier to further extend the BCs.

nfehn:
I think it is "consistent" with ExaDG. However, when looking at the acoustics module in an isolated way, I wanted to raise the question whether Dirichlet/Neumann naming is clear/intuitive, in particular with two different boundary types and the absense of "Neumann"-type boundary conditions.

> I have the code for a velocity admittance boundary condition (locally) in place. So we can not use the same boundary type later. I would like to keep both types to make it easier to further extend the BCs.

For this velocity addmittance boundary condition, what is the boundary condition for the pressure? Is one free in choosing the type of pressure boundary condition? If a certain type of pressure boundary condition has to be used for such a velocity addmittance boundary condition, this would be a reason for a single `BoundaryType` in my opinion.

Another difference that I see is that the pressure is not evaluated in an isolated way for acoustics (like separate spatial operators for velocity and pressure) as is the case for the pressure Poisson operator in `IncNS`.

jh66637:
The admittance BCs are always combined with pressure "Neumann BCs".

Thanks for clarification. With this, I think I understood what you mean:
[aa211e7](https://github.com/exadg/exadg/pull/621/commits/aa211e73a36f95d4b33ad97820adcfe0c99b2377)
I like this a lot :)

nfehn (approved):
Great!

Issue #615, opened 2023-12-01
Towards robust h-multigrid for high-order mappings


nfehn:
PR #614 introduces an application test case where a high-order mapping leads to invalid grid/mappings on coarser multigrid h-levels.

Generally speaking, there is no guarantee that h-multigrid is robust in combination with high-order mappings. Boundary layer mappings as described in PR #614 inherently lead to problems and can be seen as a practically relevant example. One could argue to use a linear mapping on the fine grid to solve the problem, but think of flow around curved bodies (in the most simple case a sphere or cylinder) where one wants to approximate the geometry at least by a quadratic mapping and - at the same time - use a grid stretching in walll-normal direction to resolve boundary layers. 

Note that the same problem can occur when describing the fine level mapping by `MappingQCache` (e.g. some complex geometry in biomedical engineering) and when constructing the mappings on coarser grids by standard multigrid transfer operations.

I suggest to introduce in ExaDG the flexibility to choose the degree of the  mappings on coarser multigrid levels independently of the degree of the fine level mapping. Regarding the parameters, there occur different questions:
- On the fine multigrid level, should we generally use the same mapping degree as used for the actual "PDE operator"? I would say "yes". We can safely assume that the fine level mapping prescribed by an application has to be valid. Otherwise we could not solve on that geometry/grid at all.
- On the coarser levels, is one single parameter (`unsigned int mapping_degree_coarse_grids`) for all the coarse grids enough, or do we even want to have more flexibility? I would say one parameter for all the coarse levels is enough. I could imagine that going to a linear mapping on coarser grids recovers robustness for many problems.

Of course, another and very general question concerns the implications of such a "variational crime" on multigrid convergence rates.

PR #622, opened 2023-12-13
enable coarse multigrid mappings of low mapping degree

nfehn:
addresses #615.

In upcoming commits, I want to introduce the capability to use a different mapping degree for the coarse multigrid mappings.

### DIFF
--- a/include/exadg/incompressible_navier_stokes/precursor/user_interface/application_base.h
+++ b/include/exadg/incompressible_navier_stokes/precursor/user_interface/application_base.h
@@ -97,7 +97,7 @@ public:
 
     // grid
     GridUtilities::create_mapping(mapping, param.grid.element_type, param.mapping_degree);
-    std::shared_ptr<std::vector<std::shared_ptr<dealii::Mapping<dim>>>> coarse_mappings;
+    std::shared_ptr<dealii::Mapping<dim>> coarse_mappings;
     multigrid_mappings = std::make_shared<MultigridMappings<dim, Number>>(mapping, coarse_mappings);
 
     grid = std::make_shared<Grid<dim>>();
--- a/include/exadg/incompressible_navier_stokes/precursor/user_interface/application_base.h
+++ b/include/exadg/incompressible_navier_stokes/precursor/user_interface/application_base.h
@@ -96,10 +96,6 @@ public:
     param.print(pcout, "List of parameters:");
 
     // grid
-    GridUtilities::create_mapping(mapping, param.grid.element_type, param.mapping_degree);
-    std::shared_ptr<dealii::Mapping<dim>> coarse_mappings;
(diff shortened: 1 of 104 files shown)


nfehn:
The problem for the test case from https://github.com/exadg/exadg/pull/614 with boundary layer manifold, which is critical in terms of the mapping degrees, can be overcome by setting `this->param.mapping_degree_coarse_grids = 1` using the capabilities introduced in the present PR. 

@kronbichler @peterrum Regarding the "variational crime" mentioned here https://github.com/exadg/exadg/issues/615#issue-2020601817: As we have seen, mapping degree of 2 for all levels does not work. Interestingly, I have just realized that mapping degree 3 on all levels seems to be fine for this setup. This gives a point of reference regrading the degradation in iterations counts when using linear mappings for the coarser multigrid levels. Mapping degree 3 for all levels requires 24.8 iterations, while mapping degree 3 for the fine level and a linear mapping for all coarser levels requires 33.3 iterations, always using pure h-MG! So the increase in iterations counts seems to be substantial here. For cph-MG, iterations counts are 6.9 for both mapping setups, so no difference is visible!

kronbichler:
These numbers are an interesting observation. I think there are two different mechanisms at work regarding the multigrid convergence here:
- Changing the mapping degree changes the approximated operators, and thus, they have (somewhat) different eigenvalue spectra, such that the decomposition in different scales with smoothers damping the "high-frequency" content has a different meaning. In the extreme case, having a bad mapping (like the degree 2 case you mention) means that the eigenvalue spectrum of the matrix A gets bad.
- The mapping can also make the standard embedding of the multigrid transfer, that is defined in the reference element, a poor choice. This means that even if the coarse-scale solution might be good "correction", the embedding of it into the fine level might distort it. I would intuitively say that this should be a fine-scale/high-frequency perturbation, so the fine-level smoother should be good at removing it again, but eigenmodes resulting from polynomial clipping might do badly.

Since the difference you observe is substantial, I think it would instructive to try to separate the two issues. I think we should be able to use the non-matching transfer operator for testing, which would make the embedding the natural one in the expanded base. That would then level my first item.

include/exadg/incompressible_navier_stokes/precursor/user_interface/application_base.h
```diff
@@ -97,7 +97,8 @@ class Domain
 
     // grid
     GridUtilities::create_mapping(mapping, param.grid.element_type, param.mapping_degree);
-    multigrid_mappings = std::make_shared<MultigridMappings<dim, Number>>(mapping);
+    std::shared_ptr<dealii::Mapping<dim>> coarse_mappings;
+    multigrid_mappings = std::make_shared<MultigridMappings<dim, Number>>(mapping, coarse_mappings);
```
nfehn, line 101: this can actually be removed now

Issue #588, opened 2023-10-31
Add Roe type flux for acoustics


jh66637:
As suggested by @kronbichler in https://github.com/exadg/exadg/pull/587#issuecomment-1785281887, we should implement the Roe type flux in a separate PR.

PR #625, opened 2023-12-14
Add Roe-type flux

jh66637:
Closes https://github.com/exadg/exadg/issues/588


include/exadg/acoustic_conservation_equations/spatial_discretization/operators/operator.h
```diff
@@ -136,6 +136,22 @@ class Kernel
...
+  inline DEAL_II_ALWAYS_INLINE //
+    scalar
+    calculate_roe_type_flux_mass(scalar const & pm,
+                                 scalar const & pp,
+                                 Number const & tau,
+                                 vector const & um,
+                                 vector const & up,
+                                 vector const & n) const
+  {
+    return Number{0.5} * (pm + pp) + tau * (um * n - up * n);
```
jh66637, line 153: @kronbichler: If I read your comment https://github.com/exadg/exadg/pull/587#issuecomment-1785281887 correctly, this is the Roe-type flux you were describing. Is this correct?

kronbichler: I must admit that I had to check carefully how the two fluxes are defined. I looked into http://dx.doi.org/10.1016/j.cma.2015.12.003 and formulas (18)-(20) in particular. What they call HDG flux should be the Roe flux. In your formula, I think both the local Lax-Friedrichs (Rusanov) flux and the Roe flux are the same, because it does not matter whether you switch the order of the subtraction and the multiplication `u * n`. Looking at the flux, they are the Roe flux both. For a Lax-Friedrichs that dissipates in all directions, you would need to take the outer product of `u \otimes n` and return that to the place of use instead of `flux_mass * n`. In other words, right now you contract between `u` and `n`, before you multiply by `n` again, but for Lax-Friedrichs you would need to contract between `n` and `n`, and afterwards multiply by `u`.

As these formulas are intricate and I might have missed some important detail, I recommend you to have another look in terms of literature and formulas, and decide how you want to proceed.

I guess right now you don't see a difference in any of the results, do you? While I would expect that for `dim > 1`, one should see a difference between Roe and Rusanov.

jh66637: I agree, that the fluxes are equivalent and the results are exactly the same. 

Thank you for pointing out the paper, I didn't find helpful literature with the keyword Roe-type flux. With this, I can go through the equations in detail :+1:  

I will convert the PR to a draft until I find the time to do so.

PR #626, opened 2023-12-14
Add sound-energy calculator

jh66637:
This makes it possible to compute the sound energy in the computational domain

```math
E = \int \frac{p^2}{2\rho c^2}+\frac{\rho \mathbf{u}\cdot\mathbf{u}}{2} \mathrm{d}\Omega
```

### DIFF
--- a/include/exadg/acoustic_conservation_equations/postprocessor/sound_energy_calculator.h
+++ b/include/exadg/acoustic_conservation_equations/postprocessor/sound_energy_calculator.h
@@ -77,7 +77,7 @@ class SoundEnergyCalculator
 {
   using This = SoundEnergyCalculator<dim, Number>;
 
-  using BlockVectorType = dealii::LinearAlgebra::distributed::BlockVector<Number>;
+  using VectorType = dealii::LinearAlgebra::distributed::Vector<Number>;
 
   using scalar = dealii::VectorizedArray<Number>;
   using vector = dealii::Tensor<1, dim, scalar>;
@@ -92,12 +92,7 @@ public:
       matrix_free(nullptr),
       dof_index_pressure(0),
       dof_index_velocity(1),
-      quad_index(0),
-      block_index_pressure(0),
-      block_index_velocity(1),
-      rho(-1.0),
-      c(-1.0),
-      rhocc_inv(-1.0)
+      quad_index(0)
   {
   }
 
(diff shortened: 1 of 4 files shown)


include/exadg/acoustic_conservation_equations/postprocessor/sound_energy_calculator.h
```diff
@@ -0,0 +1,263 @@
...
+    dof_index_pressure   = dof_index_pressure_in;
+    dof_index_velocity   = dof_index_velocity_in;
+    quad_index           = quad_index_in;
+    block_index_pressure = block_index_pressure_in;
+    block_index_velocity = block_index_velocity_in;
+
+    rho       = static_cast<Number>(data_in.density);
+    c         = static_cast<Number>(data_in.speed_of_sound);
+    rhocc_inv = static_cast<Number>(
+      1.0 / (data_in.density * data_in.speed_of_sound * data_in.speed_of_sound));
```
nfehn, line 127: I think this should be in the `cell_loop()` in `calculate_sound_energy()`

include/exadg/acoustic_conservation_equations/postprocessor/sound_energy_calculator.h
```diff
@@ -0,0 +1,263 @@
...
+    // sum over all MPI processes
+    double const sound_energy = dealii::Utilities::MPI::sum(energy, mpi_comm);
+
+    return sound_energy;
+  }
+
+  void
+  cell_loop(dealii::MatrixFree<dim, Number> const &       matrix_free_in,
+            Number &                                      dst,
+            BlockVectorType const &                       src,
```
nfehn, line 204: I believe the implementation would be simpler and more flexible overall when implementing this class by using separate velocity and pressure DoF-vectors.

jh66637: I was thinking about this as well. I decided to do it this way because otherwise we would have to store one of the two DoF vectors as a member (instead of directly passing one `BlockVector` to the loop. On the other hand we have to provide the block indices. 

I think both versions have their pros and cons and can adapt to whatever you prefer.

include/exadg/acoustic_conservation_equations/postprocessor/sound_energy_calculator.h
jh66637: [2f33618](https://github.com/exadg/exadg/pull/626/commits/2f336186a4229565ad0cccf9a46851d3e668fea6)

include/exadg/acoustic_conservation_equations/postprocessor/sound_energy_calculator.h
nfehn: I see a difference between the two variants because as far as I know the transformation between normal DoFVectors and BlockVectors does not work equivalently in both directions from the dealii-side.  So I consider the variant with the two dof vectors more flexible. Or is there functionality to create a block-vector out of two dof-vectors "on-the-fly" without copying the vectors (a constructor like `BlockVector(std::vector<DealiiDofVector *>)`)?

> because otherwise we would have to store one of the two DoF vectors as a member

I think technically there is a solution to this problem. You can pass to `cell_loop()` an argument like `std::vector<VectorType const*> const &  src`, which you can initiliaze prior to calling cell-loop via `std::vector<VectorType const *> src{&velocity, &pressure};`. This way, we should be able to write the interface of this postprocessing routine in a flexible way with `velocity` and `pressure` dof-vectors and without block-indices. Choosing this design, the indexing can be done purely local and would then be a detail of the implementation of  this postprocessing routine.

nfehn:
There seems to be a compile error.

Please also note the warning regarding an "unused variable", `Number const c   = static_cast<Number>(data.speed_of_sound);`.

nfehn (approved):

PR #628, opened 2023-12-14
Don't store MPI_Comm as reference

jh66637:
This is not urgent, but I think we should not store `MPI_Comm` as a reference.


kronbichler (approved):
Correct, `MPI_Comm` is a C-type pointer/struct already, so storing the value (rather than the reference) is the right approach.

PR #630, opened 2023-12-18
Add apply_scale for inverse mass matrix operator

jh66637:
This makes it possible to compute 
```
dst = scalling_factor * M^-1(src)
```
In case of `InverseMassType::MatrixfreeOperator` streaming the vector from memory twice is avoided.

### DIFF
--- a/include/exadg/operators/inverse_mass_operator.h
+++ b/include/exadg/operators/inverse_mass_operator.h
@@ -171,6 +171,7 @@ public:
     }
   }
 
+  // dst = M^-1 * src
   void
   apply(VectorType & dst, VectorType const & src) const
   {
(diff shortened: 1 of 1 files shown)


include/exadg/operators/inverse_mass_operator.h
```diff
@@ -186,6 +187,38 @@ class InverseMassOperator
     }
   }
 
+  // dst = scaling_factor * (M^-1 * src)
+  void
+  apply_scale(VectorType & dst, double const scaling_factor, VectorType const & src) const
+  {
+    dst.zero_out_ghost_values();
```
nfehn, line 194: In the else-branch below, the function `apply()` is called, which internally calls `dst.zero_out_ghost_values();`. So I think we should shift this line in order to avoid misunderstandings and/or redundant function calls. Additionally, a comment why we need this line would be helpful in my opinion.

nfehn (approved):
looks good from my side

PR #632, opened 2023-12-18
Make HDF5 an optional dependency and KineticEnergySpectrumCalculator an optional postprocessing routine

### LINKED ISSUE #419: Tests are failing if HDF5 (optional dependency) is not installed
HDF5 is declared an optional dependency. However, the tests are failing without hdf5 (at least in my case) with the message "deal.II is not compiled with HDF5!". I consider this a contradiction that needs to be resolved  (since testing is essential/mandatory, rendering hdf5 a required dependency).

I suggest to exclude hdf5-related content from testing.

richardschu:
closes #419

in `incompressible_navier_stokes/postprocessor/postprocessor.cpp` in the `PostProcessor::setup()`, we call among others `pointwise_output_generator.setup()` and `kinetic_energy_spectrum_calculator.setup()`.
The former requires hdf5, which is only an optional dependency as far as I know and the latter requires `elementType::Hypercube`. I moved the asserts behind a `time_control.is_active`, since the setup is called in any case, but the postprocessors might not even be active.
After the setup, the `evaluate`on the postprocessors is behind a `pointwise_output_generator.time_control.needs_evaluation()` for example. So not setting them up should be fine.

### DIFF
--- a/include/exadg/postprocessor/kinetic_energy_spectrum.cpp
+++ b/include/exadg/postprocessor/kinetic_energy_spectrum.cpp
@@ -597,7 +597,7 @@ KineticEnergySpectrumCalculator<dim, Number>::setup(
   {
     data        = data_in;
     clear_files = data.clear_file;
-    time_control.setup(data_in.time_control_data);
+    time_control.setup(data.time_control_data);
 
     dof_handler = &dof_handler_in;
     AssertThrow(
--- a/include/exadg/postprocessor/kinetic_energy_spectrum.cpp
+++ b/include/exadg/postprocessor/kinetic_energy_spectrum.cpp
@@ -596,17 +596,17 @@ KineticEnergySpectrumCalculator<dim, Number>::setup(
   data        = data_in;
   dof_handler = &dof_handler_in;
 
-  AssertThrow(
-    dof_handler->get_triangulation().all_reference_cells_are_hyper_cube(),
-    dealii::ExcMessage(
-      "This postprocessing utility is only available for meshes consisting of hypercube elements."));
-
   clear_files = data.clear_file;
 
   time_control.setup(data_in.time_control_data);
 
   if(data_in.time_control_data.is_active)
   {
+    AssertThrow(
+      dof_handler->get_triangulation().all_reference_cells_are_hyper_cube(),
+      dealii::ExcMessage(
+        "This postprocessing utility is only available for meshes consisting of hypercube elements."));
+
     if(data.write_raw_data_to_files)
     {
       AssertThrow(data.n_cells_1d_coarse_grid == 1,
--- a/include/exadg/postprocessor/kinetic_energy_spectrum.cpp
+++ b/include/exadg/postprocessor/kinetic_energy_spectrum.cpp
@@ -593,15 +593,13 @@ KineticEnergySpectrumCalculator<dim, Number>::setup(
   dealii::DoFHandler<dim> const &         dof_handler_in,
   KineticEnergySpectrumData const &       data_in)
 {
-  data        = data_in;
-  dof_handler = &dof_handler_in;
-
-  clear_files = data.clear_file;
-
-  time_control.setup(data_in.time_control_data);
-
   if(data_in.time_control_data.is_active)
   {
+    data        = data_in;
+    clear_files = data.clear_file;
+    time_control.setup(data_in.time_control_data);
+
+    dof_handler = &dof_handler_in;
     AssertThrow(
       dof_handler->get_triangulation().all_reference_cells_are_hyper_cube(),
       dealii::ExcMessage(
--- a/include/exadg/postprocessor/pointwise_output_generator_base.cpp
+++ b/include/exadg/postprocessor/pointwise_output_generator_base.cpp
@@ -145,8 +145,10 @@ PointwiseOutputGeneratorBase<dim, Number>::setup_base(
 #else
   (void)triangulation_in;
   (void)mapping_in;
-  (void)pointwise_output_data_in;
-  AssertThrow(false, dealii::ExcMessage("deal.II is not compiled with HDF5!"));
+  if(pointwise_output_data_in.time_control_data.is_active)
+  {
+    AssertThrow(false, dealii::ExcMessage("deal.II is not compiled with HDF5!"));
+  }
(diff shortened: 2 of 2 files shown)


jh66637 (approved):

include/exadg/postprocessor/pointwise_output_generator_base.cpp
```diff
@@ -145,8 +145,10 @@ PointwiseOutputGeneratorBase<dim, Number>::setup_base(
 #else
   (void)triangulation_in;
   (void)mapping_in;
-  (void)pointwise_output_data_in;
-  AssertThrow(false, dealii::ExcMessage("deal.II is not compiled with HDF5!"));
+  if(pointwise_output_data_in.time_control_data.is_active)
```
nfehn, line 148: Is the code in lines 102-114 actually placed correctly? Why not inside the if-statement (line 116)?

I think the `#ifdef` should also appear inside the if-statement above since this would make the coder easier to understand for the `#else` branch.

include/exadg/postprocessor/kinetic_energy_spectrum.cpp
```diff
@@ -596,17 +596,17 @@ KineticEnergySpectrumCalculator<dim, Number>::setup(
...
-  AssertThrow(
-    dof_handler->get_triangulation().all_reference_cells_are_hyper_cube(),
-    dealii::ExcMessage(
-      "This postprocessing utility is only available for meshes consisting of hypercube elements."));
-
   clear_files = data.clear_file;
 
   time_control.setup(data_in.time_control_data);
 
   if(data_in.time_control_data.is_active)
```
nfehn, line 603: This if-statement uses `data_in`, so I wonder whether we should actually shift everything before this if-statement inside the if-statement.

nfehn: Then, the assert on the dof_handler/triangulation would again appear next to the assignment of the dof-handler.

richardschu: Yep!

include/exadg/postprocessor/pointwise_output_generator_base.cpp
richardschu: Yep; I think you are right. I think we should move 102-114 and the `#ifdef' inside the if.

include/exadg/postprocessor/kinetic_energy_spectrum.cpp
```diff
@@ -593,20 +593,18 @@ KineticEnergySpectrumCalculator<dim, Number>::setup(
...
-
-  clear_files = data.clear_file;
-
-  time_control.setup(data_in.time_control_data);
-
   if(data_in.time_control_data.is_active)
   {
+    data        = data_in;
+    clear_files = data.clear_file;
+    time_control.setup(data_in.time_control_data);
```
nfehn, line 600: one line above, `data` is used, so I would also use `data` here instead of `data_in`.

nfehn:
Looks good from my side, apart from the minor aspect whether the variable with `_in` or without is used.

nfehn:
@richardschu @jh66637 could you please have a look at issue #419 and make sure that this problem is resolved as well? Adding a reference to this issue in the description of the present PR will then automatically close that issue.

jh66637:
I can have a look

include/exadg/postprocessor/pointwise_output_generator_base.cpp
```diff
@@ -141,13 +140,13 @@ PointwiseOutputGeneratorBase<dim, Number>::setup_base(
       auto group = hdf5_file->create_group("PhysicalInformation");
       add_time_dataset(group, "Time");
     }
-  }
+
 #else
-  (void)triangulation_in;
-  (void)mapping_in;
-  (void)pointwise_output_data_in;
-  AssertThrow(false, dealii::ExcMessage("deal.II is not compiled with HDF5!"));
```
nfehn, line 149: @jh66637 isn't the main point of the present PR that this assert is currently always triggered independently of whether we want to use this postprocessing routine or not?

I understood the present PR in the way that this line currently enforces HDF5 to be a mandatory TPL for ExaDG. And this way I see a link to issue #419.

nfehn:
@richardschu Have you encountered these problems because you have not installed HDF5 on the machine you are currently working? (I think this would help regarding the discussions in issue #419.)

include/exadg/postprocessor/pointwise_output_generator_base.cpp
jh66637: I agree, but I can not reproduce it on my laptop.

nfehn: It looks like you need to find out in more detail what your local setup is doing exactly.

richardschu:
> @richardschu Have you encountered these problems because you have not installed HDF5 on the machine you are currently working? (I think this would help regarding the discussions in issue #419.)

That was precicely the problem. I now compiled deal.II with hdf5, but I thought I might have a look at why it requires me to have it...
I am currently trying to reproduce the Assertion same way Johannes did and report back as soon as I can say more.
If this does not help, I will uninstall hdf5 again and see where the problem was. The changes proposed in this PR were what I thought the problem was, but I fixed the problem at my end in 2 ways at the same time so I cannot guarantee, that the changes proposed here (while they do make sense) do the trick.

I will upadte you guys asap.

nfehn:
@richardschu I think there is no more inspection required from your side. I think we are now all convinced that the present PR fixes the issue.

richardschu:
> @richardschu @jh66637 could you please have a look at issue #419 and make sure that this problem is resolved as well? Adding a reference to this issue in the description of the present PR will then automatically close that issue.

and how did that work now? It appeared on the side after you mentioned it, but do we have to specify, that this closes #419?

nfehn:
I changed the description of the PR manually. Writing `closes #NUMBER` will then automatically close the issue once the PR is merged.

richardschu:
> I changed the description of the PR manually. Writing `closes #NUMBER` will then automatically close the issue once the PR is merged.

oh, thanks :) !

nfehn (approved):

PR #610, opened 2023-11-28
Shift grid to driver

nfehn:
Similar to PR #596, now for the rest of ExaDG.

PR #633, opened 2023-12-18
add grid and mapping to flow set_boundary_descriptor

richardschu:
in #610 and #596, we moved the `Grid` and `Mapping` from the application to the driver for all applications. In some flow simulations, one might want to create an inflow profile on a surface flagged by some boundary ID, which, however, might not have a specific shape like a circle or rectangle. In such cases, the grid and mapping are needed in the `set_boundary_descriptor()` functions the detect the flagged surface and construct an inflow profile.

Hence I introduce this functionality for flow problems only, but we can have the same arguments for all the applications (Poisson, Acoustic, ConvectionDiffusion etc ...).

### DIFF
--- a/applications/compressible_navier_stokes/couette/application.h
+++ b/applications/compressible_navier_stokes/couette/application.h
@@ -247,8 +247,12 @@ private:
   }
 
   void
-  set_boundary_descriptor() final
+  set_boundary_descriptor(Grid<dim> const &                             grid,
+                          std::shared_ptr<dealii::Mapping<dim>> const & mapping) final
   {
+    (void)grid;
+    (void)mapping;
+
     typedef typename std::pair<dealii::types::boundary_id, std::shared_ptr<dealii::Function<dim>>>
                                                                                    pair;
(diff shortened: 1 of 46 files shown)


nfehn:
> the grid and mapping are needed in the `set_boundary_descriptor()` functions the detect the flagged surface and construct an inflow profile

I did not fully understand this. The boundary IDs are set in `create_grid()` normally?

From a software design perspective, I think we should not extend the interfaces this way. What one can always do in a specific application is to store a pointer to const-Grid / const-Mapping at the end of the `create_grid()` function.

Which information is available exactly for this type of boundary conditions? (I am wondering whether out `Cached`-type boundary conditions would be suitable for this purpose, or do you use them already for this purpose?)

richardschu:
Yes, that is the approach I am taking.
Ok; I will follow the route you suggested since it is way less annoying - I will close this PR since it is not needed.

Yes, one could fill some caches, but I am actually recomputing the inflow profile on the fly. The data that describes the profile is updated in every postprocessor call. As far as I have seen, the postprocessor does not take too much time so for now this does the job.

richardschu:
.

PR #634, opened 2023-12-20
Implement MappingDoFVector (ALE) for ElementType::Simplex

nfehn:
To enable ALE methods for `ElementType::Simplex`, we need to extend

1. class `MappingDoFVector`:

- function `fill_grid_coordinates_vector()` (in order to translate a `Mapping` into a `DoFVector`)
- function `initialize_mapping_from_dof_vector()`  (in order to translate a `DoFVector` into a `Mapping`)

2. utility functions in namespace `MappingTools`:
- functions `initialize_coarse_mappings_from_mapping_dof_vector()` to construct coarse mappings of type `MappingDoFVector` for use in h-multigrid methods
- There are two functions with this name, differing in the number of arguments. The variant with only one triangulation object for all h-levels is not relevant for `ElementType::Simplex`, right?


nfehn:
The first commit shows which places we need to extend / re-implement for simplex by introducing if-else-branches depending on the `ElementType` and by introducing asserts for element types not supported so far.

nfehn:
@kronbichler as discussed in person, the current implementation of `MappingDoFVector` might be problematic, because `dof_handler` (and the underlying FE and degree) passed to the function `fill_grid_coordinates_vector()` (and `initialize_mapping_from_dof_vector()`) can be independent of the `mapping_degree` passed to the constructor of `MappingDoFVector`. I removed the constructor argument and used `dof_handler.get_fe().degree` instead; the hierarchic/lexicographic indexing becomes a local variable instead of the member variable used previously. Could you please check commits https://github.com/exadg/exadg/pull/634/commits/32efd75f0fb9bffe735d130ad490e094c2804ba8 and https://github.com/exadg/exadg/pull/634/commits/b6e3210096d7305bc1ea7492c906220f024840dc?

kronbichler:
I had a close look at the code now. I think the new approach looks good. For the simplices, we do not have the equivalent of `MappingQCache` anyway, so all routines are more expensive. I think the longer-term solution will be to let all mapping-related code use the evaluators of solution fields, but that needs modification in deal.II as well; that is not part of the present PR in either case.

nfehn:
@kronbichler thanks for your review. 

We need some more refactoring in order to integrate Simplex elements easily without having to touch several classes (see https://github.com/exadg/exadg/pull/634/commits/923958b715e0fa06167a80788ed9fe265523fc9f and https://github.com/exadg/exadg/pull/634/commits/7fbcbe8f8cb2a1bb77006ce4b98418c0b93c31df).

include/exadg/grid/mapping_dof_vector.h
```diff
@@ -32,22 +32,89 @@
...
+{
 /**
- * A mapping class based on dealii::MappingQCache equipped with practical interfaces that can be
- * used to initialize the mapping.
+ * Extract the grid coordinates for a given external mapping and fill a
+ * dof-vector given a corresponding dealii::DoFHandler object.
+ */
+template<int dim, typename Number>
+void
+fill_grid_coordinates_vector(dealii::Mapping<dim> const &                         mapping,
```
nfehn, line 55: @kronbichler similar to comment https://github.com/exadg/exadg/pull/634/files#r1457362322, I wonder whether we could realize this function as `dealii::VectorTools::interpolate()` using an "identity function" as `dealii:Function`.

include/exadg/grid/mapping_dof_vector.h
```diff
@@ -105,215 +232,311 @@ class MappingDoFVector
...
+
+            for(unsigned int i = 0; i < fe_fe_field->dofs_per_cell; ++i)
             {
-              grid_coordinates(dof_indices[component_to_system_index[i][d]]) = point[d];
+              unsigned int const d        = fe_fe_field->system_to_component_index(i).first;
+              unsigned int const i_scalar = fe_fe_field->system_to_component_index(i).second;
+
+              dealii::Point<dim> const point = fe_values.quadrature_point(i_scalar);
+
+              displacement(dof_indices[i]) = displacement_function->value(point, d);
```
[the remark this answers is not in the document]
kronbichler: More or less; this code should be considerably more efficient than `interpolate`, because of the de-duplicated support points in the call `fe_values.reinit(cell)`, and because the deal.II function works with some weights for the more general cases of generalized support points. I would leave this code in for now (which is considerably simpler than before), but we could think about using the deal.II function and possibly improving the performance there, see also https://github.com/dealii/dealii/issues/15650 that also lists `interpolate` as poorly implemented.

include/exadg/grid/mapping_dof_vector.h
kronbichler: Do you mean `VectorTools::get_position_vector`, https://www.dealii.org/current/doxygen/deal.II/namespaceVectorTools.html#afcb40e50d2fa4807762faefed206f5e3

I did not look closely, but it might work, and we might possibly want to optimize the function there with the optimization we developed for this PR.

nfehn: Originally, I meant `VectorTools::interpolate()`, using a function `f(x) = x` (identity).

applications/convection_diffusion/sine_wave/input.json
```diff
@@ -5,8 +5,8 @@
         "IsTest": "false"
     },
     "SpatialResolution": {
-        "DegreeMin": "7",
-        "DegreeMax": "7",
+        "DegreeMin": "2",
+        "DegreeMax": "2",
```
nfehn, line 9: This change was not actually intended and might cause the test to fail.

include/exadg/grid/mapping_dof_vector.h
```diff
@@ -478,88 +726,108 @@ initialize_coarse_mappings_from_mapping_dof_vector(
...
+    {
+      coarse_mappings[h_level] =
+        std::make_shared<MappingDoFVector<dim, Number>>(*coarse_triangulations[h_level]);
+
+      // grid_coordinates_all_levels describes absolute coordinates -> use an uninitialized mapping
+      // in order to interpret the grid coordinates vector as absolute coordinates and not as
+      // displacements.
+      std::shared_ptr<dealii::Mapping<dim> const> mapping_dummy;
+      coarse_mappings[h_level]->initialize_mapping_from_dof_vector(
+        mapping_dummy, grid_coordinates_all_levels[h_level], dof_handlers_all_levels[h_level]);
```
nfehn: @kronbichler `dof_handler_all_levels` is a local variable in this function. This will cause the code to terminate in the simplex case (because of `MappingFEField` using smart pointers to the `DoFHandler`) when using these mapping objects inside the multigrid preconditioner, where the code crashes when initializing the MatrixFree objects for all the multigrid levels with the error message 

```bash
The violated condition was: 
    pointed_to_object_is_alive
Additional information: 
    The object pointed to is not valid anymore.
```
I could change the interfaces in ExaDG towards `std::shared_ptr` and store such a shared pointer to the dof-handler in the class `MappingDoFVector`. Since these requires some changes I want to ask beforehand what you would suggest to solve this problem?  (Regarding dealii, I could imagine it might be a better software design (on the dealii-side) to use arguments of type `std::shared_ptr<DoFHandler>` in the constructor of  `dealii::MappingFEField`. This would in my opinion more clearly communicate to the user that the class `MappingFEField` still needs the constructor arguments after construction of the object.)

kronbichler: Yes, I would also suggest to keep the `DoFHandler` in `MappingDoFVector`, as discussed in person.

kronbichler (approved):

PR #635, opened 2023-12-22
Add minimal aero-acoustic solver

jh66637:
This PR introduces a minimal aero-acoustic solver. There will be two follow-up PRs to be able to use this in realistic settings. Most of the added lines correspond to a mesh file.

1) Make it possible to blend in the source terms
2) Use sub-time-stepping for the acoustic solver (including printing the performance metrics, which are currently printed only partially)
3) Use analytically known source terms

Below are some screenshots of the attached test case

Acoustic pressure:
Screenshot from 2023-12-22 14-28-29

Fluid pressure:
Screenshot from 2023-12-22 14-28-59


include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.h
```diff
@@ -147,12 +149,21 @@ class SpatialOperator : public Interface::SpatialOperator<Number>
...
+  initialize_dof_vector_pressure(VectorType & dst) const;
+
   /*
    * Prescribe initial conditions using a specified analytical/initial solution function.
    */
   void
   prescribe_initial_conditions(BlockVectorType & dst, double const time) const final;
 
+  /*
+   * Set right-hand side vector that has been integrated externally.
```
nfehn, line 162: What do you mean by "integrated"? I find this wording quite often in the present PR.

include/exadg/aero_acoustic/calculators/source_term_calculator.h
```diff
@@ -0,0 +1,137 @@
...
+      matrix_free->cell_loop(&This::compute_integrated_source_term<false>,
+                             this,
+                             dst,
+                             pressure_cfd_time_derivative_in,
+                             true);
+    }
+  }
+
+private:
+  template<bool consider_convection>
```
nfehn, line 89: Personally, I do not see the reasoning for such templates. Here we compute a coupling term, which is done rarely compared to "matrix-vector products" / operator evaluations done often in the single field solvers. So from a performance perspective, I see even less motivation here to make it a template parameter. A few lines above one can also see that using templates for this purpose make the code harder to read (a procedure, that can in my opinion not really be extended to cases where we have more than one such parameter).

include/exadg/aero_acoustic/single_field_solvers/fluid.h
```diff
@@ -0,0 +1,156 @@
...
+    pde_operator->initialize_vector_pressure(pressure_time_derivative);
+  }
+
+  void
+  advance_one_timestep_with_pressure_time_derivative(bool const update_dpdt)
+  {
+    time_integrator->advance_one_timestep_pre_solve(true);
+    time_integrator->advance_one_timestep_solve();
+
+    // The pressure time derivative has to be compute before the push back
```
nfehn, line 106: typo -> computed

include/exadg/aero_acoustic/single_field_solvers/fluid.h
```diff
@@ -0,0 +1,156 @@
...
+                                                                 is_test);
+
+    time_integrator->setup(application->get_parameters().restarted_simulation);
+
+    // initialize vector that stores the pressure time derivative
+    pde_operator->initialize_vector_pressure(pressure_time_derivative);
+  }
+
+  void
+  advance_one_timestep_with_pressure_time_derivative(bool const update_dpdt)
```
nfehn, line 101: `_with_` -> `_and_compute_`

include/exadg/aero_acoustic/single_field_solvers/fluid.h
```diff
@@ -0,0 +1,156 @@
...
+  // Postprocessor
+  std::shared_ptr<IncNS::PostProcessorBase<dim, Number>> postprocessor;
+
+  // Computation time (wall clock time).
+  std::shared_ptr<TimerTree> timer_tree;
+
+private:
+  // The aeroacoustic source term needs the pressure time derivative.
+  // The update of the vector is performed using
+  // advance_one_timestep_with_pressure_time_derivative(true).
```
nfehn, line 147: see comment https://github.com/exadg/exadg/pull/635/files#r1457442798

include/exadg/aero_acoustic/volume_coupling.h
```diff
@@ -0,0 +1,123 @@
...
+
+namespace ExaDG
+{
+namespace AeroAcoustic
+{
+/**
+ * A class that handles the volume coupling between fluid and acoustic.
+ */
+template<int dim, typename Number>
+class VolumeCoupling
```
nfehn, line 38: Is this class intended to handle both directions? Will you add another function `acoustic_to_fluid()` later? (I somewhat deduce this from the naming `fluid_to_acoustic`.)

If this is the case, maybe this is something to discuss in more detail, so that I better understand the motivation behind. From a software design perspective, I would currently argue to aim at a design as modular as possible, which could lead to a solution where the different coupling directions are realized independently.

include/exadg/aero_acoustic/user_interface/parameters.h
```diff
@@ -0,0 +1,105 @@
...
+#include <deal.II/base/parameter_handler.h>
+
+namespace ExaDG
+{
+namespace AeroAcoustic
+{
+enum class FluidToAcousticCouplingStrategy
+{
+  Undefined,
+  NonNestedMGRestriction
```
nfehn, line 36: I do not like the `MG` naming here. We might have this naming in dealii, but here we should speak to the ExaDG user and not to the dealii expert in my opinion. Why not something simple like `NonNestedGridTransfer`?

include/exadg/aero_acoustic/volume_coupling.h
```diff
@@ -0,0 +1,123 @@
...
+
+private:
+  Parameters parameters;
+
+  // Single field solvers
+  std::shared_ptr<SolverAcoustic<dim, Number>> acoustic_solver;
+  std::shared_ptr<SolverFluid<dim, Number>>    fluid_solver;
+
+  // Transfer operator
+  dealii::MGTwoLevelTransferNonNested<dim, VectorType> non_nested_mg;
```
nfehn, line 109: mg naming as above.

include/exadg/aero_acoustic/volume_coupling.h
```diff
@@ -0,0 +1,123 @@
...
+  setup(Parameters const &                           parameters_in,
+        std::shared_ptr<SolverAcoustic<dim, Number>> acoustic_solver_in,
+        std::shared_ptr<SolverFluid<dim, Number>>    fluid_solver_in)
+  {
+    parameters      = parameters_in;
+    acoustic_solver = acoustic_solver_in;
+    fluid_solver    = fluid_solver_in;
+
+    acoustic_solver_in->pde_operator->initialize_dof_vector_pressure(source_term_acoustic);
+    fluid_solver_in->pde_operator->initialize_vector_pressure(source_term_fluid);
```
nfehn, line 53: It would be good to unify these namings. Of course, something that is independent of the present PR.

include/exadg/aero_acoustic/volume_coupling.h
```diff
@@ -0,0 +1,123 @@
...
+
+      non_nested_mg.restrict_and_add(source_term_acoustic, source_term_fluid);
+    }
+    else
+    {
+      AssertThrow(false, dealii::ExcMessage("FluidToAcousticCouplingStrategy not implemented."));
+    }
+
+    acoustic_solver->pde_operator->set_integrated_rhs(source_term_acoustic);
+  }
```
nfehn, line 99: I like that the coupling term calculations happen outside the single field solver implementation (SpatialOperator)!

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
```diff
@@ -321,6 +338,9 @@ SpatialOperator<dim, Number>::evaluate(BlockVectorType &       dst,
   if(param.right_hand_side)
     rhs_operator.evaluate_add(dst.block(block_index_pressure), time);
 
+  if(integrated_rhs_set)
```
nfehn, line 341: Here I wonder whether we should be more explicit in order to improve understanding and also to avoid potential mistakes). I think there should be a parameter in `Acoustics::Parameters` (probably section on mathematical model) describing whether the acoustic solver receives a source term from the fluid problem. If this parameter is `true`, I would then check inside the body of the if-statement that the pointer is valid. Outside this class, I think one could do everything with this vector/pointer, so having the state `integrated_rhs_set = true` inside this class (just because the function `set_integrated_rhs()` has been set once) might not imply that this pointer is valid. Put differently, the meaning of the variable used currently is `integrated_rhs_has_been_set_once` in my opinion.

The current implementation looks very generic (`integrated_rhs` can be anything, must not come from a fluid solver), but is this really the case? What if I want to add another source term in the future. I would need some logic outside this class that adds the two vectors together and then sets the vector once. many mistakes are possible: (i) the vectors are not added, (ii) different modules calls `set_integrated_rhs()` independently and overwrite each other. Using a more modular design with separate parameters as suggested above is easier to understand and less error prone in my opinion. Put differently, I do not consider an "explicit dependency/reference" of the acoustics module on/to the fluid module (because of a parameter like `acoustic_param.use_source_term_xyz`) a disadvantage.

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
```diff
@@ -307,6 +315,15 @@ SpatialOperator<dim, Number>::prescribe_initial_conditions(BlockVectorType & dst
   dst.block(block_index_velocity) = velocity_double;
 }
 
+template<int dim, typename Number>
+void
+SpatialOperator<dim, Number>::set_integrated_rhs(VectorType const & integrated_rhs_in)
+{
+  integrated_rhs_set = true;
+  integrated_rhs.reset(integrated_rhs_in);
+  integrated_rhs->update_ghost_values();
```
nfehn, line 324: here I do not understand why we need an update of ghost values?

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.h
jh66637: It means, that the DoF vector contains 
```math
\int q f \mathrm{d}\Omega
```
where $q$ is an ansatz function and $f$ is the typical RHS "density". If I recall correctly, in other places the RHS is applied from a DoF vector, where the DoF vector stores the RHS density. I decided to use this naming to make this clear but can adapt to another convention. If you agree, I will extend the documentation to ensure this becomes more clear.

include/exadg/aero_acoustic/single_field_solvers/fluid.h
jh66637: [6777b10](https://github.com/exadg/exadg/pull/635/commits/6777b10addeed63fdb28fe0064abc5871f63ce2c)

include/exadg/aero_acoustic/calculators/source_term_calculator.h
jh66637: [fb6bc6d](https://github.com/exadg/exadg/pull/635/commits/fb6bc6d5665cee32521fb32710b671ebd67953b2)

include/exadg/aero_acoustic/single_field_solvers/fluid.h
jh66637: [44116d0](https://github.com/exadg/exadg/pull/635/commits/44116d0dc158887a19a64116013703abed445b00)

include/exadg/aero_acoustic/single_field_solvers/fluid.h
jh66637: https://github.com/exadg/exadg/pull/635#discussion_r1457442798

include/exadg/aero_acoustic/volume_coupling.h
jh66637: Yes, the class is intended to be used for both directions. The advantage to keep both directions in the same class is that, e.g. the setup of the classes from `deal.II` (such as the non-nested multigrid class or others like a method where the meshes are intersected) are quite expensive and can be used to transfer vectors in both directions. In the present example, one could use the `prolongate_and_add()` method for acoustic->fluid coupling. 

Besides that, I think it is natural to handle the transfer of solutions in both directions in one class since it is similar to the multigrid classes in `deal.II`.

include/exadg/aero_acoustic/user_interface/parameters.h
jh66637: In the aero-acoustic community, this transfer is called conservative interpolation. Would this name also be OK for you?

include/exadg/aero_acoustic/volume_coupling.h
jh66637: :)

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
jh66637: Good catch. We don't need it since we simply add vectors and do not use this in a matrix free loop.

jh66637: [a029183](https://github.com/exadg/exadg/pull/635/commits/a0291830aa127f73a2535204b3b90af41e508557)

include/exadg/aero_acoustic/volume_coupling.h
jh66637: I would like to call this `non_nested_grid_transfer` independent on the discussion above. This is because `ConservativeInterpolation` is only related to fluid->acoustic coupling, and we can do both with this class.

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
jh66637: I resolved this in two commits:
[1d93518](https://github.com/exadg/exadg/pull/635/commits/1d9351801f6339023f4ba84e2507634407dc4921)
[cabe4ae](https://github.com/exadg/exadg/pull/635/commits/cabe4aebb84ce3048032eee61e142eaa49eef43f)

jh66637:
@nfehn Thank you for going through this large PR! I resolved most of the comments. I think the things that are not resolved are only related to naming and documentation.

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
jh66637: There is one problem remaining: A `lazy_ptr` can never be a `nullptr`. It is initially the address to a default constructed object. Hence, we can not check if `lazy_ptr` is set by checking if it is a `nullptr`.

include/exadg/aero_acoustic/volume_coupling.h
nfehn: let's do this

include/exadg/aero_acoustic/user_interface/parameters.h
nfehn: yes

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.h
nfehn: I would expect a `rhs` vector in ExaDG is always an "integrated" vector. Could you point me to the place where this is used differently in ExaDG?

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
nfehn: I think the idea of the `lazy_ptr` is that it can either be data or a pointer to data. In the present case, it seems to be used always in the sense of a pointer to external data? In that case, it would be reasonable to use a raw pointer (or shared_ptr) in my opinion instead of the `lazy_ptr`.

nfehn:
There is one aspect I was thinking about the last days. It is the geometry of this example. While I consider this a very nice example, we generally only include the most simple academic setups in the ExaDG code, and I think we should do the same here. I would see a value in collecting grids somewhere, like a **grid gallery** for example, but probably not in the primary `github.com/exadg/exadg` project. (Maybe a project `github.com/exadg/grid_gallery`, or something completely independent of ExaDG.)

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
jh66637: In that case, lets keep the raw pointer :)

include/exadg/aero_acoustic/user_interface/parameters.h
jh66637: [f6b4340](https://github.com/exadg/exadg/pull/635/commits/f6b43402d57c58301f9346d1adfdcf96e0f8fd91)

include/exadg/aero_acoustic/volume_coupling.h
jh66637: [ba3f9e1](https://github.com/exadg/exadg/pull/635/commits/ba3f9e1662ced8b8d259656b9c7f20a05d9db7d7)

jh66637:
I can replace the geometry. :) 
Just let me know if you set up a `grid_gallery` project, so I can upload the grid there :)

nfehn:
Thanks again, @jh66637 !

PR #637, opened 2024-01-03
AMR: explicit timestepping for convection diffusion module

richardschu:
Enables explicit time integration in the convection diffusion module together with adaptive mesh refinement and adds a little test case.

The test uses a static mapping already (similar to the test with the implicit time stepping). The error is a bit lower compared to the implicit test, which computes the same scenario.

I did not add `get_solution_np()` to `TimeIntBase`, which requires some dynamic casts of the `time_integrator`in the driver.

### DIFF
--- a/include/exadg/convection_diffusion/driver.cpp
+++ b/include/exadg/convection_diffusion/driver.cpp
@@ -227,7 +227,7 @@ Driver<dim, Number>::do_adaptive_refinement()
     }
     else
     {
-      AssertThrow(false, dealii::ExcNotImplemented());
+      AssertThrow(false, dealii::ExcMessage("Adaptive mesh refinement available only for unsteady problems."));
     }
   }
 }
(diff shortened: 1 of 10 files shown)


include/exadg/convection_diffusion/driver.cpp
```diff
@@ -252,16 +254,28 @@ Driver<dim, Number>::solve()
...
-          std::shared_ptr<TimeIntBDF<dim, Number>> bdf_time_integrator =
-            std::dynamic_pointer_cast<TimeIntBDF<dim, Number>>(time_integrator);
-
-          AssertThrow(bdf_time_integrator.get(),
-                      dealii::ExcMessage("Adaptive mesh refinement only implemented"
-                                         " for implicit time integration."));
-
-          mark_cells_coarsening_and_refinement(*grid->triangulation,
-                                               bdf_time_integrator->get_solution_np());
+          // Mark cells for coarsening and refinement based on the most recent solution.
```
richardschu, line 257: introducing `TimeIntBase::get_solution_np()` would lead to:
`
mark_cells_for_coarsening_and_refinement(*grid->triangulation, time_integrator->get_solution_np());
`

richardschu:
Any opinions? I have to look at the `solution_transfer.interpolate()` due to the changes in deal.II.
Would be good if we merge the open and not conflicting PRs in that direction first.

richardschu:
I have been looking through old pull requests, seeing what might be easily merged.
This one is a trivial extension of #586
With the current PR, the convection diffusion module would be fully capable of adaptivity: for implicit *and explicit* time integration.
Let me know if you have any comments here.

richardschu:
tested this yet again, restart also works as intended and the test here shows that AMR works for the explicit integrator.
I will merge once the tests are through.

Issue #594, opened 2023-11-03
Test Performance of different versions of the acoustic operator


jh66637:
As discussed here: https://github.com/exadg/exadg/pull/587#discussion_r1377399579

jh66637:
On my laptop, I observed performance differences up to 2-5%. I think this has to be benchmarked on a cluster to get reliable results. However, I don't have access to one atm.

PR #644, opened 2024-01-25
Blend-in aero-acoustic source term

jh66637:
As indicated in https://github.com/exadg/exadg/pull/635 a few small follow-up PRs are necessary to use the aero-acoustic solver in realistic settings. 

This PR introduces a way to blend in source terms. Mostly this blend-in is performed in time: I.e. the CFD solver is running and at some point, the acoustic solver starts to run as well. To avoid artificial waves due to the sudden application of the source term these source terms are typically slowly increased over time in the aero-acoustic community. 

While this is mostly enough and I didn't have problems in my test cases, in literature it is often reported that source terms are spatially faded out by a blending as well. I.e., to avoid large gradients at CFD mesh boundaries. The CFD mesh is typically smaller than the acoustic mesh and both meshes are coupled arbitrarily, so one can imagine this becoming a potential error source.


include/exadg/aero_acoustic/calculators/source_term_calculator.h
```diff
@@ -108,16 +116,30 @@ class SourceTermCalculator
...
+      if(data.blend_in && data.blend_in_function->needs_evaluation_at_time(time))
+      {
+        for(unsigned int q = 0; q < dpdt.n_q_points; ++q)
+        {
+          auto const scale = FunctionEvaluator<0, dim, Number>::value(*data.blend_in_function,
+                                                                      dpdt.quadrature_point(q),
+                                                                      time);
+          dpdt.submit_value(scale * dpdt.get_value(q), q);
+        }
+      }
```
nfehn, line 129: I am not sure if this is implemented correctly. There is already an if-else-statement above, so `dpdt.submit_value()` has been already been called before. In case you exploit intentionally that the code is still correct, I am against this programming style, because it would rely on a detail of the `MatrixFree` implementation and hard to understand on the ExaDG-level.

In my opinion, the code should be written like `dpdt.submit_value(flux, q);`, where `flux` is computed depending on the parameter settings and different formulations, e.g.

```C++
// compute "flux"
flux = ...

if(data.blend_in)
{
  ...
  flux *= scale;
}

dpdt.submit_value(flux, q);
```

include/exadg/aero_acoustic/calculators/source_term_calculator.h
```diff
@@ -108,16 +116,30 @@ class SourceTermCalculator
           dpdt.submit_value(-rho * dpdt.get_value(q), q);
       }
 
+      // apply blend in function before integration
+      if(data.blend_in && data.blend_in_function->needs_evaluation_at_time(time))
```
nfehn, line 120: 1. `&&` -> `and`

2. shared_ptr `data.blend_in_function` might be invalid, so this line could be problematic in my opinion. What is your intention? Should a user always have to define a blend_in_function, or only when `data.blend_in = true`? I think one should write two consecutive if-statements. The assert on `data.blend_in_function.get()` could be done outside the cell loop, or even during setup.

include/exadg/utilities/optional_function.h
```diff
@@ -0,0 +1,54 @@
...
+
+#include <deal.II/base/function.h>
+
+namespace ExaDG
+{
+namespace Utilities
+{
+/**
+ * This is a @c dealii::Function, that provides information if it has
+ * to be evaluated at a given time.
```
nfehn, line 33: Meaning is unclear if one does not know the implementation.

include/exadg/aero_acoustic/user_interface/field_functions.h
```diff
@@ -0,0 +1,53 @@
...
+template<int dim>
+struct FieldFunctions
+{
+  /*
+   * The function source_term_blend_in is mainly used to blend in the source
+   * term over time. In some rare cases it might also be beneficial to fade
+   * out the source term in space: If large source term contributions are
+   * placed at the boundary of the CFD domain, there can be problems when
+   * interpolating to the non-matching acoustic domain (that is mostly
+   * much larger than the CFD domain).
```
nfehn, line 40: The meaning of "CFD" could be unclear here. I would write something like "the fluid field to which the acoustic field is coupled".

include/exadg/aero_acoustic/user_interface/field_functions.h
```diff
@@ -0,0 +1,53 @@
...
+{
+namespace AeroAcoustic
+{
+template<int dim>
+struct FieldFunctions
+{
+  /*
+   * The function source_term_blend_in is mainly used to blend in the source
+   * term over time. In some rare cases it might also be beneficial to fade
+   * out the source term in space: If large source term contributions are
```
nfehn, line 37: I did not fully understand "fade out in space": The `Function` would shrink the area in which the source term is active originally, or enlarge that area?

nfehn:
I had a detailed look at the code. I understood the basic idea. However, I think we need to improve readability of the implementation, i.e. how the source term gets multiplied by a blending function, and the documentation.

include/exadg/utilities/optional_function.h
```diff
@@ -0,0 +1,54 @@
...
+public:
+  using time_type = typename dealii::Function<dim, Number>::time_type;
+
+  OptionalFunction(const unsigned int n_components = 1, const time_type initial_time = 0.0)
+    : dealii::Function<dim, Number>(n_components, initial_time)
+  {
+  }
+
+  virtual bool
+  needs_evaluation_at_time(double const new_time) const = 0;
```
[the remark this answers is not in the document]
nfehn: In case of "fade out in space", this function needs to return `true` throughout the simulation, right? The more I think about it, I am less sure whether the design is good currently.

Overall, I do not understand the trade-offs made regarding computational efficiency, by which the current design seems to be influenced. You introduced `needs_evaluation_at_time()` in order to evaluate the function not too often if I understood it correctly. You also write above that "fade out in space" is used in rare cases. This would mean that in the majority of cases, the blending function is called in very quadrature point, even though a global call (once for the whole mesh) would be enough? Performance-wise, I consider calling the function unnecessarily for every q-point more relevant than calling it un-necessarily once every time step. Against this background, it would argue we need a function `bool varies_in_space()` (which is the conditional on callling `Function::value()`) and a function `double get_time_factor()` (returing the temporal factor by which to multiply the source term) for optimal performance (but no function `needs_evaluation_at_time()` combined with `Function::value()` always evaluating spatial and temporal dimensions). The alternative is option 3 described above (one single `Function::value()` for all spatial and temporal dimensions; note that this function will only be called if `data.blend_in = true`, so the cases without blending in space or time will not be slowed down by this additional functionality).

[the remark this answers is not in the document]
jh66637: > In case of "fade out in space", this function needs to return true throughout the simulation, right? 

Yes

I see what you mean. What do you think about completely seperating the spatial and temporal blending? We could have a function that deals with the spatial rampup and one that deals with the spatial blending. It is not intuitive to have a field function that returns a constant time factor for me. The question would however be, where the function for the temporal blending should live?

include/exadg/aero_acoustic/calculators/source_term_calculator.h
jh66637: I tried to implement a compromise here. I don't want to introduce an `if` statement in the quadrature point loop. Normally I would go for a templated function and `if constexpr`, but we had that discussion multiple times and seem to have different opinions (with good arguments on both sides) on the use of templates for those scenarios.

I can restructure to what you are suggesting for now (i.e., if statement in the innermost loop), but it would be great if someone with access to a cluster could benchmark the performance difference using a templated version with `if constexpr` and the version you are suggesting, similarly to https://github.com/exadg/exadg/issues/594.

include/exadg/aero_acoustic/calculators/source_term_calculator.h
jh66637: The order of the comparisons is crucial: If `data.blend_in==false`, the second condition will not be evaluated. I can see how this is not optimal during reading the code and am happy to change it once we agree on the further procedure in the other comments.

include/exadg/aero_acoustic/user_interface/field_functions.h
jh66637: It "shrinks" the area in the sense that one would apply a smooth function that is 1 in most of the domain and goes to 0 at boundaries.

jh66637:
You raised some good questions. Let's decide how we proceed with https://github.com/exadg/exadg/pull/644#discussion_r1471420495 and https://github.com/exadg/exadg/pull/644#discussion_r1471529906 before I resolve all the issues :)

include/exadg/utilities/optional_function.h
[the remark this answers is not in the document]
nfehn: Basically, a `dealii::Function` treats both spatial and temporal components. The interface of `Function::value()` does not look like this, and I can only guess where this comes from. Maybe one does not want to see a time argument for steady-state problems, In any case, this has been a design decision on the dealii side we have to live with. Nevertheless, this implies that we basically (can/should) use the same function to treat both spatial and temporal variations, unless we consider this too inefficient because in the majority of cases this function would be called unnecessarily. However, I believe this will also restrict functionality.

I think separating the spatial and temporal blending might be two restrictive. I think about two source terms in two different parts of the domain that become active at different times. How to realise this? The final source term is not simply `time_factor * spatial_factor`, while a normal `dealii::Function::value()` in combination with `dealii::Function::set_time()` would have this capability naturally.

include/exadg/aero_acoustic/user_interface/field_functions.h
nfehn: Could you add this explanation (if not already the case).

include/exadg/aero_acoustic/calculators/source_term_calculator.h
nfehn: In the current situation, I consider it most important that any other third person can easily understand the code and is able to modify it without introducing bugs. This implies in my opinion that we should not use "special tricks" that we know are  correct (in the current code version of dealii) but are not the intended usage of FEEvaluation (as far as I know). My opinion is we should introduce that if-statement, and since there have been no other suggestions, we should proceed here in the interest of making progress.

include/exadg/aero_acoustic/calculators/source_term_calculator.h
nfehn: Personally, I would like that we do not expect ExaDG developers to have this detailed knowledge about C++. Also C++ books like "Einstieg in C++" von Arnold Willemer (4. Auflage, S. 91/92) clearly do not recommend to realize some non-trivial logic inside if statements.

kronbichler: @nfehn I agree that non-trivial logic should be avoided inside `if` statements, but I have a pretty strong opinion that checking if a pointer is valid (first condition) and if yes, evaluate the data structure behind the pointer (second condition) is among the self-explanatory conditions of the C++ programming language. In my opinion, having two nested `if` statements is more difficult to read.

I would however suggest to write the check in a different form if possible,
```suggestion
      if(data.blend_in_function.get() != nullptr && data.blend_in_function->needs_evaluation_at_time(time))
```
because this more clearly documents the intent of the two checks.

include/exadg/aero_acoustic/calculators/source_term_calculator.h
kronbichler: To understand the internal behavior mentioned by @nfehn, let me ask one question regarding this loop: You use `dpdt.get_value()` here in line 127 after you called `dpdt.submit_value()` in line 111/116. When you call `dpdt.submit_value()` now in line 127, you end up multiplying with the Jacobian determinant and quadrature weight twice for the content added in line 111/116. Is this the intended behavior, and if yes, what is the underlying algorithm for it? I suggest to either document this better or write it more explicitly in the code, in agreement with @nfehn's comment.

kronbichler: > it would be great if someone with access to a cluster could benchmark the performance difference using a templated version with if constexpr and the version you are suggesting, similarly to https://github.com/exadg/exadg/issues/594.

I agree with this desire, we should definitely do that. I think after finishing the main acoustic PRs we should run the benchmarks and see how the performance is affected. Since HPC is the primary mission of this project, we should make sure we hit the right balance in terms of performance and code appearance.

kronbichler: I can run the benchmarks, if I get told what case to run.

jh66637: > you end up multiplying with the Jacobian determinant and quadrature weight twice for the content added in line 111/116. Is this the intended behavior
This is not the intended behavior. I will restructure as requested.
> I can run the benchmarks
I will add a PR with templated versions of some operators and ping you :) I don't know when I will find the time though.

include/exadg/utilities/optional_function.h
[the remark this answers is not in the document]
jh66637: I had a hard time to separate the comments above. I tried to resolve every issue in [refracture application of scaling factor](https://github.com/exadg/exadg/pull/644/commits/1026d662abd8a15b3a60b22b307db977ea36f10b)

include/exadg/aero_acoustic/user_interface/field_functions.h
jh66637: [605ba8a](https://github.com/exadg/exadg/pull/644/commits/605ba8a104f745ab2a0b65d3aedcd93134b7ea8a)

jh66637:
I think I addressed all the comments.

jh66637:
@nfehn I rebased to the latest version of master. The failing pipeline is not related to this PR.

nfehn:
I will look over the changes again.

The failing pipeline comes from deal.II, but this should now already be fixed in ExaDG.

include/exadg/aero_acoustic/calculators/source_term_calculator.h
```diff
@@ -87,6 +95,25 @@ class SourceTermCalculator
 
     Number rho = static_cast<Number>(data.density);
 
+    // In case we blend in the source term we check, if the scaling is space dependent. Only in that
+    // case we have to evaluate the function in every equadrature point. Otherwise the scaling
+    // purely temporal and constant during this function.
```
nfehn, line 100: Just a minor comment: The comma in the first sentence should be after "source term". In the second setence, "is" seems to be missing.

include/exadg/aero_acoustic/calculators/source_term_calculator.h
```diff
@@ -87,6 +95,25 @@ class SourceTermCalculator
...
+    bool const space_dependent_scaling =
+      data.blend_in_function != nullptr ? data.blend_in_function->varies_in_space() : false;
+    Number const pure_temporal_scaling_factor =
+      data.blend_in_function != nullptr ? data.blend_in_function->compute_time_factor() : 1.0;
+
+    auto apply_scaling = [&](auto & flux, auto const & q) {
+      if(space_dependent_scaling)
+        flux *= FunctionEvaluator<0, dim, Number>::value(*data.blend_in_function, q, time);
+      else
+        flux *= pure_temporal_scaling_factor;
```
nfehn, line 114: It took me some time to understand the complete logic behind this code.

I suggest to write

```C++
Number const pure_temporal_scaling_factor =
     (not space_dependent_scaling) ? data.blend_in_function->compute_time_factor() : 1.0;
```

It think this would be somewhat more intuitive and I think this would make clear that the variable `pure_temporal_scaling_factor` is expected to be used only for one particular case.

include/exadg/aero_acoustic/user_interface/field_functions.h
```diff
@@ -0,0 +1,55 @@
...
+   * placed at the boundary of the CFD domain, there can be problems when
+   * interpolating to the non-matching acoustic domain (that is mostly
+   * much larger than the CFD domain. A typical spatial fade out function
+   * "shrinks" the area of the source terms by a smooth function that is one
+   * in large parts of the domain and reduces to zero at boundaris.
```
nfehn, line 42: "e" missing in boudaris.

include/exadg/utilities/spatial_aware_function.h
```diff
@@ -0,0 +1,62 @@
...
+public:
+  using time_type = typename dealii::Function<dim, Number>::time_type;
+
+  SpatialAwareFunction(const unsigned int n_components = 1, const time_type initial_time = 0.0)
+    : dealii::Function<dim, Number>(n_components, initial_time)
+  {
+  }
+
+  /**
+   * Varies in space, i.e. is not constant over space at every time.
```
nfehn, line 47: It seems as if "at every time" is in conflict with the comment above "if the function varies in space for a given time" where you define the function in `FieldFunctions`.

To be more explicit, we  could add `double const time` to the interface of this function (and for the function `compute_time_factor()` below). I think this would make some of the comments superfluous.

include/exadg/aero_acoustic/calculators/source_term_calculator.h
```diff
@@ -87,6 +95,25 @@ class SourceTermCalculator
...
+    // case we have to evaluate the function in every equadrature point. Otherwise the scaling
+    // purely temporal and constant during this function.
+    if(data.blend_in)
+      AssertThrow(data.blend_in_function != nullptr,
+                  dealii::ExcMessage("No blend-in function provided."));
+
+    bool const space_dependent_scaling =
+      data.blend_in_function != nullptr ? data.blend_in_function->varies_in_space() : false;
+    Number const pure_temporal_scaling_factor =
+      data.blend_in_function != nullptr ? data.blend_in_function->compute_time_factor() : 1.0;
```
nfehn, line 108: For the function `compute_time_factor()` to be evaluated correctly, where do you call `function->set_time()` in the code? As written in comment https://github.com/exadg/exadg/pull/644/files#r1528630370, my current understanding is that we should add `double const time` to the interface of this function. However, I might misunderstand something.

include/exadg/utilities/spatial_aware_function.h
```diff
@@ -0,0 +1,62 @@
...
+#define EXADG_AERO_UTILITIES_SPATIAL_AWARE_FUNCTION_H_
+
+#include <deal.II/base/function.h>
+
+namespace ExaDG
+{
+namespace Utilities
+{
+/**
+ * This is a @c dealii::Function, that provides information if its result is constant over space at
```
nfehn, line 32: "constant", while below you write "varies_in_space". I would unify this.

include/exadg/utilities/spatial_aware_function.h
```diff
@@ -0,0 +1,62 @@
...
+
+  SpatialAwareFunction(const unsigned int n_components = 1, const time_type initial_time = 0.0)
+    : dealii::Function<dim, Number>(n_components, initial_time)
+  {
+  }
+
+  /**
+   * Varies in space, i.e. is not constant over space at every time.
+   */
+  virtual bool
```
nfehn, line 49: Please add the following documentation to this function:

"If this function evaluates to true, `dealii::Function::set_time()` in combination with `dealii::Function::value()` will be used to evaluate the function's value."

include/exadg/utilities/spatial_aware_function.h
```diff
@@ -0,0 +1,62 @@
...
+  varies_in_space() const = 0;
+
+  /**
+   * Compute the value that is constant over space at every time.
```
nfehn, line 53: Please add the following documentation to this function:

"If the above function `varies_in_space()`  returns `false`, this function will be used to evaluate the function's value, bypassing `dealii::Function::value()`.

include/exadg/utilities/spatial_aware_function.h
nfehn: Please add the following documentation here:

The intention behind is to bypass a potentially expensive evaluation of `dealii::Function::value()` (which the internal implementation will call for each quadrature point) in case one knows a priori that the function does not vary in space.

applications/aero_acoustic/co_rotating_vortex_pair/application.h
```diff
@@ -636,6 +636,44 @@ class Application : public ApplicationBase<dim, Number>
...
+  bool
+  varies_in_space() const final
+  {
+    return false;
+  }
+
+  double
+  compute_time_factor() const final
+  {
+    double const time = this->get_time();
```
nfehn, line 663: As written in another comment, I would not rely here on the dealii design choice of `dealii::Function` with "hidden" time variable. The function `compute_time_factor()` appears to be intrinsically time-dependent, which we should indicate by the interface of the function.

applications/aero_acoustic/co_rotating_vortex_pair/application.h
```diff
@@ -636,6 +636,44 @@ class Application : public ApplicationBase<dim, Number>
...
+public:
+  BlendInFunction(double const blend_in_start, double const blend_in_end)
+    : Utilities::SpatialAwareFunction<dim>(1, 0.0), start(blend_in_start), end(blend_in_end)
+  {
+  }
+
+  double
+  value(dealii::Point<dim> const &, unsigned int const) const final
+  {
+    return compute_time_factor();
```
nfehn, line 651: Do I understand the code correctly that we will never run into this function, because `varies_in_space()` is implemented here in this example such that it returns `false`?

If so, I consider this implementation of `Function::value()` irritating. Users who want to implement something similar will not know what they have to write here in my opinion. So I would leave this function empty with a short comment like "will not be used for this example" or similar.

applications/aero_acoustic/co_rotating_vortex_pair/application.h
nfehn: I mean this comment here https://github.com/exadg/exadg/pull/644#discussion_r1528639377

include/exadg/aero_acoustic/calculators/source_term_calculator.h
jh66637: [033b96f](https://github.com/exadg/exadg/pull/644/commits/033b96fe0ecdd3fda9a0e7f11e8d4c3420cfd04b)

include/exadg/aero_acoustic/calculators/source_term_calculator.h
jh66637: [f763625](https://github.com/exadg/exadg/pull/644/commits/f763625ed2c96df3c3b3e917c4c1085b58158e44)

include/exadg/aero_acoustic/user_interface/field_functions.h
jh66637: [1ce00f7](https://github.com/exadg/exadg/pull/644/commits/1ce00f78f859dd47018004e634729b5bbcd36841)

include/exadg/utilities/spatial_aware_function.h
jh66637: [ef52a6e](https://github.com/exadg/exadg/pull/644/commits/ef52a6ec89f956fff89d3ecea0b015c02f4e9a40)

include/exadg/aero_acoustic/calculators/source_term_calculator.h
jh66637: [ef52a6e](https://github.com/exadg/exadg/pull/644/commits/ef52a6ec89f956fff89d3ecea0b015c02f4e9a40)

include/exadg/utilities/spatial_aware_function.h
jh66637: [4776f56](https://github.com/exadg/exadg/pull/644/commits/4776f5622e28a7d092692b74ed7fe858ab539238)

include/exadg/utilities/spatial_aware_function.h
jh66637: [4776f56](https://github.com/exadg/exadg/pull/644/commits/4776f5622e28a7d092692b74ed7fe858ab539238)

include/exadg/utilities/spatial_aware_function.h
jh66637: [4776f56](https://github.com/exadg/exadg/pull/644/commits/4776f5622e28a7d092692b74ed7fe858ab539238)

jh66637:
@nfehn I think I addressed all your concerns. Let me know if this needs more work to get it merged :)

nfehn:
Great! I see nothing that would hinder us from merging :)

PR #645, opened 2024-01-25
Enable sub-time-stepping for aero-acoustics

jh66637:
As indicated in https://github.com/exadg/exadg/pull/635 a few small follow-up PRs are necessary to use the aero-acoustic solver in realistic settings.

This PR adds sub-time-stepping for the aero-acoustic solver. I.e.: The fluid solver runs with its normal time step size and the acoustic solver runs multiple time steps for one fluid time-step. This heavily reduces the work that has to be performed by the fluid solver.

### DIFF
--- a/include/exadg/acoustic_conservation_equations/time_integration/time_int_abm.h
+++ b/include/exadg/acoustic_conservation_equations/time_integration/time_int_abm.h
@@ -57,7 +57,8 @@ public:
         is_test_in),
       param(param_in),
       postprocessor(postprocessor_in),
-      pcout(std::cout, dealii::Utilities::MPI::this_mpi_process(mpi_comm_in) == 0)
+      pcout(std::cout, dealii::Utilities::MPI::this_mpi_process(mpi_comm_in) == 0),
+      initial_time_step_size(std::numeric_limits<double>::max())
   {
   }
 
@@ -75,38 +76,36 @@ private:
   {
     pcout << std::endl << "Calculation of time step size:" << std::endl << std::endl;
 
-    double time_step = std::numeric_limits<double>::max();
-
     if(param.calculation_of_time_step_size == TimeStepCalculation::UserSpecified)
     {
-      time_step = calculate_const_time_step(param.time_step_size, param.n_refine_time);
+      initial_time_step_size = calculate_const_time_step(param.time_step_size, param.n_refine_time);
 
-      print_parameter(pcout, "time step size", time_step);
+      print_parameter(pcout, "time step size", initial_time_step_size);
     }
     else if(param.calculation_of_time_step_size == TimeStepCalculation::CFL)
     {
       double const cfl = param.cfl / std::pow(2.0, param.n_refine_time);
 
-      time_step = cfl * this->get_underlying_operator().calculate_time_step_cfl();
+      initial_time_step_size = cfl * this->get_underlying_operator().calculate_time_step_cfl();
 
       this->pcout << std::endl
                   << "Calculation of time step size according to CFL condition:" << std::endl
                   << std::endl;
       print_parameter(this->pcout, "CFL", cfl);
-      print_parameter(this->pcout, "time step size", time_step);
+      print_parameter(this->pcout, "time step size", initial_time_step_size);
     }
 
-    return time_step;
+    return initial_time_step_size;
   }
 
   double
   recalculate_time_step_size() const final
   {
-    AssertThrow(not(param.calculation_of_time_step_size == TimeStepCalculation::UserSpecified),
-                dealii::ExcMessage(
-                  "Adaptive time step is not implemented for user specified time step size."));
-
-    return {};
+    // Currently the time step sice can not vary since the
+    // it depends only on the speed of sound that is
+    // constant over time. This changes once ALE is used
+    // or additional convective terms are considered.
+    return initial_time_step_size;
   }
 
   void
@@ -127,6 +126,11 @@ private:
   std::shared_ptr<PostProcessorInterface<Number>> postprocessor;
 
   dealii::ConditionalOStream pcout;
+
+  // Store initially compute time step size. We do this because the value
+  // will not change, but we need adaptive time-stepping for efficient
+  // sub-time stepping methods.
+  double initial_time_step_size;
 };
 
 } // namespace Acoustics
--- a/include/exadg/aero_acoustic/driver.cpp
+++ b/include/exadg/aero_acoustic/driver.cpp
@@ -139,17 +139,30 @@ Driver<dim, Number>::solve()
       timer.second.restart();
     }
 
+    // To check if acoustics starts during the following sub-stepping sweep we
+    // can not simply check acoustic->time_integrator->started(). Instead we
+    // compute this information as follows:
+    const bool acoustic_starts_during_sub_stepping =
+      fluid->time_integrator->get_next_time() + fluid->time_integrator->get_time_step_size() >
+      application->acoustic->get_parameters().start_time;
+
     // The acoustic simulation uses explicit time-stepping while the fluid solver
     // uses implicit time-stepping. Therefore, we advance the acoustic solver to
     // t^(n+1) first and directly use the result in the fluid solver.
-    if(acoustic->time_integrator->started())
+    if(acoustic_starts_during_sub_stepping)
       couple_fluid_to_acoustic();
     acoustic->advance_multiple_timesteps(fluid->time_integrator->get_time_step_size());
 
-    // TODO: we have to interpolate the source term also if the acoustic solver will be started
-    // during the sub-time-stepping sweep!
-    fluid->advance_one_timestep_and_compute_pressure_time_derivative(
-      acoustic->time_integrator->started());
+    // We can not simply check acoustic->time_integrator->started() since acoustic might start
+    // during a sub-stepping sweep. To check if the acoustic might be started during the next
+    // sub-stepping BEFORE performing the current fluid time step we have to check if dt_{n+2}
+    // is larger than the acoustic start time. We need this information BEFORE the fluid
+    // time-step since we have to know if we have to compute dp/dt.
+    const bool acoustic_might_start =
+      fluid->time_integrator->get_next_time() + fluid->max_next_time_step_size() >
+      application->acoustic->get_parameters().start_time;
+
+    fluid->advance_one_timestep_and_compute_pressure_time_derivative(acoustic_might_start);
   }
 
   time_solvers_side_by_side = timer.second.wall_time();
--- a/include/exadg/aero_acoustic/driver.cpp
+++ b/include/exadg/aero_acoustic/driver.cpp
@@ -146,6 +146,8 @@ Driver<dim, Number>::solve()
       couple_fluid_to_acoustic();
     acoustic->advance_multiple_timesteps(fluid->time_integrator->get_time_step_size());
 
+    // TODO: we have to interpolate the source term also if the acoustic solver will be started
+    // during the sub-time-stepping sweep!
     fluid->advance_one_timestep_and_compute_pressure_time_derivative(
       acoustic->time_integrator->started());
   }
--- a/include/exadg/aero_acoustic/driver.cpp
+++ b/include/exadg/aero_acoustic/driver.cpp
@@ -36,7 +36,8 @@ Driver<dim, Number>::Driver(MPI_Comm const &                              comm,
     is_test(is_test),
     application(app),
     acoustic(std::make_shared<SolverAcoustic<dim, Number>>()),
-    fluid(std::make_shared<SolverFluid<dim, Number>>())
+    fluid(std::make_shared<SolverFluid<dim, Number>>()),
+    time_solvers_side_by_side(std::numeric_limits<double>::min())
 {
   print_general_info<Number>(pcout, mpi_comm, is_test);
 }
@@ -108,19 +109,6 @@ Driver<dim, Number>::set_start_time() const
 
 template<int dim, typename Number>
 void
-Driver<dim, Number>::synchronize_time_step_size() const
-{
-  // TODO: to avoid sophisticated methods for the minimal solver we simply
-  // use the smallest time step.
-  double const dt = std::min(fluid->time_integrator->get_time_step_size(),
-                             acoustic->time_integrator->get_time_step_size());
-
-  acoustic->time_integrator->set_current_time_step_size(dt);
-  fluid->time_integrator->set_current_time_step_size(dt);
-}
-
-template<int dim, typename Number>
-void
 Driver<dim, Number>::couple_fluid_to_acoustic()
 {
   dealii::Timer sub_timer;
@@ -135,9 +123,9 @@ template<int dim, typename Number>
 void
 Driver<dim, Number>::solve()
 {
-  set_start_time();
+  std::pair<bool, dealii::Timer> timer = std::make_pair(false, dealii::Timer());
 
-  synchronize_time_step_size();
+  set_start_time();
 
   AssertThrow(std::abs(application->fluid->get_parameters().end_time -
                        application->acoustic->get_parameters().end_time) < 1.0e-12,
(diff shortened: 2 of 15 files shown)


jh66637:
I have to replace the output files for the acoustic tests due to a new parameter that is printed (adaptive timestepping). Other than that this is ready :)

jh66637:
@nfehn can we proceed here once we agree on https://github.com/exadg/exadg/pull/644

nfehn:
sure. So I wait with the review of this PR until #644 is merged?

jh66637:
I think this is completely independent, so you could review it already if you find the time :)

include/exadg/acoustic_conservation_equations/time_integration/time_int_abm.h
```diff
@@ -127,6 +126,11 @@ class TimeIntAdamsBashforthMoulton
   std::shared_ptr<PostProcessorInterface<Number>> postprocessor;
 
   dealii::ConditionalOStream pcout;
+
+  // Store initially compute time step size. We do this because the value
+  // will not change, but we need adaptive time-stepping for efficient
+  // sub-time stepping methods.
+  double initial_time_step_size;
```
nfehn, line 133: I did not understand this comment when lookint at this in an isolated way. Maybe it becomes clear once I have seen the rest of the code.

include/exadg/aero_acoustic/driver.cpp
```diff
@@ -135,26 +123,49 @@ template<int dim, typename Number>
...
+      timer.first = true;
+      timer.second.restart();
+    }
+
+    // To check if acoustics starts during the following sub-stepping sweep we
+    // can not simply check acoustic->time_integrator->started(). Instead we
+    // compute this information as follows:
+    const bool acoustic_starts_during_sub_stepping =
+      fluid->time_integrator->get_next_time() + fluid->time_integrator->get_time_step_size() >
+      application->acoustic->get_parameters().start_time;
```
nfehn, line 147: I think we can improve the naming of `acoustic_starts_during_sub_stepping` and `acoustics_might_start`. The "sub-stepping" nomenclature should not be necessary on this level. I would suggest

- `acoustic_starts_during_present_timestep`
- `acoustic_might_start_during_next_timestep`

include/exadg/aero_acoustic/driver.cpp
```diff
@@ -135,26 +123,49 @@ template<int dim, typename Number>
...
                        application->acoustic->get_parameters().end_time) < 1.0e-12,
               dealii::ExcMessage("Acoustic and fluid simulation need the same end time."));
 
   while(not acoustic->time_integrator->finished())
   {
+    if(timer.first == false && acoustic->time_integrator->started())
+    {
+      timer.first = true;
+      timer.second.restart();
+    }
```
nfehn, line 140: Could you explain this? Does this mean that a pure solution of the fluid problem is ignored in terms of the timings?

`&&` -> `and`

From a software perspective, the statement `acoustic->time_integrator->started()` inside a while-loop `while(not acoustic->time_integrator->finished())` looks somewhat unclear to me. This means the while-loop is executed even though the `acoustic` solver might not even have started!? It does seem to be fine here, but theoretically, can we really rely on `acoustic->time_integrator->finished() == false` in case `acoustic->time_integrator->started() = false`? Suggestion: Could we maybe use the `fluid` object for the outer while loop? I think this would improve code readability and explain more clearly the circumstance that `acoustic` is switched on at some point.

include/exadg/aero_acoustic/single_field_solvers/acoustics.h
```diff
@@ -68,6 +68,43 @@ class SolverAcoustic
     time_integrator->setup(application->get_parameters().restarted_simulation);
   }
 
+  void
+  advance_multiple_timesteps(double const global_dt)
+  {
+    double const dt = compute_sub_time_step_size(global_dt);
+
+    // define epsilon dependent on time-step size to avoid
+    // wrong terminations of the sub-timestepping loop.
```
nfehn, line 77: "... due to round-off errors." ?

include/exadg/aero_acoustic/single_field_solvers/acoustics.h
```diff
@@ -68,6 +68,43 @@ class SolverAcoustic
...
+
+    if(time_integrator->started())
+    {
+      ++sub_time_steps.first;
+      sub_time_steps.second += n_sub_time_steps;
+    }
+  }
+
+  double
+  get_average_sub_time_steps() const
```
nfehn, line 97: please extend name to  `_average_number_of_`

include/exadg/aero_acoustic/single_field_solvers/acoustics.h
```diff
@@ -80,6 +117,26 @@ class SolverAcoustic
...
+      return global_dt;
+    }
+    else
+    {
+      // otherwise, ensure that sub-timestepping ends exactly at global_dt
+      return adjust_time_step_to_hit_end_time(0.0, global_dt, current_dt);
+    }
+  }
+
+  std::pair<unsigned int /* calls */, unsigned long long /* sub_dt */> sub_time_steps;
```
nfehn, line 139: Related to comment https://github.com/exadg/exadg/pull/645/files#r1561011435, "calls" and "sub_dt" looks somewhat misleading to me. What about "number of macro time steps" and "total number of sub time steps"?

include/exadg/aero_acoustic/driver.cpp
```diff
@@ -188,18 +197,28 @@ Driver<dim, Number>::print_performance_results(double const total_time) const
...
 
+  // Throughput in DoFs/s per time step per core (during the time both
+  // solvers ran side by side)
+  dealii::Utilities::MPI::MinMaxAvg time_solvers_side_by_side_data =
+    dealii::Utilities::MPI::min_max_avg(time_solvers_side_by_side, mpi_comm);
+  double const time_solvers_side_by_side_avg = time_solvers_side_by_side_data.avg;
+
+  dealii::types::global_dof_index const DoFs =
+    fluid->pde_operator->get_number_of_dofs() +
+    acoustic->get_average_sub_time_steps() * acoustic->pde_operator->get_number_of_dofs();
```
nfehn, line 208: Related to comments https://github.com/exadg/exadg/pull/645/files#r1561011435 and https://github.com/exadg/exadg/pull/645/files#r1561016787, I asked myself if we should actually calculate the throughput like this. If the number of acoustic sub steps is very large, the throughput is dominated by the acoustic problem, right? But I am not sure if this is the intuitive interpretation of "throughput" from a user perspective. From such a user perspective, I am not sure if I would actually like to see the throughput per macro time step with the acoustic DoFs counted only once. So I am arguing that choosing sub-stepping is a detail of the implementation (as an alternative to implicit time stepping). A high throughput should in my opinion indicate a fast computation per macro time step, but is this really the case if e.g. the fluid solver is slow but the number of acoustic sub steps is very large?

include/exadg/aero_acoustic/single_field_solvers/acoustics.h
```diff
@@ -80,6 +117,26 @@ class SolverAcoustic
 
   // postprocessor
   std::shared_ptr<Acoustics::PostProcessorBase<dim, Number>> postprocessor;
+
+private:
+  double
+  compute_sub_time_step_size(double const global_dt)
+  {
+    double const current_dt = time_integrator->get_time_step_size();
```
nfehn, line 125: Does `time integrator->get_time_step_size()` return a sub-time step here, not a global time step? If so, I think the name `current_dt` should rather be `sub_dt`. I also think implementing the content of this function at the place where this function is called would actually make the code easier to understand. The reason is that I originally thought the function name `compute_sub_time_step_size()` implies this function computes a sub-time step given a global time step, like `sub_dt = global_dt / n`, which does not seem to be the case here. I rather understand this function as `get_and_potentially_adjust_sub_time_step()`. It seems as if this function is only used once, hence, the suggestion to maybe move the code there.

nfehn:
This looks good overall! As written in the review comments, I see two major points of discussion: i) computation of the sub time step, ii) throughput calculation

include/exadg/acoustic_conservation_equations/time_integration/time_int_abm.h
nfehn: This comment is still not clear to me.

jh66637: For adaptive time stepping, we need `recalculate_time_step_size()`.

Since the time step size can not change as long as there is no background convection, `recalculate_time_step_size()` will always return the same value. To avoid computing the CFL time step size every time step, we simply store it as a member and return it.

include/exadg/aero_acoustic/single_field_solvers/acoustics.h
jh66637: [a132791](https://github.com/exadg/exadg/pull/645/commits/a132791a4dc04a4b4bfd4286c0c6f9c2638ed074)

include/exadg/aero_acoustic/single_field_solvers/acoustics.h
jh66637: [e4a3c21](https://github.com/exadg/exadg/pull/645/commits/e4a3c212a070b3a5001a68fb3583f88667ec37dc)

include/exadg/aero_acoustic/single_field_solvers/acoustics.h
jh66637: [0c319a8](https://github.com/exadg/exadg/pull/645/commits/0c319a8489f99e5a4850613c92030e09e39c27e1)

jh66637:
Thank you for the review :)

I resolved all your comments besides the throughput computation. When writing the code I was not sure how to handle this. Ignoring the pure fluid computations seemed legit to me since we are most certainly interested in the results of the aero-acoustic computations.

For me, it was interesting to see the throughput/macro time step but I see your point. I am happy to change this, but I am not sure in which direction you are targeting.

include/exadg/aero_acoustic/driver.cpp
```diff
@@ -135,26 +123,50 @@ template<int dim, typename Number>
...
+    // compute this information as follows:
+    const bool acoustic_starts_during_present_timestep =
+      fluid->time_integrator->get_next_time() + fluid->time_integrator->get_time_step_size() >
+      application->acoustic->get_parameters().start_time;
+
     // The acoustic simulation uses explicit time-stepping while the fluid solver
     // uses implicit time-stepping. Therefore, we advance the acoustic solver to
     // t^(n+1) first and directly use the result in the fluid solver.
-    if(acoustic->time_integrator->started())
+    if(acoustic_starts_during_sub_stepping)
```
nfehn, line 152: This name probably needs to be adjusted?

include/exadg/aero_acoustic/single_field_solvers/acoustics.h
```diff
@@ -68,6 +68,48 @@ class SolverAcoustic
...
   }
 
+  void
+  advance_multiple_timesteps(double const macro_dt)
+  {
+    // in case the macro time step is smaller than the acoustic timestep, simply run
+    // with macro time step, otherwise, ensure that n_sub_time_steps * sub_dt = macro_dt
+    double sub_dt = time_integrator->get_time_step_size();
+    if(macro_dt > sub_dt)
+      sub_dt = adjust_time_step_to_hit_end_time(0.0, macro_dt, sub_dt);
```
nfehn, line 78: Do we need `else{ sub_dt = macro_dt;}` to make sure that fluid and acoustic stay in sync regarding macro time steps?

include/exadg/aero_acoustic/driver.cpp
nfehn: This is still open. Maybe let me reformulate this comment: 

Current throughput definition: $T = \frac{N_{DoF, fluid} + n_{sub} N_{Dof, acoustic}}{t_{fluid} + t_{acoustic}} = \frac{N_{DoF, fluid}/n_{sub} + N_{Dof, acoustic}}{\frac{t_{fluid} + t_{acoustic}}{n_{sub}}}$

Doesn't this mean we currently compute the throughput of a single acoustic sub step? From an application perspective, I would currently argue we should compute the throughput of a macro time step $T = \frac{N_{DoF, fluid} + N_{Dof, acoustic}}{t_{fluid} + t_{acoustic}}$.

One can maybe argue for both variants. However, I am not sure if the screen output "Throughput while both solvers ran side by side:" is correct right now. For the present implementation, I would call this quantity "Throughput related to one acoustic sub step.".

There would be yet another solution, namely to simply compute and print both metrics.

What do you think?

include/exadg/aero_acoustic/single_field_solvers/acoustics.h
jh66637: Yes, this makes it more clear: [fix sumtimestep computation](https://github.com/exadg/exadg/pull/645/commits/e8d1525c2ec742cb5520c89ea43c19fcc5c9ea31) :)

include/exadg/aero_acoustic/driver.cpp
jh66637: [fix compilation](https://github.com/exadg/exadg/pull/645/commits/8be5473532ecc37cd1fcb15a56207898a8759850)

include/exadg/aero_acoustic/driver.cpp
jh66637: Thank you for the clarification. I agree in parts. I would say, the first definition should be related to the macro time step and the second should be related to the sub time step. I.e., $N_{DoF,f}+n_{sub}N_{DoF,a}$ is the number of DoFs that is solved (in average) during one macro time step while $N_{DoF,f}+N_{DoF,a}$ DoFs are solved during one sub time step. 

I added the corresponding lines in [222c01b](https://github.com/exadg/exadg/pull/645/commits/222c01bac0b8ad70844d6f5b71110f27ffc215ca). In case you do not agree, could you explain where you think above reasoning is not correct? :)

Since $n_{sub}$ is in our case the average number of sub time steps per macro time step, the average number of DoFs per macro time step are represented by a floating point number. To account for this, I added an overload  for the function that prints the stats [05b188d](https://github.com/exadg/exadg/pull/645/commits/05b188db6abe7a93d73e9c6b4a9d376c335b3247).

nfehn: I think it should be the other way round. The first equation in my comment above shows the throughput related to a sub-step (where I have added the second version of this equation in order to highlight that this refers indeed to a sub-step). I don't agree that $N_{Dofs, f} + N_{Dofs, a}$ are the DoFs solved during one sub time step. This would be $N_{Dofs, f}/n_{sub} + N_{Dofs, a}$, and in this case one has to divide the wall time by $n_{sub}$. So the two versions of the first equation are equivalent and just reflect "different perspectives".

The second equation in my comment above refers to a macro time step, where the sub-stepping done for acoustics is kind of an irrelevant detail. Hence, acoustic dofs are only counted once. Think of using implicit time stepping for acoustics as an alternative. A reasonable throughput definition should allow in my opinion to compare these two timestepping variants quantitatively (higher throughput = faster solution for the same macro time step size), meaning they have to use the same throughput definition (related to one macro time step). The first equation instead refers to a sub-step and, hence, does not allow a reasonable comparison of these two timestepping variants. It is a metric that is only relevant for the "special case" using sub-stepping. So maybe it helps if you think how to implement the printing of the throughput metrics for these two time stepping variants and which if-statements we would need for both cases. Put differently, we can look which of the equations depends on the parameter $n_{sub}$ and which does not.

nfehn:
> For me, it was interesting to see the throughput/macro time step but I see your point. I am happy to change this, but I am not sure in which direction you are targeting.

I think my point is that I do not agree with the first sentence and that this would then lead to misunderstandings for users.

include/exadg/aero_acoustic/driver.cpp
jh66637: Thank you for the clarification. I think I finally understood your reasoning :) [6a024cd](https://github.com/exadg/exadg/pull/645/commits/6a024cd4c59a5c7d5e2266b34be6adc6e621e569)

include/exadg/aero_acoustic/driver.cpp
```diff
@@ -214,14 +214,14 @@ Driver<dim, Number>::print_performance_results(double const total_time) const
...
+                            DoFs_f + DoFs_a,
                             time_solvers_side_by_side_avg,
                             acoustic->get_number_of_macro_time_steps(),
                             N_mpi_processes);
 
   pcout << std::endl << "Throughput related to one sub time step:";
   print_throughput_unsteady(pcout,
-                            DoFs_f + DoFs_a,
+                            (double)DoFs_f / sub_dt_per_macro_dt + (double)DoFs_a,
                             time_solvers_side_by_side_avg,
```
nfehn, line 225: I just had a brief look again at this PR. I have given two equivalent formulas in comment https://github.com/exadg/exadg/pull/645#discussion_r1562664109 for the throughput per time step, and I currently believe if we compute the dofs like this, we have to divide the wall time by the number of substeps. Alternatively, use the equation for the number of dofs which you have deleted above in line 217 in the last commit 6a024cd4c59a5c7d5e2266b34be6adc6e621e569, or use the old code and just switch the output "macro time step" and "sub time step" in the `pcout` commands.

nfehn (approved):

PR #649, opened 2024-02-22
Move revision file into build folder

bergbauer:
We should move generated files to the include folder like deal.II also does.


kronbichler (approved):

nfehn:
no concerns from my side :)

PR #654, opened 2024-03-26
Make sure that Structure::MultigridPreconditioner is updated

### LINKED ISSUE #652: Preconditioner setup for steady linear elasticity solver wrong
(this is more a reminder that I want to fix this in the future, but I do not have time rn)
If I remember correctly, the idea was that the solvers themselves execute a setup of the preconditioner when `solve()` is called on them. For the nonlinear (`QuasiStatic`, `Unsteady` and `Steady`) solvers, all works fine. In the linear case -- the only one that calls `solve_linear()` I believe -- the Chebyshev smoother tells me hasn't got a preconditioner. When using `Unsteady`, the setup works, so its some case we did not consider yet.

--------------------------------------------------------
An error occurred in line <3545> of file </home/richard/dealii-candi/dealii/include/deal.II/lac/precondition.h> in function
    dealii::PreconditionChebyshev<MatrixType, VectorType, PreconditionerType>::EigenvalueInformation dealii::PreconditionChebyshev<MatrixType, VectorType, PreconditionerType>::estimate_eigenvalues(const VectorType&) const [with MatrixType = ExaDG::MultigridOperatorBase<3, float>; VectorType = dealii::LinearAlgebra::distributed::Vector<float, dealii::MemorySpace::Host>; PreconditionerType = ExaDG::JacobiPreconditioner<ExaDG::MultigridOperatorBase<3, float> >]
The violated condition was: 
    data.preconditioner.get() != nullptr
Additional information: 
    (none)

Stacktrace:
-----------
#0  /home/richard/dealii-candi/exadg/build/libexadg.so: dealii::PreconditionChebyshev<ExaDG::MultigridOperatorBase<3, float>, dealii::LinearAlgebra::distributed::Vector<float, dealii::MemorySpace::Host>, ExaDG::JacobiPreconditioner<ExaDG::MultigridOperatorBase<3, float> > >::estimate_eigenvalues(dealii::LinearAlgebra::distributed::Vector<float, dealii::MemorySpace::Host> const&) const
#1  /home/richard/dealii-candi/exadg/build/libexadg.so: dealii::PreconditionChebyshev<ExaDG::MultigridOperatorBase<3, float>, dealii::LinearAlgebra::distributed::Vector<float, dealii::MemorySpace::Host>, ExaDG::JacobiPreconditioner<ExaDG::MultigridOperatorBase<3, float> > >::vmult(dealii::LinearAlgebra::distributed::Vector<float, dealii::MemorySpace::Host>&, dealii::LinearAlgebra::distributed::Vector<float, dealii::MemorySpace::Host> const&) const
#2  /home/richard/dealii-candi/exadg/build/libexadg.so: ExaDG::ChebyshevSmoother<ExaDG::MultigridOperatorBase<3, float>, dealii::LinearAlgebra::distributed::Vector<float, dealii::MemorySpace::Host> >::vmult(dealii::LinearAlgebra::distributed::Vector<float, dealii::MemorySpace::Host>&, dealii::LinearAlgebra::distributed::Vector<float, dealii::MemorySpace::Host> const&) const
#3  /home/richard/dealii-candi/exadg/build/libexadg.so: ExaDG::MultigridAlgorithm<dealii::LinearAlgebra::distributed::Vector<float, dealii::MemorySpace::Host>, ExaDG::MultigridOperatorBase<3, float>, ExaDG::SmootherBase<dealii::LinearAlgebra::distributed::Vector<float, dealii::MemorySpace::Host> > >::v_cycle(unsigned int, bool) const
#4  /home/richard/dealii-candi/exadg/build/libexadg.so: ExaDG::MultigridPreconditionerBase<3, double, float>::vmult(dealii::LinearAlgebra::distributed::Vector<double, dealii::MemorySpace::Host>&, dealii::LinearAlgebra::distributed::Vector<double, dealii::MemorySpace::Host> const&) const
#5  /home/richard/dealii-candi/exadg/build/libexadg.so: void dealii::SolverFGMRES<dealii::LinearAlgebra::distributed::Vector<double, dealii::MemorySpace::Host> >::solve<ExaDG::Structure::LinearOperator<3, double>, ExaDG::PreconditionerBase<double> >(ExaDG::Structure::LinearOperator<3, double> const&, dealii::LinearAlgebra::distributed::Vector<double, dealii::MemorySpace::Host>&, dealii::LinearAlgebra::distributed::Vector<double, dealii::MemorySpace::Host> const&, ExaDG::PreconditionerBase<double> const&)
#6  /home/richard/dealii-candi/exadg/build/libexadg.so: ExaDG::Krylov::SolverFGMRES<ExaDG::Structure::LinearOperator<3, double>, ExaDG::PreconditionerBase<double>, dealii::LinearAlgebra::distributed::Vector<double, dealii::MemorySpace::Host> >::solve(dealii::LinearAlgebra::distributed::Vector<double, dealii::MemorySpace::Host>&, dealii::LinearAlgebra::distributed::Vector<double, dealii::MemorySpace::Host> const&) const
#7  /home/richard/dealii-candi/exadg/build/libexadg.so: ExaDG::Structure::Operator<3, double>::solve_linear(dealii::LinearAlgebra::distributed::Vector<double, dealii::MemorySpace::Host>&, dealii::LinearAlgebra::distributed::Vector<double, dealii::MemorySpace::Host> const&, double, double, double, bool) const
#8  /home/richard/dealii-candi/exadg/build/libexadg.so: ExaDG::Structure::DriverSteady<3, double>::do_solve()
#9  /home/richard/dealii-candi/exadg/build/libexadg.so: ExaDG::Structure::DriverSteady<3, double>::solve()
#10  /home/richard/--------------------------------------------------------------------------

richardschu:
fixes #652
The important bit is in
https://github.com/richardschu/exadg/blob/04428487cb27f3faf8ad74e64fdd99cb8af3a92d/include/exadg/structure/preconditioners/multigrid_preconditioner.cpp#L125
where we now have:
``` 
  // In case that the operators have been updated, we also need to update the smoothers and the
  // coarse grid solver. This is generic functionality implemented in the base class.
  if(nonlinear or data.unsteady or this->update_needed)
  {
    this->update_smoothers();
    this->update_coarse_solver();
  }
```

The `this->update_needed()` was missing before, but the block behind the if is to be executed as well.

and also adds some tests to check some of the most important preconditioners.
The application was modified to get cheaper tests.


nfehn:
Thanks for figuring this out. In my opinion, there are way too many new test cases in this PR. Shouldn't there actually be one test case enough for this tiny change in the code?

include/exadg/structure/preconditioners/multigrid_preconditioner.cpp
```diff
@@ -122,7 +122,7 @@ MultigridPreconditioner<dim, Number>::update()
 
   // In case that the operators have been updated, we also need to update the smoothers and the
   // coarse grid solver. This is generic functionality implemented in the base class.
-  if(nonlinear or data.unsteady)
+  if(nonlinear or data.unsteady or this->update_needed)
```
nfehn, line 125: I agree that in the linear steady case we currently never run into this part of the code, which is a problem. However, this if-statement looks somewhat irritating to me. Isn't the consequence of what you found out that we should remove the if-statement completely?  In which cases, where we run into `preconditioner->update()`, do we not want to update the smoother/coarse solver. When comparing e.g. with the ConvDiff MG preconditioner, we do not have any if-statement before these update-calls? Could you please check this.

include/exadg/solvers_and_preconditioners/solvers/iterative_solvers_dealii_wrapper.h
```diff
@@ -282,7 +282,9 @@ class SolverGMRES : public SolverBase<VectorType>
     }
 
     AssertThrow(std::isfinite(solver_control.last_value()),
-                dealii::ExcMessage("Solver contained NaN of Inf values"));
+                dealii::ExcMessage("Convergence value of last iteration step "
+                                   "for which check was called by the solver "
```
nfehn, line 286: I do not understand what you mean by "for which check was called" and "convergence value".

I think the present PR should just fix the setup-problem (potentially with a single new test case).

I found three instances of this exc message in that file. So why did you change only some of them? Let's keep these changes out of the present PR.

richardschu: This is from the deal.II docu:
"Return the convergence value of last iteration step for which check was called by the solver."
(https://www.dealii.org/current/doxygen/deal.II/classSolverControl.html#a4a3b4da9834a3a5f1da12fa8fcc074f0)
Acc. to the docu, it is not always the residual that is checked
"To this end, the virtual function [check()](https://www.dealii.org/current/doxygen/deal.II/classSolverControl.html#a56fdf4c8d099edaca6fad443d7c76649) is called in each iteration with the current iteration step and the value indicating convergence (usually the residual)."
Within ExaDG, I think residual would actually be correct.

I only changed these because I did not see the other one.
Initially, I only saw the typo.

richardschu:
> Thanks for figuring this out. In my opinion, there are way too many new test cases in this PR. Shouldn't there actually be one test case enough for this tiny change in the code?

I would vote for testing what we do not test elsewhere, I can reduce the number of tests therefore, but I think quite some combinations are never tested.

kronbichler:
> In my opinion, there are way too many new test cases in this PR.

I think that we should considerably increase the number of tests, given that many combinations and steps are not tested. While it is hard to have definite numbers, I would consider 500-2000 tests for a project of this size not unreasonable. Of course, it would be necessary to structure our tests in a way to ensure rapid development cycles, e.g. by having a subset of quick tests or similar actions we can elaborate on. Having uncovered features is much more concerning in the longer run, which is why I am very much in favor of adding all tests that address previously untested features.

include/exadg/structure/preconditioners/multigrid_preconditioner.cpp
richardschu: I will check soon, if `(nonlinear or data.unsteady or this->update_needed)` actually ever evaluates to `false` and then push the new version.

nfehn:
In my opinion its too complex to test all combinations of parameter in ExaDG. Also the code does not have the complexity that all parameters are coupled and need to be tested in all the combinations. You found that the linear steady case is not tested, right? So why not just test this one?

Also related to your comment, @kronbichler: What I do not understand here is why we discuss the "big philosophical problem of testing" and not the problem we currently have here. It should also be part of the discussion that many many tests take part of our (expensive) development time and require resources for maintenance. Also right now as a reviewer, I actually need to invest too much time into such a tiny change. I am not complaining here, it's just that I believe there are more arguments to the discussion.

nfehn:
As an addition: It might turn out that the if-statement can be removed entirely. Shouldn't we then actually rethink the addition of many new tests here? So as a suggestion for the future: Creating a pull request with changing just the line of the if-statement, briefly discussing this (and whether we need tests for this or not; as I said, I am not sure if we already have the final variant regarding the if-statement; it might be that the code can be "simplified" compared to the previous version), and then continue (based on this discussion) could be more effective overall.

richardschu:
> As an addition: It might turn out that the if-statement can be removed entirely. Shouldn't we then actually rethink the addition of many new tests here? So as a suggestion for the future: Creating a pull request with changing just the line of the if-statement, briefly discussing this (and whether we need tests for this or not; as I said, I am not sure if we already have the final variant regarding the if-statement; it might be that the code can be "simplified" compared to the previous version), and then continue (based on this discussion) could be more effective overall.

Yes, this is what I will check: https://github.com/exadg/exadg/pull/654#discussion_r1538940917
The cases tested elsewhere, I will also remove!
Then we can have a look at whats left :)

include/exadg/structure/preconditioners/multigrid_preconditioner.cpp
richardschu: the condition is now completely removed, since it always gave true for any of the combinations tested (Steady, Unsteady, QuasiStatic with finite or infinitesimal strains (and AMG, PointJacobi and Multigrid preconditioners)).

nfehn:
For me, the reasoning is not clear why to have "redundant" tests e.g. for AMG and "multigrid". Note that "Multigrid" still has many parameters and that "Multigrid" internally contains AMG, see e.g.:

```bash
  Multigrid type:                            phMG
  p-sequence:                                Bisect
  Smoother:                                  Chebyshev
  Preconditioner smoother:                   PointJacobi
  Iterations smoother:                       5
  Smoothing range:                           2.0000e+01
  Iterations eigenvalue estimation:          20
  Coarse grid solver:                        CG
  Coarse grid preconditioner:                AMG
```
How would you motivate the need for this from the code? If I understand the implementation correctly, the Structure module uses a "generic" AMG preconditioner, not a "structure-specific" AMG preconditioner, or do I misunderstand something? Assuming this is correct, the structure-specific parameters "steady/quasistatic/unsteady" and "small/large strain" are not really seen by the implementation of the AMG preconditioner used when running Structure simulations. Then, many tests could be considered superfluous. Generally, I consider it a con of the procedure suggested here that this gives somewhat exponential complexity in the number of tests w.r.t. parameters, and I am not sure if this really helps (given that the main purpose of most of the tests here seems to be "no segmentation fault occurred" while most of the parameter combinations are actually still untested (and where I argue that we cannot avoid this) or while the implementation seems to be somewhat insensitive to the variation of parameters being tested).

In my opinion, added tests will not be removed again in the future (because one might have to expect to do harm if removing them, arguing that the person who introduced them might have had "good reasons to do this"). We are now evaluating "good reasons to do this", hence I would like to ask you for your understanding that I would like to understand the reasoning for adding certain tests.

richardschu:
@nfehn sorry I forgot about this, was focusing on teaching, conference, etc. ... we talked about this some time ago; as far as I remember, the conclusion was to remove redundant tests. In the current case: remove all the AMG tests, since the coarse grid solver of the multigrid is actually AMG.
For the remaining combinations, I have to check was is actually tested in the other applications.

richardschu:
OK update: tests in structure module are:
manufactured: Unsteady
weak_damping: Unsteady
Therefore I am also pro removing the Unsteady tests added here.

nfehn:
Yes, I voted for removing tests as much as possible. (I would also be willing to merge the changes in the code (removal of one if-statement) without modifications in applications/tests, just to have an improved version of the code (without the identified bug) asap.)

richardschu:
I kept the small strain unsteady example, since this is not tested elsewhere.
The remaining tests are:
manufactured:  unsteady, large strain
weak_damping: unsteady, large strain + weak damping terms
bar:
1) quasistatic solver + large strain + Multigrid/AMG coarse grid preconditioner
2) steady solver + small AND large strain + Multigrid/AMG coarse grid preconditioner
3) unsteady solver + small strain + Multigrid/AMG coarse grid preconditioner

Then we have all the solver options tested with no extra shenanigans! :1st_place_medal:

PR #657, opened 2024-03-26
reinit lapack_matrix in OperatorBase::internal_compute_factorized_additive_schwarz_matrices()

### LINKED ISSUE #655: AdditiveSchwarz preconditioner not working in structure module
I stumbled upon the issue regarding the `AdditiveSchwarz` preconditioner when looking at #652
Issue #652 (steady linear elasticity case) is fixed via #654, but the problem with the AS prec is that after one initial use it becomes invalid.
See the table in
https://github.com/exadg/exadg/issues/652#issuecomment-2018784445
and the stacktrace in
https://github.com/exadg/exadg/issues/652#issuecomment-2018785493

Probably the update is not done correctly as well here, or the internal state variables are not updated correctly.
@bergbauer could you comment on that guess of mine, I think you implemented that, right?
... I can search for the bug if you agree that something like this could be the reason.
I further assume that in the Poisson module this did not come up, since we do a single solve (similar to the working steady linear case, as seen in https://github.com/exadg/exadg/issues/652#issuecomment-2018784445).

richardschu:
fixes #655
Thanks @bergbauer 
did not want to bother you with the request for opening a PR :)

### DIFF
--- a/include/exadg/operators/operator_base.cpp
+++ b/include/exadg/operators/operator_base.cpp
@@ -2215,6 +2215,7 @@ OperatorBase<dim, Number, n_components>::internal_compute_factorized_additive_sc
 
         // store the cell matrix and renumber lexicographic
         auto & lapack_matrix = matrices[cell * vectorization_length + v];
+        lapack_matrix.reinit(dofs_per_cell);
 
         auto const & lex_to_hier =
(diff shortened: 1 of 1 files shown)


bergbauer (approved):
This should not allocate any new memory but will set all entries to zero. In contrast to factorizing a matrix, these costs should be negligible.

PR #659, opened 2024-05-09
Aero-acoustics from analytical CFD solutions or an analytical expression of the source term

jh66637:
The co-rotating vortex test case is quite nice to benchmark different coupling algorithms for aero-acoustics since an analytical CFD solution is known in addition to an analytical solution for the acoustics in the far field. Additionally, it is possible to derive an analytical expression of the aero-acoustic source term. 

We have used this procedure successfully to gain insight into the advantages and disadvantages of different coupling strategies.

This PR introduces two methods to compute the acoustic propagation from analytical solutions.
- Given analytical CFD pressure and velocity, simply interpolate the solution onto the CFD grid and compute the acoustic as usual. For this interpolation, I introduced an `AnalyticalTimeIntegrator`. This way, we do not have to check at numerous places if the analytical solution should be used instead of a computed one. 
- In case we know the expression of the source term we also know the analytical CFD solution. To be able to compute realistic time step sizes via the CFL condition, we make use of the `AnalyticalTimeIntegrator` in this case as well. However, the source term is now computed from the analytical function of the source term instead of the analytical CFD solution. 

In my opinion, this PR is a prerequisite for some follow-up PRs that introduce different volume coupling algorithms.

### DIFF
--- a/include/exadg/aero_acoustic/calculators/source_term_calculator.h
+++ b/include/exadg/aero_acoustic/calculators/source_term_calculator.h
@@ -197,9 +197,6 @@ private:
                   dealii::ExcMessage("No blend-in function provided."));
     }
 
-    if(data.blend_in_function)
-      data.blend_in_function->set_time(time);
-
     bool const space_dependent_scaling =
       data.blend_in_function != nullptr ? data.blend_in_function->varies_in_space(time) : false;
     Number const pure_temporal_scaling_factor =
@@ -208,7 +205,7 @@ private:
     if(space_dependent_scaling)
     {
       return [&](qpoint const & q) {
-        return FunctionEvaluator<0, dim, Number>::value(*data.blend_in_function, q);
+        return FunctionEvaluator<0, dim, Number>::value(*data.blend_in_function, q, time);
       };
     }
     else
--- a/include/exadg/aero_acoustic/calculators/source_term_calculator.h
+++ b/include/exadg/aero_acoustic/calculators/source_term_calculator.h
@@ -113,7 +113,7 @@ private:
 
     Number rho = static_cast<Number>(data.density);
 
-    auto apply_scaling = get_scaling_function();
+    auto get_scaling_factor = get_scaling_function();
 
     for(unsigned int cell = cell_range.first; cell < cell_range.second; ++cell)
     {
@@ -132,7 +132,7 @@ private:
           scalar flux = -rho * dpdt.get_value(q) + u.get_value(q) * p.get_gradient(q);
 
           if(data.blend_in)
-            apply_scaling(flux, dpdt.quadrature_point(q));
+            flux *= get_scaling_factor(dpdt.quadrature_point(q));
 
           dpdt.submit_value(flux, q);
         }
@@ -144,7 +144,7 @@ private:
           scalar flux = -rho * dpdt.get_value(q);
 
           if(data.blend_in)
-            apply_scaling(flux, dpdt.quadrature_point(q));
+            flux *= get_scaling_factor(dpdt.quadrature_point(q));
 
           dpdt.submit_value(flux, q);
         }
@@ -164,7 +164,7 @@ private:
 
     Number rho = static_cast<Number>(data.density);
 
-    auto apply_scaling = get_scaling_function();
+    auto get_scaling_factor = get_scaling_function();
 
     for(unsigned int cell = cell_range.first; cell < cell_range.second; ++cell)
     {
@@ -176,7 +176,7 @@ private:
                                                                       dpdt.quadrature_point(q));
 
         if(data.blend_in)
-          apply_scaling(flux, dpdt.quadrature_point(q));
+          flux *= get_scaling_factor(dpdt.quadrature_point(q));
 
         dpdt.submit_value(flux, q);
       }
@@ -185,7 +185,7 @@ private:
     }
   }
 
-  std::function<void(scalar &, qpoint const &)>
+  std::function<scalar(qpoint const &)>
   get_scaling_function() const
   {
     // In case we blend in the source term, we check if the scaling is space dependent. Only in that
@@ -207,16 +207,15 @@ private:
 
     if(space_dependent_scaling)
     {
-      return [&](scalar & flux, qpoint const & q) {
-        flux *= FunctionEvaluator<0, dim, Number>::value(*data.blend_in_function, q);
+      return [&](qpoint const & q) {
+        return FunctionEvaluator<0, dim, Number>::value(*data.blend_in_function, q);
       };
     }
     else
     {
       // capture scaling factor by copy
-      return [pure_temporal_scaling_factor](scalar & flux, qpoint const &) {
-        flux *= pure_temporal_scaling_factor;
-      };
+      return
+        [pure_temporal_scaling_factor](qpoint const &) { return pure_temporal_scaling_factor; };
     }
   }
 
--- a/include/exadg/aero_acoustic/calculators/source_term_calculator.h
+++ b/include/exadg/aero_acoustic/calculators/source_term_calculator.h
@@ -45,6 +45,9 @@ class SourceTermCalculator
   using CellIntegratorScalar = CellIntegrator<dim, 1, Number>;
   using CellIntegratorVector = CellIntegrator<dim, dim, Number>;
 
+  using scalar = dealii::VectorizedArray<Number>;
+  using qpoint = dealii::Point<dim, dealii::VectorizedArray<Number>>;
+
 public:
   SourceTermCalculator() : matrix_free(nullptr), time(std::numeric_limits<double>::min())
   {
@@ -57,6 +60,21 @@ public:
     matrix_free = &matrix_free_in;
     data        = data_in;
   }
+
+  void
+  evaluate_integrate(VectorType &            dst,
+                     dealii::Function<dim> & analytical_source_term,
+                     double const            evaluation_time)
+  {
+    time = evaluation_time;
+
+    dst.zero_out_ghost_values();
+
+    analytical_source_term.set_time(time);
+
+    matrix_free->cell_loop(&This::compute_source_term, this, dst, analytical_source_term, true);
+  }
+
 
   void
   evaluate_integrate(VectorType &       dst,
@@ -95,24 +113,7 @@ private:
 
     Number rho = static_cast<Number>(data.density);
 
-    // In case we blend in the source term, we check if the scaling is space dependent. Only in that
-    // case we have to evaluate the function in every equadrature point. Otherwise the scaling
-    // is purely temporal and constant during this function.
-    if(data.blend_in)
-      AssertThrow(data.blend_in_function != nullptr,
-                  dealii::ExcMessage("No blend-in function provided."));
-
-    bool const space_dependent_scaling =
-      data.blend_in_function != nullptr ? data.blend_in_function->varies_in_space(time) : false;
-    Number const pure_temporal_scaling_factor =
-      (not space_dependent_scaling) ? data.blend_in_function->compute_time_factor(time) : 1.0;
-
-    auto apply_scaling = [&](auto & flux, auto const & q) {
-      if(space_dependent_scaling)
-        flux *= FunctionEvaluator<0, dim, Number>::value(*data.blend_in_function, q, time);
-      else
-        flux *= pure_temporal_scaling_factor;
-    };
+    auto apply_scaling = get_scaling_function();
 
     for(unsigned int cell = cell_range.first; cell < cell_range.second; ++cell)
     {
@@ -128,7 +129,7 @@ private:
 
         for(unsigned int q = 0; q < dpdt.n_q_points; ++q)
         {
-          auto flux = -rho * dpdt.get_value(q) + u.get_value(q) * p.get_gradient(q);
+          scalar flux = -rho * dpdt.get_value(q) + u.get_value(q) * p.get_gradient(q);
 
           if(data.blend_in)
             apply_scaling(flux, dpdt.quadrature_point(q));
@@ -140,7 +141,7 @@ private:
       {
         for(unsigned int q = 0; q < dpdt.n_q_points; ++q)
         {
-          auto flux = -rho * dpdt.get_value(q);
+          scalar flux = -rho * dpdt.get_value(q);
 
           if(data.blend_in)
             apply_scaling(flux, dpdt.quadrature_point(q));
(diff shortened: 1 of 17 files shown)


include/exadg/aero_acoustic/calculators/source_term_calculator.h
```diff
@@ -153,6 +154,73 @@ class SourceTermCalculator
...
+    {
+      dpdt.reinit(cell);
+
+      for(unsigned int q = 0; q < dpdt.n_q_points; ++q)
+      {
+        scalar flux = -rho * FunctionEvaluator<0, dim, Number>::value(analytical_source_term,
+                                                                      dpdt.quadrature_point(q));
+
+        if(data.blend_in)
+          apply_scaling(flux, dpdt.quadrature_point(q));
```
nfehn, line 179: Is there a reason why you do not write this code like `flux *= get_scaling_factor(qpoint);`. Right now I could imagine this would simplify the code (why a lambda function; why two function arguments even though the first one (the flux) seems to be passed to this function only to be multiplied by a scaling factor?) and make the code easier to understand.

include/exadg/aero_acoustic/calculators/source_term_calculator.h
```diff
@@ -153,6 +154,73 @@ class SourceTermCalculator
...
+    // case we have to evaluate the function in every equadrature point. Otherwise the scaling
+    // is purely temporal and constant during this function.
+    if(data.blend_in)
+    {
+      AssertThrow(data.blend_in_function != nullptr,
+                  dealii::ExcMessage("No blend-in function provided."));
+    }
+
+    if(data.blend_in_function)
+      data.blend_in_function->set_time(time);
```
nfehn, line 201: Is this code new? Is it correct that this code could be shifted to line 209/210 (`if(space_dependent_scaling)`)? Could you explain the reason for this change?

include/exadg/aero_acoustic/single_field_solvers/analytical_time_int_fluid.h
```diff
@@ -0,0 +1,198 @@
...
+    velocity[0].swap(velocity_np);
+
+    push_back(pressure);
+    pressure[0].swap(pressure_np);
+  }
+
+  void
+  solve_steady_problem() final
+  {
+    AssertThrow(false, dealii::ExcMessage("Aero-Acustic not implemented for steady problem."));
```
nfehn, line 146: This class lives in namespace `ExaDG::IncNS`, so I think the message should be formulated accordingly (e.g. replace "Aero-acoustic" by TimeIntAnalytic).

include/exadg/aero_acoustic/single_field_solvers/fluid.h
```diff
@@ -86,12 +90,58 @@ class SolverFluid
...
+                                           velocity_double);
+
+          dealii::VectorTools::interpolate(*pde_operator->get_mapping(),
+                                           pde_operator->get_dof_handler_p(),
+                                           *(analytical_cfd_solution_pressure),
+                                           pressure_double);
+
+          velocity = velocity_double;
+          pressure = pressure_double;
+        };
```
nfehn, line 124: I think this should be functionality of `IncNS::SpatialOperatorBase`, similar to `IncNS::SpatialOperatorBase::prescribe_initial_conditions()`. `IncNS::FieldFunctions` already has `analytical_solution_pressure`, so one would have to add `analytical_solution_velocity`.

include/exadg/aero_acoustic/single_field_solvers/fluid.h
```diff
@@ -57,8 +58,11 @@ class SolverFluid
 
   void
   setup(std::shared_ptr<FluidAeroAcoustic::ApplicationBase<dim, Number>> application,
-        MPI_Comm const                                                   mpi_comm,
-        bool const                                                       is_test)
+        bool const                             compute_acoustic_from_analytical_cfd_solution,
+        std::shared_ptr<dealii::Function<dim>> analytical_cfd_solution_velocity,
+        std::shared_ptr<dealii::Function<dim>> analytical_cfd_solution_pressure,
```
nfehn, line 63: According to comment https://github.com/exadg/exadg/pull/659/files#r1599659156, these parameters can be removed in my opinion because they are available via the `pde_operator`.

include/exadg/aero_acoustic/single_field_solvers/fluid.h
```diff
@@ -86,12 +90,58 @@ class SolverFluid
...
+          pressure = pressure_double;
+        };
+
+
+      time_integrator =
+        std::make_shared<IncNS::TimeIntAnalytic<dim, Number>>(pde_operator,
+                                                              nullptr /*no ALE*/,
+                                                              postprocessor,
+                                                              application->get_parameters(),
+                                                              interpolate_analytical_solution,
```
nfehn, line 132: This lambda-function parameter could be removed according to comment https://github.com/exadg/exadg/pull/659/files#r1599659156 because `TimeIntAnalytic` already receives the `pde_operator` as first argument.

include/exadg/aero_acoustic/user_interface/field_functions.h
```diff
@@ -47,6 +47,24 @@ struct FieldFunctions
...
+   * The analytical solution of the CFD velocity, that can be used to compute
+   * the aero acoustic source term.
+   */
+  std::shared_ptr<dealii::Function<dim>> analytical_cfd_solution_velocity;
+
+  /*
+   * The analytical solution of the CFD pressure, that can be used to compute
+   * the aero acoustic source term.
+   */
+  std::shared_ptr<dealii::Function<dim>> analytical_cfd_solution_pressure;
```
nfehn, line 61: This is a duplication with `IncNS::FieldFunctions`, see comment https://github.com/exadg/exadg/pull/659/files#r1599659156.

include/exadg/aero_acoustic/user_interface/parameters.h
```diff
@@ -54,6 +56,14 @@ class Parameters
...
     AssertThrow(fluid_to_acoustic_coupling_strategy != FluidToAcousticCouplingStrategy::Undefined,
                 dealii::ExcMessage("Coupling strategy has to be set."));
+
+    if(compute_acoustic_from_analytical_cfd_solution)
+    {
+      AssertThrow(
+        not compute_acoustic_from_analytical_source_term,
+        dealii::ExcMessage(
+          "Only one of the following can be true: compute_acoustic_from_analytical_cfd_solution, compute_acoustic_from_analytical_source_term."));
+    }
```
nfehn, line 66: Doesn't this suggest the usage of only a single boolean parameter, or the usage of an enum parameter (e.g. `AcousticSourceTermComputation`) instead (`Patterns::Enum` should allow this for easy use in applications :))?

include/exadg/functions_and_boundary_conditions/evaluate_functions.h
```diff
@@ -137,13 +136,22 @@ struct FunctionEvaluator<0, dim, Number>
...
 
+  static inline DEAL_II_ALWAYS_INLINE //
+    dealii::Tensor<0, dim, dealii::VectorizedArray<Number>>
+    value(dealii::Function<dim> &                                     function,
+          dealii::Point<dim, dealii::VectorizedArray<Number>> const & q_points,
+          double const &                                              time)
+  {
+    function.set_time(time);
+    return value(function, q_points);
+  }
```
nfehn, line 153: Generally, I would like to see such changes in separate PRs (the change affects very general functionality of ExaDG), but I also do not want to insist on this here in this PR.

include/exadg/aero_acoustic/user_interface/parameters.h
nfehn: Looking at the rest of the code, I agree we might indeed need two parameter. What about the following set of parameters:

- one boolean parameter `use_analytical_fluid_solution`
- one enum parameter `AcousticSourceTermComputation` with two values like `Analytical` vs. `FromFluidSolution`

I think with these parameters I would understand the logic of the code more easily. According to my current understanding of the code, the combination `use_analytical_fluid_solution = true` and  `AcousticSourceTermComputation::Analytical` should be possible (I would understand this option as using `TimeIntAnalytic` for the fluid problem, but "overwriting/replacing" the usual source term for acoustic by an analytical function). However, I also understand to exclude this option from a user perspective. 

What I do not like regarding the current parameter name `compute_acoustic_from_analytical_cfd_solution` is that this name does not seem to reflect that this parameter also affects the solution of the fluid problem (and the fluid time integrator) and not just the way how the acoustic source term is computed (i.e. a user could understand this parameter as solving the fluid problem as usual but only replacing the simulated fluid solution by the analytical one when it comes to the computation of the source term).

I hope this comment clarifies my thoughts. It might be that I still did not fully understand the code.

include/exadg/aero_acoustic/single_field_solvers/analytical_time_int_fluid.h
```diff
@@ -0,0 +1,198 @@
...
+namespace ExaDG
+{
+namespace IncNS
+{
+/**
+ * A class that sets the analytical solution for the next time instead of computing it.
+ * To be able to compute temporal derivatives etc. this class is based on TimeIntBDF.
+ */
+template<int dim, typename Number>
+class TimeIntAnalytic : public TimeIntBDF<dim, Number>
```
nfehn, line 36: Did you think about placing this class/code in a more central place where the other fluid time integrators live? What is your opinion on this?

I am not completely happy with the name `TimeIntAnalytic` (since "analytic time integrator" might be understood differently than what is done here, but together with the class description I am principally also fine with the current name. I am thinking of `TimeIntInterpolateAnalyticalSolution` or `TimeIntInjectAnalyticalSolution`. Maybe this is clearer; unfortunately it is also much longer.

nfehn:
I looked at the code and I think the "time integrator" suggested here is an elegant way to achieve what you need here.

As detailed in the comments, I see some points of discussion regarding the usage of lambda functions and the design/naming of parameters.

include/exadg/aero_acoustic/calculators/source_term_calculator.h
jh66637: [2980d74](https://github.com/exadg/exadg/pull/659/commits/2980d74bc7b7357a560a7688772160c3078d18b2)

include/exadg/aero_acoustic/single_field_solvers/analytical_time_int_fluid.h
jh66637: [41de64d](https://github.com/exadg/exadg/pull/659/commits/41de64d7d97686575dd2b736d9e6f32c301fdd2f)

include/exadg/aero_acoustic/calculators/source_term_calculator.h
jh66637: I wanted to get rid of the member `time`, but decided it is better to set the time during setup of the scaling function and forgot to move `time` back to `value()`. [522b33b](https://github.com/exadg/exadg/pull/659/commits/522b33bf7e1f2d0831b10aad13c814bba7db8346)

include/exadg/aero_acoustic/single_field_solvers/fluid.h
jh66637: [b7860ce](https://github.com/exadg/exadg/pull/659/commits/b7860ce30c3cbaeee80a8cfc4f1a4c3247d59845)

include/exadg/aero_acoustic/user_interface/field_functions.h
jh66637: [b7860ce](https://github.com/exadg/exadg/pull/659/commits/b7860ce30c3cbaeee80a8cfc4f1a4c3247d59845)

include/exadg/aero_acoustic/single_field_solvers/analytical_time_int_fluid.h
jh66637: [954d775](https://github.com/exadg/exadg/pull/659/commits/954d775f6c6149875632294d9d98b68d9604e32f)

include/exadg/aero_acoustic/single_field_solvers/fluid.h
jh66637: [5efc732](https://github.com/exadg/exadg/pull/659/commits/5efc7327890f5a84d6d550396244124f94633f33)

include/exadg/aero_acoustic/single_field_solvers/fluid.h
jh66637: [5efc732](https://github.com/exadg/exadg/pull/659/commits/5efc7327890f5a84d6d550396244124f94633f33)

include/exadg/aero_acoustic/user_interface/parameters.h
jh66637: Moving the time integrator to the IncNSE module made it comparably simple. The proposed version is a mix of your idea and the parameter that is now loaced in IncNSE [5efc732](https://github.com/exadg/exadg/pull/659/commits/5efc7327890f5a84d6d550396244124f94633f33)

include/exadg/functions_and_boundary_conditions/evaluate_functions.h
jh66637: I can do this in a separate PR. I will not have the time for that in the next few days, though.

jh66637:
@nfehn Thank you for the review. I adjusted the code. Introducing a version of `value` that does not take `time` and thus can work on `const Function` will be introduced in a separate PR. Please let me know what you think about the rest of the changes :)

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
```diff
@@ -980,6 +980,45 @@ SpatialOperatorBase<dim, Number>::prescribe_initial_conditions(VectorType & velo
...
+                                   *(field_functions->analytical_solution_velocity),
+                                   velocity_double);
+
+  dealii::VectorTools::interpolate(*get_mapping(),
+                                   get_dof_handler_p(),
+                                   *(field_functions->analytical_solution_pressure),
+                                   pressure_double);
+
+  velocity = velocity_double;
+  pressure = pressure_double;
```
nfehn, line 1019: This code is more or less the same as above for the function `prescribe_initial_conditions()`. Hence, I would suggest to write one private member function (that takes the velocity and pressure functions as arguments and that implements these lines of code), which is then called by the public member functions `prescribe_initial_conditions()` and `interpolate_analytical_solution()`.

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
```diff
@@ -980,6 +980,45 @@ SpatialOperatorBase<dim, Number>::prescribe_initial_conditions(VectorType & velo
...
+                                                                  VectorType & pressure,
+                                                                  double const time) const
+{
+  AssertThrow(field_functions->analytical_solution_velocity,
+              dealii::ExcMessage("FieldFunctions::analytical_solution_velocity not set"));
+  AssertThrow(field_functions->analytical_solution_pressure,
+              dealii::ExcMessage("FieldFunctions::analytical_solution_pressure not set"));
+
+  velocity.zero_out_ghost_values();
+  pressure.zero_out_ghost_values();
```
nfehn, line 995: It is not clear to me why we have these lines of code here, while we do not have them e.g. in the function `prescribe_initial_conditions()` above.

nfehn:
@jh66637 Many thanks for your updates (and sorry for my late response, I was on holidays ...). I like the changes. I have one more comment regarding code duplication.

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
jh66637: I will move the line to the analytical time integrator:)

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
jh66637: Instead let's finally introduce a free function to remove the boilerplate code and avoid some unnecessary copy operations in numerous places of ExaDG. I would like to do this in a follow up PR if it is OK for you :)

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
nfehn: I am fine with this. I think a comment explaining why we have/need these lines of code would help me to better understand the code (e.g.: what would happen if one forgot to write these lines of code? segmentation-fault or performance penalty or ...?).

We can also do this as a follow-up so that we can merge the present PR soon.

jh66637: `dealii::VT::interpolate()` simply throws an exception with zeroed out ghost values. During time stepping the vectors are not zeroed out though, so we have to care about that ourselves.

PR #661, opened 2024-05-16
Use new `dealii::MatrixFreeTools` functions for diagonal computation and matrix assembly

peterrum:
uses https://github.com/dealii/dealii/pull/17029

### DIFF
--- a/include/exadg/operators/operator_base.cpp
+++ b/include/exadg/operators/operator_base.cpp
@@ -430,33 +430,24 @@ OperatorBase<dim, Number, n_components>::add_diagonal(VectorType & diagonal) con
     }
     else
     {
-      matrix_free->loop(&This::cell_loop_diagonal,
-                        &This::face_loop_diagonal,
-                        &This::boundary_face_loop_diagonal,
-                        this,
-                        diagonal,
-                        diagonal);
+      dealii::MatrixFreeTools::compute_diagonal(*matrix_free,
+                                                diagonal,
+                                                &This::cell_compute_diagonal_matrix,
+                                                &This::face_compute_diagonal_matrix,
+                                                &This::boundary_face_compute_diagonal_matrix,
+                                                this,
+                                                data.dof_index,
+                                                data.quad_index);
     }
   }
   else
   {
-    dealii::MatrixFreeTools::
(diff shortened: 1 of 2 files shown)


include/exadg/operators/operator_base.cpp
```diff
@@ -1302,155 +1295,58 @@ OperatorBase<dim, Number, n_components>::face_loop_empty(
...
-  Range const &                           range) const
+OperatorBase<dim, Number, n_components>::cell_compute_diagonal_matrix(
+  IntegratorCell & integrator) const
 {
-  (void)src;
-
-  IntegratorCell integrator =
-    IntegratorCell(matrix_free, this->data.dof_index, this->data.quad_index);
+  // TODO: this is currently done for every column, but would only be necessary
+  // once per cell
```
nfehn, line 1302: Just to be sure: Is this comment still correct?

nfehn:
Looks good as far as I can judge this, but I am certainly not an expert here. @kronbichler could you have a look at this PR as well?

nfehn:
@peterrum Is this ready to be merged from your side?

peterrum:
> @peterrum Is this ready to be merged from your side?

Yes

include/exadg/operators/operator_base.cpp
```diff
@@ -822,26 +813,28 @@ void
...
-                      &This::boundary_face_loop_calculate_system_matrix,
-                      this,
-                      system_matrix,
-                      system_matrix);
+    dealii::MatrixFreeTools::compute_matrix(*matrix_free,
+                                            matrix_free->get_affine_constraints(),
+                                            system_matrix,
+                                            &This::cell_compute_diagonal_matrix,
+                                            &This::face_compute_diagonal_matrix,
+                                            &This::boundary_face_compute_diagonal_matrix,
```
nfehn, line 823: Here we want to compute the full matrix, not just the diagonal. So I assume this is a bug here?

nfehn: Ah, I think I am just irritated by the name. Nevertheless, it seems as if the code does not work correctly and I get a matrix with `n_nonzero_elements() = 0`.

nfehn:
Should we revert this PR due to #702?

PR #665, opened 2024-06-04
Fix preprocessor directive of conditional PETSc code in operator base

davidscn:
... to make the code compile with PETSc, but without Trilinos.

The `DEAL_II_WITH_` defines are not equal to any true or false per se, just PP defines.


kronbichler (approved):
Excellent, thank you!

Issue #664, opened 2024-05-29
Flow past sphere / initial coarsened meshes


jh66637:
I am having trouble using global coarsened meshes (be e.g. setting AdditionalCFDRefinementsAroundSource to 1 in co_rotating_vortex_pair/input.json). This worked some time ago.

To check where this stems from, I tried to run the flow past sphere, i.e. the only application I am aware of, that uses global coarsened initial meshes (independent of preconditioners). This application already crashes during mesh generation, but I am relatively sure there will be the same issues I have in the example given above. If someone that is familiar with these parts of the code could take a look, it would be great. Once everything works, I can add tests for both test cases :)

kronbichler:
I think I can reproduce. Do you mean this error:
```
--------------------------------------------------------
An error occurred in line <12236> of file </home/martin/deal/deal.II/source/grid/tria.cc> in function
    void dealii::Triangulation<3, 3>::set_all_manifold_ids(const types::manifold_id) [dim = 3, spacedim = 3]
The violated condition was: 
    n_cells() > 0
Additional information: 
    Error: set_all_manifold_ids() can not be called on an empty
    Triangulation.
```
I will investigate, I think I know what to do.

jh66637:
This is the first error I get. I fear this might not be the only problem though...

kronbichler:
I did fix the other error in FSI applications and I see different errors at least there, which is related more closely to what you describe:
```
--------------------------------------------------------
An error occurred in line <2441> of file </home/martin/deal/deal.II/include/deal.II/lac/precondition.h> in function
    EigenvalueInformation dealii::internal::estimate_eigenvalues(const EigenvalueAlgorithmAdditionalData<PreconditionerType> &, const MatrixType *, VectorType &, VectorType &, const unsigned int) [MatrixType = ExaDG::MultigridOperatorBase<2, float>, VectorType = dealii::LinearAlgebra::distributed::Vector<float>, PreconditionerType = ExaDG::JacobiPreconditioner<ExaDG::MultigridOperatorBase<2, float>>]
The violated condition was: 
    data.preconditioner.get() != nullptr
Additional information: 
    (none)
```
I will check that one as well, here it seems to be a different kind of problem, where some diagonal is not compute throughout the hierarchy. Not sure if that is the actual problem, but at least it's one problem we currently have :wink:

jh66637:
[...]

PR #667, opened 2024-06-05
Fix manifold ids for flow past sphere

kronbichler:
Related to #664: This fixes the initial setup of the flow-past-sphere case. However, I believe this does not address the issue that @jh66637 saw in the first place.

### DIFF
--- a/applications/fluid_structure_interaction/cylinder_with_flag/application.h
+++ b/applications/fluid_structure_interaction/cylinder_with_flag/application.h
@@ -423,8 +423,6 @@ private:
 
       create_triangulation(tria);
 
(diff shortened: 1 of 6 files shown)


kronbichler:
There are additional problems in other cases, I have to investigate if we really need to set the manifold id to `dealii::numbers::flat_manifold_id` in all cases, or if there is a more fundamental issue.

kronbichler:
This needs https://github.com/dealii/dealii/pull/17110 to work properly.

jh66637:
I can confirm this fixes the flow past sphere test case :+1:

PR #669, opened 2024-06-10
fix initially coarsened cfd mesh

jh66637:
This is a quick fix for https://github.com/exadg/exadg/issues/664. As promised I will add some tests in a separate PR once this is merged.
Thank you @richardschu :)

### DIFF
--- a/applications/aero_acoustic/co_rotating_vortex_pair/application.h
+++ b/applications/aero_acoustic/co_rotating_vortex_pair/application.h
@@ -424,20 +424,6 @@ public:
 
     // pressure Poisson equation
     this->param.solver_pressure_poisson              = SolverPressurePoisson::CG;
-    this->param.solver_data_pressure_poisson         = SolverData(1e4, 1.e-12, 1.e-6, 100);
-    this->param.preconditioner_pressure_poisson      = PreconditionerPressurePoisson::Multigrid;
-    this->param.multigrid_data_pressure_poisson.type = MultigridType::cphMG;
-    this->param.multigrid_data_pressure_poisson.smoother_data.smoother =
-      MultigridSmoother::Chebyshev;
-    this->param.multigrid_data_pressure_poisson.smoother_data.preconditioner =
-      PreconditionerSmoother::PointJacobi;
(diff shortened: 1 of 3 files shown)


kronbichler (approved):

applications/aero_acoustic/co_rotating_vortex_pair/application.h
```diff
@@ -536,13 +536,26 @@ class Application : public ApplicationBase<dim, Number>
...
                                                             this->mpi_comm,
                                                             this->param.grid,
                                                             this->param.involves_h_multigrid(),
                                                             lambda_create_triangulation,
                                                             {});
 
+    // triangulation for hanging nodes
+    if(additional_refinements_around_source > 0)
+    {
+      GridUtilities::create_coarse_triangulations_after_coarsening_and_refinement(
```
nfehn, line 550: Looking into the implementation, I believe this function does essentially the same (for `TriangulationType::Distributed`) as `GridUtilities::create_coarse_triangulations()`, a function that should actually already be called automatically by the above function called in line 540, `GridUtilities::create_triangulation_with_multigrid<dim>()`, as long as the two parameters `this->param.grid.create_coarse_triangulations` and `this->param.involves_h_multigrid()` contain valid data. Hence, I believe we need to better understand the behavior.

@jh66637 Could you check if the function `GridUtilities::create_triangulation_with_multigrid<dim>()` called in line 540 enters the two if-statements `if(involves_h_multigrid)` and (inside this if-statement) `if(data.create_coarse_triangulations)`?

jh66637:
@nfehn I think shifting some code from the h_multigrid branch should work as well. Both boils down to the following: We have to create coarse triangulations if we have coarsened meshes, no matter if we have a h multigrid or not.

jh66637:
This works, I think this should be ready to merge.

nfehn:
I am still irritated why we need to always have the coarse grids, see also comment https://github.com/exadg/exadg/issues/664#issuecomment-2210468117 in the related issue. Maybe the implementation in the multigrid preconditioner is not correct.

jh66637:
Closed in favor of https://github.com/exadg/exadg/pull/678

PR #671, opened 2024-06-14
Implement dual-splitting scheme with fully implicit convective term

nfehn:
Preparation:
- unify implementation of BDFDualSplitting and BDFPressureCorrection regarding momentum step
- unify IncNS parameters
- fix some if-statements regarding implicit/explicit or viscous/convective checks and avoid unnecessary computations that we had in the previous implementation

**New scheme**: The viscous step of the dual splitting scheme becomes

$\gamma_0/\Delta t \ u^{n+1} + F(u^{n+1}) - \sum_i \beta_i F(u_{n-i}) - \nu \Delta u^{n+1} = \gamma_0/\Delta t \ \hat{\hat{u}}$

where we add the implicit convective term and subtract again the extrapolated explicit convective term (taken into account in the first substep of the dual splitting scheme).

PR #672, opened 2024-06-26
make sure that variable viscosity is updated

nfehn:
This PR fixes two things:

- Make sure that variable viscosity is updated if we solve a nonlinear problem but the variable viscosity is treated explicitly in case of pressure-correction scheme and dual-splitting scheme.  (If the variable viscosity is treated implicitly, the update happens in `evaluate_nonlinear_residual()` (with an if-statement) and should be fine.)

- In the coupled solver, an `if(this->param.viscous_problem() ...)` was missing in my opinion

### DIFF
--- a/include/exadg/incompressible_navier_stokes/time_integration/driver_steady_problems.cpp
+++ b/include/exadg/incompressible_navier_stokes/time_integration/driver_steady_problems.cpp
@@ -142,7 +142,7 @@ DriverSteadyProblems<dim, Number>::do_solve(double const time, bool unsteady_pro
   }
 
   // explicit viscosity update
-  if(this->param.viscosity_is_variable() and
+  if(this->param.viscous_problem() and this->param.viscosity_is_variable() and
      this->param.treatment_of_variable_viscosity == TreatmentOfVariableViscosity::Explicit)
   {
     dealii::Timer timer_viscosity_update;
--- a/include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_coupled_solver.cpp
+++ b/include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_coupled_solver.cpp
@@ -177,7 +177,7 @@ TimeIntBDFCoupled<dim, Number>::do_timestep_solve()
   }
 
   // explicit viscosity update
-  if(this->param.viscosity_is_variable() and
+  if(this->viscous_problem() and this->param.viscosity_is_variable() and
      this->param.treatment_of_variable_viscosity == TreatmentOfVariableViscosity::Explicit)
   {
     dealii::Timer timer_viscosity_update;
--- a/include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_coupled_solver.cpp
+++ b/include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_coupled_solver.cpp
@@ -177,7 +177,7 @@ TimeIntBDFCoupled<dim, Number>::do_timestep_solve()
   }
 
   // explicit viscosity update
-  if(this->viscous_problem() and this->param.viscosity_is_variable() and
+  if(this->param.viscous_problem() and this->param.viscosity_is_variable() and
      this->param.treatment_of_variable_viscosity == TreatmentOfVariableViscosity::Explicit)
   {
     dealii::Timer timer_viscosity_update;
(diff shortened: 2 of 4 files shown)


include/exadg/incompressible_navier_stokes/time_integration/driver_steady_problems.cpp
```diff
@@ -142,7 +142,7 @@ DriverSteadyProblems<dim, Number>::do_solve(double const time, bool unsteady_pro
   }
 
   // explicit viscosity update
-  if(this->param.viscosity_is_variable() and
+  if(this->param.viscous_problem() and this->param.viscosity_is_variable() and
```
richardschu, line 145: :+1:

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_coupled_solver.cpp
```diff
@@ -177,7 +177,7 @@ TimeIntBDFCoupled<dim, Number>::do_timestep_solve()
   }
 
   // explicit viscosity update
-  if(this->param.viscosity_is_variable() and
+  if(this->param.viscous_problem() and this->param.viscosity_is_variable() and
```
richardschu, line 180: :+1:

richardschu: I am starting to think we changed the bahevior of `viscosity_is_variable()` at some point and I completely overlooked that!

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_dual_splitting.cpp
```diff
@@ -712,6 +712,24 @@ TimeIntBDFDualSplitting<dim, Number>::viscous_step()
...
+
+      pde_operator->update_viscosity(velocity_np);
+
+      if(this->print_solver_info() and not(this->is_test))
+      {
+        this->pcout << std::endl << "Update of variable viscosity:";
+        print_wall_time(this->pcout, timer_viscosity_update.wall_time());
+      }
+    }
+
```
richardschu, line 732: just to double-check:
in case we execute the above block (viscosity update),
the first branch below is **not** taken since `nonlinear_problem_has_to_be_solved()` is false given the explicit treatment.
So actually these changes now just pull out the update, adding another condition.
You think it is easier to read this way, right? The actual behavior is the same still, is it (if the `viscous_problem()` were added in the old version)?

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_pressure_correction.cpp
```diff
@@ -357,6 +357,24 @@ TimeIntBDFPressureCorrection<dim, Number>::momentum_step()
       velocity_np = velocity_momentum_last_iter;
     }
 
+    /*
```
richardschu, line 360: Good, the nonlinear path with implicit convection and extrapolated viscous term now also has a viscosity update! Good that you caught that, this was actually wrong in old code versions also prior to #671 !!

richardschu (approved):
Good, thanks for the bugfix!

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_dual_splitting.cpp
```diff
@@ -712,6 +712,24 @@ TimeIntBDFDualSplitting<dim, Number>::viscous_step()
...
     }
 
+    /*
+     *  update variable viscosity
+     */
+    if(this->param.viscous_problem() and this->param.viscosity_is_variable() and
+       this->param.treatment_of_variable_viscosity == TreatmentOfVariableViscosity::Explicit)
+    {
+      dealii::Timer timer_viscosity_update;
+      timer_viscosity_update.restart();
```
kronbichler, line 722: This line is not necessary (and was not in the code where this was moved from), because the constructor already starts the timer,
```suggestion
```

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_pressure_correction.cpp
```diff
@@ -357,6 +357,24 @@ TimeIntBDFPressureCorrection<dim, Number>::momentum_step()
...
+     *  update variable viscosity
+     */
+    if(this->param.viscous_problem() and this->param.viscosity_is_variable() and
+       this->param.treatment_of_variable_viscosity == TreatmentOfVariableViscosity::Explicit)
+    {
+      dealii::Timer timer_viscosity_update;
+      timer_viscosity_update.restart();
```
kronbichler, line 367:
```suggestion
```

kronbichler (approved):
This looks good.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_dual_splitting.cpp
nfehn: Thanks for this hint! We might have this line quite often in the code, so it could be best to address this separately.

PR #673, opened 2024-06-26
Linearly implicit convective term for incompressible flows

nfehn:
This PR introduces another formulation of the convective term called `TreatmentOfConvectiveTerm::LinearlyImplicit`, i.e. for this new formulation 

- the nonlinear convective term $F(u,u)$ is replaced by $F(\hat{u},u)$, where $\hat{u}=\sum_i \beta_i u_{n-i}$ is an extrapolation of the velocity from previous time steps. 

The motivation is
- to overcome the CFL-type time step restriction for explicit convective terms, and
- to avoid an outer iteration needed for nonlinear problems, e.g. Newton (which for fully implicit formulations of the convective term often causes a noticeable increase in costs).

PR #674, opened 2024-06-28
Use overintegration only for implicit convective problem

nfehn:
... and not for nonlinear viscous problem.

@richardschu I think when we extended the meaning of `Parameters::nonlinear_problem_has_to_be_solved()` (which now becomes also true for nonlinear viscous problems), we didn't realize that this implies that an over-integration quad-rule will be used for the momentum operator according to the current implementation (even if no convection is involved).

Indeed, we have an assert in the `Parameter` class currently ensuring that we use the standard quad-rule for variable-viscosity problems, but still I consider the implementation in `SpatialOperatorBase` a "bug", i.e. removing the assert in `Parameters` could lead to (unrecognized) unexpected behavior.

Of course, one could also debate if `param.quad_rule_linearization` will define once and for all the chosen quad-rule, i.e. to actually remove the if-else-statement in `SpatalOperatorBase`

```bash
    if(param.implicit_convective_problem())
      return get_quad_index_velocity_nonlinear();
    else
      return get_quad_index_velocity_linear();
```
by
```bash
      return get_quad_index_velocity_nonlinear();
```

However, the downside would be that if one switches between an explicit and an implicit version of the convective term in an application (which happens quite often in my opinion), one would always have to adjust the parameter `param.quad_rule_linearization`. Hence, the current implementation provides somewhat more automation and will not use over-integration if one can determine from the parameters that the convective term is actually not involved in the momentum operator. So I would keep this if-statement in the implementation. The current documentation has not been clear about this, so I also extend the documentation in the present PR, so that the documentation is in line with the implementation.

I encountered this "bug" during PR #673, so it would be good to discuss/merge the present PR soon and then use this in #673.

### DIFF
--- a/include/exadg/incompressible_navier_stokes/user_interface/parameters.cpp
+++ b/include/exadg/incompressible_navier_stokes/user_interface/parameters.cpp
@@ -412,7 +412,7 @@ Parameters::check(dealii::ConditionalOStream const & pcout) const
     }
   }
 
-  if(viscosity_is_variable() and nonlinear_problem_has_to_be_solved())
+  if(nonlinear_viscous_problem())
     AssertThrow(quad_rule_linearization == QuadratureRuleLinearization::Standard,
                 dealii::ExcMessage(
                   "Only the standard integration rule is supported for variable viscosity. "
--- a/include/exadg/incompressible_navier_stokes/user_interface/parameters.cpp
+++ b/include/exadg/incompressible_navier_stokes/user_interface/parameters.cpp
@@ -412,11 +412,13 @@ Parameters::check(dealii::ConditionalOStream const & pcout) const
     }
   }
 
-  if(nonlinear_viscous_problem())
+  if(viscous_problem() and viscosity_is_variable())
+  {
     AssertThrow(quad_rule_linearization == QuadratureRuleLinearization::Standard,
                 dealii::ExcMessage(
                   "Only the standard integration rule is supported for variable viscosity. "
                   "Variable viscosity with multiple integration rules is not yet implemented."));
+  }
 
   // PROJECTION METHODS
   if(preconditioner_momentum == MomentumPreconditioner::Multigrid)
--- a/applications/incompressible_navier_stokes/fda_benchmark/include/postprocessor.h
+++ b/applications/incompressible_navier_stokes/fda_benchmark/include/postprocessor.h
@@ -101,7 +101,7 @@ public:
         mean_velocity_calculator.reset(
           new MeanVelocityCalculator<dim, Number>(pde_operator.get_matrix_free(),
                                                   pde_operator.get_dof_index_velocity(),
-                                                  pde_operator.get_quad_index_velocity_linear(),
+                                                  pde_operator.get_quad_index_velocity_standard(),
                                                   pp_data_fda.mean_velocity_data,
                                                   this->mpi_comm));
       }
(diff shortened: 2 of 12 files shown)


include/exadg/incompressible_navier_stokes/user_interface/parameters.cpp
```diff
@@ -412,7 +412,7 @@ Parameters::check(dealii::ConditionalOStream const & pcout) const
     }
   }
 
-  if(viscosity_is_variable() and nonlinear_problem_has_to_be_solved())
+  if(nonlinear_viscous_problem())
```
nfehn, line 415: This new version is in my opinion equivalent to the old code, just simpler. However, I am not sure if this is what we want!? Imagine we have a variable viscosity problem, but the variable viscosity is treated explicitly. In that case, we would currently allow over-integration for the momentum operator if `param.quad_rule_linearization` is set accordingly. Operators are then initialized with different quad-rules, which does not seem to be supported.

@richardschu If you agree, I would change the code accordingly in the present PR, i.e. write just `if(viscosity_is_variable())` and thereby broaden the cases where this Assert is checked.

nfehn: In this comment, I consider the case of an implicit convective problem and an explicitly treated variable viscosity. The Code in `SpatialOperatorBase`  would then assign an over-integration quad-rule for the momentum operator for both the old code version and the new version suggested in this PR.

nfehn: I decided to change the if-statement as described above for now.

nfehn:
I changed in the present PR (https://github.com/exadg/exadg/pull/674/commits/882b942c4e595879393269bf7a8f1052663f59cb) also the naming of some quad-rules because the  previous "linear" vs. "nonlinear" naming would become irritating when introducing a "linearly implicit" convective term (where we might want to use a "nonlinear" quad-rule, but the system of equations is "linear", so the meaning of "linear" vs. "nonlinear" would be unclear).

include/exadg/incompressible_navier_stokes/user_interface/parameters.cpp
richardschu: I did not use over-integration for non-Newtonian fluids yet and I am not sure how the experiences from turbulent flow (only? idk!) transfer to the cases where the non-Newtonian models make most sense (much lower Re, usually). This is why I did not extend towards over-integration with variable viscosity.

Extend the `if` and broaden the cases this is checked, but please use
`if(viscous_problem() and viscosity_is_variable())`
since we might not have a viscous term, but just the `turbulence_model_data.is_active` or `generalized_newtonian_model_data.is_active` set wrong in an application code.

nfehn: yes, above I forgot `viscous_problem()`, but the current code should have that if-statement already.

richardschu (approved):
passt, then merge :)

kronbichler:
> I did not use over-integration for non-Newtonian fluids yet and I am not sure how the experiences from turbulent flow (only? idk!) transfer to the cases where the non-Newtonian models make most sense (much lower Re, usually). This is why I did not extend towards over-integration with variable viscosity.

I think the general behavior can be expected to be quite different as compared to the over-integration on convective terms, because viscous terms can be discretized in self-adjoint form, making the general operator positive (semi-) definite also with less accurate integration. Thus, the mathematical community is typically happy with classifying this as a "variational crime" that might show up in error constants, but not in the form of a possibly unstable contribution as is the case for hyperbolic terms (that crucially depend on properties like skew-symmetry). Of course, it then also depends on other factors of the discretization whether the end result really is stable (like the form of the variable viscosity and possible feedback mechanisms), but at least I would consider them less harmful.

PR #677, opened 2024-07-04
Enable DualSplitting and Pressure Correction for Simplex Elements

dominiktassilostill:
This PR adds simplex support for the Dirichlet boundary functions of the `DualSplitting` and the `PressureCorrection` time integrators. Therefore a new Quadrature is introduced to (see [here](https://github.com/dealii/dealii/pull/17205)) which has support on the unit support points of the face, so the boundary function can be evaluated at the correct locations. The `integrate` function is used to make sure that the values are mapped to the corresponding entries in the DoF vector.
I am open for suggestions how to test this (and after testing I would like to clean up the duplicated code).

### DIFF
--- a/include/exadg/incompressible_navier_stokes/spatial_discretization/operator_dual_splitting.cpp
+++ b/include/exadg/incompressible_navier_stokes/spatial_discretization/operator_dual_splitting.cpp
@@ -596,10 +596,7 @@ OperatorDualSplitting<dim, Number>::local_interpolate_velocity_dirichlet_bc_boun
   Range const & face_range) const
 {
   unsigned int const dof_index = this->get_dof_index_velocity();
-  AssertThrow(
-    matrix_free.get_dof_handler(dof_index).get_triangulation().all_reference_cells_are_hyper_cube(),
-    dealii::ExcMessage("This function is only implemented for hypercube elements."));
-  unsigned int const quad_index = this->get_quad_index_velocity_gauss_lobatto();
+  unsigned int const quad_index = this->get_quad_index_velocity_nodal_points();
 
   FaceIntegratorU integrator(matrix_free, true, dof_index, quad_index);
 
--- a/include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
+++ b/include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
@@ -360,11 +360,23 @@ SpatialOperatorBase<dim, Number>::fill_matrix_free_data(
                                      field + quad_index_u_overintegration);
 
   // TODO create these quadrature rules only when needed
-  matrix_free_data.insert_quadrature(dealii::QGaussLobatto<1>(param.degree_u + 1),
-                                     field + quad_index_u_gauss_lobatto);
-  matrix_free_data.insert_quadrature(dealii::QGaussLobatto<1>(param.get_degree_p(param.degree_u) +
-                                                              1),
-                                     field + quad_index_p_gauss_lobatto);
+  if(param.grid.element_type == ElementType::Hypercube)
+  {
+    matrix_free_data.insert_quadrature(dealii::QGaussLobatto<1>(param.degree_u + 1),
+                                       field + quad_index_u_nodal_points);
+    matrix_free_data.insert_quadrature(dealii::QGaussLobatto<1>(param.get_degree_p(param.degree_u) +
+                                                                1),
+                                       field + quad_index_p_nodal_points);
+  }
+  else if(param.grid.element_type == ElementType::Simplex)
+  {
+    matrix_free_data.insert_quadrature(dealii::Quadrature<dim>(
+                                         dof_handler_u_scalar.get_fe().get_unit_support_points()),
+                                       field + quad_index_u_nodal_points);
+    matrix_free_data.insert_quadrature(dealii::Quadrature<dim>(
+                                         dof_handler_p.get_fe().get_unit_support_points()),
+                                       field + quad_index_p_nodal_points);
+  }
 }
 
 template<int dim, typename Number>
@@ -377,7 +389,7 @@ SpatialOperatorBase<dim, Number>::initialize_dirichlet_cached_bc()
     std::vector<unsigned int> quad_indices;
     quad_indices.emplace_back(get_quad_index_velocity_standard());
     quad_indices.emplace_back(get_quad_index_velocity_overintegration());
-    quad_indices.emplace_back(get_quad_index_velocity_gauss_lobatto());
+    quad_indices.emplace_back(get_quad_index_velocity_nodal_points());
 
     interface_data_dirichlet_cached = std::make_shared<ContainerInterfaceData<1, dim, double>>();
     interface_data_dirichlet_cached->setup(*matrix_free,
@@ -796,16 +808,16 @@ SpatialOperatorBase<dim, Number>::get_quad_index_velocity_overintegration() cons
 
 template<int dim, typename Number>
 unsigned int
-SpatialOperatorBase<dim, Number>::get_quad_index_velocity_gauss_lobatto() const
+SpatialOperatorBase<dim, Number>::get_quad_index_velocity_nodal_points() const
 {
-  return matrix_free_data->get_quad_index(field + quad_index_u_gauss_lobatto);
+  return matrix_free_data->get_quad_index(field + quad_index_u_nodal_points);
 }
 
 template<int dim, typename Number>
 unsigned int
-SpatialOperatorBase<dim, Number>::get_quad_index_pressure_gauss_lobatto() const
+SpatialOperatorBase<dim, Number>::get_quad_index_pressure_nodal_points() const
 {
-  return matrix_free_data->get_quad_index(field + quad_index_p_gauss_lobatto);
+  return matrix_free_data->get_quad_index(field + quad_index_p_nodal_points);
 }
 
 template<int dim, typename Number>
@@ -1823,7 +1835,7 @@ SpatialOperatorBase<dim, Number>::local_interpolate_stress_bc_boundary_face(
 {
   unsigned int const dof_index_u = this->get_dof_index_velocity();
   unsigned int const dof_index_p = this->get_dof_index_pressure();
-  unsigned int const quad_index  = this->get_quad_index_velocity_gauss_lobatto();
+  unsigned int const quad_index  = this->get_quad_index_velocity_nodal_points();
 
   FaceIntegratorU integrator_u(matrix_free, true, dof_index_u, quad_index);
   FaceIntegratorP integrator_p(matrix_free, true, dof_index_p, quad_index);
--- a/include/exadg/incompressible_navier_stokes/spatial_discretization/operator_pressure_correction.cpp
+++ b/include/exadg/incompressible_navier_stokes/spatial_discretization/operator_pressure_correction.cpp
@@ -227,10 +227,7 @@ OperatorPressureCorrection<dim, Number>::local_interpolate_pressure_dirichlet_bc
   Range const & face_range) const
 {
   unsigned int const dof_index = this->get_dof_index_pressure();
-  AssertThrow(
-    matrix_free.get_dof_handler(dof_index).get_triangulation().all_reference_cells_are_hyper_cube(),
-    dealii::ExcMessage("This function is only implemented for hypercube elements."));
-  unsigned int const quad_index = this->get_quad_index_pressure_gauss_lobatto();
+  unsigned int const quad_index = this->get_quad_index_pressure_nodal_points();
 
   FaceIntegratorP integrator(matrix_free, true, dof_index, quad_index);
 
--- a/include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.h
+++ b/include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.h
@@ -169,10 +169,10 @@ protected:
   get_quad_index_velocity_overintegration() const;
 
   unsigned int
-  get_quad_index_velocity_gauss_lobatto() const;
+  get_quad_index_velocity_nodal_points() const;
 
   unsigned int
-  get_quad_index_pressure_gauss_lobatto() const;
+  get_quad_index_pressure_nodal_points() const;
 
   unsigned int
   get_quad_index_velocity_linearized() const;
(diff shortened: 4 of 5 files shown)


include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
```diff
@@ -360,11 +360,22 @@ SpatialOperatorBase<dim, Number>::fill_matrix_free_data(
...
+    matrix_free_data.insert_quadrature(dealii::QGaussLobatto<1>(param.degree_u + 1),
+                                       field + quad_index_u_gauss_lobatto);
+    matrix_free_data.insert_quadrature(dealii::QGaussLobatto<1>(param.get_degree_p(param.degree_u) +
+                                                                1),
+                                       field + quad_index_p_gauss_lobatto);
+  }
+  else if(param.grid.element_type == ElementType::Simplex)
+  {
+    matrix_free_data.insert_quadrature(dealii::QSupportPointsSimplex<dim>(param.degree_u),
+                                       field + quad_index_u_gauss_lobatto);
```
kronbichler, line 374: As suggested in the deal.II PR, can't you use `dof_handler_u.get_fe().get_unit_support_points()`?

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_dual_splitting.cpp
```diff
@@ -596,64 +596,126 @@ OperatorDualSplitting<dim, Number>::local_interpolate_velocity_dirichlet_bc_boun
...
 
-  FaceIntegratorU integrator(matrix_free, true, dof_index, quad_index);
-
-  for(unsigned int face = face_range.first; face < face_range.second; face++)
+  if(matrix_free.get_dof_handler(dof_index)
+       .get_triangulation()
+       .all_reference_cells_are_hyper_cube())
   {
-    dealii::types::boundary_id const boundary_id = matrix_free.get_boundary_id(face);
+    unsigned int const quad_index = this->get_quad_index_velocity_gauss_lobatto();
```
nfehn, line 604: By introducing a member function `get_quad-index_velocity_nodal_points()` we can unify the hypercube/simplex case and write this line outside the if-else-statement for hypercube vs. simplex.

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
dominiktassilostill: Unfortunately not, `MappingInfo` is trying to create a face quadrature, which fails for `Qudarature<dim>(dof_handler_u.get_fe().get_unit_support_points())` as the quadrature doesn't know any rule how to create the face quadrature.

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_dual_splitting.cpp
```diff
@@ -596,64 +596,126 @@ OperatorDualSplitting<dim, Number>::local_interpolate_velocity_dirichlet_bc_boun
...
+            g = FunctionEvaluator<1, dim, Number>::value(*bc, face, q, quad_index);
+          }
+          else
+          {
+            AssertThrow(false, dealii::ExcMessage("Not implemented."));
+          }
+          // we do not need JxW so just write directly into values
+          auto * values = integrator.begin_values() + q;
+          for(unsigned int comp = 0; comp < integrator.n_components; ++comp)
+            values[comp * integrator.n_q_points] = g[comp];
```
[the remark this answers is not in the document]
dominiktassilostill: Yes, I think it should also work for hypercube elements, leading to the exact same code for both cases.

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_dual_splitting.cpp
dominiktassilostill: Good idea, I like the name.

nfehn:
I like the new version a lot!

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
```diff
@@ -1865,8 +1871,11 @@ SpatialOperatorBase<dim, Number>::local_interpolate_stress_bc_boundary_face(
         vector traction =
           param.density * (viscosity * (grad_u + transpose(grad_u)) * normal - p * normal);
 
-        integrator_u.submit_dof_value(traction, index);
+        auto * values = integrator_u.begin_values() + q;
+        for(unsigned int comp = 0; comp < integrator_u.n_components; ++comp)
+          values[comp * integrator_u.n_q_points] = traction[comp];
```
nfehn, line 1876: I wonder whether we should add a comment here, maybe a reference to a class/function/data structure in deal.II providing some information or explanation. As mentioned in another comment, why should we have a guarantee that this will never change in deal.II?

By looking at this code only, one might argue that the `n_components` values of one quadrature point are not next to each other in the array `values`. For a very large array `values`, this would probably be inefficient (data locality). Depending on the background/perspective, this might look counter-intuitive to a user/developer of ExaDG, so a comment providing some reasoning could be helpful.

Finally, I wonder whether this functionality wouldn't actually call for a public member function of FEEvaluation/Integrator classes in deal.II? (interface e.g. like `write_to_values(integrator, q, data)`)

What do you think @dominiktassilostill @kronbichler ?

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_dual_splitting.cpp
```diff
@@ -641,10 +630,13 @@ OperatorDualSplitting<dim, Number>::local_interpolate_velocity_dirichlet_bc_boun
         {
           AssertThrow(false, dealii::ExcMessage("Not implemented."));
         }
-
-        integrator.submit_dof_value(g, index);
+        // we do not need JxW so just write directly into values
+        auto * values = integrator.begin_values() + q;
+        for(unsigned int comp = 0; comp < integrator.n_components; ++comp)
+          values[comp * integrator.n_q_points] = g[comp];
```
nfehn, line 636: see comment https://github.com/exadg/exadg/pull/677/files#r1666513454

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
dominiktassilostill: Adding a comment and a reference to `submit_value` would certainly make things more clear. As the code is reused multiple times in the library now (and I think there are even more places where it is applicable), creating a separate function (with more in detail descriptions) for it definitely makes sense but I do not know if creating a function in `FEEvaluation` (or its base class) is necessary or not.

kronbichler: > By looking at this code only, one might argue that the n_components values of one quadrature point are not next to each other in the array values. For a very large array values, this would probably be inefficient (data locality).

The reason the data fields are not next to each other is because the computation kernels to do the interpolation/integration (`FEEvaluation::evaluate` and `integrate`) are set up component by component, with the component being a run-time parameter. However, run-time parameters cannot be used as strides without making the code significantly slower. Note also that data locality is not at all a problem here, as there is a loop around it. Caches and associated hardware prefetchers are pretty good at handling access to multiple "streams". There is a corner case of cache conflict misses, and you might ask the question under which parameters for our variables these appear. I recommend you to read https://en.wikipedia.org/wiki/Cache_performance_measurement_and_metric and check the structure of set-associative caches that are the standard on CPU today. I'd be happy to discuss more about this and possible mitigation strategies if you find problematic cases.

> Finally, I wonder whether this functionality wouldn't actually call for a public member function of FEEvaluation/Integrator classes in deal.II? (interface e.g. like `write_to_values(integrator, q, data)`)

If you want to make a suggestion @nfehn I'd be happy to discuss it. I am not entirely sure if the present code is an important enough use case, as I will elaborate in a later message, but having the code at hand would definitely help the decision making.

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_dual_splitting.cpp
```diff
@@ -641,10 +630,13 @@ OperatorDualSplitting<dim, Number>::local_interpolate_velocity_dirichlet_bc_boun
...
       }
-
+      // integration reorders the integration points to the corresponding DoFs
+      integrator.integrate(dealii::EvaluationFlags::values);
```
kronbichler, line 639: I think to a reader of the code it might not be entirely clear what kind of reordering is needed here. The important point is that, in a slight misuse of the computational interfaces, the "quadrature points" on the faces get used to embed the information into the cell DoFs by this call, to later be able to set the respective values on the cell. However, after having thought about the used concept I believe that this concept is extremely brittle: Think about the case of boundary conditions for two faces of an element. When we work on the second face, only the values along that face will be non-zero after `integrate`, the other values are zero, include the ones we tried to set on the other face. But we then call `set_dof_values`, thus overwriting the other values. For Navier-Stokes, this only affects Raviart-Thomas elements where only one component is zero (but I wonder if we take the Piola transformation properly into account), but for the structure and Poisson cases I definitely wonder if this is the right approach. I might misunderstand something here, but it definitely looks like a very dangerous approach to me.

kronbichler:
The code in question makes me wonder if we take the right direction when we use the matrix-free evaluators to impose boundary conditions. I already mentioned the issue about possible correctness in corners of the mesh due to `set_dof_values` and interpolation from the face to the cell DoFs that I either don't understand yet or that might even be wrong now. But even apart from this correctness issue, I think that adding a quadrature formula via `cell->get_unit_support_points()` is not a good strategy, for several reasons:
- The `MappingInfo` field of `MatrixFree` evaluates all geometric information (Jacobians of the transformation, Jacobian determinants, quadrature points, possibly second derivatives) for all cells and faces. All of these need to be computed, stored, and possibly accessed. Since the inhomogeneous Dirichlet boundary conditions are accessed relatively rarely and only on the Dirichlet boundary, I think that evaluating them on the fly might be the more economic decision.
- There is a lot of code added here, to create a new quadrature formula (with associated index handling in our basic infrastructure classes), the evaluation and sorting into the `FaceIntegrator` classes with the possible issues discussed in other comments of this PR. However, interpolating boundary conditions has a very intuitive interface in deal.II,
```c++
  std::map<dealii::types::global_dof_index, double> boundary_values;
  dealii::VectorTools::interpolate_boundary_values(mapping, dof_handler, function_map, boundary_values);
  for (auto const bc_pair : boundary_values)
    dst(bc_pair.first) = bc_pair.second;
```
This is much more compact code and self-documenting as our users need to be well-familiar with deal.II in any case. One could raise a concern of how efficient the evaluation of the mapping and associated algorithms to loop over the mesh boundary are in deal.II. I believe that we should rather concentrate our efforts to improve that function in case we identify it as a bottleneck, rather than trying to do it with a function that looks very brittle (the primary reason is, in my opinion, not the order of values/components behind the pointer returned by `FaceIntegrator::begin_values()`, albeit the argument that we depend on deal.II not making changes is relevant here).

To summarize: Should we try to remove all code that runs `MatrixFree::loop()` to impose Dirichlet boundary conditions strongly, and see if we can express all places with `VectorTools::interpolate_boundary_values`? There might be cases where we are missing something, but I think we should try to avoid to fill the matrix-free object with a quadrature formula defined in the support points of the finite element.

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_dual_splitting.cpp
```diff
@@ -613,15 +609,8 @@ OperatorDualSplitting<dim, Number>::local_interpolate_velocity_dirichlet_bc_boun
     if(boundary_type == BoundaryTypeU::Dirichlet or boundary_type == BoundaryTypeU::DirichletCached)
     {
       integrator.reinit(face);
-      integrator.read_dof_values(dst);
-
```
kronbichler, line 617: I had another look at this code and correctness issue I talked about. In the old code, this line (now deleted) made sure that we would keep the old DoF values set on other faces, as we only wrote the values for those entries located on the face via `face_to_cell_index_nodal`. But once we start interpolating from the face quadrature points to the cell, we lose the old values, so the proposed change will indeed run into the trap I mentioned in the other comment. I suggest to discuss this more generally, ideally by going to `VectorTools::interpolate_boundary_values`. As a side note, that function got considerably more efficient in the past few years when we implemented better interpolators for `MappingQ` via the infrastructure also used by `FEPointEvaluation`, and we can certainly have another look if necessary.

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_dual_splitting.cpp
```diff
@@ -613,15 +609,8 @@ OperatorDualSplitting<dim, Number>::local_interpolate_velocity_dirichlet_bc_boun
...
       for(unsigned int q = 0; q < integrator.n_q_points; ++q)
       {
-        unsigned int const local_face_number = matrix_free.get_face_info(face).interior_face_no;
-
-        unsigned int const index = matrix_free.get_shape_info(dof_index, quad_index)
-                                     .face_to_cell_index_nodal[local_face_number][q];
```
nfehn, line 623: In case we want to generalize rather this code towards simplex elements: We might want to take the class `ContainerInterfaceData` as a "template" to write a similar data structure for the mapping between q-points and degrees of freedom, see https://github.com/exadg/exadg/blob/5f70dfbcc5f8daac94e73859fb5d216dd49f8d76/include/exadg/functions_and_boundary_conditions/container_interface_data.h#L44-L58

Note that here in `ContainerInterfaceData`, the data type `MapVectorIndex` maps `Id = {face, q, v}` to the vector-index in the corresponding "data-arrays" of `ContainerInterfaceData`. Note that `dealii::types::global_dof_index` is not really used as a "dof-index" here.

For the present problem of mapping q-indices to dof-indices, maybe a similar data type, e.g. `MapDofIndices = std::map<Id, dealii::types::global_dof_index>` could be used for the mapping from q-indices to dof-indices? We could create this data structure in the setup phase and then read from that data structure during the simulation (very similar to the setup and use of `ContainerInterfaceData`).

We probably do not want to integrate this functionality into `ContainerInterfaceData`, because (i) not all use cases of `ContainerInterfaceData` need this information according to my understanding, and because (ii) we also need this information in cases where we do not have a corresponding `ContainerInterfaceData` (see function `local_interpolate_stress_bc_boundary_face()` below).

nfehn:
A summary of the discussion we had in person is that we try to use the previous interface with `integrator.read_dof_values()` and `integrator.submit_dof_value(g, index)` also for the simplex case. We need to properly map indices in this case, see comment https://github.com/exadg/exadg/pull/677#discussion_r1668344355 in this context.

Another solution could be to try to avoid  matrix-free loops at all in this context, as suggested by @kronbichler. We could use `dealii::interpolate_boundary_values()` for "standard" boundary conditions, but would need a corresponding solution also for boundaries of type `DirichletCached`. The function `local_interpolate_stress_bc()` fills boundary dofs neither from `dealii::Function` nor from `ContainerInterfaceData` (`DirichletCached` bc), but from some "user-defined" function as a function of other dof-vectors. We postponed this approach since boundary conditions of type `DirichletCached` are right now easily integrated into the existing infrastructure and would require a special treatment in the other case.

dominiktassilostill:
@nfehn could you have a look at this again? I think this is the minimal changes to be done with the current interface.

nfehn (approved):

PR #678, opened 2024-07-05
Fix parameter misuse for TemporalDiscretization::InterpolateAnalyticalSolution

nfehn:
This replaces PR #669. The present PR is also related to issue #664.

I see the problem here: https://github.com/exadg/exadg/blob/d216d419a25726843b7ac34b21bac391945ba6cd/include/exadg/incompressible_navier_stokes/spatial_discretization/create_operator.h#L74-L82

Creating a pressure-correction spatial operator wants to create multigrid preconditioners. However, the function `Parameters::involves_h_multigrid()` returns false for `TemporalDiscretization::InterpolateAnalyticalSolution` in the current implementation. I would currently argue that this is kind of a misuse/hack of `IncNS::Parameters` in combination with the `Driver` class and cannot be expected to work as expected.

The present PR aims to make the code consistent with the "hack" of creating `OperatorPressureCorrection` if `TemporalDiscretization::InterpolateAnalyticalSolution`. To have a clean solution of this problem, we need to implement a new class `OperatorInterpolateAnalyticalSolution` (deriving from `SpatialOperatorBase`) that does not want to create/initialize any preconditioners (including multigrid).

### DIFF
--- a/include/exadg/incompressible_navier_stokes/user_interface/parameters.cpp
+++ b/include/exadg/incompressible_navier_stokes/user_interface/parameters.cpp
@@ -637,8 +637,14 @@ Parameters::involves_h_multigrid() const
       }
     }
   }
+  // TODO: we currently treat TemporalDiscretization::InterpolateAnalyticalSolution as a splitting
+  // scheme, since the implementation creates a spatial operator related to the pressure-correction
+  // scheme in case of InterpolateAnalyticalSolution. This is only a temporary solution and we need
+  // to write a separate class SpatialOperatorInterpolateAnalyticalSolution that does not create
+  // preconditioners (including multigrid)
   else if(temporal_discretization == TemporalDiscretization::BDFDualSplittingScheme or
-          temporal_discretization == TemporalDiscretization::BDFPressureCorrection)
+          temporal_discretization == TemporalDiscretization::BDFPressureCorrection or
+          temporal_discretization == TemporalDiscretization::InterpolateAnalyticalSolution)
   {
     // pressure step is the same for both time discretization schemes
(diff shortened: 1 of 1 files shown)


jh66637:
I cannot test this at the moment, but if I remember correctly the problem did not only occur for the analytical time integrator.

kronbichler (approved):
I'm ok with this change, but let us wait for @jh66637 to do the check.

jh66637:
The fix works, thank you :)

Could you set 
```
"AdditionalCFDRefinementsAroundSource": "1"
```
in `input.json` and replace the lines 589 to 592 in the application file by
```
    this->boundary_descriptor->velocity->neumann_bc.insert(
      std::make_pair(1, std::make_shared<dealii::Functions::ZeroFunction<dim>>(dim)));
    this->boundary_descriptor->pressure->dirichlet_bc.insert(
      std::make_pair(1, std::make_shared<AnalyticalSolutionPressure<dim>>(intensity, r_0, r_c)));
```
? Currently `do_initialize_dof_handler_and_constraints()` fails with `is_singular == false` for `BDFDualSplitting` and `BDFPressureCorrection`. I am also open to other suggestions to circumvent this issue. 

Besides this, everything seems to work :+1:

nfehn:
Do I understand the above comment correctly that there is another problem if you really want to solve the fluid problem (with dual-splitting or pressure-correction time stepping)? If so, let's create another issue and related PRs. I think it is important that we address this problem and do not only circumvent the problem by other parameter settings.

nfehn:
I think I will merge this PR now so that we have this problem (with the use of parameters) fixed.

PR #679, opened 2024-07-05
Only print relevant parameters for InterpolateAnalyticalSolution

nfehn:
Currently, many parameters of the incompressible flow solver are printed for `TemporalDiscretization::InterpolateAnalyticalSolution`, even though these parameters are irrelevant. The present PR suggests to only print those that are relevant or are being used.

According to my understand, these parameters are (i) the start/end time, (ii) parameters on the computation of the time step size, (iii) and the spatial discretization (mapping degree, polynomial degrees). @jh66637 please check if this is correct. 

I decided to slightly switch the order of some parameters in the output, in order to keep the implementation and readability of the code as simple as possible.

### DIFF
--- a/applications/incompressible_navier_stokes/free_stream/tests/coupled_bdf2_moving_mesh.output
+++ b/applications/incompressible_navier_stokes/free_stream/tests/coupled_bdf2_moving_mesh.output
@@ -42,7 +42,6 @@ Physical quantities:
 
 Temporal discretization:
   Temporal discretization method:            BDFCoupledSolution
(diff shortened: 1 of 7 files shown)


kronbichler (approved):
I think early `return` statements can be confusing to some people looking at these functions, because it essentially wraps all the code inside of the remaining part of the function into an `if` statement with the negated clause of the `return`, without making this obvious at other places in the function, but I also do not see a simpler solution, so let us go with this choice here.

PR #680, opened 2024-07-05
rename parameter in IncNS::Parameters

nfehn:
When working on issues/PRs #664 , #669, #678, I was irritated by the name `use_global_coarsening` in `IncNS::Parameters`, which is going to be renamed in the present PR.


richardschu (approved):
I grepped 'use_global_coarsening', these are the last remnants of the time we supported local_smoothing and global_coarsening, huh?
Anyways, thanks for the PR, makes it clearer!

kronbichler (approved):

PR #681, opened 2024-07-16
introduce matrix-based vmult in OperatorBase

### LINKED ISSUE #682: Not all variants of MGCoarseKrylov use projected rhs vector (with nullspace projection)
See here https://github.com/exadg/exadg/blob/04c253618e9cca70ec839f6a57be75b839f91f00/include/exadg/solvers_and_preconditioners/multigrid/coarse_grid_solvers.h#L194-L196

Only the Trilinos ML branch below uses the vector `r`, while PETsc Boomer uses `src`, and also other preconditioners use `src`.

Do we all agree that this is not correct?

nfehn:
This PR aims to introduce the option to use a matrix-based `vmult()` operation in `OperatorBase`, e.g. in order to perform throughput studies with sparse matrices (instead of the usual matrix-free evaluation used in ExaDG).

This option is enabled in the present PR for the `Structure` module.

~~@richardschu Please note that with this PR, the system matrix will **not** be correctly updated when simulating a time-dependent/nonlinear problem. It will only calculate the system matrix once during setup, e.g. for throughput studies.~~

For nonlinear problems, the system matrix of the linearized operator is computed in `set_solution_linearization()` (a function that is called by the Newton solver).

For linear problems, we compute the system matrix prior to solving the linear system of equations. For time-dependent problems with constant parameters, the matrix might be calculated more often than actually needed. Upcoming PRs would have to address this in case one wants to avoid the re-assembly of the system matrix. No work is done for the usual matrix-free implementation that is used by default.

closes #682

### DIFF
--- a/include/exadg/operators/operator_base.cpp
+++ b/include/exadg/operators/operator_base.cpp
@@ -418,6 +418,11 @@ OperatorBase<dim, Number, n_components>::apply_matrix_based(VectorType &       d
           dealii::LinearAlgebra::distributed::Vector<double> const & src_double) {
         system_matrix_trilinos.vmult(dst_double, src_double);
       });
+#else
+    AssertThrow(
+      false,
+      dealii::ExcMessage(
+        "Make sure that DEAL_II_WITH_TRILINOS is activated if you want to use SparseMatrixType::Trilinos."));
 #endif
   }
   else if(this->data.sparse_matrix_type == SparseMatrixType::PETSc)
@@ -434,6 +439,11 @@ OperatorBase<dim, Number, n_components>::apply_matrix_based(VectorType &       d
                               system_matrix_petsc.vmult(petsc_dst, petsc_src);
                             });
     }
+#else
+    AssertThrow(
+      false,
+      dealii::ExcMessage(
+        "Make sure that DEAL_II_WITH_PETSC is activated if you want to use SparseMatrixType::PETSc."));
 #endif
   }
   else
--- a/include/exadg/operators/operator_base.cpp
+++ b/include/exadg/operators/operator_base.cpp
@@ -30,8 +30,8 @@
 #include <exadg/operators/operator_base.h>
 #include <exadg/solvers_and_preconditioners/utilities/block_jacobi_matrices.h>
 #include <exadg/solvers_and_preconditioners/utilities/invert_diagonal.h>
-#include <exadg/solvers_and_preconditioners/utilities/verify_calculation_of_diagonal.h>
 #include <exadg/solvers_and_preconditioners/utilities/linear_algebra_utilities.h>
+#include <exadg/solvers_and_preconditioners/utilities/verify_calculation_of_diagonal.h>
 
 namespace ExaDG
 {
--- a/include/exadg/operators/operator_base.cpp
+++ b/include/exadg/operators/operator_base.cpp
@@ -376,23 +376,13 @@ OperatorBase<dim, Number, n_components>::apply_matrix_based(VectorType &       d
   if(this->data.sparse_matrix_type == SparseMatrixType::Trilinos)
   {
 #ifdef DEAL_II_WITH_TRILINOS
-    if constexpr(std::is_same_v<Number, double>)
-    {
-      system_matrix_trilinos.vmult(dst, src);
-    }
-    else
-    {
-      // create temporal vectors of type double
-      dealii::LinearAlgebra::distributed::Vector<double> dst_double, src_double;
-      dst_double.reinit(dst, false);
-      src_double.reinit(src, true);
-      src_double = src;
-
-      system_matrix_trilinos.vmult(dst_double, src_double);
-
-      // convert: double -> Number
-      dst.copy_locally_owned_data_from(dst_double);
-    }
+    apply_function_in_double_precision(
+      dst,
+      src,
+      [&](dealii::LinearAlgebra::distributed::Vector<double> &       dst_double,
+          dealii::LinearAlgebra::distributed::Vector<double> const & src_double) {
+        system_matrix_trilinos.vmult(dst_double, src_double);
+      });
 #endif
   }
   else if(this->data.sparse_matrix_type == SparseMatrixType::PETSc)
--- a/include/exadg/operators/operator_base.cpp
+++ b/include/exadg/operators/operator_base.cpp
@@ -368,13 +368,30 @@ OperatorBase<dim, Number, n_components>::apply_matrix_based(VectorType &       d
   if(this->data.sparse_matrix_type == SparseMatrixType::Trilinos)
   {
 #ifdef DEAL_II_WITH_TRILINOS
-    system_matrix_trilinos.vmult(dst, src);
+    if constexpr(std::is_same_v<Number, double>)
+    {
+      system_matrix_trilinos.vmult(dst, src);
+    }
+    else
+    {
+      // create temporal vectors of type double
+      dealii::LinearAlgebra::distributed::Vector<double> dst_double, src_double;
+      dst_double.reinit(dst, false);
+      src_double.reinit(src, true);
+      src_double = src;
+
+      system_matrix_trilinos.vmult(dst_double, src_double);
+
+      // convert: double -> Number
+      dst.copy_locally_owned_data_from(dst_double);
+    }
 #endif
   }
   else if(this->data.sparse_matrix_type == SparseMatrixType::PETSc)
   {
 #ifdef DEAL_II_WITH_PETSC
     if(system_matrix_petsc.m() > 0)
+    {
       apply_petsc_operation(dst,
                             src,
                             petsc_vector_dst,
@@ -383,6 +400,7 @@ OperatorBase<dim, Number, n_components>::apply_matrix_based(VectorType &       d
                                 dealii::PETScWrappers::VectorBase const & petsc_src) {
                               system_matrix_petsc.vmult(petsc_dst, petsc_src);
                             });
+    }
 #endif
   }
   else
@@ -396,7 +414,8 @@ void
 OperatorBase<dim, Number, n_components>::apply_matrix_based_add(VectorType &       dst,
                                                                 VectorType const & src) const
 {
-  VectorType tmp(dst);
+  VectorType tmp;
+  tmp.reinit(dst, false);
 
   apply_matrix_based(tmp, src);
 
--- a/include/exadg/operators/operator_base.cpp
+++ b/include/exadg/operators/operator_base.cpp
@@ -146,14 +146,20 @@ template<int dim, typename Number, int n_components>
 void
 OperatorBase<dim, Number, n_components>::vmult(VectorType & dst, VectorType const & src) const
 {
-  this->apply(dst, src);
+  if(this->data.use_matrix_based_vmult)
+    this->apply_matrix_based(dst, src);
+  else
+    this->apply(dst, src);
 }
 
 template<int dim, typename Number, int n_components>
 void
 OperatorBase<dim, Number, n_components>::vmult_add(VectorType & dst, VectorType const & src) const
 {
-  this->apply_add(dst, src);
+  if(this->data.use_matrix_based_vmult)
+    this->apply_matrix_based_add(dst, src);
+  else
+    this->apply_add(dst, src);
 }
 
 template<int dim, typename Number, int n_components>
@@ -304,6 +310,20 @@ OperatorBase<dim, Number, n_components>::apply_add(VectorType & dst, VectorType 
       dst.local_element(constrained_index) += src.local_element(constrained_index);
     }
   }
+}
+
+template<int dim, typename Number, int n_components>
+void
+OperatorBase<dim, Number, n_components>::apply_matrix_based(VectorType &       dst,
+                                                            VectorType const & src) const
+{
+}
+
+template<int dim, typename Number, int n_components>
+void
+OperatorBase<dim, Number, n_components>::apply_matrix_based_add(VectorType &       dst,
+                                                                VectorType const & src) const
+{
 }
 
 template<int dim, typename Number, int n_components>
--- a/include/exadg/operators/operator_base.cpp
+++ b/include/exadg/operators/operator_base.cpp
@@ -30,8 +30,8 @@
 #include <exadg/operators/operator_base.h>
 #include <exadg/solvers_and_preconditioners/utilities/block_jacobi_matrices.h>
 #include <exadg/solvers_and_preconditioners/utilities/invert_diagonal.h>
-#include <exadg/solvers_and_preconditioners/utilities/petsc_operation.h>
 #include <exadg/solvers_and_preconditioners/utilities/verify_calculation_of_diagonal.h>
+#include <exadg/solvers_and_preconditioners/utilities/linear_algebra_utilities.h>
 
 namespace ExaDG
 {
--- a/include/exadg/operators/operator_base.cpp
+++ b/include/exadg/operators/operator_base.cpp
@@ -85,53 +85,6 @@ OperatorBase<dim, Number, n_components>::reinit(
     this->matrix_free->get_dof_handler(this->data.dof_index);
 
   n_mpi_processes = dealii::Utilities::MPI::n_mpi_processes(dof_handler.get_communicator());
-
-  if(this->data.use_matrix_based_vmult)
-  {
-    if(this->data.sparse_matrix_type == SparseMatrixType::Trilinos)
-    {
-#ifdef DEAL_II_WITH_TRILINOS
-      init_system_matrix(system_matrix_trilinos, dof_handler.get_communicator());
-      calculate_system_matrix(system_matrix_trilinos);
-#else
-      AssertThrow(
-        false,
-        dealii::ExcMessage(
-          "Make sure that DEAL_II_WITH_TRILINOS is activated if you want to use SparseMatrixType::Trilinos."));
-#endif
-    }
-    else if(this->data.sparse_matrix_type == SparseMatrixType::PETSc)
-    {
-#ifdef DEAL_II_WITH_PETSC
-      init_system_matrix(system_matrix_petsc, dof_handler.get_communicator());
-      calculate_system_matrix(system_matrix_petsc);
-
-      if(system_matrix_petsc.m() > 0)
-      {
-        // get vector partitioner
-        dealii::LinearAlgebra::distributed::Vector<Number> vector;
-        initialize_dof_vector(vector);
-        VecCreateMPI(system_matrix_petsc.get_mpi_communicator(),
-                     vector.get_partitioner()->locally_owned_size(),
-                     PETSC_DETERMINE,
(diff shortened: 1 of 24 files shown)


include/exadg/operators/operator_base.cpp
```diff
@@ -306,6 +360,68 @@ OperatorBase<dim, Number, n_components>::apply_add(VectorType & dst, VectorType
...
+    if(system_matrix_petsc.m() > 0)
+    {
+      apply_petsc_operation(dst,
+                            src,
+                            petsc_vector_dst,
+                            petsc_vector_src,
+                            [&](dealii::PETScWrappers::VectorBase &       petsc_dst,
+                                dealii::PETScWrappers::VectorBase const & petsc_src) {
+                              system_matrix_petsc.vmult(petsc_dst, petsc_src);
+                            });
```
nfehn, line 402: @peterrum, @kronbichler This code compiled without float/double conversion as needed for the Trilinos case. Does this mean that the PETSc case computes the mat-vec product in float, or is the float/double conversion just done at a different place (e.g. within deal.II)?

include/exadg/operators/operator_base.cpp
```diff
@@ -306,6 +360,68 @@ OperatorBase<dim, Number, n_components>::apply_add(VectorType & dst, VectorType
...
+#ifdef DEAL_II_WITH_TRILINOS
+    if constexpr(std::is_same_v<Number, double>)
+    {
+      system_matrix_trilinos.vmult(dst, src);
+    }
+    else
+    {
+      // create temporal vectors of type double
+      dealii::LinearAlgebra::distributed::Vector<double> dst_double, src_double;
+      dst_double.reinit(dst, false);
```
nfehn, line 379: I adapted this code from another place in ExaDG. My question is whether we really need `false` here as the second argument, because `dst` is overwritten anyway in `vmult()`, isn't it?

nfehn:
It would be great if someone feeling familiar with matrix-based implementations could have a look at the code.

What I still have to do in the present PR - but this should not affect the implementation of the matrix-vec product - is to introduce parameters e.g. in `Structure::Parameters`, such that the variant of matrix-based `vmult()` can be used in applications.

richardschu:
Hey @nfehn thanks for the PR, that would definitely make for much stronger arguments regarding performance comparisons matrix-free vs. matrix-based. Due to the fact that we want to use this for throughput studies only (I would say),  the limitation regarding inital assembly + never update is fine. Also, something like this would be super useful for future investigations. I think for the current paper, we could add results with matrix-based MG with low effort having this.
I could have a look at this PR, but am super busy these days due to the upcoming conference, teaching duties, etc. - so if no one else has time, I can have a look, but that would likely be on Friday or the weekend!

nfehn:
For matrix-based MG, we need to update the matrix properly (also on coarser levels), which I could address e.g. in a second PR.

nfehn:
Please note the updated description of this PR.

include/exadg/operators/operator_base.cpp
kronbichler: Do you mean places like https://github.com/exadg/exadg/blob/c62fd0a241beab24a91cea6fd64680cbfb2c17cd/include/exadg/operators/operator_base.cpp#L319-L328 Indeed, we should be able to replace it by `true` here (where true means "omit initializing the vector entries to zero"). The reason why the other function needs it is because that other place runs a matrix-free loop that adds into the vector, so we must start from zero. (Or we could ask the loop to set it to zero, it does not matter too much for that particular function.)

include/exadg/operators/operator_base.cpp
kronbichler: PETSc is not templated on the number type, so it is (in the standard compilation of PETSc) double-precision numbers in the matrix. I say standard compilation because you can configure PETSc with `--with-precision=single`, albeit it has been a long time ago I tried this the last time so I cannot say what precisely it entails today. In consequence, the number types get converted as we copy in/out of the PETSc vector, e.g. here: https://github.com/exadg/exadg/blob/c62fd0a241beab24a91cea6fd64680cbfb2c17cd/include/exadg/solvers_and_preconditioners/utilities/petsc_operation.h#L66-L69

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
```diff
@@ -259,8 +259,8 @@ class PreconditionerBoomerAMG : public PreconditionerBase<Number>
   BoomerData boomer_data;
 
   // PETSc vector objects to avoid re-allocation in every vmult() operation
-  mutable VectorTypePETSc petsc_vector_src;
-  mutable VectorTypePETSc petsc_vector_dst;
+  mutable Vec petsc_vector_src;
+  mutable Vec petsc_vector_dst;
```
kronbichler, line 263: Is there a reason you change this? I think we introduced the more verbose name https://github.com/exadg/exadg/blob/c62fd0a241beab24a91cea6fd64680cbfb2c17cd/include/exadg/solvers_and_preconditioners/utilities/petsc_operation.h#L31-L34 to ensure that the ambiguous name `Vec` is not ending up all over the place. We could move that to a more central location in ExaDG if you do not want to include `petsc_operator.h` in `operator_base.h`, or some similar setting.

kronbichler (approved):
OK with me, apart from the question I posted before regarding the `Vec` type (that I find confusing), but I am fine either way.

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
kronbichler: OK, let's keep this for now. For matrices one can probably create a simple interface with the basic methods to add entries, mat-vec and similar. The problem is that interfaces to deal.II are going to be a bit harder to adjust because several of those are templated, rather than accessed via virtual functions. One could see if there is a way to make that uniform there, but it would probably be a bigger project.

nfehn: I tried the variant above (`ExaDG::SparseMatrix`) but did not find a reasonable solution to hide the ifdef's in `OperatorBase` and similar files/classes. The problem is in my opinion the circumstance that the classes `MultigridOperatorBase` and `MultigridOperator` use virtual functions where we cannot use a template for the sparse matrix (for this reason, also these two classes implement certain functions for Trilinos and PETSc in the current code). `ExaDG::SparseMatrix` worked well just for the `vmult()` function, which is however only one of several functions where we distinguish between Trilinos and PETSc, which is why I think it is not worth introducing `ExaDG::SparseMatrix` just for this function and still have the ifdefs for the other occurrences.

It could work better if we derive `ExaDG::SparseMatrix` from a sparse matrix type/class in deal.II and then integrate multiple sparse matrix types (like Trilinos, PETSc) via composition and ifdef's in the member functions of such a class. The funtions `internal_init/calculate_system_matrix()` in `OperatorBase` currently use a template `SparseMatrix`, meaning that we rely on a common interface that both Trilinos and PETSc fulfill. Hence, deriving `ExaDG::SparseMatrix` from a deal.II sparse matrix fulfilling this interface could imply that we can just remove these template parameters from certain functions in `OperatorBase`, `MultigridOperatorBase`, and `MultigridOperator`. However, I did not try this, so potential problems (which I am not aware of right now) might show up there as well.

include/exadg/operators/operator_base.cpp
```diff
@@ -306,6 +313,113 @@ OperatorBase<dim, Number, n_components>::apply_add(VectorType & dst, VectorType
...
+        initialize_dof_vector(vector);
+        VecCreateMPI(system_matrix_petsc.get_mpi_communicator(),
+                     vector.get_partitioner()->locally_owned_size(),
+                     PETSC_DETERMINE,
+                     &petsc_vector_dst);
+        VecCreateMPI(system_matrix_petsc.get_mpi_communicator(),
+                     vector.get_partitioner()->locally_owned_size(),
+                     PETSC_DETERMINE,
+                     &petsc_vector_src);
+      }
```
nfehn, line 356: @kronbichler I am not sure whether I use this correctly, because the present function `assemble_matrix_if_necessary()` is potentially called again and again during a simulation, while `VecDestroy` appears only in the destructor of this class (`OperatorBase`).

include/exadg/operators/operator_base.cpp
nfehn: The `vmult()` function for AMG preconditioners is implemented in the class `PreconditionerAMG` used by both Trilinos/PETSc AMG preconditioners. So should we distinguish the implementation of this function for the Trilinos case and the PETSc case and do the float->double conversion of the deal.II vectors only in the Trilinos case, while we only call `amg->vmult()` for PETSc?

include/exadg/operators/operator_base.cpp
nfehn: I probably changed this code so that the above reference might have disappeared. Sorry.

I meant the `reinit()` before we do the `vmult()`. So the question would be whether we need a `dst_double=0` prior to calling `vmult()`. I would argue no because `vmult()` overwrites the values anyway. But I am asking nevertheless because we also had in the past special aspects e.g. related to ghost values where I am not completely sure whether such a change might invalidate something.

include/exadg/solvers_and_preconditioners/utilities/linear_algebra_utilities.h
```diff
@@ -133,6 +128,40 @@ apply_petsc_operation(
...
+{
+  if constexpr(std::is_same_v<Number, double>)
+  {
+    operation(dst, src);
+  }
+  else
+  {
+    // create temporal vectors of type double
+    dealii::LinearAlgebra::distributed::Vector<double> dst_double, src_double;
+    dst_double.reinit(dst, false); // zero entries
```
nfehn, line 154: related to comment https://github.com/exadg/exadg/pull/681#discussion_r1680581114

@kronbichler I introduced this free function to avoid code duplication at different places in ExaDG. `reinit(dst, false)` is as in the previous implementation. So my questions from the other comment would be whether we should write `reinit(dst, true)` here?

nfehn:
I checked that throughput studies and a nonlinear problem (with preconditioner update and re-assembly of the level matrices) runs without errors. This should be fine now.

Let me mention the implications/limitations of the present implementation:

1. If `use_matrix_based_implementation = true`, we assemble and store the matrix on the fine level twice, due to the "matrix-free driven design" of ExaDG with separate operators for the multigrid levels and the fine-level operator in the Krylov solver mat-vec product.
2. If `use_matrix_based_implementation = true` and if AMG is used as a coarse grid solver or as a preconditioner for the fine-level problem, the AMG preconditioner has its own matrix, so we would again assemble/store twice the same matrix on the coarse-level or on the fine level.

In my opinion, the second point is easier to resolve than the first point. We could remove the matrices from the AMG preconditioner implementations for Trilinos/PETSc and use the matrix from the underlying operator, which is derived from `OperatorBase`. If we want to realize this, we need to make the implementation of `OperatorBase` more flexible: 
- In contrast to the present implementation, assembly of the matrices in `OperatorBase` would become independent of the usage of matrices in `vmult()` and no longer be driven by a single parameter `use_matrix_based_implementation`, i.e. an operator may assemble the matrix (because AMG needs it) but still use a matrix-free implementation in the Krylov mat-vec. If a matrix-based implementation is to be used in the Krylov mat-vec product (e.g. without any AMG preconditioner involved), we would of course have to assert that the matrix has indeed been initialized/assembled.

Please let me know what you think about these implications/limitations. Please let me also know if I should generalize the implementation so that we can use an AMG preconditioner together with a matrix-based Krylov mat-vec, but only assemble and store the matrix once.

richardschu:
> I checked that throughput studies and a nonlinear problem (with preconditioner update and re-assembly of the level matrices) runs without errors for `applications/structure/manufactured`. This should be fine now.
> 
> Let me mention the implications/limitations of the present implementation:
> 
>     1. If `use_matrix_based_implementation = true`, we assemble and store the matrix on the fine level twice, due to the "matrix-free driven design" of ExaDG with separate operators for the multigrid levels and the fine-level operator in the Krylov solver mat-vec product.
> 
>     2. If `use_matrix_based_implementation = true` and if AMG is used as a coarse grid solver or as a preconditioner for the fine-level problem, the AMG preconditioner has its own matrix, so we would again assemble/store twice the same matrix on the coarse-level or on the fine level.
> 
> 
> In my opinion, the second point is easier to resolve than the first point. We could remove the matrices from the AMG preconditioner implementations for Trilinos/PETSc and use the matrix from the underlying operator, which is derived from `OperatorBase`. If we want to realize this, we need to make the implementation of `OperatorBase` more flexible:
> 
>     * In contrast to the present implementation, assembly of the matrices in `OperatorBase` would become independent of the usage of matrices in `vmult()` and no longer be driven by a single parameter `use_matrix_based_implementation`, i.e. an operator may assemble the matrix (because AMG needs it) but still use a matrix-free implementation in the Krylov mat-vec. If a matrix-based implementation is to be used in the Krylov mat-vec product (e.g. without any AMG preconditioner involved), we would of course have to assert that the matrix has indeed been initialized/assembled.
> 
> 
> Please let me know what you think about these implications/limitations. Please let me also know if I should generalize the implementation so that we can use an AMG preconditioner together with a matrix-based Krylov mat-vec, but only assemble and store the matrix once.

Thank you for your efforts Niklas!
I think the matrix-based approach in general will be used for comparison studies mainly.
Therefore I would argue that the limitations you mention are not influencing the metrics we measure in the throughput studies at least. The memory consumption is ofc impacted, ok, but as long as we stay in that throughput test setting and do not go to an actual BVP, we do not see any impact of this, right?
If one would want to test this approach in an actual BVP, then we would need to address both I think.
Moving the matrix out of the AMG prconditioner is then the right choice I think. It could simply 'ask' the operator for it.
Before I give this a spin in the throughput tests for the 'hyperelasticity paper', could you @kronbichler, confirm my thoughts here? 
Thanks a ton you two! :+1:

nfehn:
The new commits remove all ifdefs from the implementation of the multigrid coarse grid solvers. Only `PreconditionerAMG` is used there now, while all the ifdefs now appear within a single class `PreconditionerAMG`.

nfehn:
Similar as we have one class `PreconditionerAMG` for different Linear Algebra implementations, we should aim at a unified sparse matrix class in ExaDG so that we can get rid of many ifdefs in different classes/files of ExaDG. I think this will make future extensions much simpler. Changes have to be made only in a single file and are then available for different use cases in ExaDG. (Let me summarize here that the present PR required more changes than expected in the sense that we "only use functionality" (sparse matrices) that was "already there" for another purpose (AMG coarse grid solver) ... also because we had "code duplications" at many places, which then results in a lot of effort when making changes or extending the functionality without duplicating certain code snippets again.)

nfehn:
@richardschu thanks for your immediate response to my issues raised above! I think for the present PR we should not resolve all the "limitations" I mentioned above but rather aim to merge the state we have now. 

Let me mention here once more that: If we use the code of this PR to run geometric multigrid with a matrix-based implementation, the performance results are not fully optimized in the sense that the fine-level matrix is assembled/stored twice. In addition to that, the matrices used in multigrid are in double-precision even if our geometric multigrid preconditioner is of number type `Number = {float, double}`.

include/exadg/operators/operator_base.cpp
kronbichler: Yes, in fact the conversion should not be necessary for PETSc, as we do an additional copy operation in the `petsc_operation` function.

include/exadg/operators/operator_base.cpp
kronbichler: I agree, for matrix-based `vmult` one should not need a `operator=(0.0)` because the matrix-based solver will overwrite the content in their implementations (as far as I know). The matrix-free code is special in the sense of ghost entries, where we had bugs in the past (hopefully fixed today), but we don't expose the ghost range to PETSc/Trilinos, so I think this should be fine.

kronbichler (approved):

kronbichler:
Regarding the layout, I also agree that moving the matrix out of the AMG preconditioner would be nice. It does not have to happen immediately, because the performance of the iterative solver should not/only marginally be affected, but of course skipping the extra assembly work would be nice to have in the future. Regarding two matrices in the fine-level operator, I would not be concerned if this remains a limitation (unless we feel it as a bottleneck), because I would expect to have more refactoring work to be done.

richardschu:
> Let me mention here once more that: If we use the code of this PR to run geometric multigrid with a matrix-based implementation, the performance results are not fully optimized in the sense that the fine-level matrix is assembled/stored twice. In addition to that, the matrices used in multigrid are in double-precision even if our geometric multigrid preconditioner is of number type `Number = {float, double}`.

Thanks for pointing that out again; so I will do the throughput tests for the matrix-based GMG only in double precision, and compare to a matrix-free GMG running in only double precision as well.

include/exadg/solvers_and_preconditioners/utilities/linear_algebra_utilities.h
nfehn: done

include/exadg/solvers_and_preconditioners/utilities/linear_algebra_utilities.h
```diff
@@ -133,6 +128,40 @@ apply_petsc_operation(
...
+    src_double.reinit(src, true);  // do not zero entries
+    src_double = src;
```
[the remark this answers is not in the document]
nfehn: done

nfehn:
I hope I addressed everything now. I will wait for the checks to complete and then merge. I already created new issues that point to current limitations of our implementation.

include/exadg/solvers_and_preconditioners/multigrid/coarse_grid_solvers.h
```diff
@@ -197,98 +163,19 @@ class MGCoarseKrylov : public CoarseGridSolverBase<Operator>
...
-          std::dynamic_pointer_cast<PreconditionerML<Operator, NumberAMG>>(preconditioner_amg);
-
-        dealii::ReductionControl solver_control(additional_data.solver_data.max_iter,
-                                                additional_data.solver_data.abs_tol,
-                                                additional_data.solver_data.rel_tol);
-
-        if(additional_data.solver_type == KrylovSolverType::CG)
-        {
-          dealii::SolverCG<VectorTypeAMG> solver(solver_control);
-          solver.solve(coarse_operator->system_matrix, dst_tri, src_tri, *preconditioner_amg);
```
richardschu, line 220: here we use the `coarse_operator->system_matrix`.

nfehn: `coarse_operator` in the old code is probably a misleading name, since this object comes from a dynamic cast of the AMG preconditioner. So the system matrix of the AMG preconditioner is used for the vmult() in the Krylov solver, isn't it?

richardschu: that's what I meant, yes I think so too.

Issue #689, opened 2024-08-02
matrix-based operator.apply() lacks BCs for continuous Galerkin case


richardschu:
In #681, we introduced matrix-based vmult().
In the matrix-free [OperatorBase::apply(), line 275-282](https://github.com/exadg/exadg/blob/b60088eddddb04fe6ddfa7f54a9512dd0fa11e0b/include/exadg/operators/operator_base.cpp#L275-L282), constrained degrees of freedom are enforced on the `dst` vector after applying the `vmult()` for the CG case.
For the matrix-based version this is missing and turned out to cause problems in my tests. 
What I saw: matrix-based solver did not converge once I used any GMG. AMG on the fine grid was good, but even matrix-based GMG + coarse solver:Chebyshev+PointJacobi did not converge. Once I added the constraints as in the matrix-free case and CG case, everything worked as expected.
Why did we initially skip it? Was it because PR #681 started off thinking of throughput tests only?

nfehn:
> Why did we initially skip it? Was it because PR https://github.com/exadg/exadg/pull/681 started off thinking of throughput tests only?

I probably expected that things are "automatically" done correctly in the matrix-based case without the need for "manual manipulation". Another reason could be that I have not seen in the code an explicit treatment of BCs for other use cases of matrices in ExaDG such as AMG preconditioners. Do we do something special there? Why or why not?

PR #690, opened 2024-08-02
constraints in OperatorBase::apply_matrix_based() for CG elements

richardschu:
fixes #681 
We enforce the constraints on the `dst` vector in the CG case as in the matrix-free version of `OperatorBase::apply()`.
On the fine level, my tests showed expected results (AMG solver), which means we enforce the constraints on the fine level at another point.

### DIFF
--- a/include/exadg/operators/operator_base.cpp
+++ b/include/exadg/operators/operator_base.cpp
@@ -450,6 +450,19 @@ OperatorBase<dim, Number, n_components>::apply_matrix_based(VectorType &       d
   {
     AssertThrow(false, dealii::ExcMessage("not implemented."));
   }
+
+  if(not is_dg)
+  {
+    // Constrained degree of freedom are not removed from the system of equations.
+    // Instead, we set the diagonal entries of the matrix to 1 for these constrained
+    // degrees of freedom. This means that we simply copy the constrained values to the
+    // dst vector.
+    for(unsigned int const constrained_index :
+        matrix_free->get_constrained_dofs(this->data.dof_index))
+    {
+      dst.local_element(constrained_index) = src.local_element(constrained_index);
+    }
+  }
 }
 
 template<int dim, typename Number, int n_components>


kronbichler (commented):
This looks ok in principle; however I wonder why the matrix does not have dummy diagonal entries in the constrained dofs. I do not have the time to look at the code that fills the matrix, but I was curious of why this was the case. I note that if the matrix adds some dummy entries via `AffineConstraints::distribute_local_to_global` (we should check which version exactly we're using, and if it is a version that adds a diagonal to the matrix or not in the deal.II documentation), then the magnitude of those entries will be different from 1, which could cause the preconditioner/matrix interaction done in the present way to result in ill-conditioning.

richardschu:
Thank you for your comment @kronbichler
The line in ExaDG is [this one](https://github.com/exadg/exadg/blob/b60088eddddb04fe6ddfa7f54a9512dd0fa11e0b/include/exadg/operators/operator_base.cpp#L2029):
`constraint_double.distribute_local_to_global(matrices[v], dof_indices, dst);`

The deal.II docu of that function says:
"With this scheme, we never write into rows or columns of constrained degrees of freedom. [...] Thus, if a degree of freedom in `local_dof_indices` is constrained, we distribute the corresponding entries in the matrix, but also add the absolute value of the diagonal entry of the local matrix to the corresponding entry in the global matrix."
 see here [affine_constraints.h](https://github.com/dealii/dealii/blob/e36c901b66ec82824ff80ef0c1f79972d65d5089/include/deal.II/lac/affine_constraints.h#L1579-L1614)

So we have dummy entries on the diagonal.

The `src` is the same for matrix-free and matrix-based implementations.

For the matrix-free version we set constrained entries after `apply()` to correct the corresponding entries.

So replacing the matvec using actual matrices, we need to "recover" correct values in the constrained DoFs in the very same manner, don't we?

kronbichler:
> The src is the same for matrix-free and matrix-based implementations.
>
> For the matrix-free version we set constrained entries after apply() to correct the corresponding entries.
>
> So replacing the matvec using actual matrices, we need to "recover" correct values in the constrained DoFs in the very same manner, don't we?

The matrix-free version does not touch constrained entries at all, so the added values in the matrix-free case are necessary to make the operator positive definite (assuming that the rest of the matrix is). I do not particularly like this setup and have been thinking about creating a code path that removes those entries altogether when solving linear systems, see https://github.com/dealii/dealii/issues/15224 for this reason and for the reduced memory transfer (we should probably tackle this in the near future).

But for the matrix-based setup, things are different in the sense that the matrix already provides this "singularity-avoiding" mechanism via `AffineConstraints`, so any preconditioner that extracts the information from the matrix will be incompatible with our `vmult` function. However, what I believe happened in your case was that in the GMG case you describe in #689 the diagonal was computed assuming `1` on the diagonal (instead of the actual values stored in the matrix). As a result, you got a really bad spectrum in the matrix because the preconditioned system now has far-off entries for the actual diagonal compared to the one in the preconditioner. I think we should set the diagonal to the sparse matrix to 1 as well, after calling the assembly function, to make sure all preconditioners in ExaDG remain valid. I would use code like this right after we call `system_matrix.compress(VectorOperation::add);`:
```
  for (auto const &line : constraint_double)
    if (dof_handler.locally_owned_dofs().is_element(line.index))
      system_matrix.set(line.index, line.index, 1.0);
  system_matrix.compress(dealii::VectorOperation::insert);
```
I would not manually set the DoFs again, but at least that way we could be consistent. Could you please try if that works?

peterrum:
> I would not manually set the DoFs again, but at least that way we could be consistent. Could you please try if that works?

That should work. I have use the same trick in an other code.

However, I was wondering if we should not change `MatrixFreeTools::comute_matrix()` so that it "directly" puts ones on the diagonal entries of constrained indices? The typical workflow of vmults in the context of matrix-free application is that the diagonal entries of constrained indices are 0 or 1. Having other values is not that easy to accomplish. I would be fine to make the change in particular because it is the `MatrixFreeTools` namespace so functions there should be somehow consistent to the matrix-free operator. @kronbichler What do you think?

nfehn:
If I understand this discussion correctly, the current suggestion fixes the action of the mat-vec product, but preconditioners would be wrong or inconsistent? So if I get it right, we should not merge the current version? For this reason, let me make this PR a draft for now, but please correct me if I misunderstood the discussion.

kronbichler:
> I would be fine to make the change in particular because it is the MatrixFreeTools namespace so functions there should be somehow consistent to the matrix-free operator. @kronbichler What do you think?

Here you mean that you would introduce the code I stated here (and you used before) in `MatrixFreeTools::compute_matrix()` rather than doing it afterwards? I think that could be a good option, even though it would need to be documented, since the behavior is different from the classical sparse matrix assembly via `AffineConstraints` of deal.II. Whenever one uses a preconditioner that includes the diagonal, that would be fine, but there could be cases where it is not. On the other hand, it is helpful to have 1 on the diagonal.

> So if I get it right, we should not merge the current version?

Indeed, we should fix the entries actually placed on the matrix diagonal, and check if that verifies the convergence issues. If yes, I would suggest to remove the modification of the `vmult` function, because set the entries twice can be confusing.

nfehn:
@richardschu It would be great if you could provide an update on the current status here or how we plan to address this.

richardschu:
1) merged w. recent ExaDG master.
2) replaced setting the vector entries with setting the diagonal entries in the matrix-based case.
3) With the recent master (the `structure/bar` applciation serves as testbed), I can better test the differences matrix-based/matrix-free; I am sure it is only a visualization issue. **This is still open.** (I will push once I have a complete solution for this)
4) Also @nfehn: I was too stupid to set the refinement in the planned(?) way by e.g. `vector_local_refinements={1}`, can you tell me how that would work? (I was setting `colorized=true`, overwrote the bdry_ids to match the testcase definitions, but it would only segfault?)

include/exadg/structure/spatial_discretization/operator.cpp
```diff
@@ -961,6 +961,8 @@ Operator<dim, Number>::solve_nonlinear(VectorType &       sol,
 
   //  elasticity_operator_nonlinear.set_inhomogeneous_boundary_values(sol);
 
+  affine_constraints_periodicity_and_hanging_nodes.distribute(sol); // <- these should be always enforced through matrix_free!?
```
richardschu, line 964: this should not be necessary, right? @nfehn 
When adding it, the visualization is correct *and* the iteration count decreses from 34.33 on average to 33.0.
So I think there is some inconsistency regarding the hangin gnode constraints, could that be?

richardschu: (these numbers are from the matrix-free case that behaves exactly the same as the matrix-based cased iteration-wise, when the diagonal entries in matrix-based are set to 1.0)

kronbichler: The fact that the iteration counts change is strange, if this happens I believe we should look over the code again. The point is that `MatrixFree` should take care of the constraints whenever it needs it. Of course, we have the topic of how to apply the inhomogeneous/Dirichlet constraints discussed elsewhere, that needs a proper resolution to create a working code. But even that should not affect the iterations.

The `distribute` function is necessary when going to other functions that do not read out the solution vector via `FEEvaluation`, i.e., visualization/IO, specific functions, that do not resolve the constraints on the fly.

richardschu: Update: for the linear problems (unsteady + steady), the iterations are not changing (up to 2 decimal places) over 100 time steps for that problem:
image
So it is a visualization problem for the linear case. For the nonlinear case, the linearization vector might be wrong (with the zero hanging node constraints enforced).

richardschu: I think we should at least keep the vector in a consistent state regarding constraints after each timestep. Ideally, the constraints would be in the vector directly after solving the system. I do not understand why this is not the case already ... even for linear problems.

richardschu:
also, I think we might have a similar problem for the Poisson solver.
and what about CG discretizations within e.g. cph multigrid? -- if the operator is nonlinear, we might have the same issues (slightly increasing iteration counts and we wouldn't even know).

include/exadg/structure/spatial_discretization/operator.cpp
richardschu: just as a reminder:
for a linear steady problem, commenting line
https://github.com/exadg/exadg/blob/6c8d7ca09cdf81dbfbd1f79980d38182e89225a2/include/exadg/structure/spatial_discretization/operator.cpp#L1010
gives:
image

where clearly, the boundary conditions are enforced (since that is the only thing nonzero), but the nodal values are zero constrained in the vector.

richardschu:
@kronbichler do we prefer some version of
`    for(unsigned int const constrained_index :`
`        matrix_free->get_constrained_dofs(this->data.dof_index))`
`      dst.local_element(constrained_index) = src.local_element(constrained_index);`
over the 
`constraints.distribute(vec)`
I have now?
Note that we have the `affine_constraints_periodicity_and_hanging_nodes`
and a respective `dof_index_inhomogeneous` in `matrix_free`, which we might use in the snippet above.

richardschu:
I am looking at this yet again, kind of a "Frühjahrsputz" trying to resolve some open issues/PRs and this one does not consider a new feature, but actually a bugfix.
I can also deliver an adapted Poisson case to show the improvement also in that module.

richardschu:
this is the grep of the relevant functions:
image
I think we simply forgot to treat the hanging node and periodicity constraints within `operator->set_inhomogeneous_boundary_values()`.
I think renaming `OperatorBase::set_inhomogeneous_boundary_values()` and the derived functions to `OperatorBase::set_inhomogeneous_constrained_values()` would make sense.
If we set the constrained DoFs to zero, solve and then apply them again, I think it actually makes sense, that the hanging node (and periodicity) constraints are to be enforced also here again, since they are of the form `x_1 = .5 * (x_2 + x_3)` and not `x_1 = 0.0`. The we we actually achieve is still to be discussed, I will push here.

include/exadg/poisson/spatial_discretization/laplace_operator.cpp
```diff
@@ -399,89 +399,99 @@ LaplaceOperator<dim, Number, n_components>::do_boundary_integral_continuous(
...
-                                                     boundary_values,
-                                                     mask);
-  }
-
-  // set Dirichlet values in solution vector
-  for(auto m : boundary_values)
-    if(dst.get_partitioner()->in_local_range(m.first))
-      dst[m.first] = m.second;
+  // periodicity and hanging node constraints
+  unsigned int const dof_index_inhomogeneous = this->get_dof_index_inhomogeneous();
```
richardschu, line 407: these 4 lines are new, the rest is just the old code moved behind an `if`-statement.
This `if`statemanet is introduced to allow for updating only the hanging node values.
Every time when solving a system, the constraints at the hanging nodes are set to zero.
This is ok for Dirichlet constraints in the nonlinear solver.
The current version fixes the incorrect behavior we had for hanging nodes before.
Open:
i) why are zero constraints on hanging nodes (and maybe periodic constraints) enforced via MatrixFree? (probably stemming from the DG approach being standard)
ii) can we swap, if needed, the `constraints.distribute()` for something more efficient?

include/exadg/operators/operator_base.cpp
```diff
@@ -1001,6 +1011,20 @@ OperatorBase<dim, Number, n_components>::internal_calculate_system_matrix(
...
+  dealii::DoFHandler<dim> const & dof_handler =
+    this->matrix_free->get_dof_handler(this->data.dof_index);
+  for(auto const & line : this->constraint->get_lines())
+  {
+    if(dof_handler.locally_owned_dofs().is_element(line.index))
+    {
+      std::cout << "system_matrix.el(i,i) = " << system_matrix.el(line.index, line.index)
+                << std::endl;
+      system_matrix.set(line.index, line.index, 1.0);
+    }
```
richardschu, line 1025: this `std::cout` will ofc be removed.
yields something like:
image
so we do write entries on the diagonal, but they are not 1.0.
1.0 is what we expect in other parts of the code afaik.

richardschu (commented):
added optional adaptive refinement for the structure/bar and poisson/gaussian applications to test the hanging nodes infrstructure. We could also come up with a test for this.

include/exadg/poisson/spatial_discretization/operator.cpp
```diff
@@ -501,7 +501,8 @@ Operator<dim, n_components, Number>::solve(VectorType &       sol,
   if(param.spatial_discretization == SpatialDiscretization::CG)
   {
     laplace_operator.set_time(time);
-    laplace_operator.set_inhomogeneous_boundary_values(sol);
+    laplace_operator.set_inhomogeneous_constrained_values(
+      sol, false /* periodicity_and_hanging_node_constraints_only */);
   }
```
richardschu, line 507: I think from this code here it is clear that we enforce constraints again because they are zero otherwise in the CG case. I think this is rooted in the DG case being the default in ExaDG and we try to be compatible with that using a CG approach. Hence, I think that considering for hanging node and periodicity constraints here as well is not too far off.

include/exadg/operators/operator_base.cpp
richardschu: [obviously this is why tests fail]

include/exadg/poisson/spatial_discretization/laplace_operator.cpp
kronbichler: I am not entirely sure I got the direction of the question right, but the reason why those are handled via `MatrixFree` is because it is more efficient to do so there than to do `constraints.distribute`. In fact, with the algorithms implemented in deal.II over the years hanging node constraints are relatively cheap, and it would be essentially impossible for a separate object like `constraints.distribute()` to reach this performance. For periodicity, we do the natural thing and read the actual dof values on the side which carries the degree of freedom already. (I would go the opposite way and say that the deal.II design for periodicity is not really straight-forward and it is a pity that we need to carry separate degrees of freedom on two sides; for the simple cases you never need them, only in more complicated periodicity settings.)

include/exadg/operators/operator_base.cpp
kronbichler: The code that resolves the constraints and write the entries is this one here:
https://github.com/exadg/exadg/blob/6d3cb9706eccaaa8a41f26e3890bdc8ced8bf5b0/include/exadg/operators/operator_base.cpp#L2024-L2029

This function uses the average size of the diagonal entry in the local matrix if I remember things correctly. In most cases, the entry on the diagonal does not matter, because every preconditioner derived from the matrix picks up the right order of magnitude (and it is the better range of eigenvalues in any case). It is also clear why we do things as we do. I would recommend to `set(i, i, 1.0)` for the constrained dofs `i` in this function, that should fix things.

richardschu:
for the record: talked to Martin and we figured we can replace
`constraints.distribute()`
with a matrix free loop reading (resolving constraints) and distributing to the vector if we see these calls lead to major performance penalties in the future.

richardschu:
just checked if the nonlinear structure solver works if we skip that additional call
```
elasticity_operator_nonlinear.set_inhomogeneous_constrained_values(
  sol, true /* periodicity_and_hanging_node_constraints_only */);
```
because it is the only spot where we use the second argument as `true`.
And it seems we do not have to use it at all:
image
So I am removing that argument again, pushing this much closer to the original state.
I will keep a commit though, in case we want to test that again in the future.

include/exadg/structure/spatial_discretization/operator.cpp
```diff
@@ -922,17 +923,6 @@ Operator<dim, Number>::solve_nonlinear(VectorType &       sol,
   // solve nonlinear problem
   auto const iter = newton_solver->solve(sol, update);
 
-  // TODO
```
richardschu: I removed this completely as it is commented, we looked at the constraints once again, and we tested matrix-free and matrix-based variants. All works.

include/exadg/structure/spatial_discretization/operators/mass_operator.h
```diff
@@ -57,8 +57,18 @@ class MassOperator : public ExaDG::MassOperator<dim, dim, Number>
   }
 
   void
-  set_inhomogeneous_boundary_values(VectorType & dst) const final
+  set_inhomogeneous_constrained_values(VectorType & dst) const final
   {
+    // periodicity and hanging node constraints are enforced strongly for continuous spaces
+    if(not this->is_dg)
```
richardschu: only for CG, as in DG, we have face integrals enforcing continuity.

include/exadg/poisson/spatial_discretization/laplace_operator.cpp
```diff
@@ -399,9 +399,18 @@ LaplaceOperator<dim, Number, n_components>::do_boundary_integral_continuous(
 
 template<int dim, typename Number, int n_components>
 void
-LaplaceOperator<dim, Number, n_components>::set_inhomogeneous_boundary_values(
+LaplaceOperator<dim, Number, n_components>::set_inhomogeneous_constrained_values(
   VectorType & dst) const
 {
+  // periodicity and hanging node constraints are enforced strongly for continuous spaces
+  if(not this->is_dg)
```
richardschu: only for CG, as in DG, we have face integrals enforcing continuity.

PR #692, opened 2024-08-03
remove PreconditionerAMG::apply_krylov_solver_with_amg_preconditioner()

### LINKED ISSUE #691: coarse grid solver w. AMG uses system_matrix.vmult()
Even pre #681, we used the `coarse_operator.system_matrix` for the Krylov solver on the coarse level.
See [here](https://github.com/exadg/exadg/pull/681/files#r1702718208).

with #681, `precondition_amg.h` also includes [`PreconditionerAMG::apply_krylov_solver_with_amg_preconditioner()`](https://github.com/exadg/exadg/blob/b60088eddddb04fe6ddfa7f54a9512dd0fa11e0b/include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h#L393C3-L393C46) 
which is only used in [MGCoarseKrylov::operator()(dst,src)](https://github.com/exadg/exadg/blob/b60088eddddb04fe6ddfa7f54a9512dd0fa11e0b/include/exadg/solvers_and_preconditioners/multigrid/coarse_grid_solvers.h#L169)

With this setup, the type conversion coarsest level: double ---> double (in case of Trilinos::ML) ---> float
happens explicitly when we solve, once, using [`apply_function_in_double_precision()`](https://github.com/exadg/exadg/blob/b60088eddddb04fe6ddfa7f54a9512dd0fa11e0b/include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h#L418-L426).

However, we could also just use `underlying_opartor` directly as shown in #691.
This way, we would convert single--->double--->single implicitly all the time in each iteration (which are hopefully <30 ?).

So there could be performance loss.
However, I was having trouble using the current implementation using `preconditioner_amg.apply_krylov_solver_with_amg_preconditioner()` which just gave me NaN after one iteration, but *not* in the first application of the solver. This is still a mistery to me, especially since it does not happen on two other machines, only on Fritz.
I will only investigate this issue further, if we decide to keep `PreconditionerAMG::apply_krylov_solver_with_amg_preconditioner()`. Is it a-priori clear, that type conversion is better/worse than `system_matrix.vmult()` vs. the matrix-free `operator.vmult()`? What am I not seeing? #681 specifically introduced this functionality to use it only in a single place, so there is a reason for that, which I do not see yet.

Thanks for any comments in advance :+1:

richardschu:
fixes #691 
this is WIP just to show how easy that would be, but as I said, I am overlooking something here probably.

### DIFF
--- a/include/exadg/solvers_and_preconditioners/multigrid/coarse_grid_solvers.h
+++ b/include/exadg/solvers_and_preconditioners/multigrid/coarse_grid_solvers.h
@@ -161,78 +161,67 @@ public:
     if(additional_data.operator_is_singular)
       dealii::VectorTools::subtract_mean_value(r);
 
-    if(additional_data.preconditioner == MultigridCoarseGridPreconditioner::AMG)
-    {
-      std::shared_ptr<PreconditionerAMG<Operator, Number>> preconditioner_amg =
-        std::dynamic_pointer_cast<PreconditionerAMG<Operator, Number>>(preconditioner);
+    std::shared_ptr<Krylov::SolverBase<VectorType>> solver;
 
-      preconditioner_amg->apply_krylov_solver_with_amg_preconditioner(dst,
-                                                                      r,
-                                                                      additional_data.solver_type,
-                                                                      additional_data.solver_data);
-    }
-    else
+    if(additional_data.solver_type == MultigridCoarseGridSolver::CG)
     {
-      std::shared_ptr<Krylov::SolverBase<VectorType>> solver;
+      Krylov::SolverDataCG solver_data;
+      solver_data.max_iter             = additional_data.solver_data.max_iter;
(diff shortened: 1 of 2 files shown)


nfehn:
I think the old code runs the Krylov solver on the coarse grid in double precision for the ML or BoomerAMG preconditioners.

The code suggested here would run the Krylov solver in single precision, even though the matrices are represented in double precision. So the present PR changes the code in my opinion. Running the Krylov solver in single-precision could potentially lead to convergence problems.

In #681, I did not want to change the implementation or what the code does, but just do a refactoring, which is why I introduced a function like `apply_krylov_solver_with_amg_preconditioner()`, in order to have a clearer design with the implementation happening in the respective Trilinos/PETSc implementations of the AMG preconditioners.

PR #693, opened 2024-08-05
add near null space for PreconditionerML

richardschu:
~requires https://github.com/dealii/dealii/pull/17396~
requires https://github.com/dealii/dealii/pull/17477
[PR in deal.II incoming from Peter]
[that is why tests fail, and that is why I set this to draft]

we use 
`dealii::DoFTools::extract_level_elasticity_modes()`
and
`dealii::DoFTools::extract_elasticity_modes()`

TESTS:
previously, in the iliac bifurcation example done for the recent paper, I did not really see differences > 20%.
Now, I tested this again on the `structure/bar` example using 
p=1, 
5 uniform refinements (421443 DoFs), 
`QuasiStatic` solver, 
CG linear solver with AMG preconditioner and relative tolerance of 1e-3 :
iteration counts_________________: Default | ScalarLaplace | VectorLaplace | Elasticity
ILU 0, 1 sweep__________________: 277.57__| 280.14________| 65.57_________| 29.57
Chebyshev, 6 sweeps, 2 overlap_: 46.86___| 46.86_________| 11.43_________| 11.14
 
I was using the latter smoother settings for the iliac bifurcation and a time dependent problem there, so it has the same trend. BUT using a cheap smoother, we see *huge* differences. That is what I wanted to see!
(last commit [31de486](https://github.com/exadg/exadg/pull/693/commits/31de4867d1345a2c16eefe9053955a83b17a0cb0) removed solver settings to default in `structure/bar`)


include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
```diff
@@ -154,17 +274,201 @@ class PreconditionerML : public PreconditionerBase<double>
...
+      // Initialize with the Teuchos::ParameterList.
+      amg.initialize(system_matrix, parameter_list);
+    }
 
     this->update_needed = false;
   }
 
 private:
+  Teuchos::ParameterList
+  get_parameter_list()
```
nfehn, line 306: I think this function should be realized as a free function with `MLData` as a parameter. This function seems to be largely independent of the implementation of `PreconditionerML`, which should be reflected in the software design in my opinion.

Next, I would think about integraging `Teuchos::ParameterList` into `MLData`.

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
```diff
@@ -154,17 +274,201 @@ class PreconditionerML : public PreconditionerBase<double>
...
+    else
+    {
+      parameter_list.set("ML output", 0);
+    }
+
+    return parameter_list;
+  }
+
+  std::vector<VectorType>
+  get_constant_modes()
```
nfehn, line 374: I think we should also realize this function as free function so that the dependencies become clearer. I see mapping, on_coarse_grid, n_constant_modes, mpi-communicator and vector partitioner as input to this function. Hence, this function looks like a very general dealii utility function.

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
```diff
@@ -154,17 +274,201 @@ class PreconditionerML : public PreconditionerBase<double>
...
+                                       constant_modes[i],
+                                       dealii::ComponentMask(),
+                                       on_coarse_grid ? 0 : dealii::numbers::invalid_unsigned_int);
+    }
+
+    return constant_modes;
+  }
+
+  void
+  set_operator_nullspace(Teuchos::ParameterList &              parameter_list,
```
nfehn, line 419: I currently do not see dependencies of this function on member variables of the class `PreconditionerML`. This would also call for realizing this function as a free function.

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
```diff
@@ -319,22 +623,29 @@ class PreconditionerBoomerAMG : public PreconditionerBase<Number>
...
     (void)initialize;
+    (void)mapping;
     this->data = data;
 
     if(data.amg_type == AMGType::BoomerAMG)
     {
 #ifdef DEAL_II_WITH_PETSC
+      AssertThrow(data.amg_operator_type == AMGOperatorType::Default,
+                  dealii::ExcMessage("AMGType::BoomerAMG incorporates "
+                                     "only AMGOperatorType::Default."));
```
nfehn, line 648: This check implies that one has to change `amg_operator_type` in the input file when switching from `Trilinos` to `PETSc` for AMG. Principally, I prefer if one does not have to do this. Alternatives would be (i) to simply ignore this parameter for `PETSc` and document in the code that this parameter is only used for `Trilinos`, or (ii) to indicate by the parameter name that this parameter is for `Trilinos` only, e.g. `ConstantModeTypeTrilinos`.

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
```diff
@@ -154,17 +274,201 @@ class PreconditionerML : public PreconditionerBase<double>
...
+    {
+      // Default setup.
+      amg.initialize(system_matrix, ml_data);
+    }
+    else
+    {
+      // get Teuchos::ParameterList to provide custom near null space basis
+      Teuchos::ParameterList parameter_list = get_parameter_list();
+
+      std::vector<VectorType> constant_modes = get_constant_modes();
```
nfehn, line 287: Is there a reason why this function is called outside `set_operator_nullspace()`? Does the comment below "has to stay alive" also hold for this? I think the code would be easier to understand when integrating this into `set_operator_nullspace()`, because dependencies are not clear in the present code in my opinion.

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
```diff
@@ -2,7 +2,7 @@
  *
  *  ExaDG - High-Order Discontinuous Galerkin for the Exa-Scale
  *
- *  Copyright (C) 2021 by the ExaDG authors
+ *  Copyright (C) 2021 by the ExaDG authors and Marco Feder
```
nfehn, line 5: I think we should not introduce this style. If you want, you can add this name to the documentation of the particular function that was inspired by this developer. However, a link to deal.II code or some other publicly available repo would be better in my opinion.

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
```diff
@@ -154,17 +274,201 @@ class PreconditionerML : public PreconditionerBase<double>
...
+      n_constant_modes = dim;
+    }
+    else if(operator_type == AMGOperatorType::Elasticity)
+    {
+      n_constant_modes = dim == 1 ? 1 : (dim == 2 ? 3 : 6);
+    }
+    else
+    {
+      AssertThrow(false, dealii::ExcMessage("AMGOperatorType not defined"));
+    }
```
nfehn, line 392: Related to comment https://github.com/exadg/exadg/pull/693/files#r1705187688, this seems to be duplicated code. To avoid this, I would let the class `ConstantModes<dim>` take over the responsibillity for `n_constant_modes`.  `ConstantModes<dim>` receives the operator type in the constructur and has a member function `n_modes()` to perform a loop. When looping over all the modes to compute the vectors, one calls another member function `mode_generator.set_mode(i)`, similar to `dealii::Function::set_time()`.

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
```diff
@@ -154,17 +274,201 @@ class PreconditionerML : public PreconditionerBase<double>
...
+    // multigrid.
+    bool const on_coarse_grid = pde_operator.get_matrix_free().get_dof_handler().has_level_dofs();
+
+    for(unsigned int i = 0; i < n_constant_modes; ++i)
+    {
+      constant_modes[i].reinit(pde_operator.get_matrix_free().get_vector_partitioner(),
+                               system_matrix.get_mpi_communicator());
+
+      // Fill vector with rigid body modes.
+      ConstantModes<dim> const mode_generator(operator_type, i);
```
nfehn, line 406: What about defining this variable outside the for-loop according to comment https://github.com/exadg/exadg/pull/693/files#r1705195474? This should allow a design according to the "single-responsibility principle".

nfehn: Here in this line, one would then write `mode_generator.set_mode(i)`.

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
```diff
@@ -154,17 +274,201 @@ class PreconditionerML : public PreconditionerBase<double>
...
+    else if(operator_type == AMGOperatorType::Elasticity)
+    {
+      n_constant_modes = dim == 1 ? 1 : (dim == 2 ? 3 : 6);
+    }
+    else
+    {
+      AssertThrow(false, dealii::ExcMessage("AMGOperatorType not defined"));
+    }
+
+    std::vector<VectorType> constant_modes(n_constant_modes);
```
nfehn, line 394: for a single responsibility design, write `mode_generator.n_modes()` instead of computing `n_constant_modes` by the lines of code above.

nfehn:
I think this is a very important issue that you address here! Thanks!

The code is somewhat "verwoben" currently. I took me some time to understand the flow of information and which class/function **depends on** which parameters/objects. I think we need to improve this.

applications/structure/bar/application.h
```diff
@@ -319,6 +319,27 @@ class Application : public ApplicationBase<dim, Number>
...
+#ifdef DEAL_II_WITH_TRILINOS
+    this->param.multigrid_data.coarse_problem.amg_data.ml_data.elliptic = true;
+    bool const mg_involves_p = (this->param.multigrid_data.type == MultigridType::pMG or
+                                this->param.multigrid_data.type == MultigridType::hpMG or
+                                this->param.multigrid_data.type == MultigridType::phMG);
+    this->param.multigrid_data.coarse_problem.amg_data.ml_data.higher_order_elements =
+      (preconditioner == Preconditioner::AMG or
+       (preconditioner == Preconditioner::Multigrid and not mg_involves_p)) ?
+        this->param.degree > 1 :
+        false;
```
kronbichler, line 334: I suggest to remove this parameter. I did experiment with it in 2008, and with the state of ML at that time it was useful. However, improvements in ML around that time made the parameter obsolete, and the parameter just sets the parameter "aggregation: type" to "Uncoupled" which is at that value anyway: https://github.com/dealii/dealii/blob/14276d45ac74825eb1a3116f73bda03680855465/source/lac/trilinos_precondition_ml.cc#L75-L84

include/exadg/solvers_and_preconditioners/multigrid/coarse_grid_solvers.h
```diff
@@ -58,7 +58,7 @@ class CoarseGridSolverBase
   update() = 0;
 };
 
-template<typename Operator>
+template<int dim, typename Operator>
```
kronbichler, line 61: Is the additional template parameter necessary? I would prefer if it could be deduced from `Operator` - this also relates to your question of how to deduce the mapping type. Does it work to call `pde_operator.get_matrix_free().get_mapping_info().mapping` (which returns a pointer to a mapping). Alternatively, I would prefer to deduce the template parameter `dim` from `Operator::dimension` (to be introduced as a `static constexpr` in field as I believe we don't currently do it) or by templating the respective constructors on `dim` to be able to accept a mapping (I also did not check this). My main point is that `Operator` already contains a template on `dim`, so it feels it should give us the respective information.

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
kronbichler: I agree. Also, the code has evolved over many places if I recall the history. There are several pieces originally from deal.II, that were later extended to manual null spaces in 2015/16 in https://github.com/kronbichler/adaflo/blob/6329eeec998ce7f4d527f4fed96b9043e9ca2106/source/navier_stokes_preconditioner.cc#L474-L532 and additional evolution towards structural problems by different people. But I would definitely support listing contributions by Marco in the authors page.

include/exadg/solvers_and_preconditioners/multigrid/coarse_grid_solvers.h
nfehn: I agree. I did not mention this so far because having done the refactoring suggested in other comments, we might want to re-discuss which objects should be passed to the constructor of `PreconditionerML` and which should actually be initialized outside and then handed over to `PreconditionerML`. So my question is: what are the implications of having the data structures that require `dim` actually outside the class `PreconditionerML`?

nfehn:
I have a general question: For a particular PDE model problem (say hyperelasticity, module `Structure`), do we want/need the freedom to select from scalar/vector Laplace vs. Elasticity? An alternative perspective would be to invoke a call like `pde_operator->get_constant_modes_amg()` (similar to e.g. `pde_operator->is_singular()`), i.e. only the PDE operator can and should provide information about constant modes. This design might naturally avoid the problem with the mapping, since a PDE operator is close to data structures such as mapping.

richardschu:
Thanks all for your comments! I will get back to this asap :+1: 
Depending on how fast @fdrmrc and @peterrum can do that, we ofc happily integrate what you come up with -- we are meaking heavy use of that micture of codes already :smile: 
I will re-design and come back at you with a new proposal!

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
```diff
@@ -154,17 +274,201 @@ class PreconditionerML : public PreconditionerBase<double>
...
+
+    // The AMG Preconditioner might be used as a coarse-level solver/preconditioner within
+    // multigrid.
+    bool const on_coarse_grid = pde_operator.get_matrix_free().get_dof_handler().has_level_dofs();
+
+    for(unsigned int i = 0; i < n_constant_modes; ++i)
```
richardschu, line 400: NOTE TO @richardschu :
this cannot work if the `dof_index` is not 0 (e.g. pressure in Navier-Stokes).
You need something along the lines of   
`

      // Fill vector with rigid body modes.
      unsigned int const dof_index = pde_operator.get_dof_index();
	  unsigned int const n_components = pde_operator.get_matrix_free().get_dof_handler(dof_index).get_fe().n_components();
	  dealii::ComponentMask component_mask(n_components, true);
	  if(n_components == 1)
	  {
		  ConstantModes<dim, 1> const mode_generator(operator_type, i);
		  dealii::VectorTools::interpolate(mapping,
										   pde_operator.get_matrix_free().get_dof_handler(dof_index),
										   mode_generator,
										   constant_modes[i],
										   component_mask,
										   operating_on_fine_else_coarsest_level ?
											 dealii::numbers::invalid_unsigned_int :
											 0);
	  }
	  else if(n_components == dim)
	  {
		  ConstantModes<dim, dim> const mode_generator(operator_type, i);
		  dealii::VectorTools::interpolate(mapping,
										   pde_operator.get_matrix_free().get_dof_handler(dof_index),
										   mode_generator,
										   constant_modes[i],
										   component_mask,
										   operating_on_fine_else_coarsest_level ?
											 dealii::numbers::invalid_unsigned_int :
											 0);
	  }
`

applications/structure/bar/application.h
```diff
@@ -319,6 +319,27 @@ class Application : public ApplicationBase<dim, Number>
     this->param.preconditioner      = preconditioner;
     this->param.multigrid_data.type = MultigridType::phMG;
 
+    this->param.multigrid_data.coarse_problem.amg_data.amg_operator_type =
```
richardschu, line 322: for reference :) @peterrum

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
richardschu: Since the stuff he and Peter provided is now in that unknown dealii branch that has DoFTools::extract_level_elasticity_modes() and  DoFTools::extract_elasticity_modes(), I think his contributions are then to be valued by deal.II, does that make sense?

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
```diff
@@ -79,8 +82,121 @@ create_subcommunicator(dealii::DoFHandler<dim, spacedim> const & dof_handler)
...
+    else
+    {
+      AssertThrow(mode_index < 6,
+                  dealii::ExcMessage("AMGOperatorType::Elasticity has six constant modes in 3D."));
+    }
+  }
+  else
+  {
+    AssertThrow(false, dealii::ExcMessage("AMGOperatorType not defined."));
+  }
```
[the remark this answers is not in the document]
richardschu: this is gone completely anyways, I will push as soon as I have the branch name from Peter.

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
richardschu: ok. I moved `Teuchos::ParameterList` to AMG_Data as I have "no" control over ML_Data in deal.II.

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
richardschu: I will remove the assert and adjust the name.

include/exadg/solvers_and_preconditioners/multigrid/coarse_grid_solvers.h
richardschu: If i can use the `pde_operator.get_matrix_free().get_mapping_info()mapping` as Martin suggests, then we do not need the `dim` since I only need to `get_ML_parameter_list(ml_data, dim)`.
Let's see, I will ask Peter for the name of the branch to compile this and then push after testing.

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
richardschu: its not needed anymore, thanks for your effort! :)

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
richardschu: its not needed anymore, thanks for your effort! :)

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
richardschu: No I just thought that its easier to see what happens if I block the code into some parts that do only a limited thing instead of having 1 blob of code.
Now it is outside since we use deal.II functionality :)

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
richardschu: its not needed anymore, thanks for your effort! :)

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
richardschu: its not needed anymore, thanks for your effort! :)

applications/structure/bar/application.h
richardschu: I just removed it for now. `aggregation: type` is always uncoupled by default anyways? For `elliptic` and `not higher_order`, we set it to decoupled, but I read the comment in the deal.II part you sent, as if it was the default anyways? And that part just remains as kind of a reminder? I just set the parameter to not make anything wrong, you know. I have no clue about the ML intrinsics/settings etc. I just know, that it had barely any or even no effect setting this parameter.

kronbichler: Yes, this is what I read from https://ftp.mcs.anl.gov/pub/pdetools/nightlylogs/xsdk/xsdk-configuration-tester/packages/trilinos/packages/ml/doc/mlguide.pdf (section 6.4.5, page 23) as being the default anyway. It is there as a reminder indeed. This is why I suggest removing it.

richardschu:
much better now :100: thanks @peterrum 
requires [peterrum/dealii/tree/extract_elasticity_modes](https://github.com/peterrum/dealii/tree/
so lets wait a bit, but the other troubles are gone.
Thanks everyone for your help :+1:

peterrum:
> much better now 💯 thanks @peterrum

https://github.com/dealii/dealii/pull/17477

richardschu:
requires #695 
(in the sense that tests fail)

richardschu:
\rebuild

nfehn:
I agree that the code looks now very clean in ExaDG.

I could not find an answer to the question from this comment: https://github.com/exadg/exadg/pull/693#issuecomment-2271107815

Do we want to set th parameter `ml_operator_type` in the application-file, or could we set it internally (in preconditioner classes in ExaDG for a particular module like `Structure`)? The latter suggestion would have the advantage that a user can not forget to set this parameter in an application-file.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_parameters.h
```diff
@@ -104,6 +110,7 @@ struct AMGData
     ml_data.smoother_sweeps = 1;
     ml_data.n_cycles        = 1;
     ml_data.smoother_type   = "ILU";
+    ml_operator_type        = MLOperatorType::Laplace;
```
nfehn, line 113: Related to comment https://github.com/exadg/exadg/pull/693#issuecomment-2315343733, it is in my opinion hard to understand for a user what the code actually does internally. Do I have to set this parameter in an application-file or not? The declaration of the type `MLOperatorType` does currently not provide documentation. My opinion is currently that this is an internal parameter that ExaDG can deduce from the PDE model problem (Poisson, Structure) that is considered.

If not realized as an internal parameter (but as a user parameter as done currently), we typically use the enum value `Undefined` so that the code is more transparent to the user. When not setting the parameter and using the default value `Undefined`, the user gets an error message stating that the parameter has not been defined so far, ideally only for those cases  where this parameter is going to be used by the code. I like this style because this means for a user that one does not have to study all the internal implementations in ExaDG in order to be sure that the code really does what one expects it to do. I understand that the present implementation requires minimal changes to the code and in the sense that a user only has to set this parameter for Elasticity problems and do/change nothing for e.g. Poisson. At the same time I have concerns whether we really benefit from this style because of the implicit knowledge/information involved.

richardschu:
> I have a general question: For a particular PDE model problem (say hyperelasticity, module `Structure`), do we want/need the freedom to select from scalar/vector Laplace vs. Elasticity? An alternative perspective would be to invoke a call like `pde_operator->get_constant_modes_amg()` (similar to e.g. `pde_operator->is_singular()`), i.e. only the PDE operator can and should provide information about constant modes. This design might naturally avoid the problem with the mapping, since a PDE operator is close to data structures such as mapping.

I do not think so. For testing purposes that might be interesting to some users, but then those can change that in their code versions. I would rather say use it as internal variable!

include/exadg/solvers_and_preconditioners/multigrid/multigrid_parameters.h
richardschu: I would like to go for: introduce `Undefined` to print it to screen (we can also skip this, I have no preference, but for safety in the tests it would be good to make sure it is set correctly).
And then later set it during setup.

richardschu: I would rather set it to `Laplace` by default (internal variable), where `Elasticity` is then the exception.

richardschu:
Ok so a few comments on the new version:
1) keep in parameters, actually allowing the user to modify, but set default to `MLOperatorType:.Elasticity` in `Structure::ApplicaitonBase::setup()`.
2) default is `MLOperatorType::Laplace`, no annoying change of default for other operators, this is also the default for ML and Boomer AMG anyways
3) I couldnt `std::is_same<Operator,ExaDG::Structure::LinearOperator<dim,Number>::value` in PreconditionerAMG (which would be my favoured option tbh), since there is also `MultigridOperatorBase` passed in as `Operator` ... so the info on the type of operator is lost. I could introduce virtual `OperatorBase::is_elasticity_problem()` returning `false` by default and overwrite in derived (Elasticity) classes to return `true`; mirror this is in `MultigridOperatorBase` ... which is a bit much for just getting a bool for the TrilinosML AMG preconditioner. Hence the current solution.

include/exadg/structure/user_interface/application_base.h
```diff
@@ -70,12 +70,14 @@ class ApplicationBase
   {
     parse_parameters();
 
-    // parameters
-    set_parameters();
 #ifdef DEAL_II_WITH_TRILINOS
-    // Overwrite default MLOperatorType::Laplace to include rotational rigid body modes.
+    // Overwrite default MLOperatorType::Laplace to include rotational rigid body modes;
```
richardschu, line 74: sets default for structure applications to `Elasticity`, but later the user can overwrite

applications/structure/bar/application.h
```diff
@@ -319,9 +319,6 @@ class Application : public ApplicationBase<dim, Number>
     this->param.preconditioner      = preconditioner;
     this->param.multigrid_data.type = MultigridType::phMG;
 
-    this->param.multigrid_data.coarse_problem.amg_data.ml_operator_type =
```
richardschu: not required in application

include/exadg/structure/driver.h
```diff
@@ -97,9 +97,6 @@ class Driver
 
   std::shared_ptr<MultigridMappings<dim, Number>> multigrid_mappings;
 
-  // user parameters
```
richardschu, line 100: why was this actually in there? just something someone forgot to delete? Probably ownership of parameters went from the driver to the application?

include/exadg/structure/user_interface/application_base.h
```diff
@@ -72,6 +72,10 @@ class ApplicationBase
 
     // parameters
     set_parameters();
+#ifdef DEAL_II_WITH_TRILINOS
```
richardschu, line 73: moved this in the next commit

include/exadg/structure/user_interface/application_base.h
nfehn: In my opinion this is the same style as before ("implicitly" defined default value with adjustments in only some modules; is e.g. the FSI module correct now?). The above comment "overwrite default MLOperatorType::Laplace" is not safe, one can easily change the default-value in ExaDG (which happens at a completely different place and affects all the modules) and this comment will be falsified (and also the code no longer works as expected). I also think we should not start with such problem specific adjustments and ifdefs related to Trilinos on such a high level (application class). Note that the structure solver is used in other modules such as FSI: can a user understand (for the present code) what will happen there? Hence, I think we need to set this parameter at a more central place in `Structure::Operator::setup_preconditioner()` for example, see https://github.com/exadg/exadg/blob/711cb015f0a56047dcd4c2f91a9b3d349dbcdd6e/include/exadg/structure/spatial_discretization/operator.cpp#L323 (To be on the safe side, the respective preconditioner implementations could add asserts such that the parameter value received by the preconditioner class agrees with the expected value for a certain PDE model problem.) Then, there is no need to print this parameter to the screen output, because it is an internal parameter.

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.h
```diff
@@ -210,6 +210,12 @@ class MultigridPreconditionerBase : public PreconditionerBase<Number>
   unsigned int
   get_number_of_levels() const;
 
+  /**
+   * Returns the number of h-levels.
+   */
+  unsigned int
+  get_number_of_h_levels() const;
+
```
nfehn, line 218: This function seems to be protected. Did you check whether derived classes make use of this functionality (and then could also call this member function) or - otherwise - whether this function can be made private?

applications/poisson/gaussian/tests/gaussian.output
```diff
@@ -59,6 +59,7 @@ Solver:
       Smoother sweeps:                       1
       Number of cycles:                      1
       Smoother type:                         ILU
+      AMG operator type:                     Laplace
```
nfehn, line 62: If we consider `Laplace` the correct value for the `ExaDG::Poisson` module and `Elasticity` the correct value for the `ExaDG::Structure` module (developments happening actually in Trilinos/deal.II), there is no need print this parameter in my opinion. I rather tend to add asserts deep in the code in ExaDG making sure that the preconditioner implemenetations receive a certain, expected  value. (If one wants to change this parameter for experimentation, one would have to go to this place in ExaDG with the assert, comment the assert, and overwrite the value. I consider this justified right now because I consider the development of these preconditioners settings and optimal parameter values as developments happening outside ExaDG.)

richardschu: ok

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.h
richardschu: made private! To be clear: I introduced this function with this PR. We use it twice currently. Used it four times in previous versions of this PR.

include/exadg/structure/user_interface/application_base.h
richardschu: I will remove it and not print it. and set it in the suggested place.

include/exadg/structure/driver.h
richardschu: @nfehn did you see this?

richardschu:
@nfehn I could do the copy of `amg_data` also only `#ifdef Trilinos` (introduces more lines changed, and the data structures are not too heavy I think, so I thought it doesnt really matter. Note also that `PreconditionerAMG` copies the amg_data it gets - I did not introduce this)
Also, please help me think if this covers all the appearances of `PreconditionAMG`, I also covered the `MultigridPreconditioner` children. (I tested iNS, Structure, Poisson, FSI manually; only structure and Poisson with `Preconditioner::AMG`, others only with coarse problem solver/preconditioner AMG).

richardschu:
@nfehn I addressed the comments, could you have another look?
I had to copy the data for the preconditioner at some points since it is `const`, but we want to change it depending on the operator.

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
```diff
@@ -96,8 +162,11 @@ class PreconditionerML : public PreconditionerBase<double>
   dealii::TrilinosWrappers::PreconditionAMG amg;
 
 public:
-  PreconditionerML(Operator const & op, bool const initialize, MLData ml_data = MLData())
-    : pde_operator(op), ml_data(ml_data)
+  PreconditionerML(Operator const &     op,
+                   bool const           initialize,
+                   MLOperatorType const ml_operator_type,
```
nfehn, line 167: I think we should avoid copying the data and the related if-defs to set the operator type. Making `MLOperatorType` a constructor argument of `PreconditionerAMG` has the problem that in the multigrid preconditioner (and coarse grid solver) implementation we do not know the operator type, since this is generic code so far. So we either have to extend the interface of such generic multigrid classes, or instead query the required information from the PDE operator. Regarding the latter, different options are possible in my opinion:

i) pde_operator->get_pde_operator_type()
ii) pde_operator->get_rigid_body_modes()

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
```diff
@@ -154,8 +223,54 @@ class PreconditionerML : public PreconditionerBase<double>
...
+          pde_operator.get_matrix_free().get_dof_handler(),
+          dealii::ComponentMask(dimension, true));
+      }
+      else
+      {
+        rigid_body_modes = dealii::DoFTools::extract_rigid_body_modes(
+          *pde_operator.get_matrix_free().get_mapping_info().mapping,
+          pde_operator.get_matrix_free().get_dof_handler(),
+          dealii::ComponentMask(dimension, true));
+      }
```
nfehn, line 257: It looks like the natural place for this code would be the implementation of the Elasticity operator.

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
```diff
@@ -154,8 +223,54 @@ class PreconditionerML : public PreconditionerBase<double>
...
+    if(ml_operator_type == MLOperatorType::Laplace)
+    {
+      // Default setup; add near null space for Laplace operator.
+      amg.initialize(system_matrix, ml_data);
+    }
+    else if(ml_operator_type == MLOperatorType::Elasticity)
+    {
+      // get Teuchos::ParameterList to provide custom near null space basis
+      unsigned int const     dimension      = pde_operator.get_matrix_free().dimension;
+      Teuchos::ParameterList parameter_list = get_ML_parameter_list(ml_data, dimension);
```
nfehn, line 235: Is there a reason why the parameter list is defined here and not below in line 262/263? It doesn't look like there is any dependency with the code below that computes the rigid body modes?

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
```diff
@@ -154,8 +223,54 @@ class PreconditionerML : public PreconditionerBase<double>
     // re-calculate matrix
     pde_operator.calculate_system_matrix(system_matrix);
 
-    // initialize Trilinos' AMG
-    amg.initialize(system_matrix, ml_data);
+    if(ml_operator_type == MLOperatorType::Laplace)
+    {
+      // Default setup; add near null space for Laplace operator.
+      amg.initialize(system_matrix, ml_data);
```
nfehn, line 229: Is `dimension` (used in the other branch below) irrelevant for Laplace? (scalar vs. vector Laplace)
Is there a reason why we do not unify the call `amg.initialize()` and use the same code for all the variants of the operator type?

include/exadg/structure/spatial_discretization/operator.cpp
```diff
@@ -453,19 +453,20 @@ Operator<dim, Number>::setup_preconditioner()
   }
   else if(param.preconditioner == Preconditioner::AMG)
   {
+    AMGData amg_data = param.multigrid_data.coarse_problem.amg_data;
+#ifdef DEAL_II_WITH_TRILINOS
+    amg_data.ml_operator_type = MLOperatorType::Elasticity;
+#endif
```
nfehn, line 459: We have this code for elasticity multiple times, and multiple times for Laplace. My hope with the design described in other comments would be that we only have to implement this once for each operator.

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
nfehn: So I mean option ii) described here: https://github.com/exadg/exadg/pull/693#discussion_r1795030659

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
nfehn: I tend to prefer ii), but then it would be good if we do not need `get_pde_operator_type()` in addition. Currently, we do not compute rigid body modes for Laplace and use somewhat different code paths instead. So my question is here probably again why we have the two options of AMG-initialization with `ml_data` or `parameter_list`.

richardschu: I unified the call now for an operator by setting up a `Teuchos:.ParameterList` in any case. This might impact some performance results (I guess only mildly), since setting up the `PreconditionerML` through `ml_data` did not per default allow repartitioning afaik.

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
richardschu: we now have `OperatorBase::get_constant_modes()`, which creates default modes, and the derived classes then can overwrite this function as done in `ElasticityOperatorBase`.
The modes can have different types though, which is a bit of a bummer ...
there are `std::veector<std::vector<bool>> constant_modes` for cases when only identifying a component is enough (Poisson problems for example), while there are also
`std::vector<std:.vector<double>> constant_modes_values`, which is the more general case, describing modes comprised of non-trivial entries.
I think the former still exists for performance/storage reasons and backward compatibility. I decided to keep them both and the `constant_modes_values` is checked for values first, which mirrors the behavior within deal.II. On our side, we have to provide one of the two, otherwise we assert.

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
richardschu: there was no reason, I think I rewrote this part and forgot to move it

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
richardschu: It is in general not, now this question pops up in `OperatorBase::get_constant_modes()`:
Do we use a scalar-valued Poisson in `dim`-dimensional space the default, or the related `dim`-component problem? I went for the former, since it is exactly what we did before as it replicates the default from deal.II.
My experiments within the structure module showed, that it did not really matter if the operator is more complex, and results were actually slightly better using the scalar-valued version (so what we have selected now).

richardschu: Plus the default fits exactly the pressure Poisson problem we often solve, for the momentum or diffusion operator, we are quite wrong in any way ... but one could try to use the dim-component version. But since I did not test all the examples I went with replicating the default from before :)

include/exadg/structure/spatial_discretization/operator.cpp
richardschu: removed!

include/exadg/structure/spatial_discretization/operators/elasticity_operator_base.h
```diff
@@ -99,6 +99,39 @@ class ElasticityOperatorBase : public OperatorBase<dim, Number, dim>
   OperatorData<dim> const &
   get_data() const;
 
+  /*
+   * Provide near null space basis vectors used e.g. in AMG setup. deal.II allows providing constant
+   * modes as `std::vector<std::vector<double>> constant_modes_values` or
+   * `std::vector<std::vector<bool>>   constant_modes`,
+   * where `constant_modes_values` is currently checked first (within deal.II), and if not provided,
+   * `constant_modes` is used. Provide rigid body modes, override implementation in `OperatorBase`.
```
nfehn, line 107: I have only one minor comment here: I would skip the part of the documentation that is duplicated from `OperatorBase`. Things might change in dealii, so this way we might avoid that comments in ExaDG become outdated in the future.

nfehn:
The code looks clean in my opinion. Thanks for your efforts! I think we can then merge soon.

PR #695, opened 2024-08-23
Remove serial SolutionTransfer

richardschu:
Don't like daily mails regarding the `ExaDG` build failing? I have something for you ; )
We touch this once the `p::d::SolutionTransfer` and `Solutiontransfer` have been merged in deal.II I suggest.

### DIFF
--- a/include/exadg/operators/solution_transfer.h
+++ b/include/exadg/operators/solution_transfer.h
@@ -82,7 +82,8 @@ public:
     }
     else
     {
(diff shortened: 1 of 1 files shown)


peterrum (approved):
Fine for now. But all the serial specialization code can be removed now.

richardschu:
Ok then lets use this to repair the compile and I will come up with a more general solution!

richardschu:
Ok I also removed the serial code now as it is not needed anymore.
The wrapper class `ExadG::SolutionTransfer()` only adds an update of the ghost values on top of `dealii::p::d::SolutionsTransfer`.

PR #700, opened 2024-09-17
fix compile warnings

richardschu:
updated compiler ... and now finally fixed some compile warnings.

### DIFF
--- a/include/exadg/incompressible_navier_stokes/spatial_discretization/operator_coupled.cpp
+++ b/include/exadg/incompressible_navier_stokes/spatial_discretization/operator_coupled.cpp
@@ -1208,24 +1208,30 @@ OperatorCoupled<dim, Number>::apply_inverse_negative_laplace_operator(VectorType
   {
     // solve a linear system of equations for negative Laplace operator to given (relative)
     // tolerance using the PCG method
-    VectorType const * pointer_to_src = &src;
+    dst = 0.0;
     if(this->is_pressure_level_undefined())
     {
-      VectorType vector_zero_mean;
-      vector_zero_mean = src;
-
+      VectorType vector_zero_mean = src;
       if(laplace_operator->operator_is_singular())
       {
         dealii::VectorTools::subtract_mean_value(vector_zero_mean);
       }
-
-      pointer_to_src = &vector_zero_mean;
+      else
+      {
+        AssertThrow(false,
+                    dealii::ExcMessage(
+                      "Undefined pressure level always leads to singular Laplace operator."));
+      }
+      // Note that the preconditioner is not updated here since it has
+      // already been updated in the function update_block_preconditioner().
+      solver_pressure_block->solve(dst, vector_zero_mean);
     }
-
-    dst = 0.0;
-    // Note that the preconditioner is not updated here since it has
-    // already been updated in the function update_block_preconditioner().
-    solver_pressure_block->solve(dst, *pointer_to_src);
+    else
+    {
+      // Note that the preconditioner is not updated here since it has
+      // already been updated in the function update_block_preconditioner().
+      solver_pressure_block->solve(dst, src);
+    }
   }
 }
 
--- a/include/exadg/incompressible_navier_stokes/spatial_discretization/operator_coupled.cpp
+++ b/include/exadg/incompressible_navier_stokes/spatial_discretization/operator_coupled.cpp
@@ -1206,32 +1206,24 @@ OperatorCoupled<dim, Number>::apply_inverse_negative_laplace_operator(VectorType
   }
   else // exact_inversion_of_laplace_operator == true
   {
-    // solve a linear system of equations for negative Laplace operator to given (relative)
-    // tolerance using the PCG method
+    // solve a linear system of equations related to the negative Laplace operator to given
+    // (relative) tolerance using an iterative method
+    VectorType const * pointer_to_src = &src;
+
+    // manipulate src-vector if operator is singular
+    VectorType vector_zero_mean;
+    if(laplace_operator->operator_is_singular())
+    {
+      vector_zero_mean = src;
+      dealii::VectorTools::subtract_mean_value(vector_zero_mean);
+      pointer_to_src = &vector_zero_mean;
+    }
+
     dst = 0.0;
-    if(this->is_pressure_level_undefined())
-    {
-      VectorType vector_zero_mean = src;
-      if(laplace_operator->operator_is_singular())
-      {
-        dealii::VectorTools::subtract_mean_value(vector_zero_mean);
-      }
-      else
-      {
-        AssertThrow(false,
-                    dealii::ExcMessage(
-                      "Undefined pressure level always leads to singular Laplace operator."));
-      }
-      // Note that the preconditioner is not updated here since it has
-      // already been updated in the function update_block_preconditioner().
-      solver_pressure_block->solve(dst, vector_zero_mean);
-    }
-    else
-    {
-      // Note that the preconditioner is not updated here since it has
-      // already been updated in the function update_block_preconditioner().
-      solver_pressure_block->solve(dst, src);
-    }
+
+    // Note that the preconditioner is not updated here since it has
+    // already been updated in the function update_block_preconditioner().
+    solver_pressure_block->solve(dst, *pointer_to_src);
   }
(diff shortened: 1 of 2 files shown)


include/exadg/incompressible_navier_stokes/spatial_discretization/operator_coupled.cpp
```diff
@@ -1208,24 +1208,30 @@ OperatorCoupled<dim, Number>::apply_inverse_negative_laplace_operator(VectorType
...
+    {
+      // Note that the preconditioner is not updated here since it has
+      // already been updated in the function update_block_preconditioner().
+      solver_pressure_block->solve(dst, src);
     }
-
-    dst = 0.0;
-    // Note that the preconditioner is not updated here since it has
-    // already been updated in the function update_block_preconditioner().
-    solver_pressure_block->solve(dst, *pointer_to_src);
```
richardschu, line 1228: `vector_zero_mean` goes out of scope?
i just replicated the functionality, but I think:
1) `dst=0` gets rid of the initial guess ... so is not really a good idea?
2) an undefined pressure level (no Neumann BC) in the Navier-Stokes problem always leads to a singular Laplace operator. So why the second if? Is there something I forget? How would the operator not be singular in that case? Is the wording of the functions misleading and I dont get it?

include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.cpp
```diff
@@ -471,8 +471,7 @@ MultigridPreconditionerBase<dim, Number, MultigridNumber>::
       dealii::Functions::ZeroFunction<dim, MultigridNumber> zero_function(
         dof_handler->get_fe().n_components());
 
-      auto const & mapping_dummy =
-        dof_handler->get_fe().reference_cell().template get_default_linear_mapping<dim>();
```
richardschu, line 475: I found this syntax often on deal.II, but I do not know it. My compiler also got confused. Why not take the right one?

peterrum: Internally, the function does the same. Both won't work for mixed meshes. So we will need to change this line at most then.

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_coupled.cpp
richardschu: Yes we are in the preconditioner, specifically `OperatorCoupledapply_preconditioner_pressure_block()` (see line 1119 above) . We solve the Schur complement `S p = b`. I think `dst` constains the last pressure iterate, which might help here, but I do not know. To verify, we would need to test. Should I add a 
`TODO: test if setting dst = 0; impacts iteration counts`
?
I think for now, let us reproduce the previous behavior, as we are here just for the compile warnings ([exactly what I did in the new version]).

richardschu: Regarding the compile warning and actual problem:
.) I am not at all against calling `solver->solve()` once, that would be beautiful, but it does not work in the present form, since for `this->is_pressure_level_undefined() = true`, we define `VectorType vector_zero_mean = src;` then `pointer_to_src = &vector_zero_mean;`, BUT the `vector_zero_mean` goes out of scope. So the pointer points to freed memory?

Now that you say it: if we really want to `dst=0;` (in the current implementation this is still optional), we can use `dst` as the temporary storage since it never goes out of scope. I can come up with a version that uses `dst` that way, you prefer that over the version that I did?

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_coupled.cpp
```diff
@@ -1208,24 +1208,30 @@ OperatorCoupled<dim, Number>::apply_inverse_negative_laplace_operator(VectorType
...
       {
         dealii::VectorTools::subtract_mean_value(vector_zero_mean);
       }
-
-      pointer_to_src = &vector_zero_mean;
+      else
+      {
+        AssertThrow(false,
+                    dealii::ExcMessage(
+                      "Undefined pressure level always leads to singular Laplace operator."));
```
richardschu, line 1223: Note that this `else`-branch + `AssertThrow()` is new. I think that `this->is_pressure_level_undefined()` implies `laplace_operator->operator_is_singular()`. I would like to add the assert.

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_coupled.cpp
nfehn: Ah, thanks for the hint regarding the scope. I didn't see this. The declaration of the vector should happen outside the brackets.

nfehn: If I remember correctly, for some Krylov solvers, the preconditioner is applied to the residual, so the outcome of the preconditioner application would in general not be close to the solution of the linear system of equations?

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_coupled.cpp
nfehn: I think the additional assert is irritating, because `is_pressure_level_is_undefined()` is used when setting up the operator data for the laplace operator. I think the code should look like this:


```C++
  else
  {
     VectorType const * pointer_to_src = &src;

     // manipulate src-vector if operator is singular
      VectorType vector_zero_mean;
      if(laplace_operator->operator_is_singular())
      {
        vector_zero_mean = src;
        dealii::VectorTools::subtract_mean_value(vector_zero_mean);
        pointer_to_src = &vector_zero_mean;
      }

      // call the solver
     ...
   }
```

Does that make sense?

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_coupled.cpp
nfehn: Overall, this else branch is not intended to be used in practise. For the Laplace operator, just the multigrid V-cycle should be a good preconditioner and nested Krylov solvers should generally be avoided. Or do you have an application where you see a major difference regarding this if/else-branch?

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_coupled.cpp
richardschu: Ehm so I think if you replace `else` in your code block with `if(this->is_pressure_level_undefined())`, it is what we had above. But lets stay closer to the version we had originally. I will push a suggestion asap! I think I know what you mean :)

nfehn: The `else` in the pseudo-code above comes from the outer if/else branch of the function `apply_negative_laplace_operator()`.

So I really want to avoid the usage of `is_pressure_level_undefined()` and `operator_is_singular()`, where I agree that this is irritating.

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_coupled.cpp
richardschu: Ah now I get it. You mean that the iterative solver should not be used to solve the pressure block in general in a production run:
```
if(this->param.exact_inversion_of_laplace_operator == false)
{
  // standard path
}
else
{
  // debug path, not to be used in production
}
```
That's true, yes :)

richardschu: > > If I remember correctly, for some Krylov solvers, the preconditioner is applied to the residual, so the outcome of the preconditioner application would in general not be close to the solution of the linear system of equations?
> 
> image
> 
> Yes 👍 So `dst = 0;` makes a lot of sense now. we can use it as a temporary vector. Then we do not need an additional vector. I will adapt to the style you used before!

reply to myself: obviously one cant use `dst` as the temporary storage (like `dst = src; dst -=mean(dst);`) since one cant solve using a single vector as `dst` and `src` and the latter is `const`.

nfehn (approved):

PR #701, opened 2024-09-18
Add smartpointer header

bergbauer:
Fix build against deal.II master due to https://github.com/dealii/dealii/pull/17689


bergbauer:
This is fixed by https://github.com/dealii/dealii/pull/17714. At some point we should rename `SmartPointer` to the new name `ObserverPointer`.

Closing

PR #709, opened 2024-10-22
make sure that matrices are assembled for multigrid with matrix-based implementation

nfehn:
Currently, this PR only addresses the Poisson module.

### DIFF
--- a/include/exadg/poisson/preconditioners/multigrid_preconditioner.cpp
+++ b/include/exadg/poisson/preconditioners/multigrid_preconditioner.cpp
@@ -76,8 +76,11 @@ MultigridPreconditioner<dim, Number, n_components>::update()
 
     this->update_matrix_free_objects();
 
-    this->for_all_levels(
-      [&](unsigned int const level) { get_operator(level)->update_penalty_parameter(); });
+    // update operators
+    this->for_all_levels([&](unsigned int const level) {
+      get_operator(level)->update_penalty_parameter();
+      get_operator(level)->assemble_matrix_if_necessary();
+    });
 
     // Once the operators are updated, the update of smoothers and the coarse grid solver is generic
     // functionality implemented in the base class.
--- a/include/exadg/poisson/spatial_discretization/laplace_operator.cpp
+++ b/include/exadg/poisson/spatial_discretization/laplace_operator.cpp
@@ -44,6 +44,8 @@ LaplaceOperator<dim, Number, n_components>::initialize(
   kernel.reinit(matrix_free, data.kernel_data, data.dof_index);
 
   this->integrator_flags = kernel.get_integrator_flags(this->is_dg);
+
+  this->assemble_matrix_if_necessary();
 }
 
 template<int dim, typename Number, int n_components>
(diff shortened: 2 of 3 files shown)


nfehn:
Let me point to another issue:

The two parameters `param.use_matrix_based_implementation` are `param.multigrid_data.min_degree_matrix_free` are currently independent. This means that  `param.use_matrix_based_implementation` decides about the mat-vec product in the Krylov solver, while `param.multigrid_data.min_degree_matrix_free` about the mat-vec products in the multigrid preconditioner. 

One implication is: If `param.use_matrix_based_implementation = true` and `param.multigrid_data.min_degree_matrix_free = 1` (or not setting the parameter at all because `1` is the default value), multigrid will run in a matrix-free manner, but a user might potentially expect that everything runs in a matrix-based manner because of the "global" parameter `param.use_matrix_based_implementation = true`. This means that a user needs to set two parameters to run in a fully matrix-based manner. What do you think about this? Could this potentially be misunderstood?

The alternative would be that we replace
```C++
// in MultigridPreconditionerBase::initialize_operators()
    bool const use_matrix_based_implementation =
      level_info[level].degree() < this->data.min_degree_matrix_free;
```
by 
```C++
// in e.g. Poisson::MultigridPreconditioner::initialize_operator()
    bool const use_matrix_based_implementation =
      (level_info[level].degree() < this->data.min_degree_matrix_free) or operator_data.use_matrix_based_vmult;
```
i.e. the global parameter `use_matrix_based_implementation` has a higher priority and forces multigrid to use a matrix-based implementation if we use this on the fine level. The disadvantage of the alternative variant is that we have to implement this redundantly for every multigrid preconditioner (i.e. every module in ExaDG like Poisson, Structure, ...). Moreover, one might argue that the "standard case" is to use a high-order method with matrix-free on the fine level, but only decide for matrix-based within multigrid on some levels of lower polynomial degree. Also note that the "global parameter" `use_matrix_based_implementation` has been introduced for throughput studies and currently only exists in the `Poisson` and `Structure` module (but for example not in the `IncNS` module). For these reasons, I would use the logic as realized currently.

nfehn:
See commit https://github.com/exadg/exadg/pull/709/commits/c1b7e0bbcbf28686cf2bfb332160ab127058ae04 in order to be more precise regarding the issues discussed here https://github.com/exadg/exadg/pull/709#issuecomment-2475713630

bergbauer:
> What do you think about this? Could this potentially be misunderstood?

Yes, I think from the name `use_matrix_based_implementation` one would expect that every `vmult` is done matrix based. I see different options to solve this:
- adjust  also the name of the parameter to reflect that this only affects the global operator
- use the logic `min_degree_matrix_free` internally for everything, which means setting it to "large integer" for `use_matrix_based_implementation == true` and have both operator and multigrid run matrix-based

nfehn:
For now, I came to the conclusion that it is good to avoid internal dependencies between the parameters (e.g. one parameter overwriting the other in certain cases). This is hard to document e.g. since different classes are involved. For reasons of modularity, I believe the parameter `min_degree_matrix_free` needs to be part of `MultigridData`.

I added the new parameter(s) for the `poisson/sine` application to demonstrate usage.

richardschu:
check if we can remove the line added in
https://github.com/exadg/exadg/pull/797
(I will look into this soon)

PR #711, opened 2024-11-13
Fix compilation with current deal.II

kronbichler:
Since https://github.com/dealii/dealii/pull/17847 removed a file we rely on, we should update the internal data structures.


nfehn (approved):

nfehn:
thanks!

PR #716, opened 2024-12-31
copy dst vector entries in apply_function_in_double_precision()

### LINKED ISSUE #715: Gaussian test fails in debug mode
if I run the tests in debug mode, the Gaussian test fails giving the stacktrace in the comment below.
I traced it back to `apply_function_in_double_precision()`.
Within the `operation(dst, src)` being a `solver.solve()`, and after a `std::cout` countaining the `##+`string, I tried to just compute `dst.l2_norm()` and `src.l2_norm()` before doing anything else, which already triggers an assert since we do not copy entries in `dst`, but the ones in `src`. So some entries are not set/infinite.
For `apply_function_in_double_precision()`to be general, we should also copy entries in the `dst` vector.
PR incoming.

richardschu:
fixes #715

### DIFF
--- a/include/exadg/solvers_and_preconditioners/utilities/linear_algebra_utilities.h
+++ b/include/exadg/solvers_and_preconditioners/utilities/linear_algebra_utilities.h
@@ -153,6 +153,7 @@ apply_function_in_double_precision(
     dealii::LinearAlgebra::distributed::Vector<double> dst_double, src_double;
     dst_double.reinit(dst, true); // do not zero entries
     src_double.reinit(src, true); // do not zero entries
+    dst_double.copy_locally_owned_data_from(dst);
     src_double.copy_locally_owned_data_from(src);
 
     operation(dst_double, src_double);


kronbichler:
While I agree with making sure the `dst` argument of an iterative solver is properly initialized (because it is an input+output argument in that case), I wonder if it makes sense in other contexts where the operation overwrites the content and treats `dst` as a pure output argument. If this is never called in a performance-critical part, the proposed patch is a very good way to ensure correctness, but we should consider the use cases a bit. (One could say that the fact that we perform float-double conversions both on the `src` vector and in the end is indication enough.) I would thus say: Let us go with this solution for now, but keep in mind the possible implications in terms of efficiency.

kronbichler (approved):

nfehn:
To be explicit about the (currently) wrong behavior: The reason is that this function may add into `dst`, right?

richardschu:
> While I agree with making sure the `dst` argument of an iterative solver is properly initialized (because it is an input+output argument in that case), I wonder if it makes sense in other contexts where the operation overwrites the content and treats `dst` as a pure output argument. If this is never called in a performance-critical part, the proposed patch is a very good way to ensure correctness, but we should consider the use cases a bit. (One could say that the fact that we perform float-double conversions both on the `src` vector and in the end is indication enough.) I would thus say: Let us go with this solution for now, but keep in mind the possible implications in terms of efficiency.

`apply_function_in_double_precision()` *is* used for throughput comparisons.
Exactly the case you mention is the matrix-based vs. matrix-free comparison in

https://github.com/exadg/exadg/blob/6d3cb9706eccaaa8a41f26e3890bdc8ced8bf5b0/include/exadg/operators/operator_base.cpp#L414C1-L420C10
 
Apart from this, we use it for
1) coarse-grid solver, ML AMG multigrid cycle `precondition_ml.vmult()`
https://github.com/exadg/exadg/blob/6d3cb9706eccaaa8a41f26e3890bdc8ced8bf5b0/include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h#L471C3-L477C12

2) coarse-grid solver, Krylov-solver-wrapped AMG `preconditioner.apply_krylov_solver_with_amg_preconditioner()`
https://github.com/exadg/exadg/blob/6d3cb9706eccaaa8a41f26e3890bdc8ced8bf5b0/include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h#L519C1-L522C84

I can introduce a boolean argument `copy_dst_vector` into `apply_function_in_double_precision()` with default `true` and we use it for the ML and Krylov+ML cases. In `OperatorBase::apply_matrix_based()` we set it to `false` and hence skip the copying? What do you think @kronbichler @nfehn ?

PR #717, opened 2025-01-02
add copy_dst_entries arg to apply_function_in_double_precision()

richardschu:
this is what I meant in
https://github.com/exadg/exadg/pull/716#issuecomment-2567683586

### DIFF
--- a/include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
+++ b/include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
@@ -474,7 +474,8 @@ public:
         [&](dealii::LinearAlgebra::distributed::Vector<double> &       dst_double,
             dealii::LinearAlgebra::distributed::Vector<double> const & src_double) {
           preconditioner_ml->vmult(dst_double, src_double);
-        });
+        },
+        true /* copy_dst_entries */);
 #else
       AssertThrow(false, dealii::ExcMessage("deal.II is not compiled with Trilinos!"));
 #endif
@@ -520,7 +521,8 @@ public:
                                                                       src_double,
                                                                       solver_type,
                                                                       solver_data);
-        });
+        },
+        true /* copy_dst_entries */);
 #else
       AssertThrow(false, dealii::ExcMessage("deal.II is not compiled with Trilinos!"));
 #endif
--- a/include/exadg/operators/operator_base.cpp
+++ b/include/exadg/operators/operator_base.cpp
@@ -417,7 +417,8 @@ OperatorBase<dim, Number, n_components>::apply_matrix_based(VectorType &       d
       [&](dealii::LinearAlgebra::distributed::Vector<double> &       dst_double,
           dealii::LinearAlgebra::distributed::Vector<double> const & src_double) {
         system_matrix_trilinos.vmult(dst_double, src_double);
-      });
+      },
+      false /* copy_dst_entries */);
 #else
     AssertThrow(
       false,
--- a/include/exadg/solvers_and_preconditioners/utilities/linear_algebra_utilities.h
+++ b/include/exadg/solvers_and_preconditioners/utilities/linear_algebra_utilities.h
@@ -141,7 +141,8 @@ apply_function_in_double_precision(
   dealii::LinearAlgebra::distributed::Vector<Number> &                            dst,
   dealii::LinearAlgebra::distributed::Vector<Number> const &                      src,
   std::function<void(dealii::LinearAlgebra::distributed::Vector<double> &,
-                     dealii::LinearAlgebra::distributed::Vector<double> const &)> operation)
+                     dealii::LinearAlgebra::distributed::Vector<double> const &)> operation,
+  bool const copy_dst_entries = true)
 {
   if constexpr(std::is_same_v<Number, double>)
   {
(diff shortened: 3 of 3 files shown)


nfehn:
Thanks for these explanations.

In my opinion, the notion of "copy" is not clear from the interface of this function (e.g. no copy is done in the other branch), so one could argue that it should remain a detail of the implementation of this function. The interface of the function says in my opinion that dst might be non-zero and that this content of dst (when entering the function) can be relevant. In this sense, PR #716 is the fix needed w.r.t. the interface of this function.

nfehn:
I think I understand the motivation for the present PR and I agree that we should do the copy in general. It is not clear to me whether we should try to avoid the copy in some cases according to the present PR, for the following reasons:

1. The only place in this PR where the last argument is set to `false` is the case where the function is called with vectors of basic data type `double`, hence not entering the else branch affected by the present PR. The code in ExaDG calling this function with vectors of basic data type `float` (potentially) call the respective function with the last argument set to `true` according to the present state of this PR. So the present PR wouldn't change something in ExaDG right now, would it?
2. I think the present PR complicates somewhat the code or increases the need for documentation. Hence, apart from aspect 1 mentioned above, there should be motivation that this change results in measurable performance improvements in uses cases where this PR really changes the code. Following the goal of performance improvements, realizing the respective functions truly in `float` precision would probably allow much higher improvements.
3. Regarding the naming, I would (i) prefer a name where `true` is the default and (ii) where the meaning of the variable is clearer that the naming with `copy`. One suggestion fulfilling these properties could be `dst_is_input`.

As a code reviewer, I have to shed light on code changes from different perspectives and raise critical questions. As explained above, I do not see really high priority for this PR. However, if you want to realize the present PR, I am open to do so.

richardschu:
Thanks Niklas for taking a look!
>     As a code reviewer, I have to shed light on code changes from different perspectives and raise critical questions. As explained above, I do not see really high priority for this PR. However, if you want to realize the present PR, I am open to do so.
I totally understand and I am not taking it personally at all! I appreciate your efforts and your input! :+1:

Regarding the comments you had:

1) I am not 100% sure this is correct, but you looked deeper into the `apply_matrix_based()` pit, maybe I misunderstand whats happening. Please let me know if the following is correct or not:

. lets say the outer Krylov solver runs in `double`
. the mulrigrid preconditioner runs in `float` per default (see https://github.com/exadg/exadg/blob/6d3cb9706eccaaa8a41f26e3890bdc8ced8bf5b0/include/exadg/solvers_and_preconditioners/multigrid/multigrid_preconditioner_base.h#L60C1-L64C44)
. hence, the behavior of the code is changing if `use_matrix_based_implementation == true`, right? (we skip copying the `dst` vector in that case)?
. therefore, the matrix-free approach is not impacted by the changes (affected parts use `copy_dst_entries == true` for correctness), but the matrix-based methods are affected via `apply_matrix_based/_add()` (with Trilinos). For the affected matrix-based methods, however, this might not even be close to the actual bottleneck.

2) completely true, we would have to test to justify the additional complication and documentation effort.
3) I would be/am totally fine with renaming `copy_dst_entries` -> `dst_is_input`.

Given the reasoning under 1), we can very well close the PR given that the performance gains from skipping copying the `dst` vector are likely to be small following a matrix-based approach, if @kronbichler agrees on these points (Martin just please give a short thumbs up if you agree here; of course testing would be needed, we can try that next time we are comparing to a matrix-based variant; I just did not want to ignore your comment https://github.com/exadg/exadg/pull/716#issuecomment-2567136793)

nfehn:
You are right @richardschu, matrix-based might be used just for the throughput runs (which I had in mind yesterday), but also as the implementation variant for all operator evaluations in a solver.

I am open in all directions. If we realize the changes suggested here, I would be happy about further comments in the code (because in a function called `vmult()`, it might not be obvious why one has to copy the dst-vector / consider the dst-vector an input vector in my opinion (to be on the safe side, one would always do it, but thats exactly the point of discussion here ...): googeling "dealii vmult" and following the google search result that sounds like Trilinos (https://www.dealii.org/current/doxygen/deal.II/classTrilinosWrappers_1_1SparseMatrix.html#aac3432b6b2cd742db298d049760eedfb), I get the impression that the dst-vector is not an input. Continuing in this philosophical direction, I would ask the question if the naming and usage of `vmult` in dealii is intuitive and consistent. Of course, I should better google "dealii trilinos amg vmult" and then go to https://www.dealii.org/current/doxygen/deal.II/classTrilinosWrappers_1_1PreconditionBase.html#a56cdde99bb47cec6af2228870e5b5129, but the documentation of `vmult()` there just says "apply the preconditioner" without further links or explanations ...).

include/exadg/solvers_and_preconditioners/preconditioners/preconditioner_amg.h
```diff
@@ -474,7 +474,8 @@ class PreconditionerAMG : public PreconditionerBase<Number>
         [&](dealii::LinearAlgebra::distributed::Vector<double> &       dst_double,
             dealii::LinearAlgebra::distributed::Vector<double> const & src_double) {
           preconditioner_ml->vmult(dst_double, src_double);
-        });
+        },
+        true /* copy_dst_entries */);
```
nfehn, line 478: See also discussion here https://github.com/dealii/dealii/issues/17978

kronbichler: I would say that this interface should not need to copy entries because we call the `ApplyInverse` function of the Epetra operators (I did not look at this code closely yet), we need the initial values only if an iterative solver is called. Is any of the tests failing if we call this particular place with `false` argument as evidence that this function does not overwrite the vector?

richardschu: so the assert we ran into in #715 in debug mode was when we were trying to get the norm of the `dst` vector, see https://github.com/exadg/exadg/issues/715#issuecomment-2566520074
Note that `dst.l2_norm()` was added by me to provoke the assert.
As far as I am concerned, I voted for closing #717 completely while keeping in mind that we are copying data around in some cases. We need to check this next time we do performance measurements.

PR #722, opened 2025-01-19
Fix boost archiving of LA::d::Vector and make it debug-friendly

### LINKED ISSUE #721: Serialization does not store vectors
when using the restart mechanics, we store the `LinearAlgebra::distributed::Vector`s into a `boost::archives::binary_archive`.
The infrastructure is there, but this vector type does not support serialization.
I wrote a little test (attached) that showed me that serialization is possible for
`dealii::Vector`
`PETScWrappers::MPI::Vector`
and not possible for
`dealii::LinearAlgebra::distributed::Vector`
`TrilinosWrappers::MPI::Vector`
(for good reasons, see the documentation of the respective functions, or simply because it is not implemented yet)

For the PETSc vector also in parallel this works totally as intended, for the Trilinos Vector, this is not the case, which confused me a bit.
Also, there is no warning whatsoever, the vector entries are simply not touched.

I will do the following:
1) define a type for the `boost::archive` such that we can easier debug this stuff in `time_int_base.h`.
I found that "storing" the `dealii::LinearAlgebra::distributed::Vector` just adds a single(!) `0` in the archive, otherwise I could only guess from the size of the file that some stuff is missing. Then with the `boost::archive::text_oarchive`, I could see that indeed we are doing _something_, but it was only writing this `0`.
 
2) Serialize the local entries directly and copy into the `dealii::LinearAlgebra::distributed::Vector`.
This way, we just add a utility function to go from `dealii::LinearAlgebra::distributed::Vector` -> `Number` -> `boost::archive` and back.

PR incoming

[serialize_vectors.txt](https://github.com/user-attachments/files/18464629/serialize_vectors.txt)

richardschu:
fixes #721

essentially in 3 steps:

1) 
https://github.com/exadg/exadg/commit/7b98515f16245710bd71f3144342d04a84ef6e38
introduces a type to easily switch from the 
binary `boost::archive::binary_i/oarchive` 
to the human readable `boost::archive::text_i/oarchive`
via
image

2)
61ffe67efc1881b885371634e46ee305ec9507ca
 
introduces the utility functions
`read_distributed_vector()`
and
`write_distributed_vector()`
in
`include/utilities/boost_archive.h`

we essentially loop over the local entries and write them to the archive, same when we read them back.
IMPORTANT: we do not store any other info regarding the data structures. Hence we assume that the mesh/DoFs/parallel layout/etc. is _exactly the same_. This is similar to the already functionality, this PR is simply a bugfix.

3)
I added a test in
`tests/utilities/boost_archive.cc`
which can be run in `debug` and `release` (since debug mode currently outputs the vector norms ... this is a point to discuss) and in `serial` and `mpirun=2`.

I am pro outputting the norms in general, but this is some global communication... I guess that one scalar value doesnt hurt though? And snapshots are not created all the time but frequently enough. SO I think checking the vectors when you read them in gives you a little bonus confidence that you copied the right files!

`mpirun=2` I think is necessary since checkpointing is only really useful once we have a parallel run and the reading and writing of the archives is done per processor. These tests run in under 2 secs.


richardschu:
once we have agreed on how to apply the fix, I can squash the commits!

richardschu:
I just changed the norm output a bit to have a constant format like "1.2345e+05" to compare the read/write vectors better.
Also, I included the binary boost archive header despite the fact that it compiled before with no problems. But adding both text and binary archive types is more clear I think. We need both.
This is ready for review.

richardschu:
> I just changed the norm output a bit to have a constant format like "1.2345e+05" to compare the read/write vectors better. Also, I included the binary boost archive header despite the fact that it compiled before with no problems. But adding both text and binary archive types is more clear I think. We need both. This is ready for review.

"ready for review" and did not even compile the tests :+1:

nfehn:
It looks like you could reduce the two new functions to a single one by using a lambda. This way, I think you could also avoid the second template argument for each these two functions. If I understand the logic correctly, the purpose of both functions is "for every element of a vector, do something". From a software design perspective, one might argue that "do_something" itself is a function and should not cause to implement the same functionality essentially multiple times. In the time integration routine, the code could look like

```C++
// define lambda's read_from_archive, write_to_archive ...

for_every_vector_entry(vector, read_from_archive);
...
for_every_vector_entry(vector, write_to_archive);
```

Can't you move the debug output directly to the test program?

richardschu:
> It looks like you could reduce the two new functions to a single one by using a lambda. This way, I think you could also avoid the second template argument for each these two functions. If I understand the logic correctly, the purpose of both functions is "for every element of a vector, do something". From a software design perspective, one might argue that "do_something" itself is a function and should not cause to implement the same functionality essentially multiple times. In the time integration routine, the code could look like
> 
> ```c++
> // define lambda's read_from_archive, write_to_archive ...
> 
> for_every_vector_entry(vector, read_from_archive);
> ...
> for_every_vector_entry(vector, write_to_archive);
> ```
> 
> Can't you move the debug output directly to the test program?

Thanks for your comment Niklas!
I thought about this again; my motivation was that if a user is reading/writing checkpoint data, they probably want to know if everything worked as planned. Especially since when restarting, the time integration scheme might have a "hickup" (adaptive time stepping; _keeping_ the max CFL constant, but one has to start with lower order). When reading the output (e.g., plot solution over time for some point in Paraview) one sees that values seem ok first, but then different transients are computed (they are always ok, there is no bug, but the user might not know all the intrinsics). 

What I mean to say is that when one checks the log files, one sees that the reading was correctly done when the norms of all vectors match. Hence I would even vote for printing the norms all the time, i.e., in `debug` and `release` mode.
Then we can also write a single function using the `&` operator with the boost archive. I simply choose `<<` and `>>` here since I already had different implementations and these operators are very intuitive also if one does not know that `&` might read/write depending on the boost archive type.

nfehn:
Of course.

richardschu:
I merged now the two read/write functions using the `&` operator.
The type of operation is now inferred from the `ArchiveType`.

After talking to Martin I also removed the different behavior in `debug` and `release` mode: 
now, we always use the `boost_binary_i/oarchivetype` (as before). 
Therefore the tests in `debug`/`release` are identical (as we also went for always writing the vector norms)
--> so I simplified the tests.

include/exadg/time_integration/restart.h
```diff
@@ -66,6 +72,77 @@ write_restart_file(std::ostringstream & oss, std::string const & filename)
...
+    print_vector_l2_norm(vector);
+  }
+
+  // Depending on VectorType, we have to loop over the blocks to
+  // access the local entries via vector.local_element(i).
+  using Number = typename VectorType::value_type;
+  if constexpr(std::is_same<VectorType,
+                            dealii::LinearAlgebra::distributed::Vector<Number>>::value or
+               std::is_same<VectorType,
+                            dealii::LinearAlgebra::distributed::Vector<Number> const>::value)
```
kronbichler, line 112: You could simplify the check a bit by
```suggestion
  if constexpr(std::is_same<std::remove_cv_t<VectorType>,
                            dealii::LinearAlgebra::distributed::Vector<Number>>::value)
```
where the `remove_cv_t` removes a possible `const` keyword of the type.

include/exadg/convection_diffusion/time_integration/time_int_bdf.cpp
```diff
@@ -427,11 +428,11 @@ TimeIntBDF<dim, Number>::print_solver_info() const
 
 template<int dim, typename Number>
 void
-TimeIntBDF<dim, Number>::read_restart_vectors(boost::archive::binary_iarchive & ia)
+TimeIntBDF<dim, Number>::read_restart_vectors(BoostInputArchiveType & ia)
 {
   for(unsigned int i = 0; i < this->order; i++)
   {
-    ia >> solution[i];
+    read_write_distributed_vector(solution[i], ia);
```
kronbichler, line 435: I would personally find it more intuitive to have two different functions `read_distributed_vector` (for this case) and `write_distributed_vector` for the `write_restart_vector` function below, given that we already have two different functions in this class and all similar classes below. There is the issue of code duplication, but to me the fact that the boost implementation can deduce from the templated type whether it should read or write is in my opinion a bigger obstacle for many users than duplicating 10 lines of code (one could clean up the functions a bit if we just read or write). The other question is whether we should implement a restart functionality in the deal.II vector classes, which would simplify this code here.

kronbichler (approved):
This is good. I have two questions, see below. (We do not need to do it here, we can also postpone them to later.)

include/exadg/time_integration/restart.h
richardschu: saving some space is always good :)

include/exadg/convection_diffusion/time_integration/time_int_bdf.cpp
richardschu: thats one of the reasons why I chose this approach initially, I totally agree its easier to read `<<` and `>>` than to look up the docu of `boost` and find that it is handled automatically.
I have no strong preference as the explanations in the related dealii steps also use the `&` and explain it and we use it elsewhere. After Niklas' https://github.com/exadg/exadg/pull/722#issuecomment-2645793739  I changed it from what you are suggesting now.

Regarding putting vector serialization into dealii:
As of now, one usually goes the way through `SolutionTransfer` calling `prepare_for_serialization` and serialized the `triangulation` afterwards. Since I am working on this part at the moment, I would say we go with the current version and then look at the generalized version I come up with.

kronbichler: OK, let's keep it for now and see where we'll be going.

include/exadg/time_integration/restart.h
```diff
@@ -66,6 +72,75 @@ write_restart_file(std::ostringstream & oss, std::string const & filename)
...
+    for(unsigned int i = 0; i < vector.locally_owned_size(); ++i)
+    {
+      archive & vector.local_element(i);
+    }
+  }
+  else if constexpr(std::is_same<VectorType,
+                                 dealii::LinearAlgebra::distributed::BlockVector<Number>>::value or
+                    std::is_same<
+                      VectorType,
+                      dealii::LinearAlgebra::distributed::BlockVector<Number> const>::value)
```
kronbichler, line 121: Can we do the same here?
```suggestion
  else if constexpr(std::is_same<std::remove_cv_t<VectorType>,
                                 dealii::LinearAlgebra::distributed::BlockVector<Number>>::value)
```

richardschu:
Since we reached a final state in the discussion, I am merging now.
The entire restart/checkpointing might be refactored in the near future anyways as I am working on a generalization rn.
Thanks again for your remarks.

PR #723, opened 2025-01-23
Add direct sparse coarse grid solver

bergbauer:
What do you think about adding an interface to the sparse direct solver for the coarse grid solver?

### DIFF
--- a/include/exadg/solvers_and_preconditioners/multigrid/multigrid_parameters.h
+++ b/include/exadg/solvers_and_preconditioners/multigrid/multigrid_parameters.h
@@ -31,6 +31,7 @@
 #include <deal.II/lac/trilinos_precondition.h>
 
 // ExaDG
+#include <deal.II/lac/trilinos_solver.h>
 #include <exadg/solvers_and_preconditioners/solvers/solver_data.h>
 #include <exadg/utilities/print_functions.h>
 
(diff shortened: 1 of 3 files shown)


nfehn:
Why not? Sounds like a good extension of the ExaDG library (or, better, framework).

Maybe you could explain the meaning of "sparse" here? Is it just because the problem stems from a finite element problem, or is the direct solver specialized in any way regarding "sparsity"?

include/exadg/solvers_and_preconditioners/multigrid/multigrid_parameters.h
```diff
@@ -94,6 +96,11 @@ enum class MultigridCoarseGridPreconditioner
   AMG
 };
 
+struct SparseDirectData
+{
+  dealii::TrilinosWrappers::SolverDirect::AdditionalData trilinos_data;
+};
+
```
nfehn, line 103: Does this work without trilions-ifdef?

richardschu:
I think this can make sense for i) debugging (direct solver = you cannot mess up the coarse grid tolerance), and ii) if the conditioning is extremely bad (probably why you are doing this, huh?). In GMG, one can then shave off some DoFs going to coarse `h` or `p` levels to make them even tractable using a direct solver. Which `solver::type` have you tried in the `Amesos` ackage? I think even if we _could_ set the min coarse problem size in the ML preconditioner high enough, this still makes sense especially if one has access to MUMPS or Pardiso and not only AmesosKLU.

PR #726, opened 2025-02-20
correct indices for BlockVectorType

richardschu:
this is an add on to #722 
I missed those switched indices.
The analogue `VectorType` I tested, the `BlockVectorType` I just tested now when rewriting the de-/serialization.


richardschu:
I am merging this because it is just an obvious index error.

PR #728, opened 2025-02-22
add missing time integration vectors in de-/serialization of `time_int_abm_base.h`

### LINKED ISSUE #727: restart mechanics of `acoustic_conservation_eqns`: some data assigned wrong?
I am currently working on the serialization/deserialization and came accross what I consider unexpected:

The code is in #728. [edit: made the fix in that branch]

I am restarting the `acoustic_conservation_equations/circular_moving_gauss_pulse/application.h`.

Note that this is before I fiddled around with any of the restart mechanics.
The parallel layout is the same (in fact I run on 1 core only), the discretization is identical to what was stored to disk.

When I use uniform time steps, second-order and `start_with_low_order = false`, I get:

Image

where this little wavelet is exactly where I reload the results from disk.
But I do not expect any problems here at all, since the timesteps are even uniform.
The vectors norms at reading/writing seem to be good:

Image

Image

using a first-order scheme in time, I get (this is restarted at the same time the other run is):

Image

with the same time step size, reducing it by a factor 10 I get:

Image

This is the last step of the simulation, it is a very good approximation, no difference to the non-checkpointed run!
So the "initial wavelet" is much reduced and there are no waves generated by the pulse travelling.

I conclude that there is something wrong. But I do not see it.
Any ideas?
This is problem is linear and we do not even solve a linear system, right?
For me, this sounds like we are reading the vectors correctly from grid, but how the timestepping scheme is restarted is wrong.

richardschu:
fixes #727
... there were just some vectors missing in the de-/serialization of the time integration scheme.
I am amazed it did not completely crash when restarting and a vector was just 0 tbh ...

### DIFF
--- a/applications/acoustic_conservation_equations/circular_moving_gauss_pulse/application.h
+++ b/applications/acoustic_conservation_equations/circular_moving_gauss_pulse/application.h
@@ -153,6 +153,8 @@ public:
                         number_of_rotations,
                         "Number of pulse rotations during runtime.",
                         dealii::Patterns::Double(1.0e-12));
+      prm.add_parameter("WriteRestart", write_restart, "Should restart files be written?");
+      prm.add_parameter("ReadRestart", read_restart, "Is this a restarted simulation?");
     }
     prm.leave_subsection();
   }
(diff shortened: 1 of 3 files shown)


richardschu:
I am merging this since it is a tiny fix, essentially.

I am aware of the fact that one could reconstruct one of the vectors that are de-/serialized with this patch.

Since the `order` is variable, I went with the easy fix here, but one could think of reconstructing the most recent vector stored from the `solution` _also_ stored, then only de-/serializing the old ones.
This would lead to a little save of de-/serialized data, hence speed up the snapshotting and potentially also the deserialization.

Once we see that this strategy is indeed too wasteful, we can refactor.

PR #729, opened 2025-02-23
generalize de-/serialization

richardschu:
...for now _only_ in the `acoustic_conservation_equations` module for Adams-Bashforth-Moulton timestepper, since it is the most isolated one. The utility functions related to the former strategy are left untouched, since we want backwards-compatibility until we have morphed all the modules. 

The major contribution of this PR is a bunch of utility functions in `restart.h`, this completely follows the basic deal.II approach:
1) collect a bunch of vectors (here in the time stepper `time_int_abm_base.h`)
2) create a `std::vector<VectorType *>` of all the vectors you want serialized, 
here: ask the `SpatialOperator` to `serialize` vectors, hand them to `dealii::SolutionTransfer`, which in turn is serialized through the `Triangulation`.
3) Store the mapping as a vector and serialize it as well.

the deserialization:

1) deserialize the grid
2) load the vectors
3) assign or process the vectors: if necessary or desired, we perform a matrix-free grid-to-grid projection using `dealii::RemotePointEvaluation`.
4) If manifolds were attached, attach them again given the current triangulation. The deserialized triangulation can consider a mapping that was stored, but only for hypercubes since we need `MappingQCache` for this. If this is not available, the new run can still get the solution, we use no mappings in this case(!).

This setup allows us to:
.) change the grid, FE degree or even the FE itself (I tested `Raviart-Thomas`<->`FE_DGQ` elsewhere as well, although the former are allergic to mappings still, they can only be used without mappings)
.) change the core count
.) perform a general grid-to-grid projection to consider only some detail of some previous precursor simulation 

What is still annoying:

1) depending on the `TriangulationType` we use to deserialize, we need the coarse grid stored as well, since `parallel::distributed::Triangulation::load()` needs exactly the same coarse grid to be in the triangulation already.
I put a `triangulation.store()` in the lambda that creates the coarse grid, see the `applications/acoustic_conservation_equations/circular_moving_gauss_pulse/application.h`. This is a bit tedious, but I did not want to introduce a function `create_coarse_triangulation() = 0` to be filled for de-/serialization, which I would then need to make accessible from the point where I actually want to create the triangulation. This is ugly, but I found no better way for now.

2) currently, the user has to provide a lot of parameters to describe the parameters of the snapshot. This can later be moved towards the boost archive and can be recovered without the user interfering. 

3) The mapping is annoying. Not having it at all in the projection might give complete garbage, but de-/serializing it with this strategy is not possible for all element types. If I had a user-provided `create_grid()` available at this point in the code, that would save us a lot of trouble. I did not see how I could do that, maybe I do not see the trees here standing in the middle of the forest (we even let the user define a `create_triangulation` lambda, so maybe ...).

So it seems 1) and 3) are rooted in the  same problem and 2) is temporary. I will think more about a solution to a `create_grid` available there, for this draft now, I just wanted to "get it done" in some way.

**The cool bit** is what can we do with this now:
our favourite de-/serialization example: the `circular_moving_gauss_pulse`

When I keep the grid completely the same and use the global projection, I can compute the error we make projecting:

[left is the restarted run, right is the reference]
image

So (depending on the user-controlled settings and FE spaces used):
.) the actual projection error is not even visible at all (make sure to look at the last error computed, all digits are *identical*, the projection error is less than the approximation error)
.) the iteration counts are low, we use a matrix-free Jacobi-preconditioned CG solver, other preconditioners are in theory possible as we use the standard `ExaDG::MassOperator`.

this looks like the following after the dot did a full circle:

image

[the little waves around the red dot are actually from starting the simulation, no noise is introduced with the restart at "~10 o'clock", since this PR already considers #728 resolving #727]

NOW THE FUN PART:
we change the approximation space (in later versions of this PR also changing the cores is possible, but this is for now asserted do the the boost archive approach not handling this at all)

here the logs again [left restarted run with `p=7`,`h_lvl = 2`, right `p=5`, `h_lvl=3`]

image

The projection norms from before are omitted here, since the standard strategy only delivers garbage here (which was also not asserted before). We see, however, that the error is only slightly higher! Nice!

And the obligatory solution snapshots:
[start, before restart, after restart,  end]

image
image
image
image

### DIFF
--- a/include/exadg/acoustic_conservation_equations/spatial_discretization/interface.h
+++ b/include/exadg/acoustic_conservation_equations/spatial_discretization/interface.h
@@ -31,10 +31,11 @@ namespace Acoustics
 {
 namespace Interface
 {
-template<typename Number>
+template<int dim, typename Number>
 class SpatialOperator
 {
 public:
+  using VectorType      = dealii::LinearAlgebra::distributed::Vector<Number>;
   using BlockVectorType = dealii::LinearAlgebra::distributed::BlockVector<Number>;
 
   virtual ~SpatialOperator() = default;
@@ -51,6 +52,13 @@ public:
   // time integration: evaluate
   virtual void
   evaluate(BlockVectorType & dst, BlockVectorType const & src, double const time) const = 0;
+
+  // required for restart functionality
+  virtual void
+  serialize_vectors(std::vector<BlockVectorType const *> block_vectors) const = 0;
+
+  virtual void
+  deserialize_vectors(std::vector<BlockVectorType *> block_vectors) const = 0;
 
   virtual double
   calculate_time_step_cfl() const = 0;
--- a/include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
+++ b/include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
@@ -24,10 +24,12 @@
 
 // ExaDG
 #include <exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.h>
+#include <exadg/grid/grid_utilities.h>
 #include <exadg/grid/mapping_dof_vector.h>
 #include <exadg/operators/finite_element.h>
 #include <exadg/operators/grid_related_time_step_restrictions.h>
 #include <exadg/operators/quadrature.h>
+#include <exadg/time_integration/restart.h>
 #include <exadg/utilities/exceptions.h>
 
 namespace ExaDG
@@ -43,7 +45,7 @@ SpatialOperator<dim, Number>::SpatialOperator(
   Parameters const &                             parameters_in,
   std::string const &                            field_in,
   MPI_Comm const &                               mpi_comm_in)
-  : Interface::SpatialOperator<Number>(),
+  : Interface::SpatialOperator<dim, Number>(),
     grid(grid_in),
     mapping(mapping_in),
     boundary_descriptor(boundary_descriptor_in),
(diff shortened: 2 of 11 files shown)


richardschu:
There are 2 things I fixed elsewhere and should not clutter the discussion here:
1) we also have to consider for `CG`/`DG` element types where implemented, or likewise for `L2`/`HDIV`. This will be stored in the binary archive alongside most of the parameters for now stored in `restart_data`.
2) When integrating the right hand side of the projection we can extend towards the general case
`if(n_components == 0) {/* do something */} else {/* do something not only for dim == 3 */}`. 
I leave these changes out here for now.

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
```diff
@@ -242,6 +244,145 @@ SpatialOperator<dim, Number>::get_dof_handler_u() const
...
+  std::shared_ptr<dealii::Triangulation<dim>> checkpoint_triangulation =
+    deserialize_triangulation<dim>(this->get_dof_handler_u().get_triangulation(),
+                                   param.restart_data.filename,
+                                   param.restart_data.triangulation_type,
+                                   mpi_comm);
+
+  // Setup DoFHandlers *as checkpointed*, sequence matches `this->serialize_vectors()`.
+  dealii::DoFHandler<dim> checkpoint_dof_handler_u(*checkpoint_triangulation);
+  dealii::DoFHandler<dim> checkpoint_dof_handler_p(*checkpoint_triangulation);
+  dealii::DoFHandler<dim> checkpoint_dof_handler_mapping(*checkpoint_triangulation);
```
richardschu, line 294: to be moved behind the `if(consider_mapping)`

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
```diff
@@ -242,6 +244,145 @@ SpatialOperator<dim, Number>::get_dof_handler_u() const
...
+
+  ElementType const checkpoint_element_type = get_element_type(*checkpoint_triangulation);
+
+  std::shared_ptr<dealii::FiniteElement<dim>> checkpoint_fe_u =
+    create_finite_element<dim>(checkpoint_element_type, true, dim, param.restart_data.degree_u);
+  std::shared_ptr<dealii::FiniteElement<dim>> checkpoint_fe_p =
+    create_finite_element<dim>(checkpoint_element_type, true, 1, param.restart_data.degree_p);
+  std::shared_ptr<dealii::FiniteElement<dim>> checkpoint_fe_mapping = create_finite_element<dim>(
```
richardschu, line 302: to be moved behind the `if(consider_mapping)`

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
```diff
@@ -242,6 +244,145 @@ SpatialOperator<dim, Number>::get_dof_handler_u() const
...
+    checkpoint_element_type, true, dim, param.restart_data.mapping_degree);
+
+  checkpoint_dof_handler_u.distribute_dofs(*checkpoint_fe_u);
+  checkpoint_dof_handler_p.distribute_dofs(*checkpoint_fe_p);
+  checkpoint_dof_handler_mapping.distribute_dofs(*checkpoint_fe_mapping);
```
richardschu, line 307: to be moved behind the `if(consider_mapping)`

include/exadg/acoustic_conservation_equations/spatial_discretization/interface.h
```diff
@@ -31,10 +31,11 @@ namespace Acoustics
 {
 namespace Interface
 {
-template<typename Number>
+template<int dim, typename Number>
```
nfehn, line 34: Why `dim`?

A main reason for the existence of these interface classes is the goal to write time integration classes without template parameter `dim`.

include/exadg/acoustic_conservation_equations/time_integration/time_int_abm.h
```diff
@@ -32,18 +32,19 @@ namespace ExaDG
 {
 namespace Acoustics
 {
-template<typename Number>
+template<int dim, typename Number>
```
nfehn, line 35: `dim` is "forbidden" for time integrators.

include/exadg/time_integration/restart.h
```diff
@@ -138,6 +155,739 @@ read_write_distributed_vector(VectorType & vector, BoostArchiveType & archive)
...
+}
+
+/**
+ * Utilitiy function to project vectors from a source to a target triangulation
+ * via `dealii::RemotePointEvaluation`, matrix-free mass operator evaluation
+ * and a Jacobi-preconditioned CG solver.
+ */
+template<int dim, typename Number, int n_components, typename VectorType>
+inline void
+project_vectors(
```
nfehn, line 694: Everything in this file related to the projection should not be in `time_integration/restart.h` in my opinion, but in `include/exadg/operators` (where we have e.g. utility functions for interpolation between solutions on different grids, see https://github.com/exadg/exadg/blob/master/include/exadg/operators/solution_interpolation_between_triangulations.h). Right now, I would argue such functionality does not belong to `include/exadg/grid`, because finite elements and dof handlers are involved here (not just triangulations and mapping), hence `include/exadg/operators`.

include/exadg/time_integration/restart_data.h
```diff
@@ -99,6 +108,33 @@ struct RestartData
...
+  // mapping via a displacement vector *and* reading it back in. Hence, this parameter needs to
+  // match in serialization/deserialization runs.
+  bool consider_mapping;
+
+  // The `mapping_degree` considered when storing or reading the grid.
+  unsigned int mapping_degree;
+
+  // Parameters for `dealii::RemotePointEvaluation` used for grid-to-grid projection.
+  double rpe_tolerance_unit_cell;
+  bool   rpe_enforce_unique_mapping;
```
nfehn, line 137: From a design perspective, RestartData should contain an object of a class like `GridToGridProjectionData` (to be introduced), which can be extended, e.g. by solver tolerance parameters, etc.

include/exadg/time_integration/restart.h
```diff
@@ -138,6 +155,739 @@ read_write_distributed_vector(VectorType & vector, BoostArchiveType & archive)
...
+    double const abs_tol            = 1e-16 * system_rhs.l2_norm();
+    double constexpr rel_tol        = 1e-12;
+
+    dealii::ReductionControl     reduction_control(max_iter, abs_tol, rel_tol);
+    dealii::SolverCG<VectorType> solver_cg(reduction_control);
+
+    VectorType sol;
+    sol.reinit(system_rhs, false /* omit_zeroing_entries */);
+
+    solver_cg.solve(mass_operator, sol, system_rhs, jacobi_preconditioner);
```
nfehn, line 769: Please use ExaDG data structures for solver tolerances and linear solvers, see also https://github.com/exadg/exadg/pull/729/files#r1980170026. We want to be able to easily generalize the code, potentially set solver tolerances in applications files, etc.. Using ExaDG data structures, this can later be realized with less effort. E.g. solver tolerances could already now be part of `GridToGridProjectionData`. Then, one can set these parameters in applications files. If `RestartData` implements a print function and also `GridToGridProjectionData`, printing all parameters at the beginning of the simulation is straight-forward and simple with this design.

include/exadg/time_integration/restart.h
```diff
@@ -138,6 +155,739 @@ read_write_distributed_vector(VectorType & vector, BoostArchiveType & archive)
...
+
+    *target_vectors[i] = sol;
+
+    if(dealii::Utilities::MPI::this_mpi_process(target_dof_handler.get_communicator()) == 0)
+    {
+      std::cout << "    global projection required " << reduction_control.last_step()
+                << " CG iterations.\n";
+    }
+  }
```
nfehn, line 778: Please use ExaDG utility functions like `print_solver_info_linear()` for this (making `pcout` an argument of this function). See usage e.g. in https://github.com/exadg/exadg/blob/master/include/exadg/convection_diffusion/time_integration/time_int_bdf.cpp

nfehn:
As described in the comments, I think this PR should be split up in two PRs, starting with the topic of grid-to-grid projection and realizing this as utility functions in ExaDG indepdent of restart.

richardschu:
https://github.com/exadg/exadg/pull/729/commits/1bfca2925e55ad962d2a61c92d191fbffdc77481
addresses the changes that are required down the line. These were not obvious at this point, but I fixed some places to already fit incompressible Navier-Stokes, ALE mappings, multi components etc.
The other comments are to be addressed shortly.

include/exadg/time_integration/restart.h
richardschu: good, I will split off the projection introduced here and put it into `include/exadg/operators/solution_projection_between_triangulations.h`
I tried to use the functions in there actually first, but it did not work for Raviart-Thomas for some reason so I opted for the projection and `RemotePointEvaluation`.
For the current use case -- I think it is `FE_Q` and `FE_DGQ` only, grid-to-grid interpolation works. If I grep `SolutionInterpolationBetweenTriangulations`, I get zero hits in ExaDG, so I did not bother with this further although I would say the `ElementType` has to be asserted within `SolutionInterpolationBetweenTriangulations`. I think @jh66637 is using this so I just left it there ;)

include/exadg/time_integration/restart.h
```diff
@@ -138,6 +155,739 @@ read_write_distributed_vector(VectorType & vector, BoostArchiveType & archive)
...
+    {
+      dealii::Point<dim, dealii::VectorizedArray<Number>> const cell_batch_points =
+        fe_eval.quadrature_point(q);
+      for(unsigned int i = 0; i < dealii::VectorizedArray<Number>::size(); ++i)
+      {
+        dealii::Point<dim> p;
+        for(unsigned int d = 0; d < dim; ++d)
+        {
+          p[d] = cell_batch_points[d][i];
+        }
```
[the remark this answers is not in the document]
richardschu: yes we could do this, for me there are atm too many moving parts though.
I will open another PR where I tackle grid to grid projection via free functions such that I can later apply it here and consider the relevant comments directly there!
until then, this PR is on hold, hence on `draft?.

richardschu:
do not forget to remove line 44 in `tests/utilities/grid_to_grid_projection.cpp`
`#include <exadg/operators/solution_projection_between_triangulations.cpp> // temporary`

include/exadg/time_integration/restart.h
jh66637: I am currently not using it anymore, there are more advanced interpolation methods I am relying on, which I want to move to ExaDG when I find the time. I would still like to keep it. It is quite handy to have multiple options how to interpolate between grids :)

richardschu:
`do not merge` because I remove the lines testing the restart approaches against the former one (which is available for identical discretizations & core count only, but is used for verification)

richardschu:
@Niklas would you mind taking another look? the diff was reduced to 700 lines due to #735

richardschu:
@nfehn I dont know if I tagged you wrong; just in case tagging you again :)

richardschu:
I really really want to continue here.
If you want, @nfehn please have a look again; 
otherwise I will merge this after having a final look in the coming days.

include/exadg/operators/inverse_mass_operator.h
```diff
@@ -48,6 +50,30 @@ struct InverseMassOperatorData
...
+  {
+    if(fe.conforms(dealii::FiniteElementData<dim>::L2))
+    {
+      if(element_type == ElementType::Hypercube)
+      {
+        return InverseMassType::MatrixfreeOperator;
+      }
+      else
+      {
+        return InverseMassType::ElementwiseKrylovSolver;
```
kronbichler, line 68: If we work more frequently with simplex meshes, we should do a different approach - elementwise Krylov is likely a good option for deformed mesh elements, but affine meshes, which are the majority of meshes in practice, one should use the inverted mass matrix in reference coordinates and scaling by the Jacobian determinant. Of course, such a strategy is not the purpose of this PR and the alternative option is not available in ExaDG right now, but I thought it is good to mention it.

include/exadg/time_integration/restart.h
```diff
@@ -27,13 +27,35 @@
...
+#include <deal.II/grid/grid_generator.h>
+#include <deal.II/grid/manifold.h>
+#include <deal.II/grid/tria.h>
 #include <deal.II/lac/la_parallel_block_vector.h>
 #include <deal.II/lac/la_parallel_vector.h>
+#include <deal.II/lac/solver_cg.h>
+#include <deal.II/matrix_free/fe_point_evaluation.h>
+#include <deal.II/numerics/vector_tools.h>
+#include <deal.II/particles/data_out.h>
+#include <deal.II/particles/particle_handler.h>
```
kronbichler, line 48: Do you really need these two include files?

kronbichler (approved):
This looks good to me - if you remove the places you marked, then I think this can get merged.

include/exadg/time_integration/restart.h
richardschu: no, that is probably leftover from a previous version of this PR. Thanks for spotting it!

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
```diff
@@ -240,6 +242,151 @@ SpatialOperator<dim, Number>::get_dof_handler_u() const
...
+    create_finite_element<dim>(checkpoint_element_type, true, 1, param.restart_data.degree_p);
+
+  checkpoint_dof_handler_u.distribute_dofs(*checkpoint_fe_u);
+  checkpoint_dof_handler_p.distribute_dofs(*checkpoint_fe_p);
+
+  std::vector<dealii::DoFHandler<dim> const *> checkpoint_dof_handlers(2);
+  checkpoint_dof_handlers[block_index_velocity] = &checkpoint_dof_handler_u;
+  checkpoint_dof_handlers[block_index_pressure] = &checkpoint_dof_handler_p;
+
+  // Deserialize vectors stored in triangulation, sequence matches `this->serialize_vectors()`.
```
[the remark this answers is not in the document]
richardschu: I want to express that the vectors setup in terms of their number of blocks and their sequence (velocity and pressure) etc ahve to match in
`SpatialOperator<dim, Number>::serialize_vectors()`
and
`SpatialOperator<dim, Number>::deserialize_vectors()`
So I would rather leave it.
Do you think this is completely superfluous? Or do you have a suggestion on how to improve it?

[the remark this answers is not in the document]
kronbichler: I get this now, thank you. I would then probably suggest to slightly change the text to
```suggestion
  // Deserialize the stored vectors associated with the previous triangulation / dof handlers,
  // in the sequence if blocks (velocity/pressure) matching the one in `this->serialize_vectors()`.
```

richardschu:
I will not squash commits here since 1) I want to keep the `target_mapping` around somewhere I can remember, 2) same goes for the comparison code, 3) @kronbichler comments would be accredited to me.
I will merge this later and continue with #731 #732 #734

PR #730, opened 2025-02-24
Fix setup order of aeroacoustic application and ensure blend in start time < blend in end time

jh66637:
The setup of the aero-acoustic application was broken. We make use of the acoustic and the fluid application during the setup of the aero-acoustic field functions, so we have to set up acoustic and fluid before the setup of the field functions.

Besides this, I found that there is a typo setting the end time of the blend-in function. I fixed the end time and added asserts to ensure the end time is > the start time.

### DIFF
--- a/applications/aero_acoustic/co_rotating_vortex_pair/application.h
+++ b/applications/aero_acoustic/co_rotating_vortex_pair/application.h
@@ -793,15 +793,6 @@ public:
   {
   }
 
-private:
-  void
-  set_single_field_solvers(std::string input_file, MPI_Comm const & comm) final
-  {
-    this->acoustic =
-      std::make_shared<AcousticsAeroAcoustic::Application<dim, Number>>(input_file, comm);
-    this->fluid = std::make_shared<FluidAeroAcoustic::Application<dim, Number>>(input_file, comm);
-  }
-
   void
   set_field_functions() final
   {
@@ -812,6 +803,15 @@ private:
 
     this->field_functions->analytical_aero_acoustic_source_term =
       std::make_shared<AnalyticalSourceTerm<dim>>(source_term_with_convection, intensity, r_0, r_c);
+  }
+
+private:
+  void
+  set_single_field_solvers(std::string input_file, MPI_Comm const & comm) final
+  {
+    this->acoustic =
+      std::make_shared<AcousticsAeroAcoustic::Application<dim, Number>>(input_file, comm);
+    this->fluid = std::make_shared<FluidAeroAcoustic::Application<dim, Number>>(input_file, comm);
   }
 
   void
--- a/applications/aero_acoustic/co_rotating_vortex_pair/application.h
+++ b/applications/aero_acoustic/co_rotating_vortex_pair/application.h
@@ -658,6 +658,9 @@ public:
   BlendInFunction(double const blend_in_start, double const blend_in_end)
     : Utilities::SpatialAwareFunction<dim>(1, 0.0), start(blend_in_start), end(blend_in_end)
   {
+    AssertThrow(start < end,
+                dealii::ExcMessage(
+                  "End of blend in has to be smaller compared to start of blend in."));
   }
 
   double
@@ -798,8 +801,9 @@ public:
   {
     this->field_functions->source_term_blend_in =
       std::make_shared<BlendInFunction<dim>>(this->acoustic->get_parameters().start_time,
-                                             0.1 * (this->acoustic->get_parameters().end_time -
-                                                    this->acoustic->get_parameters().start_time));
+                                             this->acoustic->get_parameters().start_time +
+                                               0.1 * (this->acoustic->get_parameters().end_time -
+                                                      this->acoustic->get_parameters().start_time));
 
     this->field_functions->analytical_aero_acoustic_source_term =
       std::make_shared<AnalyticalSourceTerm<dim>>(source_term_with_convection, intensity, r_0, r_c);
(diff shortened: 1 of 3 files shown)


jh66637:
Yes, everything works as intended. 

I will add a test tomorrow. Just note that the test results will not look good because it requires some time to approach the analytical solution in addition to high spatial resolution. 

What is the maximum time we allow a single test case to run?

richardschu:
> Yes, everything works as intended.
> 
> I will add a test tomorrow. Just note that the test results will not look good because it requires some time to approach the analytical solution in addition to high spatial resolution.
> 
> What is the maximum time we allow a single test case to run?

Great, thanks for the fix!

please do that, I want to make sure we don't destroy functionality with new PRs.
We run in `release` and with 1 core, we have no strong limit yet, for this particular case: can you produce anything remotely useful (in terms of accuracy of coures :smile:) taking the current tests on your hardware as baseline, by going to 100% of the currently most expensive one? That would be great! If not, we can discuss once we have more concrete numbers!

include/exadg/aero_acoustic/user_interface/application_base.h
```diff
@@ -331,16 +331,17 @@ class ApplicationBase
...
+  virtual void
+  set_field_functions() = 0;
+
   virtual void
   add_parameters(dealii::ParameterHandler & prm)
   {
     parameters.add_parameters(prm, "AeroAcoustic");
-
-    acoustic->add_parameters(prm);
-    fluid->add_parameters(prm);
```
nfehn, line 343: I would expect that this change makes this function https://github.com/exadg/exadg/blob/024b1224410ddd01a12ed924e107ee84838e1d14/include/exadg/aero_acoustic/solver.h#L37 incorrect.

nfehn: it seems as if you only want to change the private member function `parse_parameters()`. If so, one would have to change the implementation of that private member function. For clarity, a comment should be added that this private member function only parses the parameters of the coupling problem, not the parameters of the single-field solvers.

nfehn:
I think I would write the class `AeroAcoustic::ApplicationBase` like this

```C++
class ApplicationBase
{
public:
  ApplicationBase(std::string const & input_file, MPI_Comm const & comm)
    : parameter_file(input_file),
      mpi_comm(comm),
      pcout(std::cout, dealii::Utilities::MPI::this_mpi_process(mpi_comm) == 0)
  {
    set_single_field_solvers(parameter_file, mpi_comm);
  }

  virtual ~ApplicationBase() = default;

  void
  setup()
  {
    parse_parameters();
    parameters.check();
    parameters.print(pcout, "List of parameters for aero-acoustic solver");

    field_functions = std::make_shared<FieldFunctions<dim>>();
    set_field_functions();
  }
```

Then, you can call the `AeroAcoustic::ApplicationBase::setup()` after the setup of the single field problems. The function `Driver::setup()`would then look like this:

```C++
  // setup acoustic solver
  {
    dealii::Timer timer_local;
    acoustic->setup(application->acoustic, mpi_comm, is_test);
    timer_tree.insert({"AeroAcoustic", "Setup", "Acoustic"}, timer_local.wall_time());
  }
  // setup fluid solver
  {
    dealii::Timer timer_local;
    fluid->setup(application->fluid, mpi_comm, is_test);
    timer_tree.insert({"AeroAcoustic", "Setup", "Fluid"}, timer_local.wall_time());
  }

  // setup application
  application->setup();
```

Then, there should be no need to make `set_field_functions()` public. This would be important in my opinion in order to not blurr responsibilities. Choosing this variant, it would become also more clear that the parameters-related part of `ApplicationBase::setup()` only affects the additional coupling parameters, not the single field solvers.

Could you please check whether this variant works correctly?

applications/aero_acoustic/co_rotating_vortex_pair/application.h
```diff
@@ -793,27 +796,28 @@ class Application : public ApplicationBase<dim, Number>
...
+                                               0.1 * (this->acoustic->get_parameters().end_time -
+                                                      this->acoustic->get_parameters().start_time));
 
     this->field_functions->analytical_aero_acoustic_source_term =
       std::make_shared<AnalyticalSourceTerm<dim>>(source_term_with_convection, intensity, r_0, r_c);
   }
 
+private:
+  void
+  set_single_field_solvers(std::string input_file, MPI_Comm const & comm) final
```
nfehn, line 814: This name should be changed. It is not the solvers that are set here, but the applications.

-> `set_single_field_applications()`

nfehn: I think other modules in ExaDG for coupled problems do not have such a function. I think the convention in ExaDG is currently that the constructor of a concrete `Application` has to create instances of concrete single-field applications, see https://github.com/exadg/exadg/blob/024b1224410ddd01a12ed924e107ee84838e1d14/applications/incompressible_flow_with_transport/differentially_heated_cavity/application.h#L504-L512. 

However, I also agree if you prefer such a private member function defined in the base class `ApplicationBase`. In this case, I nevertheless believe we should try to keep the convention that this is done in the constructor of the `Application` class as I have explained in other comments.

jh66637:
I will have another look, I hope it works the way you described it.
@richardschu The timings for the test to produce OKish results are really bad. I have to wait like 10 minutes on my machine to get quite bad results. Do we have tests without analytic solution in other apps so that we can simply compare results instead of immense errors?

jh66637:
@nfehn I adjusted as you suggested :)

nfehn (approved):

PR #731, opened 2025-02-25
Extend deserialization rollout

richardschu:
builds on #729 

-> as of now, the diff will be 1000 lines less than what is shown here.

PR #732, opened 2025-02-26
Extend deserialization rollout implicit

richardschu:
continuation of #729 and #731
This extends the generalized serialization towards **implicit** solvers in the **convection diffusion** module.
I tested with the 

`convection_diffusion/rotating_hill`: implicit time integration now and adaptive mesh refinement

application; with two cores I get:
before serialization:
image

after deserialization, where I use the same manifold, but no adaptive refinement (completely new grid and domain decomposition essentially):

image

end of the simulation:

image

error increases marginally, see:
[left the restarted run, right the solution with adaptively refined grid]
image

This is draft until we figured out how to/if we need to reshape #729

### DIFF
--- a/include/exadg/compressible_navier_stokes/spatial_discretization/operator.cpp
+++ b/include/exadg/compressible_navier_stokes/spatial_discretization/operator.cpp
@@ -28,6 +28,7 @@
 #include <exadg/operators/finite_element.h>
 #include <exadg/operators/grid_related_time_step_restrictions.h>
 #include <exadg/operators/quadrature.h>
+#include <exadg/time_integration/restart.h>
 
 namespace ExaDG
 {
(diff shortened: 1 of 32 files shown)


include/exadg/compressible_navier_stokes/spatial_discretization/operator.cpp
```diff
@@ -339,6 +349,124 @@ Operator<dim, Number>::prescribe_initial_conditions(VectorType & src, double con
...
+  // Load potentially unfitting checkpoint triangulation of TriangulationType.
+  std::shared_ptr<dealii::Triangulation<dim>> checkpoint_triangulation =
+    deserialize_triangulation<dim>(dof_handler.get_triangulation(),
+                                   param.restart_data.filename,
+                                   param.restart_data.triangulation_type,
+                                   mpi_comm);
+
+  // Setup DoFHandlers *as checkpointed*, sequence matches `this->serialize_vectors()`.
+  dealii::DoFHandler<dim> checkpoint_dof_handler(*checkpoint_triangulation);
+  dealii::DoFHandler<dim> checkpoint_dof_handler_mapping(*checkpoint_triangulation);
```
richardschu, line 391: this and related has to be moved behind a `if(consider_mapping)` (down below where it is actually used.)

richardschu:
updated with new version of #729
in commit
https://github.com/exadg/exadg/pull/732/commits/d2b995eababe47f35a1d08aa1bb2d3fed7ada109
manual tests are passing as described above

include/exadg/compressible_navier_stokes/spatial_discretization/operator.cpp
richardschu: considered w
https://github.com/exadg/exadg/pull/732/commits/d2b995eababe47f35a1d08aa1bb2d3fed7ada109

kronbichler:
Can you rebase here as well?

PR #735, opened 2025-03-11
introduce `grid_to_grid_projection()` and related test

richardschu:
This realizes the first part of #729 : grid-to-grid projection in the form of several free functions.
The utility is based on `RemotePointEvaluation`, `FEPointEvaluation` and the `MassOperator`.
The systems are solved in matrix-free fashion, where preconditioners available are: `PointJacobi`, `AdditiveSchwarz`, `AMG`, `InverseMassPreconditioner` and `None`. We can hand over a vector of `DoFHandler`s and source vectors _per_ `DoFHandler`, so that we do not set up multiple `MatrixFree` but only a single one.

The test uses the utility function for `dim=2` or `3`, `ElementType::Simplex` and `ElementType::Hypercube`. The projections (always in both directions _to_ and _from_) between overlapping spaces constructed via `FiniteElement`s considered are:
`FE_Q` <-> `FE_RaviartThomasNodal`
,
`FE_DGQ` <-> `FE_Q`
and
`FE_SimplexP` <-> `FE_SimplexDGP`

The testcase looks like this:
[source grid in white wireframe, target grid in color showing the solution]
image
and we compute the error between:
.) exact function (some `f(x) = 0.5 sin(x)` or similar)
and
.) interpolated function in source grid (`VectorTools::Interpolate`) -> projection to target grid via the proposed `grid_to_grid_projection()`

The 3D Simplex tests are disabled because `RemotePointEvaluation` is taking quite long to find all the points with the current setup. Also, the target grid is a "shrunk" source grid since `RemotePointEvaluation` sometimes did not like the corners (4 individual points on the inner circles). Maybe I am using it wrong or the mapping setup is insufficient idk.

### DIFF
--- a/include/exadg/operators/solution_projection_between_triangulations.cpp
+++ b/include/exadg/operators/solution_projection_between_triangulations.cpp
@@ -1,0 +1,361 @@
+/*  ______________________________________________________________________
+ *
+ *  ExaDG - High-Order Discontinuous Galerkin for the Exa-Scale
+ *
+ *  Copyright (C) 2025 by the ExaDG authors
+ *
+ *  This program is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *  ______________________________________________________________________
+ */
+
+// C/C++
+#include <algorithm>
+
+// deal.II
+#include <deal.II/base/conditional_ostream.h>
+#include <deal.II/base/timer.h>
+#include <deal.II/matrix_free/fe_point_evaluation.h>
+#include <deal.II/numerics/vector_tools.h>
+
+// ExaDG
+#include <exadg/matrix_free/matrix_free_data.h>
+#include <exadg/operators/inverse_mass_operator.h>
+#include <exadg/operators/mapping_flags.h>
+#include <exadg/operators/quadrature.h>
+#include <exadg/operators/solution_projection_between_triangulations.h>
+#include <exadg/postprocessor/write_output.h>
+
+namespace ExaDG
+{
+namespace GridToGridProjection
+{
+template<int dim, int n_components, typename Number>
+std::vector<dealii::Point<dim>>
+collect_integration_points(
+  dealii::MatrixFree<dim, Number, dealii::VectorizedArray<Number>> const & matrix_free,
+  unsigned int const                                                       dof_index,
+  unsigned int const                                                       quad_index)
+{
+  CellIntegrator<dim, n_components, Number> fe_eval(matrix_free, dof_index, quad_index);
+
+  // Conservative estimate for the number of points.
+  std::vector<dealii::Point<dim>> integration_points;
+  integration_points.reserve(
+    matrix_free.get_dof_handler(dof_index).get_triangulation().n_active_cells() *
+    fe_eval.n_q_points);
+
+  for(unsigned int cell_batch_idx = 0; cell_batch_idx < matrix_free.n_cell_batches();
+      ++cell_batch_idx)
+  {
+    fe_eval.reinit(cell_batch_idx);
+    for(const unsigned int q : fe_eval.quadrature_point_indices())
+    {
+      dealii::Point<dim, dealii::VectorizedArray<Number>> const cell_batch_points =
+        fe_eval.quadrature_point(q);
+      for(unsigned int i = 0; i < dealii::VectorizedArray<Number>::size(); ++i)
+      {
+        dealii::Point<dim> p;
+        for(unsigned int d = 0; d < dim; ++d)
+        {
+          p[d] = cell_batch_points[d][i];
+        }
+        integration_points.push_back(p);
+      }
+    }
+  }
+
+  return integration_points;
+}
+
+template<int dim, int n_components, typename Number, typename VectorType>
+VectorType
+assemble_projection_rhs(
+  dealii::MatrixFree<dim, Number, dealii::VectorizedArray<Number>> const & matrix_free,
+  std::vector<
+    typename dealii::FEPointEvaluation<n_components, dim, dim, Number>::value_type> const &
+                     values_source_in_q_points_target,
+  unsigned int const dof_index,
+  unsigned int const quad_index)
+{
+  VectorType system_rhs;
+  matrix_free.initialize_dof_vector(system_rhs, dof_index);
+
+  CellIntegrator<dim, n_components, Number> fe_eval(matrix_free, dof_index, quad_index);
+
+  unsigned int idx_q_point = 0;
+
+  for(unsigned int cell_batch_idx = 0; cell_batch_idx < matrix_free.n_cell_batches();
+      ++cell_batch_idx)
+  {
+    fe_eval.reinit(cell_batch_idx);
+    for(unsigned int const q : fe_eval.quadrature_point_indices())
+    {
+      dealii::Tensor<1, n_components, dealii::VectorizedArray<Number>> tmp;
+
+      for(unsigned int i = 0; i < dealii::VectorizedArray<Number>::size(); ++i)
+      {
+        typename dealii::FEPointEvaluation<n_components, dim, dim, Number>::value_type const
+          values = values_source_in_q_points_target[idx_q_point];
+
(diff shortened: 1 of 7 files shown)


tests/utilities/grid_to_grid_projection.cc
```diff
@@ -0,0 +1,657 @@
...
+#include <deal.II/numerics/vector_tools.h>
+
+// ExaDG
+#include <exadg/grid/grid_data.h>
+#include <exadg/operators/finite_element.h>
+#include <exadg/operators/solution_projection_between_triangulations.h>
+#include <exadg/postprocessor/error_calculation.h>
+#include <exadg/postprocessor/write_output.h>
+#include <exadg/utilities/numbers.h>
+#include <exadg/operators/solution_projection_between_triangulations.cpp> // temporary
```
nfehn, line 44: I think you need to add this file in the CMakeLists file.

include/exadg/solvers_and_preconditioners/solvers/iterative_solvers_dealii_wrapper.h
```diff
@@ -99,6 +99,16 @@ struct SolverDataCG
...
+  void
+  print(dealii::ConditionalOStream const & pcout) const
+  {
+    print_parameter(pcout, "Maximum number of iterations", max_iter);
+    print_parameter(pcout, "Absolut solver tolerance", solver_tolerance_abs);
+    print_parameter(pcout, "Relative solver tolerance", solver_tolerance_rel);
+    print_parameter(pcout, "Use preconditioner", use_preconditioner);
+    print_parameter(pcout, "Compute performance metrics", compute_performance_metrics);
+  }
+
```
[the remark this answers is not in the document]
nfehn: These solvers in namespace `Krylov` are meant as "internal" solvers. In many case, we want to be able to switch between different Krylov solvers, e.g. CG vs GMRES. Then, using `SolverDataCG` in applications does not work and would also require `SolverDataGMRES`, etc.

So please use https://github.com/exadg/exadg/blob/master/include/exadg/solvers_and_preconditioners/solvers/solver_data.h as in other applications in ExaDG. We could rethink this, but let's not mix again this PR with other topics.

include/exadg/operators/solution_projection_between_triangulations.cpp
```diff
@@ -0,0 +1,414 @@
...
+    write_points_in_dummy_triangulation(
+      points_not_found, "./", "points_not_found", 0, source_dof_handler.get_communicator());
+
+    AssertThrow(rpe_source.all_points_found(),
+                dealii::ExcMessage(
+                  "Could not interpolate source grid vector in target grid. "
+                  "Points exported to `./all_points.pvtu` and `./points_not_found.pvtu`"));
+  }
+
+  CellIntegrator<dim, n_components, Number> fe_eval(target_matrix_free, dof_index, quad_index);
```
nfehn, line 259: This object should be created in `assemble_projection_rhs()`, just as in `collect_integration_points()`, in my opinion.

include/exadg/operators/solution_projection_between_triangulations.cpp
```diff
@@ -0,0 +1,414 @@
...
+
+template<int dim, int n_components, typename Number, typename VectorType>
+VectorType
+assemble_projection_rhs(
+  dealii::MatrixFree<dim, Number, dealii::VectorizedArray<Number>> const & matrix_free,
+  CellIntegrator<dim, n_components, Number> &                              fe_eval,
+  std::vector<
+    typename dealii::FEPointEvaluation<n_components, dim, dim, Number>::value_type> const &
+                     values_source_in_q_points_target,
+  unsigned int const dof_index)
```
nfehn, line 96: What if `dof_index` does not fit to `fe_eval`? -> Create `CellIntegrator` object inside this function.

tests/utilities/grid_to_grid_projection.cc
nfehn: ah, you did ... but the function in that file have many template parameters.

So I guess a .cpp file does not make sense in this form (without instantiations for all possible values of the template parameters). I suggest to put everything into the .h file.

nfehn:
I think the most fundamental question for this PR is right now why you do not make use of (or extend) the inverse mass operator infrastructure we have currently, https://github.com/exadg/exadg/blob/master/include/exadg/operators/inverse_mass_operator.h.

include/exadg/operators/solution_projection_between_triangulations.cpp
```diff
@@ -0,0 +1,414 @@
...
+    fe_eval.distribute_local_to_global(system_rhs);
+  }
+  system_rhs.compress(dealii::VectorOperation::add);
+
+  return system_rhs;
+}
+
+template<int dim, typename Number, int n_components, typename VectorType>
+void
+project_vectors(
```
nfehn, line 145: I think for a modular design, this function should only compute the rhs-vector, and then make use of `InverseMassOperator` infrastructure in ExaDG.

include/exadg/operators/solution_projection_between_triangulations.cpp
```diff
@@ -0,0 +1,414 @@
...
+                                        source_dof_handler.get_communicator()) == 0));
+
+    timer.restart();
+
+    unsigned int n_iter = solver.solve(sol, system_rhs);
+
+    if(not data.is_test)
+    {
+      print_solver_info_linear(pcout, n_iter, timer.wall_time());
+    }
```
nfehn, line 293: -> `inverse_mass_operator.apply(dst, src)`

-> generic ExaDG infrastructure using the most efficient variant (e.g. explicit matrix-free inverse mass for DG + hypercube, elementwise inverse mass for DG + simplex, and Krylov solver for global problem otherwise (continuous Galerkin, RT))

richardschu:
> I think the most fundamental question for this PR is right now why you do not make use of (or extend) the inverse mass operator infrastructure we have currently, https://github.com/exadg/exadg/blob/master/include/exadg/operators/inverse_mass_operator.h.

#746 -> I think we should unify the `InverseMassOperator` first, then I would add the variable coefficients as this is less intrusive potentially and then we use the `InverseMassOperator` on a specially constructed `rhs(u_to_project)`.

tests/utilities/grid_to_grid_projection.cc
richardschu: I would rather keep the separation in `.h` and `.cpp`.
Currently, we need the `.cpp` to be included like this, but we can remove it immediately after, when focusing on the restart procedures.

include/exadg/operators/solution_projection_between_triangulations.cpp
richardschu: moved

include/exadg/operators/solution_projection_between_triangulations.cpp
richardschu: yes, makes sense!
I can soon push a new version.
thanks for that hint :)

include/exadg/operators/solution_projection_between_triangulations.cpp
richardschu: the reason I did not do this before was that I did not get that the `InverseMassOperatorHdiv` was essentially the most general case I needed. Now after the unification of the `InverseMassOperator`, I think we can mvoe forward with this in a nice way!

include/exadg/operators/solution_projection_between_triangulations.cpp
richardschu: moved

richardschu:
@nfehn please have a look again, I think I addressed your comments :+1:

include/exadg/operators/solution_projection_between_triangulations.h
```diff
@@ -0,0 +1,134 @@
...
+template<int dim, typename VectorType>
+void
+grid_to_grid_projection(
+  std::vector<std::vector<VectorType *>> const &       source_vectors_per_dof_handler,
+  std::vector<dealii::DoFHandler<dim> const *> const & source_dof_handlers,
+  std::shared_ptr<dealii::Mapping<dim> const> const &  source_mapping,
+  std::vector<std::vector<VectorType *>> &             target_vectors_per_dof_handler,
+  std::vector<dealii::DoFHandler<dim> const *> const & target_dof_handlers,
+  std::shared_ptr<dealii::Mapping<dim> const> const &  target_mapping,
+  GridToGridProjectionData<dim> const &                data);
```
kronbichler, line 129: Besides these general functions, shouldn't we also try to provide a function that can use an existing `MatrixFree` object for the mass matrix operator? The reason is that build a `MatrixFree::reinit()` object is expensive compared to a mass matrix solve in the usual case, for DG elements with cell-wise inverse mass matrices the cost difference is even larger.

richardschu: yes, good point! :+1: 
I will add another function that wraps the existing `grid_to_grid_projection()` and passes a `MatrixFree` object.

kronbichler (approved):
Looks good, thank you!

richardschu:
@nfehn any further comments?
I would like to continue with #729 and related projects :rocket:

kronbichler:
The review comments were addressed as far as I can tell, follow-up comments, if they were to arise, can be tackled in follow-up PRs.

PR #736, opened 2025-03-12
fix some application parameter copy error and some include guard

richardschu:
... improve a tiny bit at a time :)


richardschu:
I am directly merging since this is just some typos essentially and/or in the application files only.

PR #742, opened 2025-05-26
variable coefficients for InverseMassOperator

richardschu:
**note** only really considers the `InverseMassType::MatrixfreeOperator` since the `MassOperator` only features variable coefficients after #741 .
I wanted to make the smallest changes possible.
we can address then enabling variable coefficients also for the other `InverMassType`s in follow-up PRs.

### DIFF
--- a/include/exadg/operators/inverse_mass_operator.h
+++ b/include/exadg/operators/inverse_mass_operator.h
@@ -30,14 +30,21 @@
 #include <exadg/matrix_free/integrators.h>
 #include <exadg/operators/inverse_mass_parameters.h>
 #include <exadg/operators/mass_operator.h>
+#include <exadg/operators/variable_coefficients.h>
 #include <exadg/solvers_and_preconditioners/preconditioners/block_jacobi_preconditioner.h>
 #include <exadg/solvers_and_preconditioners/preconditioners/jacobi_preconditioner.h>
 
 namespace ExaDG
 {
+template<typename Number>
 struct InverseMassOperatorData
 {
-  InverseMassOperatorData() : dof_index(0), quad_index(0)
+  InverseMassOperatorData()
+    : dof_index(0),
+      quad_index(0),
+      coefficient_is_variable(false),
+      consider_inverse_coefficient(false),
+      variable_coefficients(nullptr)
   {
   }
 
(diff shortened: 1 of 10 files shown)


include/exadg/operators/inverse_mass_operator.h
```diff
@@ -46,6 +49,10 @@ struct InverseMassOperatorData
   unsigned int quad_index;
 
   InverseMassParameters parameters;
+
+  // Variable parameter.
+  bool                                                          coefficient_is_variable;
+  VariableCoefficients<dealii::VectorizedArray<Number>> const * variable_coefficients;
```
kronbichler, line 55: This stores a scalar per quadrature point; in combination with the mass operator we keep two separate arrays that need to be kept in sync. I would have implemented the operation by using the data of the mass operator (through some access functionality), and instead apply the inverse operation here by a division rather than multiplication in line 264-265. Of course, a division incurs some cost, especially when running from caches, but I would introduce that option as soon as measurements indicate we need it. Or what application settings do you envision, and would it ever be problematic to fill both separately (as far as I see, that code is not there yet)?

richardschu: I tried to make it more similar to `ViscousOperator` now, but the difference is that the `ViscousOperator` holds `ViscousOperatorData` and a shared_ptr to the `ViscousKernel`, while
the mass operator owns the Masskernel and the MassOperatorData now.

I suggest we discuss the new commit (incoming) and then check if we have to refine the design later on, once all the parts are in place!

richardschu: yes thanks, makes sense!
the code here actually uses the coefficient^-1, so is tailored to assembling
(p, 1/coefficient p)
which is not general at all, but from the present PR this is not clear.
We will name the coefficients here `variable_coefficients` and add a comment saying that one gets the action of 
(p, coefficient p)^-1
Then we add getters to the `mass_kernel`/`mass_operator` and only need to update one single place!
Then we probably add some parameter `invert_coefficient` to signal if we want `coefficient` or `1/coefficient`.

I will look into this.

richardschu:
this is ready to go, but I would like to merge #749 first. then we can directly extend the idea here in a follow-up for all `InverMassType` using the `MassOperator` with variable coefficients.

include/exadg/operators/inverse_mass_operator.h
```diff
@@ -315,14 +346,59 @@ class InverseMassOperator
...
+      {
+        integrator.reinit(cell);
+        integrator.read_dof_values(src, 0);
 
-      inverse_mass.apply(integrator.begin_dof_values(), integrator.begin_dof_values());
+        dealii::AlignedVector<dealii::VectorizedArray<Number>> inverse_JxW_times_coefficient(
+          this->matrix_free->get_dofs_per_cell(dof_index));
+        inverse_mass.fill_inverse_JxW_values(inverse_JxW_times_coefficient);
+
+        if(consider_inverse_coefficient)
```
richardschu, line 360: We could move that `if(consider_inverse_coefficient)` out of the cell loop by duplicating a few lines of code.
Since the parameter `consider_inverse_coefficient` is typically constant per object (or at least per call of `cell_loop_matrix_free_operator`), I think the branch prediction is probably handling this well enough.

PR #749, opened 2025-07-03
Unify inverse mass operators

### LINKED ISSUE #746: `InverseMassOperator` could encapsulate `InverseMassOperatorHdiv`?
Looking at the `InverseMassOperator`, in particular the implementatios depending on the `InverseMassType` I was wondering what the specific motivations were for splitting off the version for Hdiv conforming elements?
I would like to unify them, essentially introducing a `InverseMassType::Global`, which would then solve a global problem.
For Raviart-Thomas elements, this would then be the only allowed type.
But since this is so obvious, I was wondering where the catch is? I do not see it yet! Can someone comment?
Otherwise I will just try to unify tomorrow and if it works out, we might consider that potentially simpler variant.

This would make extending #742 easier, which comes before tackling #735 as they all work on the `MassOperator` and `InverseMassOperator`

richardschu:
fixes #746 
where "fixes" is a strong word ...

I still prefer this variant to not have `InverseMassOperator` and `InverseMassOperatorHdiv`, where the latter might be seen as the general case (solving the global mass operator problem iteratively) covering the actual algorithms included in the former (but is probably slower). 
The present variant checks if the requested `InverseMassType` can be used for the discretization and asserts the rest.

One can optionally provide constraints, which are not needed for a mass operator in general, but can be provided to enforce harde constraints, e.g., for adaptive-CG or Raviart-Thomas discretizations.

additionally:
.) the `InverseMassOperatorHdiv` is now only set up when actually needed
.) the new `InverseMassOperator` also has the option to use a block-jacobi preconditioner: `PreconditionerMass::BlockJacobi` (before we had `None` or `PointJacobi`)

### DIFF
--- a/include/exadg/operators/inverse_mass_operator.h
+++ b/include/exadg/operators/inverse_mass_operator.h
@@ -70,8 +70,9 @@ public:
   }
 
   void
-  initialize(dealii::MatrixFree<dim, Number> const & matrix_free_in,
-             InverseMassOperatorData const           inverse_mass_operator_data)
+  initialize(dealii::MatrixFree<dim, Number> const &   matrix_free_in,
+             InverseMassOperatorData const             inverse_mass_operator_data,
+             dealii::AffineConstraints<Number> const * constraints = nullptr)
   {
     this->matrix_free = &matrix_free_in;
     dof_index         = inverse_mass_operator_data.dof_index;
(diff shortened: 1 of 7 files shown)


richardschu:
I am merging this now to continue #742
we can split them up later if we think its better than what we have now!

include/exadg/operators/inverse_mass_operator.h
```diff
@@ -175,15 +237,39 @@ class InverseMassOperator
   void
   apply(VectorType & dst, VectorType const & src) const
   {
-    dst.zero_out_ghost_values();
-
-    if(data.implementation_type == InverseMassType::MatrixfreeOperator)
+    if(data.implementation_type == InverseMassType::GlobalKrylovSolver)
     {
-      matrix_free->cell_loop(&This::cell_loop_matrix_free_operator, this, dst, src);
```
richardschu, line 182: note how here we did this also for `&src == &dst` which was not save for the `ElementwiseKrylov`

include/exadg/operators/inverse_mass_operator.h
```diff
@@ -246,122 +332,21 @@ class InverseMassOperator
...
-  apply(VectorType & dst, VectorType const & src) const
-  {
-    Assert(solver.get() != 0, dealii::ExcMessage("Mass solver has not been initialized."));
-
-    VectorType temp;
-
-    // Note that the inverse mass operator might be called like inverse_mass.apply(dst, dst),
-    // i.e. with identical destination and source vectors. In this case, we need to make sure
-    // that the result is still correct.
-    if(&dst == &src)
```
richardschu, line 342: here we explicitly considered the case `&src == &dst`, but in my tests this is not even needed?

PR #751, opened 2025-07-04
add workflow to check include guards

### LINKED ISSUE #543: #pragma once vs. include guards
In PR #542 , a discussion regarding include tokens came up. Do we go `#pragma once`? I think it is simpler than the include guards we are currently having!
Any thoughts on this?

richardschu:
fixes #543 
(or leads to a state where we check the include guards automatically):
Make the include guards uniform and introduce a github workflow to check them just like the indentation
I only set this up for `include/exadg/` so we skip the project name that is often used.


.github/workflows/include_guards.yml
```diff
@@ -0,0 +1,19 @@
...
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
+    - uses: sbeyer/include-guards-check-action@v2.0.0
+      with:
+        path: 'include/exadg/'
+        pattern: 'INCLUDE_EXADG_{dirs}_{file}_'
+
```
richardschu, line 19: this defines the format of the include guard, which I deduced from the used definitions prior to this PR.
The `_` in the end was used in most cases, hence I considered it for the standard.

include/exadg/aero_acoustic/calculators/source_term_calculator.h
```diff
@@ -1,5 +1,26 @@
...
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *  ______________________________________________________________________
+ */
+
```
richardschu, line 21: here the copyright was missing, put it with 2025 as I do not want to back-date

include/exadg/functions_and_boundary_conditions/linear_interpolation.h
```diff
@@ -19,6 +19,9 @@
  *  ______________________________________________________________________
  */
```
richardschu, line 21: here the include guards were missing completely

richardschu:
merging since there is no fundamental change (include guards or pragma once as mentioned in #543 ) and format is essentially kept unchanged, but standardized in all headers now.

PR #752, opened 2025-07-05
test `MassOperator` and `InverseMassOperator` for variable coefficients

richardschu:
[DO NOT MERGE]

this tests the `MAssOperator` and the `InverseMassOperator` with variable coefficients using `Simplex` and `Hypercube` elements with continuous and discontinuous spaces incl. `FE_RaviartThomasNodal`.
The idea is to fill a vector 
`vec[i] = rand()/RAND_MAX`
and then perform
`vec = M^-1 * M * vec`
and compare to the initial random vector.

Why DO NOT MERGE:
tba


include/exadg/operators/inverse_mass_operator.h
```diff
@@ -266,38 +266,41 @@ class InverseMassOperator
   void
   apply(VectorType & dst, VectorType const & src) const
   {
+    // Note that the inverse mass operator might be called like inverse_mass.apply(dst, dst),
```
richardschu, line 269: I moved this out to have it for both variants.
One of the original `InverseOperatorMassHdiv` or `InverseOperatorMass` before the merge in #749 had this.
For safety, allow it for the merge one as well.
The test below actually checks the case `&dst == &src` and the regular one.

tests/operators/inverse_mass_variable_coefficient.cc
```diff
@@ -0,0 +1,549 @@
...
+    vector.local_element(i) = static_cast<Number>(std::rand()) / RAND_MAX;
+  }
+  VectorType const reference(vector);
+
+  pcout << "\n"
+        << "  ||ref||_infty = " << reference.linfty_norm() << "\n";
+
+  // Multiply vector by `MassOperator` and `InverseMassOperator`:
+  // M^-1 * M * v = I * v = v
+  {
```
richardschu, line 350: this is the important bit here

tests/operators/inverse_mass_variable_coefficient.output
```diff
@@ -0,0 +1,306 @@
...
+
+  ||ref||_infty = 0.998761
+  &src == &dst in InverseMassOperator: || vec1 - vec2||_infty = 0
+  consider_inverse_coefficient = 0   : || vec  - ref ||_infty = 1.02002e-15
+
+  ||ref||_infty = 0.968522
+  &src == &dst in InverseMassOperator: || vec1 - vec2||_infty = 0
+  consider_inverse_coefficient = 1   : || vec  - ref ||_infty = 1.44329e-15
+
+
```
richardschu, line 85: Here is the problem I found:
for the `MatrixFreeOperator`, using more than on component leads to a wrong result.
One component (which we also used for the scalar pressure mass operator with variable coefficient in the preconditioner) _does work_!

richardschu:
I am merging because this test can be added without resolving the issue #754 identified since the problem was there already before, we just did not know.

tests/operators/inverse_mass_variable_coefficient.output
richardschu: yes I think this is easier :) thank you for your comments

include/exadg/operators/inverse_mass_operator.h
richardschu: I first did not do it and found that the test as done below was failing.
Then I moved it and saw that for the `InverseMassType::MatrixfreeOperator` it was also needed for the case `$src==$dst`. But maybe this is just a side-effect of the other problem we have, I will test this again before we have a chat.

PR #756, opened 2025-07-07
fix InverseMass, -test, and ElemwiseKrylov for &src=&dst

### LINKED ISSUE #754: `InverseMassType::MatrixfreeOperator` fails for variable coefficients and `n_components>1`
This was identifed in the test added in #752.
To my current understanding, this might be an issue within deal.II as for constant coefficients perfect results are obtained.
There are two options:

**1)** 
The use in `InverseMassOperator` is correct: 
https://github.com/exadg/exadg/blob/9de8955c497248f2599ae7c1ecb4a28c8a9593d8/include/exadg/operators/inverse_mass_operator.h#L347-L388
and there is a bug in deal.II

**2)** 
We use the `dealii::MatrixFreeOperators::CellwiseInverseMassMatrix<dim, -1, n_components, Number>`
incorrectly, but our specific combination of parameters is not caught by the exceptions within
https://github.com/dealii/dealii/blob/0c9caf4a73ad25f301f4805706cc2369912b2647/include/deal.II/matrix_free/evaluation_kernels.h#L2395-L2556

richardschu:
fixes #754 
by actually setting all the coefficients, not only the ones for the first component :face_exhaling: 

.) adds the `InverseMassType::BlockMatrices` to the inverse mass test

then i tested again the case `&src == &dst`:

.) removes the copying of the vector for the case `&src=&dst` in `InverseMassOperator::apply()`

-> this needed a fix in `Elementwise::IterativeSolver`, which set also `dst = 0;` for the case when `&dst==&src`. Now, we set the initial condition on the element-local vector, since then we set `integrator.set_dof_values(dst, 0);` anyways ... so I think we did not even set the initial guess with that line before (`AlignedVector` calls default initializer on `VectorizedArray`, so the elements are uninitialized).

### DIFF
--- a/include/exadg/solvers_and_preconditioners/solvers/wrapper_elementwise_solvers.h
+++ b/include/exadg/solvers_and_preconditioners/solvers/wrapper_elementwise_solvers.h
@@ -90,8 +90,6 @@ public:
   unsigned int
   solve(VectorType & dst, VectorType const & src) const override
   {
-    dst = 0;
-
     op.get_matrix_free().cell_loop(&THIS::solve_elementwise, this, dst, src);
 
     return 0;
@@ -110,7 +108,8 @@ private:
 
     unsigned int const dofs_per_cell = integrator.dofs_per_cell;
 
-    dealii::AlignedVector<dealii::VectorizedArray<Number>> solution(dofs_per_cell);
+    dealii::AlignedVector<dealii::VectorizedArray<Number>> solution(
+      dofs_per_cell, dealii::make_vectorized_array<Number>(0.0));
 
     // setup elementwise solver
     if(iterative_solver_data.solver_type == Solver::CG)
--- a/tests/operators/inverse_mass_variable_coefficient.output
+++ b/tests/operators/inverse_mass_variable_coefficient.output
@@ -4,49 +4,49 @@
   is_dg            = 1
   RT_on_hypercube  = 0
   InverseMassType  = ElementwiseKrylovSolver
-  Number of degrees of freedom: 384
+  Number of degrees of freedom: 192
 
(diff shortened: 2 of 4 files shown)


include/exadg/solvers_and_preconditioners/solvers/wrapper_elementwise_solvers.h
```diff
@@ -90,8 +90,6 @@ class IterativeSolver
   unsigned int
   solve(VectorType & dst, VectorType const & src) const override
   {
-    dst = 0;
-
     op.get_matrix_free().cell_loop(&THIS::solve_elementwise, this, dst, src);
```
richardschu, line 93: `dst` is only used in line 148 in the new code; the `dst=0` did not even have an effect I think.

tests/operators/inverse_mass_variable_coefficient.output
```diff
@@ -66,6 +66,74 @@
   consider_inverse_coefficient = 1   : || vec  - ref ||_infty = 4.85723e-15
 
 
+  element_type     = Simplex
```
richardschu, line 69: note how all combinations now give good results for any number of components and `&dst == &src`

richardschu:
please note how for the `ElementwiseKrylov` appraoch, this was correct for `&src == &dst`, see
https://github.com/exadg/exadg/pull/749#discussion_r2190556973
and for the `InverseMassOperatorHdiv` (global system solve), we had
https://github.com/exadg/exadg/pull/749#discussion_r2190556973
but this is not needed according to the present test, could that be correct @kronbichler ?
I don't expect different behavior doing this multithreaded?

PR #759, opened 2025-07-17
add vector plot utility function

richardschu:
This replaces a part of #735 to separate concerns
It is only a tiny utility function that allows one to print a vector to `.vtu` in higher-order fashion without going through the ~80 lines.
This is particularly handy when debugging.


richardschu:
adding this to continue with #735
I do not expect anyone being against this addition, but we can remove it afterwards and I just delete the plotting functionality in #735

PR #762, opened 2025-07-29
remove .cpp file, move to header

### LINKED ISSUE #761: remove temporary `#include [].cpp`
this is only relevant after #735 is merged, and should be fixed when merging #729 and serves as a reminder for myself.

richardschu:
fixes #761 
follow-up to #735 
the explicit instantiations of the templated free functions necessary were just too ugly since the signatures are too complex and we have 3 template parameters. Hence I removed the `.cpp` file and moved the implementation to the header.


richardschu:
realizes https://github.com/exadg/exadg/pull/735#discussion_r2014998152

kronbichler (approved):

PR #766, opened 2025-08-05
Update docker to noble

bergbauer:
The deal.II `focal` Docker container has been discontinued on 31 January 2025. This PR suggests updating to `noble` directly and skipping `jammy`.

We can also use 4 threads for building the Docker container since GitHub has upgraded its Linux runners.

One complication is the newer deal.II Docker containers have `magic enum` installed and activated inside the `Patterns` machinery. I suggest working around that if we detect `DEAL_II_WITH_MAGIC_ENUM`.

Lastly, deprecation warnings in the `prototypes` folder are fixed.


richardschu:
thanks @bergbauer !
It seems to me the test itself is missing, we only have the output here, could you add it as well?

bergbauer:
> It seems to me the test itself is missing, we only have the output here, could you add it as well?

This PR adds alternative output to an existing test https://github.com/exadg/exadg/blob/master/tests/utilities/enum_parameter_handler.cc

We have to do that since we use `Patterns::Selection` while deal.II uses `Patterns::List` for enum.

richardschu:
> > It seems to me the test itself is missing, we only have the output here, could you add it as well?
> 
> This PR adds alternative output to an existing test https://github.com/exadg/exadg/blob/master/tests/utilities/enum_parameter_handler.cc
> 
> We have to do that since we use `Patterns::Selection` while deal.II uses `Patterns::List` for enum.

aaah thanks, i forgot about that test!

bergbauer:
> ... this would really only run, when the configuration would fit, right?

It runs on both configurations but passes if one of the two (where one is the new option) provided outputs is correct, see https://dealii.org/developer/developers/testsuite.html#variants.

kronbichler (approved):
Very nice! Thank you @bergbauer!

PR #768, opened 2025-08-23
update output for 2D simplices, lower error -> no other changes

richardschu:
I think this is from
https://github.com/dealii/dealii/pull/18832
note how the errors went down tremendously from
5e-3 ----> 1.6e-6
Maybe be coincidence only, I did not look into that.
Since the error between the interpolation to the exact solution reduces, no other changes in ExaDG are needed.

### DIFF
--- a/tests/utilities/grid_to_grid_projection.mpirun=4.output
+++ b/tests/utilities/grid_to_grid_projection.mpirun=4.output
@@ -10,13 +10,13 @@
             4800 (continuous)
 
 Calculate error for all fields for initial data:
-  Relative error (L2-norm): 5.34535e-03
+  Relative error (L2-norm): 1.40436e-06
 
 Calculate error for all fields for initial data:
-  Relative error (L2-norm): 5.34535e-03
+  Relative error (L2-norm): 1.40436e-06
 
 Calculate error for all fields for initial data:
-  Relative error (L2-norm): 5.34535e-03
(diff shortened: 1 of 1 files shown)


richardschu:
@drwells we are not doing convergence rates in our tests, only single errors.
Since this is the only test that was affected shows me that we do not really cover the simplex case with any of our tests.
One could run simplex cases via `exadg/applications/poisson/sine` or `exadg/applications/structure/bar` and replacing the `ElemntType::Hypercube` with `ElementType::Symplex` or similar.
 
If the error was O(h) before, we would have very likely noticed if we used that in a convergence study, so I think we never did a study rigorously enough. The O(h) error might still not have dominated as the leading constant might have been much lower as some other source of error ... unlikely though ... I think we just did not use it!

kronbichler:
I think we did some accuracy tests, but all tests I did were in 3d, where the problem did not appear. But yes, it would be good to have a basic convergence test.

richardschu:
> I think we did some accuracy tests, but all tests I did were in 3d, where the problem did not appear. But yes, it would be good to have a basic convergence test.

Sure, I meant automated convergence tests used in a test; potentially based on something like #368 
I think the rationale behind the decision not doing it at the moment might have been:
do convergence tests once we a new feature/formulation or somethingg similar is introduced _manually_, 
then make a test that documents the expected error (and hope this is then equivalent, which in most cases it would)

PR #778, opened 2025-10-02
Fix issue 770

richardschu:
this is hopefully the last refactor of the restart utilities, we now
.) store some parameters in an archive alongside the triangulation to get some basic info of the stored data
.) split up the `consider_mapping` parameter bcs we might want different things:
a) store an unmapped grid
b) use a mapped version of the grid used at restart (might be unstable, since we resort on RPE and points might lie outside the domain due to rounding)
c) use an unmapped version of the grid used at restart (this is then the more stable alternative to b) --> use both grids unmapped and it should then work much more likely if the linear representations match)


richardschu:
splitting the parameters and testing again I found some error:
If we store the unmapped grid and use an unmapped version of the current grid
```
    this->param.restart_data.discretization_identical                = false; // forces grid to grid projection
    this->param.restart_data.consider_mapping_write             = false; // write the mapping as a DoF vector when serializing 
    this->param.restart_data.consider_mapping_read_source = false; // ignore the mapping if there were any serialized
    this->param.restart_data.consider_mapping_read_target  = false; // force use of linear mapping of current grid
```
this should be the most stable version which we want to include.
However, not all points in the `target` grid (the current grid we are restarting with) can be found:
image
where it is clearly visible that the stored grid is linear, has no mapping, while ignoring the mapping used at restart is not possible.
the `bool consider_mapping_read_target` has not effect.
This I do not understand, I marked the places where we go with `##+`.
This is then even setting up an entirely new `MatrixFree` object in `grid_to_grid_projection()`.

@kronbichler can you have a look at these places? I am either missing something or the mapping is ignored ...?

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
```diff
@@ -247,14 +247,28 @@ void
...
+  // Attach vectors to triangulation and serialize.
   std::vector<dealii::DoFHandler<dim> const *> dof_handlers(2);
   dof_handlers.at(block_index_velocity) = &this->get_dof_handler_u();
   dof_handlers.at(block_index_pressure) = &this->get_dof_handler_p();
 
   std::vector<std::vector<VectorType const *>> vectors_per_dof_handler =
     get_vectors_per_block<VectorType const, BlockVectorType const>(block_vectors);
 
-  if(param.restart_data.consider_mapping)
+  if(param.restart_data.consider_mapping_write)
```
richardschu, line 271: switch if the mapping should be attached to the triangulation as a displacement vector or not.

include/exadg/acoustic_conservation_equations/spatial_discretization/spatial_operator.cpp
```diff
@@ -374,17 +396,43 @@ SpatialOperator<dim, Number>::deserialize_vectors(
...
 
-    ExaDG::GridToGridProjection::do_grid_to_grid_projection<dim, Number, VectorType>(
-      checkpoint_vectors,
-      checkpoint_dof_handlers,
-      checkpoint_mapping,
-      vectors_per_dof_handler,
-      dof_handlers,
-      *matrix_free,
-      data);
+    if(param.restart_data.consider_mapping_read_target)
```
richardschu, line 403: this is the important bit:
if `consider_mapping_read_target == true`
we reuse the matrix_free object, which has the mapping setup already.
if not, we use ~do_~`(!)`grid_to_grid_projection()`, which sets up an entire matrix_free object.
I tried also with
```
matrix_free->update_mapping(*target_mapping)
[grid-to-grid interpolation]
matrix_free->update_mapping(this->get_mapping())
```
but the matrix_free object is `const` in the present context.

richardschu:
ok I guess it can only be the manifolds at this point.
will try to stash them and re-apply after doing the projection.

richardschu:
ok this is the right direction.
I need to stash also the refinement and then re-apply

richardschu:
alternatives are:
a) serialize the grid and deserialize with ignoring the mapping.
b) use the existing funtionality:
i) serialize _uniformly_ refined, mapped grid
ii) restart with _uniformly_ refined, *un*mapped grid; serialize that once again without timestepping
iii) restart with desired, *un*mapped discretization; serialize that once again without timestepping
iv) restart with desired, mapped discretization and `discretization_idetical = true`

richardschu:
the pressure is not high enough to get this to work atm.
I will split-off the refactor of the restart data and its serialization alongside the triangulation into #x

kronbichler:
I do not really understand here. Do you want to read the solution from an undeformed case back in, but that gives the strange behavior? It would mean we do need to set all data in the read-in operation without the mapping when interpreting the data. I made a quick look but did not see the problem, I can look again. Maybe it helps to have a short meeting to discuss?

kronbichler:
Thank you for the additional explanation, I now have a better understanding. I think you can do your i) by passing  to `MatrixFree::reinit` an `MappingQ1` or `MappingP1` on hex/simplex meshes, but clearly that does not resolve the case where the triangulation already has generated new child cells with the manifold. I guess this then boils down to how the meshes/deformations are actually set up. I've come to like the setting where `Triangulation` is mostly represented by a flat manifold and every refinement done on the flat manifold, where we then deform the elements by a suitable `MappingQCache` or `MappingFEField`. However, it's not possible to do for all cases.

Regarding the robustness issue with `RemotePointEvaluation`, I agree that it could be cumbersome, but I also think that making that more robust would be desirable. So that could be another point of attack. I think we need to sit down at some point and discuss the option, I have some ideas but not a solution yet.

richardschu:
Thanks for your comment.
Yes, I agree ... the `Manifold`s are just so easy to use I guess ; )

let us see where we run into problems using the periodic hill and other more challenging setups!
Working on RPE could also be fun ... maybe even for student projects :)

PR #782, opened 2025-10-06
enbale restart in periodic_hill/application.h

richardschu:
enable restart mechanics in the periodic hill benchmark.
Note that running this with `SpatialDiscretization::HDIV` leads to an assert. We need some more fixes, see
https://github.com/kronbichler/exadg/pull/1
and
https://github.com/kronbichler/exadg/pull/2

Once we have a fully functional version there, we will bring  it back into the main repo.


richardschu:
ah wrong branch.

PR #789, opened 2025-10-23
Add Consistent Splitting Scheme

dominiktassilostill:
This PR introduces the consistent splitting scheme. The divergence and continuity penalty terms are applied in post processing. Applying them in the momentum step will follow in another PR. @richardschu


richardschu:
Thanks @dominiktassilostill, happy to see that coming to ExaDG! :rocket:  
I am super curious how it behaves compared to the existent solvers we have. 
I will have a look asap and leave some comments!

applications/incompressible_navier_stokes/vortex/tests/consistentsplitting.output
```diff
@@ -0,0 +1,193 @@
...
+User specified time step size:
+
+  time step size:                            1.2500e-01
+
+... done!
+
+Starting time loop ...
+
+Calculate error for velocity at time t = 0.0000e+00:
+  Relative error (L2-norm): 1.96754e-03
```
richardschu: relative error is a bit high
can we afford something around 1e-6 or lower even?
Also, is this only one time step?
Can you at least do some 100?
I know a test should be fast running, but it is nice to see that the methods actually lead to a low error in a reasonable test already.
If we mess up some terms in the future, I think a stricter test might help catch errors.
But this has to be balanced, so I know it can be tricky! :smile:

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_consistent_splitting.h
```diff
@@ -0,0 +1,232 @@
+/*  ______________________________________________________________________
+ *
+ *  ExaDG - High-Order Discontinuous Galerkin for the Exa-Scale
+ *
+ *  Copyright (C) 2021 by the ExaDG authors
```
richardschu: 2025 here pls :)

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_consistent_splitting.h
```diff
@@ -0,0 +1,232 @@
...
+
+  void
+  face_loop_empty(dealii::MatrixFree<dim, Number> const &,
+                  VectorType &,
+                  VectorType const &,
+                  Range const &) const
+  {
+  }
+
+  // rhs PPE
```
richardschu:
```suggestion
   /*
    * Right hand side of the PPE
    */
```

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_consistent_splitting.h
```diff
@@ -0,0 +1,232 @@
...
+  void
+  rhs_ppe_laplace_add(VectorType & dst, double const & time) const;
+
+  unsigned int
+  solve_pressure(VectorType & dst, VectorType const & src, bool const update_preconditioner) const;
+
+  /*
+   * Viscous step.
+   */
+
```
richardschu:
```suggestion
```

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_consistent_splitting.h
```diff
@@ -0,0 +1,232 @@
...
+    VectorType &                            dst,
+    VectorType const &                      src,
+    Range const &                           cell_range) const;
+  void
+  local_rhs_ppe_div_term_body_forces_inner_face(
+    dealii::MatrixFree<dim, Number> const & matrix_free,
+    VectorType &                            dst,
+    VectorType const &                      src,
+    Range const &                           face_range) const;
+  void
```
richardschu:
```suggestion

  void
```

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_consistent_splitting.h
```diff
@@ -0,0 +1,232 @@
...
+
+    void
+    local_rhs_ppe_div_term_convective_boundary_face(
+    dealii::MatrixFree<dim, Number> const & matrix_free,
+    VectorType &                            dst,
+    VectorType const &                      src,
+    Range const &                           face_range) const;
+  
+  
+  // Neumann boundary condition term
```
richardschu:
```suggestion
  /* 
   * Neumann boundary condition term
   */
```

include/exadg/incompressible_navier_stokes/user_interface/parameters.cpp
```diff
@@ -485,6 +489,25 @@ Parameters::check(dealii::ConditionalOStream const & pcout) const
...
+  // CONSISTENT SPLITTING SCHEME
+  if(temporal_discretization == TemporalDiscretization::BDFConsistentSplittingScheme)
+  {
+    AssertThrow(spatial_discretization != SpatialDiscretization::HDIV,
+                dealii::ExcMessage("Not implemented."));
+
+    AssertThrow(order_extrapolation_pressure_rhs <= order_time_integrator,
+      dealii::ExcMessage("Invalid parameter order_extrapolation_pressure_rhs!"));
+
+    AssertThrow(order_extrapolation_pressure_nbc <= order_time_integrator,
```
richardschu: Is this also used? This refers to the $J_p$ used in the paper, similar to the dual splitting scheme.
But we use it for the entire RHS of the PPE for the consistent splitting scheme.
Can you please explain this?

include/exadg/incompressible_navier_stokes/user_interface/parameters.cpp
```diff
@@ -180,6 +180,10 @@ Parameters::Parameters()
     order_extrapolation_pressure_nbc((order_time_integrator <= 2) ? order_time_integrator : 2),
     formulation_convective_term_bc(FormulationConvectiveTerm::ConvectiveFormulation),
 
+    // CONSISTENT SPLITTING SCHEME  
+    order_extrapolation_pressure_rhs(order_time_integrator),
```
richardschu:
```suggestion
    order_extrapolation_pressure_rhs(order_time_integrator),
    order_extrapolation_pressure_rhs((order_time_integrator == 1) ? order_time_integrator : order_time_integrator-1),
```

richardschu: wouldn't this be what we use in the paper as default?

include/exadg/incompressible_navier_stokes/user_interface/parameters.cpp
```diff
@@ -1141,6 +1169,47 @@ Parameters::print_parameters_dual_splitting(dealii::ConditionalOStream const & p
...
+    return;
+
+  pcout << std::endl << "Consistent splitting scheme:" << std::endl;
+
+  // Leray projection
+  if(apply_leray_projection)
+    pcout << "  Apply Leray projection in the PPE" << std::endl;
+
+  // formulations
+  print_parameter(pcout, "Order of extrapolation convective pressure rhs", order_extrapolation_pressure_rhs);
```
richardschu:
```suggestion
  print_parameter(pcout, "Order of extrapolation of convective terms in pressure rhs", order_extrapolation_pressure_rhs);
```

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_consistent_splitting.cpp
```diff
@@ -0,0 +1,776 @@
...
+
+template<int dim, typename Number>
+void
+OperatorConsistentSplitting<dim, Number>::compute_Leray_projection_term(VectorType &       dst,
+                                                                        VectorType const & src,
+                                                                        double const & time) const
+{
+  this->evaluation_time = time;
+
+
```
richardschu:
```suggestion
```

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_consistent_splitting.cpp
```diff
@@ -0,0 +1,776 @@
...
+        velocity.gather_evaluate(src, dealii::EvaluationFlags::values);
+
+        // loop over quadrature points and compute the local volume flux
+        for (const unsigned int q : pressure.quadrature_point_indices())
+          {
+            const auto u = velocity.get_value(q);
+            pressure.submit_gradient(u, q);
+          }
+
+        // multiply by nabla v^h(x) and sum
```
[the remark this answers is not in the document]
richardschu: we do this so often :smile: I think we can expect people get that part :)

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_consistent_splitting.cpp
```diff
@@ -0,0 +1,776 @@
...
+      {
+        pressure.reinit(cell);
+        velocity.reinit(cell);
+
+        velocity.gather_evaluate(src, dealii::EvaluationFlags::values);
+
+        // loop over quadrature points and compute the local volume flux
+        for (const unsigned int q : pressure.quadrature_point_indices())
+          {
+            const auto u = velocity.get_value(q);
```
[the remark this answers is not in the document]
richardschu: Also, regarding the use of `auto`: a similar taste point we discussed in he past ...
One might say that `auto` is very elegant, and easy enough to understand in most places.
But one could also say that the reader has to think what the type *actually* is.
Also here, @kronbichler, any clear guidelines? I think the codebase now is also not clearly pro or con, but is there a guideline? I like to write the type if its not like 5 lines.
In the present case you even have `scalar`, `vector` and `tensor` defined which I think are super neat. Please use them and its 0 load on my brain reading this :+1:

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_consistent_splitting.cpp
```diff
@@ -0,0 +1,776 @@
...
+  FaceIntegratorU velocity_minus(matrix_free, true, dof_index_velocity, quad_index_pressure);
+
+  for (unsigned int face = face_range.first; face < face_range.second; face++)
+    {
+      dealii::types::boundary_id const boundary_id = matrix_free.get_boundary_id(face);
+
+      BoundaryTypeU const boundary_type =
+        this->boundary_descriptor->velocity->get_boundary_type(boundary_id);
+
+  if(boundary_type == BoundaryTypeU::Dirichlet or boundary_type == BoundaryTypeU::DirichletCached)
```
richardschu: i like that the `if` is outside the loop :)

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_consistent_splitting.cpp
```diff
@@ -0,0 +1,776 @@
...
+      {
+          pressure_minus.reinit(face);
+
+          for (const unsigned int q : pressure_minus.quadrature_point_indices())
+          {
+            // This term cancels with the boundary condition
+            pressure_minus.submit_value({}, q);
+          }
+
+        pressure_minus.integrate_scatter(dealii::EvaluationFlags::values, dst);
```
richardschu: can we get away with just not calling `pressure_minus.integrate_scatter()`? I think we had the same today in 
https://github.com/exadg/exadg/pull/792#discussion_r2478369193

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_consistent_splitting.cpp
```diff
@@ -0,0 +1,776 @@
...
+
+          velocity_minus.gather_evaluate(src,
+            dealii::EvaluationFlags::values);
+
+          for (const unsigned int q : pressure_minus.quadrature_point_indices())
+            {
+              const auto normal = pressure_minus.normal_vector(q);
+                
+              const auto u      = velocity_minus.get_value(q);
+              const auto div_u = - u * normal;
```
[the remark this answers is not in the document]
richardschu: just put some random name, choose anything, but `div_u` was a bit misleading :smile:

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_consistent_splitting.cpp
```diff
@@ -0,0 +1,776 @@
...
+  if(boundary_type == BoundaryTypeU::Dirichlet or boundary_type == BoundaryTypeU::DirichletCached)
+  {
+          pressure_minus.reinit(face);
+          for (const unsigned int q : pressure_minus.quadrature_point_indices())
+            {
+              // Cancels with consistent boundary condition
+              pressure_minus.submit_value({}, q);                
+            }
+
+          pressure_minus.integrate_scatter(dealii::EvaluationFlags::values, dst);
```
richardschu: same as above, I think we can just omit this entire thing?

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_consistent_splitting.cpp
```diff
@@ -0,0 +1,776 @@
...
+
+          for (const unsigned int q : pressure_minus.quadrature_point_indices())
+            {
+              const auto normal = pressure_minus.normal_vector(q);
+
+              // auto bc = this->boundary_descriptor->velocity->neumann_bc.find(boundary_id)->second;
+              // auto q_points = pressure_minus.quadrature_point(q);
+              // This should be the boundary condition but we get only grad_u * n which we cannot use, so use the gradient instead
+              const auto  grad_u = velocity_minus.get_gradient(q); //FunctionEvaluator<1, dim, Number>::value(*bc, q_points, this->evaluation_time);
+              const auto u_plus          = velocity_minus.get_value(q);
```
richardschu: is this a very misleading name or the approximation
$\mathbf{u}^{+}$ = $\mathbf{u}^{-}$?

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_consistent_splitting.cpp
```diff
@@ -0,0 +1,776 @@
...
+          velocity_minus.gather_evaluate(src, dealii::EvaluationFlags::values | dealii::EvaluationFlags::gradients);
+
+          for (const unsigned int q : pressure_minus.quadrature_point_indices())
+            {
+              const auto normal = pressure_minus.normal_vector(q);
+
+              // auto bc = this->boundary_descriptor->velocity->neumann_bc.find(boundary_id)->second;
+              // auto q_points = pressure_minus.quadrature_point(q);
+              // This should be the boundary condition but we get only grad_u * n which we cannot use, so use the gradient instead
+              const auto  grad_u = velocity_minus.get_gradient(q); //FunctionEvaluator<1, dim, Number>::value(*bc, q_points, this->evaluation_time);
```
richardschu: this comment is a bit unclear, can you rephrase?
The point is that you cannot insert the boundary data, as we have only $\nabla p \cdot \mathbf{n}$, right?
Maybe consider something like
"The provided boundary data cannot be used to compute the flux, use the elements velocity instead."

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_consistent_splitting.cpp
```diff
@@ -0,0 +1,776 @@
...
+  FaceIntegratorP integrator(matrix_free, true, dof_index_pressure, quad_index_pressure);
+
+  for(unsigned int face = face_range.first; face < face_range.second; face++)
+  {
+    integrator.reinit(face);
+
+    BoundaryTypeU boundary_type =
+      this->boundary_descriptor->velocity->get_boundary_type(matrix_free.get_boundary_id(face));
+
+    for(unsigned int q = 0; q < integrator.n_q_points; ++q)
```
richardschu: Since we do nothing except for the Neumann case, can you reorder the `if` and the `for` loop?
Then we can skip the `integrator.submit_value(zero, q);` for the Dirichlet faces.
Also this is similar to https://github.com/exadg/exadg/pull/792#discussion_r2478369193

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_consistent_splitting.cpp
```diff
@@ -0,0 +1,776 @@
...
+    integrator_velocity.gather_evaluate(acceleration, dealii::EvaluationFlags::values);
+
+    integrator_pressure.reinit(face);
+
+    BoundaryTypeU boundary_type =
+      this->boundary_descriptor->velocity->get_boundary_type(data.get_boundary_id(face));
+
+    for(unsigned int q = 0; q < integrator_pressure.n_q_points; ++q)
+    {
+      if(boundary_type == BoundaryTypeU::Dirichlet or boundary_type == BoundaryTypeU::DirichletCached)
```
richardschu: Also here, please reorder :)

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_consistent_splitting.cpp
```diff
@@ -0,0 +1,776 @@
...
+  {
+    pressure.reinit(face);
+
+    omega.reinit(face);
+    omega.gather_evaluate(src, dealii::EvaluationFlags::gradients);
+
+    BoundaryTypeP boundary_type =
+      this->boundary_descriptor->pressure->get_boundary_type(matrix_free.get_boundary_id(face));
+
+    for(unsigned int q = 0; q < pressure.n_q_points; ++q)
```
richardschu: Also here :)

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_consistent_splitting.cpp
```diff
@@ -0,0 +1,776 @@
...
+    {
+      integrator.reinit(face);
+      integrator.read_dof_values(dst);
+
+      for(unsigned int q = 0; q < integrator.n_q_points; ++q)
+      {
+        unsigned int const local_face_number = matrix_free.get_face_info(face).interior_face_no;
+
+        unsigned int const index = matrix_free.get_shape_info(dof_index, quad_index)
+                                     .face_to_cell_index_nodal[local_face_number][q];
```
richardschu: this would not work for H(div) would it?
I grepped `face_to_cell_index_nodal` and see it is used in some places in the library.
Probably that was a discussion point before ...
I guess since it is similar to the dual splitting scheme we go with this until we port the strategy from Martin's branch over to ExaDG.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_consistent_splitting.cpp
```diff
@@ -0,0 +1,921 @@
...
+  // velocity divergence
+  for(unsigned int i = 0; i < velocity_divergence.size(); ++i)
+    pde_operator->initialize_vector_pressure(velocity_divergence[i]);
+
+  // convective term divergence
+  for(unsigned int i = 0; i < vec_convective_term_div.size(); ++i)
+  pde_operator->initialize_vector_pressure(vec_convective_term_div[i]);
+
+    
+
```
richardschu:
```suggestion
```

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_consistent_splitting.cpp
```diff
@@ -0,0 +1,921 @@
...
+}
+
+
+template<int dim, typename Number>
+void
+TimeIntBDFConsistentSplitting<dim, Number>::initialize_current_solution()
+{
+  if(this->param.ale_formulation)
+    AssertThrow(false, dealii::ExcMessage("not implemented."));
+  //  this->helpers_ale->move_grid(this->get_time());
```
richardschu: fine for now :)

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_consistent_splitting.cpp
```diff
@@ -0,0 +1,921 @@
...
+  vec_convective_term_div_np = 0.;
+  pde_operator->apply_convective_divergence_term(vec_convective_term_div_np, velocity[0], this->get_time());
+  vec_convective_term_div[0].swap(vec_convective_term_div_np);
+}
+
+template<int dim, typename Number>
+void
+TimeIntBDFConsistentSplitting<dim, Number>::initialize_former_multistep_dof_vectors()
+{
+  // note that the loop begins with i=1! (we could also start with i=0 but this is not necessary)
```
richardschu:
```suggestion
  // Note that the loop begins with i=1, where we could also start with i=0 but this is not necessary.
```

richardschu: for some reason I read this as 1 factorial $1!$ ($=1$)

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_consistent_splitting.cpp
```diff
@@ -0,0 +1,921 @@
...
+   */
+  if(this->param.apply_leray_projection)
+    for(unsigned int i = 0; i < velocity_divergence.size(); ++i)
+      rhs.add(this->bdf.get_alpha(i) / this->get_time_step_size(), velocity_divergence[i]);
+
+  /*
+  *  II. convective extrapolation
+  */
+  for(unsigned int i = 0; i < extra_pressure_rhs.get_order();
+      ++i) // TODO: get the right extrapolation order here
```
richardschu: What does this mean here? Please explain.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_consistent_splitting.cpp
```diff
@@ -0,0 +1,921 @@
...
+   *  III. forcing term
+   */
+  if(this->param.right_hand_side)
+    pde_operator->rhs_ppe_div_term_body_forces_add(rhs, this->get_next_time());
+
+  /*
+   *  IV. handle consistent boundary condition
+   */
+  /*
+   *  IV.1 extrapolate speed and compute curl curl term
```
richardschu:
```suggestion
   *  IV.1 extrapolate speed and compute curl-curl term
```

richardschu (commented):
@dominiktassilostill absolutely beautiful :rocket: 
Mainly minor things, maybe check first on the change suggestions, commit of them whichever you like, and then you can resolve the comments I just mentioned some stuff, there are a few questions open.
But in general, only very minor comments.
Did you perform rigorous convergence tests as well?
It is basically impossible to check all the code :smile: 
I always merge this bad boy locally for faster convergence studies:
https://github.com/exadg/exadg/pull/368

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_consistent_splitting.cpp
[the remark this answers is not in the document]
dominiktassilostill: Changed it to follow the convention `const type`

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_consistent_splitting.cpp
dominiktassilostill: The asserts are now in the function, which checks the validity of the parameters. The same applies to the variable viscosity.

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_consistent_splitting.cpp
richardschu: thanks for the comment, lets leave it for now, if needed we change it later for the HODS and this scheme at once.

include/exadg/incompressible_navier_stokes/user_interface/parameters.cpp
richardschu: @dominiktassilostill this one is still open :)

include/exadg/incompressible_navier_stokes/user_interface/parameters.cpp
richardschu: @dominiktassilostill this one is still open :)
please just resolve if true :+1:

richardschu (commented):
a few tiny things left! thanks!

richardschu (commented):
clear from my side! squash commits before merge pls :)
@kronbichler you want to add to that?
I think we cannot be too wrong since we are close to the dual-splitting scheme and we have a test :sunglasses: 
thanks @dominiktassilostill for the great work! :1st_place_medal:

applications/incompressible_navier_stokes/vortex/application.h
```diff
@@ -258,8 +258,8 @@ class Application : public ApplicationBase<dim, Number>
 
     // TEMPORAL DISCRETIZATION
     this->param.solver_type                  = SolverType::Unsteady;
-    this->param.temporal_discretization      = TemporalDiscretization::BDFDualSplittingScheme;
-    this->param.treatment_of_convective_term = TreatmentOfConvectiveTerm::Explicit;
+    this->param.temporal_discretization      = TemporalDiscretization::BDFConsistentSplittingScheme;
+    this->param.treatment_of_convective_term = TreatmentOfConvectiveTerm::LinearlyImplicit;
```
kronbichler: This is not urgent and must not happen in this PR, but would it make sense to control this an input file, such that we can run the test for several different time stepping methods (in the future)?

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_consistent_splitting.cpp
```diff
@@ -0,0 +1,895 @@
...
+                                                 velocity[0],
+                                                 this->get_time());
+  vec_convective_term_div[0].swap(vec_convective_term_div_np);
+}
+
+template<int dim, typename Number>
+void
+TimeIntBDFConsistentSplitting<dim, Number>::initialize_former_multistep_dof_vectors()
+{
+  // Note that the loop begins with i=1, we could also start with i=0 but this is not necessary.
```
kronbichler: Can we remove the second part this comment, I don't think it contributes to the understanding. The reason is that the data for the time value 0 is set elsewhere.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_consistent_splitting.cpp
```diff
@@ -0,0 +1,895 @@
...
+template<int dim, typename Number>
+void
+TimeIntBDFConsistentSplitting<dim, Number>::pressure_step()
+{
+  dealii::Timer timer;
+  timer.restart();
+
+  // compute right-hand-side vector
+  VectorType rhs(pressure_np);
+  rhs_pressure(rhs);
```
kronbichler: As above, do not not create `rhs` from the vector `pressure_np` unless you actually want to use the content. Otherwise, do
```suggestion
  VectorType rhs;
  rhs.reinit(pressure_np);
  rhs_pressure(rhs);
```
if you want to have `rhs` initialized to zero, or if the zeroing happens inside `rhs_pressure` do `rhs.reinit(pressure_np, true);` where `true` indicates that we omit initializing the entries.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_consistent_splitting.cpp
```diff
@@ -0,0 +1,895 @@
...
+  push_back(velocity);
+  velocity[0].swap(velocity_np);
+
+  push_back(pressure);
+  pressure[0].swap(pressure_np);
+
+  // We also have to care about the history of velocity Dirichlet boundary conditions.
+  // Note that velocity_dbc_np has already been updated.
+  push_back(velocity_dbc);
+  velocity_dbc[0].swap(velocity_dbc_np);
```
kronbichler: We should eventually (i.e., not in this PR) get rid of that global vector, and have a specialized vector for quadrature points at the Dirichlet boundaries only. It should be reasonably simple to set up, but having a brittle mechanism for the boundary dofs into the cell. Think about the corner where multiple faces are at the Dirichlet boundary; this works for most cases but if you use discontinuous data leads to trouble.

kronbichler (approved):
Great work, I really like this, thanks @dominiktassilostill and @richardschu for the intensive work. I have a number of minor comments below, and one larger one that I repeat here: I'd prefer if we could merge the cell loops for the various contributions to the right hand sides. There is a minor performance gain from doing so, especially if linear solvers are fast. The more important part is that we need quite some overhead to add code each time for cell, inner face and boundary face evaluators, and the generic parts of those functions. I admit that there is an inherent self-documentation of various terms here, but I believe it is better to try to do at the quadrature point level or similar. We can merge the code for now and postpone this to a later step, but I really would like to keep the number of loop instances small in the long run.

include/exadg/incompressible_navier_stokes/spatial_discretization/operator_projection_methods.cpp
```diff
@@ -120,6 +120,14 @@ OperatorProjectionMethods<dim, Number>::initialize_laplace_operator()
     // conditions).
     laplace_operator_data.operator_is_singular = this->is_pressure_level_undefined();
   }
+  else if(this->param.temporal_discretization ==
+          TemporalDiscretization::BDFConsistentSplittingScheme)
+  {
+    // the Krylov subspace projection is needed for the consistent splitting scheme since the linear
+    // system of equations is not consistent for this splitting method (due to the boundary
+    // conditions).
+    laplace_operator_data.operator_is_singular = this->is_pressure_level_undefined();
```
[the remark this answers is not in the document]
dominiktassilostill: Yes, this should be changed in a follow up.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_consistent_splitting.cpp
```diff
@@ -0,0 +1,895 @@
...
+{
+  pde_operator->prescribe_initial_conditions(velocity[0], pressure[0], this->get_time());
+
+  // Now compute divergence of velocity and convective term
+  VectorType velocity_divergence_np(pressure[0]);
+  velocity_divergence_np = 0.;
+  pde_operator->compute_Leray_projection_term(velocity_divergence_np,
+                                              velocity[0],
+                                              this->get_time());
+  velocity_divergence[0].swap(velocity_divergence_np);
```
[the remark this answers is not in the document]
dominiktassilostill: Applied this to a bunch of vectors, probably a lot more reasonable than it is done in the dual splitting operator.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_consistent_splitting.cpp
```diff
@@ -0,0 +1,895 @@
...
+  pde_operator->do_rhs_ppe_laplace_add(rhs, this->get_next_time());
+
+  // IV.3. pressure Neumann boundary condition: temporal derivative of velocity
+  VectorType acceleration(velocity_dbc_np);
+  if(this->param.apply_leray_projection)
+    acceleration.equ(this->bdf.get_gamma0() / this->get_time_step_size(), velocity_dbc_np);
+  else
+    compute_bdf_time_derivative(
+      acceleration, velocity_dbc_np, velocity_dbc, this->bdf, this->get_time_step_size());
+  pde_operator->rhs_ppe_nbc_numerical_time_derivative_add(rhs, acceleration);
```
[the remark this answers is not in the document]
dominiktassilostill: I combined the loops of `rhs_ppe_nbc_viscous_add` and `rhs_ppe_nbc_numerical_time_derivative_add`, we can add the others in later on. The functionallity of `do_rhs_ppe_laplace_add` would then be doubled.

include/exadg/incompressible_navier_stokes/user_interface/parameters.cpp
```diff
@@ -1141,6 +1168,51 @@ Parameters::print_parameters_dual_splitting(dealii::ConditionalOStream const & p
...
+  if(apply_leray_projection)
+    pcout << "  Apply Leray projection in the PPE" << std::endl;
+
+  // formulations
+  print_parameter(pcout,
+                  "Order of extrapolation of convective terms in pressure rhs",
+                  order_extrapolation_pressure_rhs);
+  print_parameter(pcout,
+                  "Order of extrapolation of viscous term in pressure Neumann BC",
+                  order_extrapolation_pressure_nbc);
```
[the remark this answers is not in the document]
dominiktassilostill: Made this more beautiful.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_consistent_splitting.cpp
kronbichler: We had those in some cases of FSI problems (or more generally, with moving domains), where this gets injected from another field. If I remember correctly, we had different versions while developing the various algorithms for FSI, and some of the infrastructure in ExaDG was written before the remote point evaluation of deal.II (and related functions) were created, so there are some selections that could be streamlined here I believe. So yes, I agree we should work towards proving all information we need at quadrature points, and only have that option.

include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_consistent_splitting.cpp
```diff
@@ -0,0 +1,895 @@
...
+  }
+}
+
+template<int dim, typename Number>
+void
+TimeIntBDFConsistentSplitting<dim, Number>::prepare_vectors_for_next_timestep()
+{
+  Base::prepare_vectors_for_next_timestep();
+
+  push_back(velocity);
```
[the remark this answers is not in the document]
richardschu: definitely, opening up an issue and PR after we merged this, I am trying to avoid conflicts here at this advanced stage!

applications/incompressible_navier_stokes/vortex/application.h
richardschu: defnitely, I think that would be a good way to make testing for us also faster; I often find myself changing the application setup first thing. In a follow-up, we can do this for this application as well.

PR #792, opened 2025-10-28
Fix issue 745: correct dual splitting for variable viscosity case

### LINKED ISSUE #745: correct dual splitting for variable viscosity case
the boundary term in the dual splitting scheme is wrong.
we need to add a term involving the gradient of the viscosity, based on
#740 
I can port the changes from another branch.

richardschu:
fixes #745
see https://www.arxiv.org/abs/2506.14424
image

this implements the very last bit.

The commit
50e2c2147f8d5754ff99c33f2cdcb1294b838886
contains the viscous operator added to the RHS of the PPE.
According to my tests, this only worsens the situation, but is in agreement with
https://doi.org/10.1016/0021-9991(91)90007-8
https://doi.org/10.1016/S0168-9274(99)00112-9
I wanted to keep a version in the repo in case we want to test it in the future at some point again.

### DIFF
--- a/include/exadg/incompressible_navier_stokes/postprocessor/postprocessor.cpp
+++ b/include/exadg/incompressible_navier_stokes/postprocessor/postprocessor.cpp
@@ -362,9 +362,8 @@ PostProcessor<dim, Number>::initialize_derived_fields()
     viscosity.initialize_vector = [&](VectorType & dst) {
       navier_stokes_operator->initialize_vector_velocity_scalar(dst);
     };
-    viscosity.recompute_solution_field = [&](VectorType & dst, VectorType const & src) {
+    viscosity.recompute_solution_field = [&](VectorType & dst, VectorType const &) {
       // this only *accesses* the currently stored viscosity field
-      (void)src;
       navier_stokes_operator->access_viscosity(dst);
     };
 


include/exadg/incompressible_navier_stokes/spatial_discretization/operator_dual_splitting.cpp
```diff
@@ -550,6 +550,83 @@ OperatorDualSplitting<dim, Number>::local_rhs_ppe_nbc_viscous_add_boundary_face(
...
+      {
+        integrator_pressure.submit_value(dealii::make_vectorized_array<Number>(0.0), q);
+      }
+      else
+      {
+        AssertThrow(false,
+                    dealii::ExcMessage("Boundary type of face is invalid or not implemented."));
+      }
+    }
+    integrator_pressure.integrate_scatter(dealii::EvaluationFlags::values, rhs_ppe);
```
kronbichler: I know we do this elsewhere, but my recommendation would be to only do it on the Neumann boundary and not work for the other case (you could actually move it further out, but I think you see the logic):
```suggestion
    if(boundary_type == BoundaryTypeP::Neumann)
    {
      for(unsigned int q = 0; q < integrator_pressure.n_q_points; ++q)
      {
        vector normal = integrator_pressure.normal_vector(q);

        dealii::SymmetricTensor<2, dim, dealii::VectorizedArray<Number>> sym_grad_u =
          integrator_velocity.get_symmetric_gradient(q);

        vector grad_nu = integrator_viscosity.get_gradient(q);

        scalar h = dealii::make_vectorized_array<Number>(2.0) * (normal * (sym_grad_u * grad_nu));

        integrator_pressure.submit_value(h, q);
      }
      integrator_pressure.integrate_scatter(dealii::EvaluationFlags::values, rhs_ppe);
    }
    else if(boundary_type == BoundaryTypeP::Dirichlet)
    {
    }
    else
    {
      AssertThrow(false,
                  dealii::ExcMessage("Boundary type of face is invalid or not implemented."));
    }
```

kronbichler: Yes exactly!

kronbichler (approved):

PR #806, opened 2025-11-18
use symmetric tensors in structure module

richardschu:
.) forces `material.second_piola_kirchhoff_stress()` and `material.second_piola_kirchhoff_stress_displacement_derivative()` to return a symmetric tensor
.) introduces needed helper functions in `continuum_mechancis.h`

### DIFF
--- a/include/exadg/structure/material/library/st_venant_kirchhoff.cpp
+++ b/include/exadg/structure/material/library/st_venant_kirchhoff.cpp
@@ -140,13 +140,12 @@ StVenantKirchhoff<dim, Number>::cell_loop_set_coefficients(
 }
 
 template<int dim, typename Number>
-dealii::Tensor<2, dim, dealii::VectorizedArray<Number>>
-StVenantKirchhoff<dim, Number>::second_piola_kirchhoff_stress_symmetrize(
-  dealii::Tensor<2, dim, dealii::VectorizedArray<Number>> const & strain,
-  unsigned int const                                              cell,
-  unsigned int const                                              q) const
+typename StVenantKirchhoff<dim, Number>::symmetric_tensor
+StVenantKirchhoff<dim, Number>::second_piola_kirchhoff_stress_symmetrize(tensor const &     strain,
+                                                                         unsigned int const cell,
+                                                                         unsigned int const q) const
 {
-  dealii::Tensor<2, dim, dealii::VectorizedArray<Number>> S;
+  symmetric_tensor S;
 
   if(E_is_variable)
   {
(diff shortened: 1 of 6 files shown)


include/exadg/structure/material/library/st_venant_kirchhoff.cpp
```diff
@@ -155,40 +154,36 @@ StVenantKirchhoff<dim, Number>::second_piola_kirchhoff_stress_symmetrize(
...
+  // Since we return a `symmetric_tensor`, we only need to set symmetric entries once.
+  if constexpr(dim == 3)
   {
     S[0][0] = f0 * strain[0][0] + f1 * (strain[1][1] + strain[2][2]);
     S[1][1] = f0 * strain[1][1] + f1 * (strain[0][0] + strain[2][2]);
     S[2][2] = f0 * strain[2][2] + f1 * (strain[0][0] + strain[1][1]);
     S[0][1] = f2 * (strain[0][1] + strain[1][0]);
     S[1][2] = f2 * (strain[1][2] + strain[2][1]);
     S[0][2] = f2 * (strain[0][2] + strain[2][0]);
-    S[1][0] = S[0][1];
```
richardschu: here we explicitly see that we can save some computations, but it is rather in the operations with the symmetric tensors that we automatically profit whenever `dealii::SymmetricTensor`s are can be used.

include/exadg/structure/spatial_discretization/operators/continuum_mechanics.h
```diff
@@ -30,40 +30,177 @@ namespace ExaDG
...
+  for(unsigned int i = 0; i < dim; ++i)
+  {
+    for(unsigned int j = i; j < dim; ++j)
+    {
+      result[i][j] = H[i][j] + H[j][i];
+    }
+  }
+
+  // Debug output.
+  if constexpr(false)
```
richardschu: I can also make a test alternatively that checks all the helper functions in `continuum_mechanics.h` ... that felt a bit silly now as it would do something like
```
tensor dst, src;
// fill tensor with random values
dst = compute_H_times_HT(src);
// compute error
```
If the implementation were less straight-forward, I would do it.
I would say i we ever extend it towards a possibly more optimized variant(?), we go for the test.

include/exadg/structure/spatial_discretization/operators/continuum_mechanics.h
```diff
@@ -30,40 +30,177 @@ namespace ExaDG
...
+
+  return result;
+}
+
+// Create a symmetric tensor from a tensor H times H^T.
+// Note that we have (H * H^T)^T = (H^T)^T * H^T = H * H^T.
+template<int dim, typename Number>
+inline DEAL_II_ALWAYS_INLINE //
+  dealii::SymmetricTensor<2, dim, dealii::VectorizedArray<Number>>
+  compute_H_times_HT(dealii::Tensor<2, dim, dealii::VectorizedArray<Number>> const & H)
```
richardschu: note that we have now `compute_HT_times_H()` and `compute_H_times_HT` and the only thing that changes is the position of the transpose sign. The argument is *not* aussumed symmetric, so the results of the two functions are not the same. I did not want to create a single funtion with a boolean (template) parameter `transpose_first_else_second`, because I think this is then a bit confusing, especially in the code where this is then used. This function signature is a bit nicer to read, having
```
symmetric_tensor tmp = compute_H_times_HT<dim, Number>(src);
```
instead of
```
symmetric_tensor tmp = compute_H_times_H<dim, Number,false /* transpose_first_else_second */>(src);
```
.

kronbichler (approved):
Nice, I like it.

PR #813, opened 2025-11-19
Rename consistent and dual splitting

richardschu:
based on #812 since the same lines are affected, hence `draft`.

`BDFDualSplittingScheme` --> `BDFDualSplitting`
`BDFConsistentSplittingScheme` --> `BDFConsistentSplitting`
because for the others, namely
`BDFCoupledSolution`
and
`BDFPressureCorrection`,
one could also add the `Scheme`.
Let's keep the names short.


kronbichler (approved):
I agree, making the names more compact is better in this context, I like it!

PR #815, opened 2025-11-20
move `get_constant_modes()` implementation to cpp file

richardschu:
... and nothing else :)


richardschu:
merge since does not change anything.

PR #817, opened 2025-11-21
introduce `MagnitudeCalculator` in structure module

richardschu:
this enables to add the magnitude of the displacement to the output by setting
`pp_data.output_data.write_displacement_magnitude = true;`
in the application.

I am ofc aiming for other things (Jacobian, Eigenstresses and Eigenstrains), but it is a good minimal start to extend the postprocesser of the structure module.

We did not have any calculators in the structure module before, that is why its ~500 lines.

A few notes:

1)
The postprocessor *alone* decides, if we need the scalar space.
I wanted there to be no penalty, if the user does not ask for the displacement magnitude (aka any scalar variable in postprocessing).
Therefore I had to introduce `setup_scalar_field` in the structure operator determined outside and adapt the drivers to first execute `create_postprocessor()` to return the postprocessor in the drivers, where I then added `postprocessor.requires_scalar_field()`.

2)
We re-use the Navier-Stokes calculator.
The problem is: constraints are handled slightly differently for CG and DG spaces and therefore the
```
integrator.gather_evaluate(src, flags);
```
should be replaced with
```
integrator.read_dof_values_plain(src);
integrator.evaluate(flags);
```
in all calculators for reusability. This will go in a follow-up PR, I am doing the minimum here.


include/exadg/structure/postprocessor/output_generator.cpp
```diff
@@ -73,6 +73,18 @@ OutputGenerator<dim, Number>::setup(dealii::DoFHandler<dim> const & dof_handler_
                          mpi_comm);
     }
 
+    // write grid
```
richardschu: this was not there before, but as useful as for the fluid mechanics solver:
usecase:
you want the grid with mapping, apart from that only data that is not on the grid (currently only errors writte to the log file).

include/exadg/structure/time_integration/driver_quasi_static_problems.cpp
```diff
@@ -31,17 +31,17 @@ namespace Structure
 {
 template<int dim, typename Number>
 DriverQuasiStatic<dim, Number>::DriverQuasiStatic(
-  std::shared_ptr<Interface::Operator<Number>> operator_,
```
richardschu: we have a mixed syntax here in the library.
I am changing all
`variable_` to `variable_in`
only because if my IDE underlines the names, I cannot see the `_` anymore.
Skipping it is a bit confusing if one has member variables of the same name:
`variable(variable)`
or
`this->variable = variable;`
instead
`variable(variable_in)`
I can read and one can forget the `this->` or one gets an error if one by mistake writes
`variable = variable;`
---which would likely be a bug, since it has no effect on `this->variable`.

PR #819, opened 2025-11-21
Refactor calculators

richardschu:
0) use `integrator.read_dof_values_plain()` to ignore any constraints stored in the `MatrixFree` object such that we can reuse the calculators for any kind of problem where the constraints stored in the `MatrixFree` object might not be the same as for the DG case that we used in the Navier-Stokes module. Since this is in postprocessing, the `src` vector being the solution already has all constraints enforced. This is the only functional change, the rest is cosmetics for better readability.

1) the calculators have some main method, e.g., `compute_divergence()` for the `DivergenceCalculator`.
But these methods actually compute the right-hand side of an L2 projection.
Since the output `dst` already has the right size to be a DoF vector of an appropriately sized FE space, I renamed the function to `compute_projection_rhs()`.

2) rename the `src` and `dst` vectors to make clear which function space thy belong to.
E.g. for the `DivergenceCalculator` we have:
`src_vector_valued` ... which is the velocity
`dst_scalar_valued` ... which is the divergence of the velocity

3) make the calculators and naming used more generic and not tailored to the Navier-Stokes case. Also add some docu on the quantities that are computed.

### DIFF
--- a/include/exadg/operators/navier_stokes_calculators.cpp
+++ b/include/exadg/operators/navier_stokes_calculators.cpp
@@ -45,20 +45,21 @@ DivergenceCalculator<dim, Number>::initialize(
 
 template<int dim, typename Number>
 void
-DivergenceCalculator<dim, Number>::compute_divergence(VectorType &       dst,
-                                                      VectorType const & src) const
+DivergenceCalculator<dim, Number>::compute_projection_rhs(
+  VectorType &       dst_scalar_valued,
+  VectorType const & src_vector_valued) const
 {
-  dst = 0;
+  dst_scalar_valued = 0.0;
 
-  matrix_free->cell_loop(&This::cell_loop, this, dst, src);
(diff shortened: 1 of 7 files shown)


include/exadg/operators/navier_stokes_calculators.cpp
```diff
@@ -251,60 +259,59 @@ VorticityCalculator<dim, Number>::cell_loop(
       vector omega_vector;
       for(unsigned int d = 0; d < number_vorticity_components; ++d)
         omega_vector[d] = omega[d];
+
       integrator.submit_value(omega_vector, q);
     }
 
-    integrator.integrate(dealii::EvaluationFlags::values);
-    integrator.set_dof_values(dst);
+    integrator.integrate_scatter(dealii::EvaluationFlags::values, dst_vector_valued);
```
richardschu: `integrate_scatter()`

PR #821, opened 2025-11-22
split definition and implementation of `InversemassOperator`

richardschu:
... and nothing else, no functional change.

### DIFF
--- a/include/exadg/operators/inverse_mass_operator.cpp
+++ b/include/exadg/operators/inverse_mass_operator.cpp
@@ -1,0 +1,364 @@
+/*  ______________________________________________________________________
+ *
+ *  ExaDG - High-Order Discontinuous Galerkin for the Exa-Scale
+ *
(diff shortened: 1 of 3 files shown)


include/exadg/operators/inverse_mass_operator.cpp
```diff
@@ -0,0 +1,364 @@
...
+      integrator.read_dof_values(src, 0);
+
+      inverse_mass.apply(integrator.begin_dof_values(), integrator.begin_dof_values());
+
+      integrator.set_dof_values(dst, 0);
+    }
+  }
+}
+
+template class InverseMassOperator<2, 1, float>;
```
richardschu: we need
scalar version
dim version
dim + 2 version for restart, see
https://github.com/exadg/exadg/blob/213d339016667bff7684ad58f7ea191d9209cf0c/include/exadg/operators/solution_projection_between_triangulations.h#L324C1-L361C6

PR #826, opened 2025-11-26
generalize iNS manufactured solution for Hdiv

richardschu:
introduces periodic boundary conditions for the manufactured solution case s.t. we can check the accuracy of Hdiv. Was this done anywher? 
I will also use it to test restart.

@kronbichler this will be the testbed to check the restart mechanics properly (will merge selected commits to the other branch).

### DIFF
--- a/applications/incompressible_navier_stokes/navier_stokes_manufactured/application.h
+++ b/applications/incompressible_navier_stokes/navier_stokes_manufactured/application.h
@@ -24,11 +24,23 @@
 
 // ExaDG
 #include <exadg/grid/mesh_movement_functions.h>
+#include <exadg/grid/periodic_box.h>
 
 namespace ExaDG
 {
 namespace IncNS
 {
+/*
+ * Type of boundary conditions enforced.
+ */
+enum class BoundaryCondition
+{
+  MixedDirichletNeumann,
+  PureDirichlet,
+  PureNeumann,
+  Periodic
+};
+
 /*
  * Manufactured solution for incompressible flow of a generalized Newtonian fluid in a hypercube.
  * (Navier-)Stokes equations, where the convective term may be disabled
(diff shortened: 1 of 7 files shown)


applications/incompressible_navier_stokes/navier_stokes_manufactured/application.h
```diff
@@ -605,28 +614,51 @@ class Application : public ApplicationBase<dim, Number>
...
+      // Save the *coarse* triangulation for later deserialization.
+      if(write_restart and this->param.grid.triangulation_type == TriangulationType::Serial)
+      {
+        save_coarse_triangulation<dim>(this->param.restart_data, tria);
+      }
 
-        if(global_refinements > 0)
-          tria.refine_global(global_refinements);
-      };
+      if(global_refinements > 0)
```
richardschu: we refine afterwards, since we have to store the coarse tria in case of restart. I want to write some tests for restart next.

include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
```diff
@@ -171,11 +171,6 @@ SpatialOperatorBase<dim, Number>::initialize_dof_handler_and_constraints()
   }
   print_parameter(pcout, "number of dofs per cell", fe_u->n_dofs_per_cell());
   print_parameter(pcout, "number of dofs (total)", dof_handler_u.n_dofs());
-  if(param.spatial_discretization == SpatialDiscretization::HDIV)
```
richardschu: are the DoFs that are periodic not counted for the L2 case? I think this message would apply in any case, and hence does not really help. Since we are using DG, the periodic BCs only tell the boundary cells, which cells are their neighbors on the other side of the domain, right? So there are never any DoFs reduced ..?

PR #836, opened 2025-11-30
replace `integrate() ... set_dof_values()` w `integrate_scatter()`

richardschu:
I just happened to grep for `set_dof_values` and found these while debugging #831
docu: "might use some internal optimizations"


richardschu:
HDIV for compressible NS would be unusual maybe, but for any space that does no have the Kronecker delta property (such that we can write the values into the vector directly) this is wrong. I am not sure if there is a useful combination, but we can be safe and it might be faster :)

Issue #832, opened 2025-11-29
iNS `HDIV` has temporal convergence order 1


richardschu:
with #368 I found linear convergence of `HDIV` in time. in space I did not even test yet.
But look at the error in the velocity:
L2:

<img width="1237" height="1125" alt="Image" src="https://github.com/user-attachments/assets/ad9771e6-11c7-4ebf-a56d-1aa87f773945" />

HDIV:

<img width="1237" height="1125" alt="Image" src="https://github.com/user-attachments/assets/81e9d385-7723-4fb0-a7b2-010f0ec44eb3" />

so I think there is some old data (as we do have convergence in time, but $\epsilon \sim \mathcal{O}(\Delta t^1)$ )
in some boundary condition, which is periodic though.

Maybe we did not distribute the constraints somewhere (as `MatrixFree` automatically does) so some data is not updated after solving such that it is from the last timestep.

richardschu:
this also happens also when running in serial

richardschu:
Currently I am trying to figure out if
https://github.com/exadg/exadg/blob/3fdfbc1861e30131bfd6aae9a431c54762e39dac/include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp#L125C1-L131C23 
can be wrong in some way since I do not understand the
```
direction = face.face_idx[0] / 2
```
which should be

<img width="607" height="220" alt="Image" src="https://github.com/user-attachments/assets/b95a197c-80de-4eca-9a78-2da9f41160df" />

(see https://dealii.org/current/doxygen/deal.II/structGridTools_1_1PeriodicFacePair.html#af40cb9234af9aba3a47a8994670849c3)

argument to

<img width="507" height="146" alt="Image" src="https://github.com/user-attachments/assets/2085db60-0db2-4e03-b529-e01b2913604b" />

<img width="1917" height="651" alt="Image" src="https://github.com/user-attachments/assets/61863ae2-ef1e-4010-9dce-51762e78e3c1" />

but since this works for `BDFCoupledSolution` it should be good. I rather put some random `constraints_u.distribute(solution)` to see if that helps ...

richardschu:
ok new observation: I am doing $\nu =0$ and no convective term.
I also did not add the BCs to the `inverse_mass_velocity` since they are not needed and should give something useful in any case. 
That means I have
[...]

PR #838, opened 2025-11-30
enable divergence penalty for HDIV in projection methods

richardschu:
for `BDFCoupledSolution`, the divergence penalty is asserted, but for the fractional stepping schemes it is not.
Now, we can in principle use Hdiv spaces + divergence penalty.

(I thought this might be the cause for #832
but still no luck.)

(note that HDIV is still asserted for `BDFPressureCorrection` and `BDFConsistentSplitting` )


include/exadg/incompressible_navier_stokes/spatial_discretization/spatial_operator_base.cpp
```diff
@@ -1777,10 +1776,13 @@ template<int dim, typename Number>
 void
 SpatialOperatorBase<dim, Number>::setup_projection_solver()
 {
-  // setup projection solver
+  // skip setup of projection solver
+  if(param.use_divergence_penalty == false and param.use_continuity_penalty == false)
```
richardschu: I prefer this over a wrapping `if()`, but we can do that as well.

PR #842, opened 2025-12-01
Fix Navier-Stokes calculators

kronbichler:
I tried to look into a case with H(div) and the vorticity was off, because we forgot to set the vectors to zero before adding into them in `MatrixFree::cell_loop`. We should let `MatrixFree` do that. And once there, do it uniformly for all calculators.


richardschu (changes requested):
thanks for fixing this! there is two `vec = 0;` still there, please remove them as well.

PR #845, opened 2025-12-01
enable GMRES in structure module

richardschu:
For a problem w 350k DoFs in the 3D manufactured application I get:
(Error norms are identical. )

FGMRES
image

CG
image

GMRES
image

So here it barely pays off in this simple example, but if we had closer to 30 iterations, having `GMRES` is an easily achieved performance improvement (requires less memory, has lower orthogonalization cost, has a less expensive restart, etc.).

### DIFF
--- a/applications/structure/manufactured/application.h
+++ b/applications/structure/manufactured/application.h
@@ -340,6 +340,18 @@ public:
   {
   }
 
+  void
+  add_parameters(dealii::ParameterHandler & prm) final
+  {
+    ApplicationBase<dim, Number>::add_parameters(prm);
+
+    prm.enter_subsection("Application");
+    {
+      prm.add_parameter("Solver", solver, "Krylov solver used.");
+    }
+    prm.leave_subsection();
+  }
+
 private:
   void
   set_parameters() final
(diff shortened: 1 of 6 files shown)


kronbichler:
Could you please elaborate on the statement
> So here it barely pays off in this simple example, but if we had closer to 30 iterations, having GMRES is an easily achieved performance improvement (requires less memory, has lower orthogonalization cost, has a less expensive restart, etc.).

The numbers show that CG is fastest, as expected, and indeed I would always use CG if the system is symmetric positive definite and the preconditioner is linear (i.e. no inner solvers or other things that make it non-stationary during a single linear solve). Are you referring to the difference between GMRES and FGMRES? Between those two, GMRES uses less memory, but they both use the same orthogonalization process at least in the deal.II implementation and behave similarly for the restart. In general, one prefers GMRES whenever the preconditioner is linear and stationary, but it is only a minor difference in runtime in most circumstances, like yours.

With regard to Bicgstab, you observe that it takes fewer iterations, but you have to keep in mind that it constructs two Krylov subspaces (the second sometimes being referred to as 'shadow space'), which means it takes two matrix-vector products and two preconditioner applications per iteration. You can look at the algorithm at https://en.wikipedia.org/wiki/Biconjugate_gradient_stabilized_method#Preconditioned_BiCGSTAB or the deal.II implementation that got used in the example https://github.com/dealii/dealii/blob/68708f1839313bfacf14f892ec156fb59eb1830d/include/deal.II/lac/solver_bicgstab.h#L332-L395

PR #846, opened 2025-12-02
Introduce `GMRES` and `BiCGStab` solvers in structure module

richardschu:
just by adding the conversion from enums to the `dealii::SolverSelector::name`

At a later point, we can completely remove the enums if we want to.
See #851

Iterations are identical to what is posted in https://github.com/exadg/exadg/pull/845

### DIFF
--- a/applications/structure/manufactured/application.h
+++ b/applications/structure/manufactured/application.h
@@ -340,6 +340,18 @@ public:
   {
   }
 
+  void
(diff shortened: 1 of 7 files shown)


richardschu:
sure! I will try something based on #845 and open a PR if I like it :)

richardschu:
I would say lets close this and continue with #848 ff

richardschu:
I will introduce the additional enum with this PR once #850 is merged.

PR #852, opened 2025-12-03
Fix issue 852: move `evaluate_convergence_study()` to destructor of postprocessor

### LINKED ISSUE #849: automated convergence studies require output folders to match
see https://github.com/exadg/exadg/pull/368/files#r2582266914:
the folder the convergence data is written to is a separate parameter from the one the data is read from.
unify the folders.

richardschu:
fixes #849 

in a nutshell: we move the call to the destructor of the postprocessors.
Then we do not have the problem that we do not know the `directory` that was set in the application.
We would have no other way of finding out what the user did in the application on top of what is defined in the `.json` parameter file.

The `evaluate_convergence_study()` processes all files found in the folder set.
That is why there is some logic attaching folders uniquely to not be repetitive.
In the application, activating this functionality is controlled by `error_data.compute_convergence_table`.

If each `error_data` has its own directory, the data is not merged into one convergence table and tables are printed one after the other.

### DIFF
--- a/applications/incompressible_navier_stokes/navier_stokes_manufactured/application.h
+++ b/applications/incompressible_navier_stokes/navier_stokes_manufactured/application.h
@@ -794,6 +794,7 @@ private:
     pp_data.output_data.write_higher_order = true;
 
     // calculation of velocity error
+    pp_data.error_data_u.compute_convergence_table          = false;
     pp_data.error_data_u.write_errors_to_file               = true;
     pp_data.error_data_u.time_control_data.is_active        = true;
     pp_data.error_data_u.time_control_data.start_time       = start_time;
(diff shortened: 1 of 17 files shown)


include/exadg/incompressible_navier_stokes/time_integration/time_int_bdf_dual_splitting.cpp
```diff
@@ -386,6 +386,9 @@ TimeIntBDFDualSplitting<dim, Number>::convective_step()
   // compute convective term and extrapolate convective term (if not Stokes equations)
   if(this->param.convective_problem())
   {
+    // DO NOT MERGE THIS ##+
+    velocity_np.zero_out_ghost_values();
+
```
kronbichler: I assume you want to remove it now?

kronbichler (approved):

Issue #851, opened 2025-12-03
TODOs for port to `Krylov::KrylovSolver`


richardschu:
after we have ported all modules to `KrylovSolver`, we should consider:

1) remove the enums like `Solver::CG` for the user to just set, e.g, `this->param.solver_data.name = "fgmres"` or similar in the application.

2) getting rid of structs such as `Krylov::SolverDataCG`, which where the intermediates to go from `ExaDG::SolverData` to the `dealii::SolverDataCG` structs.

PR #859, opened 2025-12-07
use `KrylovSolver` for coarse grid solver

richardschu:
I will make multiple small PRs so that it easier to review for the rollout of `KrylovSolver`


richardschu:
continuing here to finish #851

PR #865, opened 2025-12-11
remove solver wrappers and enums like `Solver::CG`

### LINKED ISSUE #851: TODOs for port to `Krylov::KrylovSolver`

richardschu:
closes #851

move completely to `Krylov::KrylovSolver` by:
.) remove deal.ii solver wrapper classes
.) _unify_ enums that control solver choice like `SolverCG`
.) move that enum to `SolverData`
.) update the applications and tests that are affected

### DIFF
--- a/include/exadg/incompressible_navier_stokes/spatial_discretization/operator_projection_methods.cpp
+++ b/include/exadg/incompressible_navier_stokes/spatial_discretization/operator_projection_methods.cpp
@@ -219,21 +219,6 @@ OperatorProjectionMethods<dim, Number>::setup_solver_pressure_poisson()
   bool constexpr compute_eigenvalues         = false;
   bool const use_preconditioner =
     this->param.preconditioner_pressure_poisson != PreconditionerPressurePoisson::None;
-  std::string name;
-  if(this->param.solver_pressure_poisson == SolverPressurePoisson::CG)
-  {
-    name = "cg";
-  }
-  else if(this->param.solver_pressure_poisson == SolverPressurePoisson::FGMRES)
-  {
-    name = "fgmres";
-  }
-  else
-  {
-    AssertThrow(false,
-                dealii::ExcMessage(
-                  "Specified solver for pressure Poisson equation is not implemented."));
-  }
 
   typedef Krylov::
     KrylovSolver<Poisson::LaplaceOperator<dim, Number, 1>, PreconditionerBase<Number>, VectorType>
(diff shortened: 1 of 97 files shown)


include/exadg/incompressible_navier_stokes/spatial_discretization/operator_projection_methods.cpp
```diff
@@ -219,21 +219,6 @@ OperatorProjectionMethods<dim, Number>::setup_solver_pressure_poisson()
   bool constexpr compute_eigenvalues         = false;
   bool const use_preconditioner =
     this->param.preconditioner_pressure_poisson != PreconditionerPressurePoisson::None;
-  std::string name;
```
richardschu: this is an example how we skip now the parsing step enum -> `std::string` required for `dealii::SolverSelector`.

include/exadg/incompressible_navier_stokes/user_interface/enum_types.h
```diff
@@ -359,14 +359,7 @@ enum class QuadratureRuleLinearization
  *  use CG (conjugate gradient) method as default. FGMRES might be necessary
  *  if a Krylov method is used inside the preconditioner (e.g., as multigrid
  *  smoother or as multigrid coarse grid solver)
- */
```
richardschu: note that all of these solvers support now the solver supported in `dealii::SolverSelector` automatically.

nfehn:
I am not sure if we should really solve it like this. I actually don't like that the string parameter now appears in the user interface of ExaDG: i) you now do not get errors at compile time in case of mis-spellings, ii) an IDE does not show the options of possible solvers (nor does the ExaDG doxygen documentation), iii) if dealii decides to change the name "fmgres" to "FlexGMRES" (fictitious, but let's assume), you will not detect this problem in ExaDG immediately, but get run time errors for various applications. (You could principally translate the string name within ExaDG to the new dealii naming, but this is hard to understand/document and introduces somehow technical debt in ExaDG).

The solution is therefore to define an enum class for the solver type in ExaDG, which becomes a parameter in ExaDG::SolverData (if you want), and which is translated once and for all to the dealii-string in ExaDG::KrylovSolver by some utility function, member function, lambda.

This suggestion resolves i) and ii) by design, and a change according to iii) implies just a single change in ExaDG.

PR #869, opened 2025-12-17
enable implicit boundary integrals for CG

richardschu:
with "implicit" I mean on the left hand side of a linear system $A x = b$.

Prior to this PR we only had `do_boundary_integral_continuous()` to integrate boundary terms in the CG case.
The boundary integrals considered did not include the solution itself, but where constant containing the Neumann BC data only.

With this PR, we enable (non-)linear boundary terms, which is the first step to consider for `param.pull_back_traction_vector = true` with a consistent Jacobian or other (non-)linear integrals that have to be added to the homogeneous operator.

This is allowed by having `do_boundary_integral_continuous()` which now also has `OperatorType` (similar to the DG case) and computes the boundary integral in the linear(-ized) operator, while `do_boundary_integral_continuous_nonlinear()` evaluates the boundary term with the most recent linearization vector in the nonlinear case.

The actual Jacobian of the nonlinear scaling of the Neumann term for `pull_back_traction = true` is not included.
This PR produces identical results to `master`.
In addition to `ctest`, I tested (non-)linear iteration counts with the `bar` and `manufactured` application in the structure module with matrix-based and matrix-free variants.

For looking at the diff: make sure to hide whitespace.

~we should merge #871 before such that we can verify a greater portion of the code to be functional.~

### DIFF
--- a/include/exadg/operators/operator_base.cpp
+++ b/include/exadg/operators/operator_base.cpp
@@ -252,7 +252,7 @@ OperatorBase<dim, Number, n_components>::calculate_inverse_diagonal(VectorType &
 {
   this->calculate_diagonal(diagonal);
 
-  if(false)
+  if constexpr(false)
   {
     verify_calculation_of_diagonal(
       *this, diagonal, matrix_free->get_dof_handler(this->data.dof_index).get_mpi_communicator());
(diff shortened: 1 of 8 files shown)


include/exadg/operators/operator_base.cpp
```diff
@@ -305,28 +313,42 @@ template<int dim, typename Number, int n_components>
...
+  if(not is_dg)
+  {
+    // Constrained degrees of freedom are not removed from the system of equations. Instead, we set
+    // the diagonal entries of the matrix to 1 for these constrained degrees of freedom. This means
+    // that we need to copy the constrained values to the `dst` vector.
     for(unsigned int const constrained_index :
         matrix_free->get_constrained_dofs(this->data.dof_index))
     {
-      dst.local_element(constrained_index) += src.local_element(constrained_index);
+      dst.local_element(constrained_index) = src.local_element(constrained_index);
```
kronbichler: I don't understand this change. The function is `apply_add`, so shouldn't we keep the code line as it was before?

kronbichler (approved):
Looks good to me, thank you.

PR #882, opened 2025-12-28
Fix issue 870: use `MatrixFreeTools` for diagonal and matrix assembly

### LINKED ISSUE #870: use `dealii::MatrixFreeTools` for diagonal and matrix assembly
I think we should revisit
[https://github.com/exadg/exadg/pull/661/files](url)
which we merged and then un-merged I think.
There were some problems with #661, but we might get rid of some custom (meaning ExaDG and not dealii) code in `OperatorBase`.

richardschu:
closes #870
based on #661, which was reverted in #704

I checked in the Poisson Gaussian application that the iteration counts are unchanged to master for DG and CG, matrix-free and matrix-based.

@peterrum I put you as coauthor here since from #661 there were only tiny changes needed :) thanks for your help!

The two things I did differently as compared to #661:
.) set diagonal values to one in assembled matrix
.) use 
```
this->do_face_integral(integrator_m, integrator_p);
```
instead of
```
this->do_int_face_integral(integrator_m, integrator_p);
```

### DIFF
--- a/applications/poisson/gaussian/input.json
+++ b/applications/poisson/gaussian/input.json
@@ -16,7 +16,8 @@
         "AdaptiveRefinement": "false",
         "UseMatrixBasedOperator": "false",
         "Preconditioner": "Multigrid",
-        "MinDegreeMatrixFree": "1"
+        "MinDegreeMatrixFree": "1",
+        "SpatialDiscretization": "CG"
     },
     "Output": {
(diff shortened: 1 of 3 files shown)


richardschu:
I am merging this now since I am quite confident that this preserves functionality since we have tests since #871.

PR #886, opened 2026-01-08
Tracer particles postprocessor

dominiktassilostill:
This PR is a proposal for a tracer particle postprocessor based on the deal.II particle infrastructure (and is largely based on the prototype application of @peterrum ). It assumes massless particles which are transported in a fluid, i.e. the particles have the same velocity as the fluid. The time integration is just a explicit Euler scheme, should be fine for now, for more involved problems with forces, this should be changed to either VV or RK.
For large problems it was faster for me to iterate over the particles compared to going over all cells.

The starting positions of the particles are given by the user in a vector, we should think of a better way to do this. @richardschu

### DIFF
--- a/include/exadg/incompressible_navier_stokes/postprocessor/postprocessor.cpp
+++ b/include/exadg/incompressible_navier_stokes/postprocessor/postprocessor.cpp
@@ -39,7 +39,8 @@ PostProcessor<dim, Number>::PostProcessor(PostProcessorData<dim> const & postpro
     div_and_mass_error_calculator(comm),
     kinetic_energy_calculator(comm),
     kinetic_energy_spectrum_calculator(comm),
-    line_plot_calculator(comm)
+    line_plot_calculator(comm),
+    particle_calculator(comm)
 {
 }
 
@@ -101,6 +102,10 @@ PostProcessor<dim, Number>::setup(Operator const & pde_operator)
   kinetic_energy_spectrum_calculator.setup(navier_stokes_operator->get_matrix_free(),
                                            navier_stokes_operator->get_dof_handler_u(),
                                            pp_data.kinetic_energy_spectrum_data);
+
+  particle_calculator.setup(navier_stokes_operator->get_dof_handler_u(),
+                            *navier_stokes_operator->get_mapping(),
+                            pp_data.particle_data);
 
   line_plot_calculator.setup(navier_stokes_operator->get_dof_handler_u(),
                              navier_stokes_operator->get_dof_handler_p(),
@@ -252,6 +257,17 @@ PostProcessor<dim, Number>::do_postprocessing(VectorType const &     velocity,
     kinetic_energy_spectrum_calculator.evaluate(velocity,
                                                 time,
                                                 Utilities::is_unsteady_timestep(time_step_number));
+  }
+
+  /*
+   *  calculation of particle trajectory
+   *  `is_active` only controls output, particles are advanced every timestep but only printed
+   * sometimes
+   */
+  if(pp_data.particle_data.time_control_data.is_active)
+  {
+    particle_calculator.evaluate(
+      velocity, time, particle_calculator.time_control.needs_evaluation(time, time_step_number));
   }
 
   /*
--- a/include/exadg/postprocessor/particle_calculation.cpp
+++ b/include/exadg/postprocessor/particle_calculation.cpp
@@ -1,0 +1,165 @@
+/*  ______________________________________________________________________
+ *
+ *  ExaDG - High-Order Discontinuous Galerkin for the Exa-Scale
+ *
+ *  Copyright (C) 2025 by the ExaDG authors
+ *
+ *  This program is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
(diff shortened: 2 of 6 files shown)


include/exadg/incompressible_navier_stokes/postprocessor/postprocessor.cpp
```diff
@@ -254,6 +259,16 @@ PostProcessor<dim, Number>::do_postprocessing(VectorType const &     velocity,
                                                 Utilities::is_unsteady_timestep(time_step_number));
   }
 
+  /*
+   *  calculation of particle tracetory
```
richardschu:
```suggestion
   *  calculation of particle trajectory
```

include/exadg/incompressible_navier_stokes/postprocessor/postprocessor.cpp
```diff
@@ -254,6 +259,16 @@ PostProcessor<dim, Number>::do_postprocessing(VectorType const &     velocity,
...
+   *  if this is active if advances every timestep but only prints sometimes
```
richardschu:
```suggestion
   *  `is_active` only controls output, particles are advanced every timestep
```

include/exadg/postprocessor/particle_calculation.cpp
```diff
@@ -0,0 +1,165 @@
+/*  ______________________________________________________________________
+ *
+ *  ExaDG - High-Order Discontinuous Galerkin for the Exa-Scale
+ *
+ *  Copyright (C) 2021 by the ExaDG authors
```
richardschu:
```suggestion
 *  Copyright (C) 2025 by the ExaDG authors
```

include/exadg/postprocessor/particle_calculation.cpp
```diff
@@ -0,0 +1,165 @@
...
+    create_directories(data.directory, mpi_comm);
+
+    dof_handler = &dof_handler_in;
+    mapping     = &mapping_in;
+
+    // init particle handler
+    particle_handler.initialize(dof_handler->get_triangulation(), *mapping);
+
+    // get bounding boxes to distribute globally
+    const auto my_bounding_box = dealii::GridTools::compute_mesh_predicate_bounding_box(
```
richardschu:
```suggestion
    auto const my_bounding_box = dealii::GridTools::compute_mesh_predicate_bounding_box(
```
...not bcs I think it is super important, but bcs we have this sequence in most places of `ExaDG`.

include/exadg/postprocessor/particle_calculation.cpp
```diff
@@ -0,0 +1,165 @@
...
+      dof_handler->get_triangulation(), dealii::IteratorFilters::LocallyOwnedCell());
+    global_bounding_boxes = dealii::Utilities::MPI::all_gather(mpi_comm, my_bounding_box);
+
+    // insert particles
+    if constexpr(dim == 2)
+      particle_handler.insert_global_particles(data.starting_points_2d, global_bounding_boxes);
```
richardschu, line 71: why didn't you do
```
template<int dim>
struct ParticleData
{ 
  // ... 
} ;
```
? Just to not have it templated, right? I do not have a strong favorite here, I just want to understand why please :smile: . Even if there was no strong reasoning behind, you can also just leave it as it is!

include/exadg/postprocessor/particle_calculation.cpp
```diff
@@ -0,0 +1,165 @@
...
+    const auto dh_cell = typename dealii::DoFHandler<dim>::cell_iterator(*cell, dof_handler);
+
+    // get velocity in the cell
+    solution_values = 0.;
+    dh_cell->get_dof_values(velocity, solution_values);
+
+    // collect all particles in the cell
+    const auto pic = particle_handler.particles_in_cell(cell);
+
+    // collect current refernce position of particles
```
richardschu:
```suggestion
    // collect current reference position of particles
```

include/exadg/postprocessor/particle_calculation.h
```diff
@@ -0,0 +1,113 @@
+/*  ______________________________________________________________________
+ *
+ *  ExaDG - High-Order Discontinuous Galerkin for the Exa-Scale
+ *
+ *  Copyright (C) 2021 by the ExaDG authors
```
richardschu:
```suggestion
 *  Copyright (C) 2025 by the ExaDG authors
```

richardschu (commented):
Thanks @dominiktassilostill, cool functionality!
Could you add it to some testcase just so that it is even easier to use?

In a follow-up we could create seed points based on boundary IDs or implement higher-order time integrators.

include/exadg/postprocessor/particle_calculation.cpp
dominiktassilostill: It is just written that way to prevent having the template but we should think of a better way to do this, maybe by giving some lambda function, which then also can be used to periodically reinsert particles into the domain.

include/exadg/postprocessor/particle_calculation.cpp
dominiktassilostill: But it is 2026:)

dominiktassilostill:
> Thanks @dominiktassilostill, cool functionality! Could you add it to some testcase just so that it is even easier to use?

Do you want an example, as now in the `vortex` application or a dedicated test?

include/exadg/incompressible_navier_stokes/postprocessor/postprocessor.cpp
dominiktassilostill: I should really learn to spell:(

include/exadg/postprocessor/particle_calculation.cpp
richardschu: ok, I would say just leave it for now :)

richardschu:
> Do you want an example, as now in the `vortex` application or a dedicated test?

No, looks good :)

richardschu (commented):
@dominiktassilostill this is ready to merge from your side?

dominiktassilostill:
@richardschu should be good to go now:)

PR #887, opened 2026-01-12
Fast inverse mass matrix for affine simplex elements

dominiktassilostill:
This adds a matrix-free inverse mass operator for simplex elements under affine mappings. It inverts the mass matrix on the reference element in the setup and then applies the inverse mass matrix on the vector with the inverted local metric terms. This assumes the the determinant of the Jacobian is constant over the cell, i.e. it only works for affine mappings.

### DIFF
--- a/include/exadg/operators/inverse_mass_operator.cpp
+++ b/include/exadg/operators/inverse_mass_operator.cpp
@@ -19,6 +19,13 @@
  *  ______________________________________________________________________
  */
 
+// deal.II
+#include <deal.II/fe/fe_simplex_p.h>
+#include <deal.II/fe/mapping_fe.h>
+#include <deal.II/grid/grid_generator.h>
+#include <deal.II/lac/la_parallel_vector.h>
+#include <deal.II/matrix_free/operators.h>
+
 #include <exadg/operators/inverse_mass_operator.h>
 
 namespace ExaDG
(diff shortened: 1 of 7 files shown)


include/exadg/operators/inverse_mass_operator.cpp
```diff
@@ -64,20 +71,79 @@ InverseMassOperator<dim, n_components, Number>::initialize(
...
+      const bool cartesian_or_affine_mapping =
+        std::all_of(matrix_free->get_mapping_info().cell_type.begin(),
+                    matrix_free->get_mapping_info().cell_type.end(),
+                    [](auto g) {
+                      return g <= dealii::internal::MatrixFreeFunctions::GeometryType::affine;
+                    });
+      AssertThrow(
+        cartesian_or_affine_mapping,
+        dealii::ExcMessage(
+          "The matrix-free cell-wise inverse mass operator can only be applied with affine mapping on non-hypercube elements."));
```
kronbichler: Can you please break the lines at around 80-90 characters (I prefer to not have text too long in lines).

include/exadg/operators/inverse_mass_operator.cpp
```diff
@@ -64,20 +71,79 @@ InverseMassOperator<dim, n_components, Number>::initialize(
...
+
+      AssertThrow(
+        !coefficient_is_variable,
+        dealii::ExcMessage(
+          "The matrix-free cell-wise inverse mass operator can only be applied with constant coefficients over a cell on non-hypercube elements, use the apply_scaling instead."));
```
kronbichler: Same here.

include/exadg/operators/inverse_mass_operator.cpp
```diff
@@ -64,20 +71,79 @@ InverseMassOperator<dim, n_components, Number>::initialize(
...
+                                          dealii::update_JxW_values);
+        fe_values.reinit(triangulation.begin_active());
+
+        mass_matrix = 0.;
+        for(unsigned int i = 0; i < dofs_per_cell; ++i)
+          for(unsigned int j = 0; j < dofs_per_cell; ++j)
+            for(unsigned int q = 0; q < quadrature_mass.size(); ++q)
+              mass_matrix(i, j) +=
+                fe_values.JxW(q) * fe_values.shape_value(i, q) * fe_values.shape_value(j, q);
+        mass_matrix.gauss_jordan();
```
kronbichler, line 139: Depending on the condition number (which might matter for higher polynomial degrees), doing Gauss--Jordan might not be the most accurate inversion in terms of roundoff. It is usually better to compute a Householder matrix and then apply it to unit vectors, but it is a bit more code, so I'd leave it in this form for now.

include/exadg/operators/inverse_mass_operator.cpp
```diff
@@ -64,20 +71,79 @@ InverseMassOperator<dim, n_components, Number>::initialize(
...
+
+        for(unsigned int i = 0; i < dofs_per_cell; ++i)
+          for(unsigned int j = 0; j < dofs_per_cell; ++j)
+            this->inverse_mass_matrix.emplace_back(mass_matrix(i, j));
```
kronbichler, line 144: I suggest to reserve the right amount of memory in `inverse_mass_matrix` before the loop, i.e., `this->inverse_mass_matrix.reserve(dofs_per_cell * dofs_per_cell);` to avoid automatic resizing.

include/exadg/operators/inverse_mass_operator.cpp
```diff
@@ -343,6 +434,112 @@ InverseMassOperator<dim, n_components, Number>::cell_loop_matrix_free_operator(
...
+  dealii::AlignedVector<dealii::VectorizedArray<Number>> * scratch_data =
+    matrix_free->acquire_scratch_data();
+  const dealii::internal::MatrixFreeFunctions::ShapeInfo<Number> & shape_info =
+    matrix_free->get_shape_info(dof_index, quad_index);
+  const unsigned int     dofs_per_cell = shape_info.dofs_per_component_on_cell;
+  constexpr unsigned int n_lanes       = dealii::VectorizedArray<Number>::size();
+
+  const auto & mapping_data = matrix_free->get_mapping_info().cell_data[quad_index];
+
+  scratch_data->resize_fast(2 * dofs_per_cell * n_components);
```
kronbichler: Is there a reason you do this manually, instead of using `FEEvaluation` to handle the reading of vectors and to return `JxW(0)`? Not that I'm against this version, because `FEEvaluation` has quite some overhead, but it could have made the code a bit shorter.

kronbichler (approved):
This is a really nice contribution, I've wanted to have this for a long time. Well done! I have a few minor comments. And one more thing: For the case when the cells are not affine or we have variable coefficients, some scaled version of this would make for an excellent preconditioner for batched solvers.

include/exadg/operators/inverse_mass_operator.cpp
dominiktassilostill: Tbh I just used the build in function, if we go to higher polynomial degrees we can change it.

include/exadg/operators/inverse_mass_operator.cpp
dominiktassilostill: This was manual because I had the code around, now it is a shorter (and more comprehensible) version using `FEEvaluation` for reading/writing. I assumed calling `JxW(0)` includes the quadrature weight of the first quadrature point, so the code uses directly the determinant.

kronbichler: You are right, `JxW(0)` includes the weight.

kronbichler (approved):

include/exadg/operators/inverse_mass_operator.cpp
```diff
@@ -64,20 +71,81 @@ InverseMassOperator<dim, n_components, Number>::initialize(
...
+      AssertThrow(cartesian_or_affine_mapping,
+                  dealii::ExcMessage(
+                    "The matrix-free cell-wise inverse mass operator can only be applied "
+                    "with affine mapping on non-hypercube elements."));
+
+      AssertThrow(
+        !coefficient_is_variable,
+        dealii::ExcMessage(
+          "The matrix-free cell-wise inverse mass operator can only be applied with constant "
+          "coefficients over a cell on non-hypercube elements, use the apply_scaling instead."));
```
richardschu:
```suggestion
          "coefficients over a cell on non-hypercube elements, use apply_scale() in this case."));
```

include/exadg/operators/inverse_mass_operator.cpp
```diff
@@ -269,6 +342,26 @@ InverseMassOperator<dim, n_components, Number>::apply_scale(VectorType &       d
       },
       dof_index);
   }
+  else if(data.implementation_type == InverseMassType::MatrixfreeOperator && !is_hypercube_element)
+  {
+    // In the InverseMassType::MatrixfreeOperator case we can avoid
+    // streaming the vector from memory twice.
+
+    // ghost have to be zeroed out before MatrixFree::cell_loop().
```
richardschu:
```suggestion
    // ghost entries have to be zeroed out before `MatrixFree::cell_loop()`.
```

richardschu (commented):
thanks @dominiktassilostill
Did you also test this in debug? :sweat_smile:

dominiktassilostill:
> thanks @dominiktassilostill Did you also test this in debug? 😅

Yes the test runs for me in debug mode.

richardschu:
perfect! thank you for this contribution : )

PR #889, opened 2026-02-10
Implementation of Standard k-epsilon with eddy viscosity coupling

SoorajBharan:
This draft pull request presents an initial implementation of the standard $k-\epsilon$ RANS turbulence model for incompressible flows within the ExaDG framework. The formulation, discretization, and solver coupling are described in the attached PDF, which serves as a reference for this PR.
The purpose of opening this PR is to obtain feedback on the implementation.

### Overview
1. Transport equations for $k-\epsilon$ are implemented using the existing `convection_diffusion` infrastructure.
2. A logarithmic formulation, $\mathcal{K}$ =ln(k) and $\mathcal{E}$=ln($\epsilon$), is used to enforce positivity.
3. The coupling between incompressible RANS equations and tranport equation for $\mathcal{K}$ and $\mathcal{E}$ is done using existing `incompressible_flow_with_transport` library.

### Test Case
The current validation case is a fully developed turbulent channel flow with:
1. periodic boundary conditions in the streamwise direction,
2. no-slip walls in the wall-normal direction.

The setup follows ERCOFTAC Case 032, with initial conditions for velocity, $k$, and $\epsilon$ derived from DNS data, as described in the attached document.

### Current Issues
At the present stage, the solver exhibits numerical instabilities for this wall-bounded flow:

1. When using a BDF time integrator, the simulation diverges after a few iterations due to a rapid growth of the eddy viscosity(dissipation is significantly lower that production).
2. When using a Runge–Kutta time integrator, the turbulence variables blow up near the walls and the instability subsequently propagates into the interior of the domain.

### Ongoing Work
Wall functions for the standard $k-\epsilon$ model are currently being implemented. The observed instabilities are believed to be related to the absence of an explicit near-wall treatment, which is known to be required for the standard $k-\epsilon$ formulation in wall-bounded flows.

### Feedback Requested
I would appreciate feedback on the following points:
- Whether the observed instabilities indicate any fundamental issues in the formulation or solver coupling.
- Whether the current behavior is consistent with expectations for a standard $k-\epsilon$ model without wall functions.
- Any recommendations regarding boundary-condition treatment or best practices for implementing wall functions in this context.

Link to DNS data for test case : http://cfd.mace.manchester.ac.uk/ercoftac/doku.php?id=cases:case032

PDF with brief details of implementation: [brief_description.pdf](https://github.com/user-attachments/files/25207110/main.pdf)


richardschu:
Dear @SoorajBharan
it is great to see that you can use ExaDG for your work.
We are also happy if you want to contribute -- after all, this is how the code and the community can grow!

However, I am not sure what to make of this PR, since reviewing ~53k lines of code takes several weeks if not months, given that none of us can work fulltime on `ExaDG`, let alone on your contribution. 
There was a recent update to deal.II's `contributing.md` (https://github.com/dealii/dealii/pull/19289/changes), which might serve as a guideline here:
```
A good rule-of-thumb is to limit patches to no more than about 1000 new or
changed lines of code. The work required to review a patch scales nonlinearly
with the amount of changes, and a series of smaller patches will make it through
the system faster than one big patch. If you are not sure how to break up your
work feel free to open an issue where we can discuss the changeset.
```

In that spirit, please break down your patch in bitesize pieces, as I think otherwise we are simply overwhelmed.
Also, please consider extending existing functionality, not copying the files, renaming them, and changing a few parts here and there. We are trying to avoid copy-paste as much as possible as it is tricky to maintain in the long run!

PR #897, opened 2026-09-16
replace depracated `MGTransferGlobalCoarsening` with `MGTransferMatrixFree`

richardschu:


richardschu:
I tested this in `release` mode locally; runs through cleanly.
I am merging directly because it is a trivial type replacement.

PR #898, opened 2026-09-16
github wokflows: fix the GitHub Actions expression syntax

richardschu:
I was trying to see why the tests did not run, stumbled upon this typo.
Fixed that in the github settings.
We just had 60 days of inactivity.

This PR fixes the GitHub Actions expression syntax.
Concurrency groups correctly include the event type, preventing runs triggered by different events on the same workflow and Git ref from interfering with each other.


richardschu:
[tests fire again, so first point not addressed in this PR worked]

richardschu:
merging since this is a tiny fix we did not even feel before.

